PR #2, opened 2015-04-04
Adds an (optional, but on by default) prompt before picking up redundant XP evokers

aegolden:
In a recent game I ended up carrying around two lamps of fire for a long time before I used one of them and saw that they *both* went inert.  It looks like that's by design, but I think it would be nice for the game to warn the player that they're about to use up an inventory slot for no benefit, so I added a prompt before picking up a redundant xp evoker.


wheals:
Looks like a35564097 and 021ab238 should have already handled this. The implementation there is a bit harder to turn off, but I don't think there's ever a reason to need more than one, so it's OK to me that you need to use clear_autopickup_funcs() to do so.

Code looks good (though we put opening braces on their own lines, and logical operators at the start of new lines rather than the end of old ones), but you got pre-empted by a dev :P

PR #3, opened 2015-04-06
Fix targeting of Dithmenos spell-shadowing.

amalloy:
If you target a floor tile with `za.`, your shadow shouldn't use `zaf`
targeting, which might, eg, hit an allied monster behind the tile
you targeted.


amalloy:
Applied as 6233edf2a1a5d1921c88d418c5815618bd2fa902.

PR #4, opened 2015-04-16
Panlord resists

alexjurkiewicz:
Treat rF and rC the same during Panlord generation, and make SInv random too.


alexjurkiewicz:
Updated this pull request. Removed SInv randomisation since nobody really liked it, and make resist distribution much closer to existing resist distribution.

PR #5, opened 2015-04-16
Gozag extended

alexjurkiewicz:
Original bug report: https://crawl.develz.org/mantis/view.php?id=9642


MarvinPA:
Pushed these, with some further edits.

PR #6, opened 2015-04-17
Surely Yiuf has collected some decent hammers.

alexjurkiewicz:
Occasionally drop a `good_item hammer` in Yiuf's hut.


alexjurkiewicz:
(The only brand hammers seem to gen with is protection. I'm not sure why.)

cbuchananhowland:
Yiuf's hammers can already generate as artefacts, and every now and then 
one of those can be quite good.

alexjurkiewicz:
Yeah, but much more rarely than this patch.

PleasingFungus:
Consensus in ##crawl-dev was that this was excessive; Yiuf already has plenty of loot.

PR #8, opened 2015-04-18
Myrmecia

TomConrad:
This is the pull request for Myrmecia as discussed on Mantis https://crawl.develz.org/mantis/view.php?id=9647.


alexjurkiewicz:
It would be good to squash all these commits into a single one. Short version: `git rebase -i 93bfbe0~1` and mark all but the first as `fixup`.

TomConrad:
I can't squash the commits today, but I will do it tomorrow.

PleasingFungus:
Since the branch was squashed, I don't think we need to keep the pull request open; further discussion can go in 9647.

PR #11, opened 2015-04-22
Gozag potion petition improvements

alexjurkiewicz:
Note: This branch is not yet ready for merge. I'm opening the pull request for feedback.

### DIFF
--- a/crawl-ref/source/godabil.cc
+++ b/crawl-ref/source/godabil.cc
@@ -4242,6 +4242,26 @@ bool gozag_setup_potion_petition(bool quiet)
     return true;
 }
 
+// Return the Gozag Potion Petition price for a potion effect
+int _potion_effect_pricing(item_def potion)
+{
+    int price = item_value(potion, true);
+    auto pot = get_potion_effect(static_cast<potion_type>(potion.sub_type));
+    // Pot is useless for us
+    if (!pot->can_quaff())
+    {
+        price /= 10; // arbitrary
+    }
+    // Reduced price if we're just extending duration.
+    // Call it a 'repeat customer discount', to encourage
+    // gratuitous potion purchasing towards the end of fights.
+    if (pot->effect_active())
+    {
+        price /= 2; // arbitrary
+    }
+    return price;
+}
+
 bool gozag_potion_petition()
 {
     CrawlVector *pots[GOZAG_MAX_POTIONS];
@@ -4269,6 +4289,7 @@ bool gozag_potion_petition()
                 you.props[key].new_vector(SV_INT, SFLAG_CONST_TYPE);
                 pots[i] = &you.props[key].get_vector();
 
+                // Add a potion combination
                 ADD_POTIONS(*pots[i], _gozag_potion_list);
                 if (coinflip())
                     ADD_POTIONS(*pots[i], _gozag_potion_list);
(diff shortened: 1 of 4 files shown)


crawl-ref/source/godabil.cc
```diff
@@ -4242,6 +4242,49 @@ bool gozag_setup_potion_petition(bool quiet)
     return true;
 }
 
+// Would a certain potion effect be immediately useful?
+// TODO: check duration for all buff-extending effects (might, haste, etc)
+int _potion_effect_pricing(item_def potion)
+{
+    int price = item_value(potion, true);
+    switch (potion.sub_type)
```
alexjurkiewicz: This switch statement is too big. I'd like to reduce the price of any potion if `!get_potion_effect(potion.sub_type)->can_quaff()` but that formulation gives me a type casting error. Not sure what I'm doing wrong.
Once that's done, only special cases will belong in this select statement (like flight & berserk).

crawl-ref/source/godabil.cc
```diff
@@ -4259,7 +4302,7 @@ bool gozag_potion_petition()
             for (int i = 0; i < GOZAG_MAX_POTIONS; i++)
             {
                 prices[i] = 0;
-                int multiplier = 25; // arbitrary
+                int multiplier = 20 + random2(10); // arbitrary
```
alexjurkiewicz: With the change to reduce prices of less-useful potions, this maybe should increase.

crawl-ref/source/potion.cc
```diff
@@ -73,7 +69,11 @@ class PotionCuring : public PotionEffect
             return true;
         }
 
-        return false;
```
alexjurkiewicz: I'm really not 100% about this logic here. Can someone double check??

amalloy: I think this makes it impossible for players to quaff a heal_wounds manually when they're at full HP and not rotted. Maybe that's a good change, but I'm not sure it's a good idea to sneak that in with a gozag change.

alexjurkiewicz: Fair call, I should have mentioned this in the issue body.

crawl-ref/source/godabil.cc
```diff
@@ -4242,48 +4244,85 @@ bool gozag_setup_potion_petition(bool quiet)
     return true;
 }
 
-bool gozag_potion_petition()
+// What's the price of a particular potion effect?
+int _gozag_price_effect(item_def potion)
```
alexjurkiewicz: godabil.cc: In function ‘int _gozag_price_effect(item_def)’:
    godabil.cc:4248:5: warning: no previous declaration for ‘int _gozag_price_effect(item_def)’ [-Wmissing-declarations]
 int _gozag_price_effect(item_def potion)

Not sure how to fix this

amalloy: @alexjurkiewicz I think you just need to add a `static` to the front: `static int ...`

MarvinPA:
I don't see any reason to guarantee healing and magic, seems very unnecessary to me.

amalloy:
I agree @MarvinPA. In my Mu^Gozag game, I was usually not looking for healing potions anyway, but for buffs. And if I did need a heal, well, most petitions included a heal anyway.

MarvinPA:
I'm not sure about the faith removal thing but I think a one-time cost increase for all three abilities would be better than just for potion petition, maybe.

alexjurkiewicz:
I've removed guaranteed healing/magic. Do you still think faith should increase the price for all abilities? I'll update the commit if so.

alexjurkiewicz:
@MarvinPA I think this is ready for merge -- any comments if not?

MarvinPA:
I'm not sure about discounting extended potions - I think having a good effect extended is still pretty useful. I don't really know about discounting useless ones either for that matter, it seems possibly just like it makes the ability unnecessarily more complex for not a huge gain.

I definitely like tying Gozag gold drops to something other than monster size, threat is probably reasonable there.

Also still unsure about faith generally, seems tricky to give it a reasonably consistent and intuitive effect on removal - it might actually be more reasonable to just have faith have no effect with Gozag, since Gozag doesn't care about anything other than gold itself.

alexjurkiewicz:
I've removed the commit that changes gold drop values since it needs further tweaking before it's ready for commit.

MarvinPA:
Pushed the variable potion pricing, thanks. Having gold drops work in some different way would be nice but that probably needs more thought and can be a separate thing.

PR #12, opened 2015-04-24
Remove aevp-based speed penalty for UC/throwing.

alexjurkiewicz:
Makes sense from a lore perspective (...sorta), but in practice just a
nerf for two attack forms that didn't need nerfing.


neilmoore:
Committed (0.17-a0-840-g666ab1d) after some discussion with elliptic, who pointed out that the AEVP > 3 cutoff meant that the delay didn't affect all that many characters anyway.  I made a few tweaks to get it to compile and to remove the now unused armour_penalty variable.

PR #15, opened 2015-04-28
Rework Powered By Death.

alexjurkiewicz:
Old PbD relied on corpses in LoS to determine regen strength, which
encouraged kiting and waiting around in corpse LoS to heal up. The bonus
was strongest at the end of a fight which wasn't very interesting.
Finally it didn't work with Gozag, which annoyed me.

New PbD works without corpses. On kill, PbD is activated for a random
number of turns (this is unchanged). However the strength is a player
property that's (possibly) incremented with each kill, rather than being
dependant on corpses in LoS. The strength degrades with time.

Note that new PbD starts and increases strength on any kill, not just
those that leave a corpse. This is a significant buff in low-corpse
areas and has been balanced by reducing the chance a kill will increase
the buff strength.

This is a simplified implementation of a stacking buff, the current
numbers are some what arbitrary and will probably need further tweaks.
They are:
duration: mutlevel*8 + 2d8
strength: increases per-kill with chance 100%->30% decreasing with
    number of stacks. Decreases by 1 on 1/4 turns, no maximum.

### DIFF
--- a/crawl-ref/source/mon-death.cc
+++ b/crawl-ref/source/mon-death.cc
@@ -2566,40 +2566,48 @@ int monster_die(monster* mons, killer_type killer,
             corpse = corpse2;
     }
 
-    if ((killer == KILL_YOU
-         || killer == KILL_YOU_MISSILE
-         || killer == KILL_YOU_CONF
-         || pet_kill)
-             && corpse >= 0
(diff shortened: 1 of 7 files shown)


alexjurkiewicz:
ping @gammafunk ? @elliptic ?

PR #16, opened 2015-04-29
Rework Spiny mutation.

alexjurkiewicz:
Previously had a 1/33-1/22 chance of doing 2-3d5 damage to each
attacker once per round. Now applies (1+mut)d5 damage after each attack.

Numbers might need tweaking, it's substantially better now. But old
spiny was very terrible, so.


MarvinPA:
Closing, since elliptic pushed a different spiny reword.

PR #18, opened 2015-04-30
Don't call followers of good gods Hellbinder/Demonologist

amalloy:
Summoner titles for followers of good gods don't really sound right when they talk about doing stuff that your god hates, and you probably don't do anyway. I changed those two titles to something more good-sounding. I tried to pick titles that make sense as a divine leader sort of thing, but I'm not at all tied to the specific titles I chose. If anyone has better non-demonic titles I'd happily use those instead.


PleasingFungus:
Pushed as 863cd0925eaba2e6cca46a3f9ec990933e490c9a, thanks!

PR #21, opened 2015-05-04
Support Yama LSM security requirements for ptrace.

alexjurkiewicz:
On modern linux systems you need explicit permission to ptrace a non-child process, so make sure we give it when starting gdb.

This patch is not ready for merge, but this area of crawl is scary to me so I'm looking for feedback.

**TODO 1**

    <+|amethyst> chequers:  needs something to make it Linux-only

(I don't know how to solve this. Also, should it be webtiles only?)

**TODO 2**

    <+|amethyst> also, I don't like this 'sleep 1 to avoid a race'

(I don't know what a better solution would be -- any suggestions?)

**TODO 3**

    <+|amethyst> and our TARGET_OS_LINUX is enabled for Cygwin for some reason??

(Is this a TODO for this PR?)

### DIFF
--- a/crawl-ref/source/crash.cc
+++ b/crawl-ref/source/crash.cc
@@ -67,6 +67,17 @@ template <typename TO, typename FROM> TO nasty_cast(FROM f)
 
 #endif // BACKTRACE_SUPPORTED
 
+// Support Yama LSM ptrace restrictions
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
fixed todo1 & todo2, still waiting on todo3

PR #25, opened 2015-05-10
Rework Powered By Pain.

alexjurkiewicz:
Similar to my spiny tweak, make the proc rate constant and more frequent. Also ensure every proc does something useful (currently 50% of procs add MP, which is useless on many characters).

Also add some more effects to the list.

### DIFF
--- a/crawl-ref/source/ouch.cc
+++ b/crawl-ref/source/ouch.cc
@@ -653,33 +653,56 @@ static void _powered_by_pain(int dam)
     const int level = player_mutation_level(MUT_POWERED_BY_PAIN);
 
     if (you.mutation[MUT_POWERED_BY_PAIN]
-        && (random2(dam) > 4 + div_rand_round(you.experience_level, 4)
-            || dam >= you.hp_max / 2))
+        && (one_chance_in(5)
+            || dam >= you.hp_max / (2 + level)))
     {
-        switch (random2(4))
+        mpr("You focus on the pain.");
+        bool success = false;
+        int tries = 0;
+        do
         {
-        case 0:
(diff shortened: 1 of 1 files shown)


crawl-ref/source/ouch.cc
```diff
@@ -653,33 +653,56 @@ static void _powered_by_pain(int dam)
...
+        bool success = false;
+        int tries = 0;
+        do
         {
-        case 0:
-        case 1:
-        {
-            if (you.magic_points < you.max_magic_points)
+            tries++;
+            switch (random2(4))
```
alexjurkiewicz: not sure if this is the best way to select between code paths here -- it's a bit fragile. Adding another option requires you to remember to increase this integer.

crawl-ref/source/ouch.cc
```diff
@@ -653,33 +653,56 @@ static void _powered_by_pain(int dam)
...
+        mpr("You focus on the pain.");
+        bool success = false;
+        int tries = 0;
+        do
         {
-        case 0:
-        case 1:
-        {
-            if (you.magic_points < you.max_magic_points)
+            tries++;
```
alexjurkiewicz: rather than trying a random case each time, is there some way to randomly order the options and try them in turn? there's an annoying case with my code where 1/4 options is available but it doesn't get tried due to RNG

MarvinPA:
Closing, this was discussed in IRC a bit - PbP is quite strong and noticeable currently. The occasional strong effects work pretty well, and it's also intended for the mutation to push the player towards using MP in some way in order to be able to benefit from the recovery effect.

PR #26, opened 2015-05-11
Make constriction messages use 'your' when appropriate

amalloy:


amalloy:
I wasn't sure of the right place to put my helper function, or whether there's some other name preferred over `article_for`.

amalloy:
This works properly for uniques, too, which I was encouraged in ##crawl-dev to check: "constricted by Vashnia" rather than "constricted by your Vashnia"

PR #30, opened 2015-05-12
Check for MUT_NO_LOVE before making Pikel's slaves neutral

amalloy:
Since the pikel-death code is shared with the code for making holy allies turn neutral when you leave TSO, I updated that message too. It will probably never come up (you get MUT_NO_LOVE, switch to TSO, summon some hostile "allies", and then abandon TSO?), but it seemed like the message should match the behavior just in case.


neilmoore:
Committed in 0.17-a0-1012-g414ce7a, thanks!

PR #35, opened 2015-05-15
Treat abandon+worship as direct switching, for E/Z/1.

amalloy:
The religion screen isn't clear about *how* you should switch
between the good gods, so we should really treat aXp the same
as just p.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -2824,7 +2824,7 @@ static spret_type _do_ability(const ability_def& abil, bool fail)
             && yesno("Are you sure you won't change your mind later?",
                      false, 'n'))
         {
-            excommunication();
+            excommunication(GOD_NO_GOD, false, true);
         }
         else
         {
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -2824,7 +2824,7 @@ static spret_type _do_ability(const ability_def& abil, bool fail)
             && yesno("Are you sure you won't change your mind later?",
                      false, 'n'))
         {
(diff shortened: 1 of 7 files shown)


MarvinPA:
I think your unmarshalling code isn't quite right, since you have the unmarshalling of saved_good_god_piety and previous_good_god disappearing on major save bump - it needs to initialise them to 0 with the tag check, or otherwise unmarshall them normally?

amalloy:
You are probably right, @MarvinPA, but I guess I don't understand the forward-compatibility design in tags.cc. It sounds to me like you are saying that if major version is changed to 35, this unmarshalling code disappears, which does look like a problem indeed. But didn't I do it similarly to `TAG_MINOR_LORC_TEMPERATURE`, which also has unmarshalling that will vanish in version 35? What should I be doing differently to be forwards-compatible here? Just take it out of the #ifdef, which I guess is not needed for backwards compatibility, so that it always happens?

MarvinPA:
Lava orc temperature is something that's being removed when the major version is bumped, so the unmarshalling code for it is also marked for removal. Since saved_good_god_piety is being added, you want to initialise it to 0 if the major version is 34 and the minor version <= TAG_MINOR_SAVED_PIETY (meaning the save comes from a game that didn't already had a saved piety field). Otherwise, you want to unmarshall it normally since the save must be from a game that already has that field.

TAG_MINOR_RU_SACRIFICES is an example that does that kind of thing (just keep an eye on where the braces and #ifdefs go, since that one was also wrong until just recently! :P). Probably |amethyst or somebody else in IRC can do a better job explaining save compat than me if this didn't make any sense.

PR #37, opened 2015-05-19
Rename Sustain Abilities to Sustain Attributes.

alexjurkiewicz:


alexjurkiewicz:
There are some other sustab referenced in `crawl-ref/source/wiz-dump.cc` which I think are old code compact so I didn't touch -- is that correct??

alexjurkiewicz:
Fixed the wiz-dump conditionals @wheals mentioned

PR #38, opened 2015-05-21
Remove Ghoul random rotting.

alexjurkiewicz:
Requires the player to butcher every corpse they find to maximise the
chance they have a chunk on hand all the time, which is simply tedious
without adding much gameplay depth outside low-corpse branches.

Also doesn't play nice with Gozag!

### DIFF
--- a/crawl-ref/source/player-reacts.cc
+++ b/crawl-ref/source/player-reacts.cc
@@ -962,22 +962,22 @@ static void _decrement_durations()
         }
     }
 
-    if (you.species == SP_GHOUL)
+    if (you.species == SP_GHOUL and you.hunger_state < HS_SATIATED)
     {
         int resilience = 400;
 
         if (you_worship(GOD_CHEIBRIADOS) && you.piety >= piety_breakpoint(0))
             resilience = resilience * 3 / 2;
 
-        // Faster rotting when hungry.
-        if (you.hunger_state < HS_SATIATED)
-            resilience >>= HS_SATIATED - you.hunger_state;
+        const int hunger = HS_SATIATED - you.hunger_state;
+
+        resilience >>= hunger;
 
         if (one_chance_in(resilience))
         {
             dprf("rot rate: 1/%d", resilience);
             mprf(MSGCH_WARN, "You feel your flesh rotting away.");
-            rot_hp(1);
+            rot_hp(random2(hunger * 2) + 1);
         }
     }
 


doy:
Not a fan. Not only does this remove pretty much the only unique feature of Ghouls, "always wanting to butcher things" is also true of, say, Trolls just because of their actual hunger.

alexjurkiewicz:
Updated so rotting only occurs when hungry (or worse) and the rot amount is higher. Might be a bit too cruel, and too likely to send hungry ghouls into a rotting death spiral.

doy:
What gameplay effects are you trying to achieve here?

alexjurkiewicz:
In short, rotting while satiated is low impact and annoying to deal with, I wrote more about it here: https://crawl.develz.org/tavern/viewtopic.php?f=8&t=16237

doy:
Most people in that thread seem to disagree with you (and I do too). It really sounds like the issues you have here are more with chunk eating in general than ghouls specifically (it's not clear from your description why eating chunks is more irritating to you for ghouls than for trolls, for instance) and so it'd be nice to have a more coherent overall design in mind.

doy:
In particular, I disagree that it is low impact, since it adds an additional motivator to keep moving, even if you have plenty of permafood. Making ghouls not rot unless hungry would eliminate that.

doy:
Also also, as a more general concept, this would remove an interesting aspect of both ghouls and non-permafood without any compensation, which makes it more likely for people to just want to remove one or both of them outright (both have been suggested in the past). This is a valid opinion and discussion to have, but if that is the actual goal, I think that is what we should actually be discussing. If that's not the actual goal, we should find a way to accomplish the actual goals that doesn't solely consist of removing interesting parts of the game.

alexjurkiewicz:
I think my first approach was not good -- as you said, removing the Gh food clock was a Bad Idea. I like Gh food clock, it's well designed. Specifically, I prefer it over the conventional food clock because the punishment for playing slow is more graduated and transparent.

The two niggles I have with it are:
* Bad effects happen even when playing ahead of the clock.
* The bad effect - one point of rot - is negligible at any point in the game without stacking, and fixing it requires manual intervention.

So this PR is an attempt to keep the Gh food clock intact but a) not punish the player when they're doing fine, and b) decrease the time between when the punishment is applied and when it becomes relevant.

Getting into your individual points:

> It really sounds like the issues you have here are more with chunk eating in general than ghouls specifically (it's not clear from your description why eating chunks is more irritating to you for ghouls than for trolls, for instance) and so it'd be nice to have a more coherent overall design in mind.

I'm fine with eating -- the `easy_eat_*` rc commands make you eat automatically when necessary (for hunger reasons). Butchering is a manual process, but most species (including Tr) nearly never need you to explicitly eat chunks. In comparison Gh does (to fix rot).

Needing manual work to fix rot as a punishment for falling behind the food clock is fine, but I think it's not good to sometimes be punished even when you're keeping up.

> In particular, I disagree that it is low impact, since it adds an additional motivator to keep moving, even if you have plenty of permafood. Making ghouls not rot unless hungry would eliminate that.

In both the current and this PR's system, you're motivated to keep progressing to get chunks to stave off rotting. The point at which rotting kicks in is different, but the motivation to keep moving is definitely still there.

> Also also, as a more general concept, this would remove an interesting aspect of both ghouls and non-permafood without any compensation, which makes it more likely for people to just want to remove one or both of them outright (both have been suggested in the past).

Yeah, agreed. I'm not suggesting removal of the Gh food clock -- just the removal of rotting while non-hungry :)

doy:
Maybe the thing we actually want here is to make `easy_eat_*` also eat chunks if you're a Ghoul with rot? I'm still not a fan of the way that the current patch makes permafood affect how much Ghouls rot.

doy:
Hmmm, alternately, I wonder how big of a change it would be to make Ghouls not able to eat anything except chunks?

doy:
Basically, I just don't want to make optimal behavior for Ghouls to be to eat chunks as little as possible, since that kind of ruins the whole Ghoul concept (and I feel like that's what this pull request would encourage, intended or not).

alexjurkiewicz:
> Hmmm, alternately, I wonder how big of a change it would be to make Ghouls not able to eat anything except chunks?

> Basically, I just don't want to make optimal behavior for Ghouls to be to eat chunks as little as possible, since that kind of ruins the whole Ghoul concept (and I feel like that's what this pull request would encourage, intended or not).

Yeah, I agree, Ghouls should want to eat most chunks they generate. My problem is the current system to encourage them to do so is mildly annoying even when you are chunk-rich and doesn't really have a gameplay impact in even 15 rune games.

Part of that is definitely because permafood helps massively in low-chunk areas, since there are only a few parts where you want to eat a lot of rations. Not being able to eat permafood would be a very cool change! I considered it but slow areas of the game (Tomb:2/3, Zot:5, PanHell lords) get very difficult if not impossible.

In fairness to the current system, if you removed permafood its design would be a perfect clock for these slow areas. Maybe the solution is to remove permafood after all?? But that alone doesn't address the other half of my concern: the current system is annoying!

If your big problem now is that this change reduces the chunk-eatyness of Gh, what do you think about adding Fast Metabolism to them? I haven't really thought properly about the impact during the problem areas (mentioned before), but for the rest of the game it wouldn't be a huge problem while increasing chunk-eating rate.

alexjurkiewicz:
One other idea: chunks go back to healing only 1hp of rot for Gh.

MarvinPA:
Closing this since it needs more thought and discussion to come up with something that would work.

PR #42, opened 2015-05-25
Explicitly reveal mindelay skill to the player.

alexjurkiewicz:
Previously base & minimum delay was revealed to the player, which
required maths to make use of. Now the base delay is hidden and instead
is shown mindelay and the skill required to reach it.

For weapons that cannot reach mindelay (eg Dark Maul), the level shown
is 28, which I think is nice :).

This commit would break for any weapons that require an odd skill level
to reach mindelay, but I don't think any such weapons do/can exist.


bhickey:
amalloy is right on this one. Showing the required skill is a great improvement, but let's retain the delay numbers as well. There's plenty of room on the weapon screen.

alexjurkiewicz:
Base delay is shown as before now

PR #47, opened 2015-05-30
Another attempt at make_fiery_weapon (volcano)

drke2:
Bumping up glaive made it overly weighted towards polearms, which bothered me, so removed spear since it had no equivalent in the other weapon classes, and tweaked the weights to be around 30 for LBl, Polearms, and Axes. Left the other types about where they were and added some basetypes I felt should be included. This shouldn't be a buff to volcano loot overall in light of the previous commit reducing lajatang weight from 10 to 1.


gammafunk:
I proposed a comment cleanup: http://sprunge.us/VXVO

Uses 72-char line length and the more direct style we prefer for commit messages. Let me know if it's ok to update the commit message with this (and feel free to propose your own update message).

PR #48, opened 2015-06-01
Tweak GC/GSC base delays

alexjurkiewicz:
These aren't used much right now since they overlap so closely with
GSCs. 16dam/-4acc, compared to 15/-1 for eveningstar/double sword,
22/-7 for GSC and 17/-4 for great mace. Delay remains unchanged.

### DIFF
--- a/crawl-ref/source/itemprop.cc
+++ b/crawl-ref/source/itemprop.cc
@@ -384,10 +384,10 @@ static const weapon_def Weapon_prop[] =
     { WPN_GREAT_MACE,        "great mace",         17, -4, 17,  9,
         SK_MACES_FLAILS, SIZE_MEDIUM,  SIZE_BIG,    MI_NONE,
         DAMV_CRUSHING, 3, 10, M_AND_F_BRANDS },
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
I basically agree with this concern:

    06:16 <+gammafunk> well I guess my concern is creating a situation where, because of Og ev penalty, you'd be silly to ever go 2h
    06:17 <+gammafunk> since dealing with bad defenses to get massive damage is kind of the only reason og exists
    06:18 <+gammafunk> currently an og can go evening star and get acceptable damage with good defenses, but not something close to 2hweapon damage.
                   Making gc just slightly better than eveningstar also doesn't sound more interesting
    06:18 <+gammafunk> again elliptic is good to query on this; maybe he can see a fit for gc or another 1h weapon
    06:19 <+gammafunk> if gc really is boring removal is always an option

Any devs with an opinion?

alexjurkiewicz:
Updated title and commit as per IRC discussion

alexjurkiewicz:
Ping. Commit message:

> Now require 18 & 24 skill respectively to reach mindelay (from 20 & 22) which should give GCs a bit more of a niche to be useful in.

PR #49, opened 2015-06-02
Add +weapon skill as artefact properties.

alexjurkiewicz:
They generate rarely on artefacts with the following restrictions:
* Only one +skill type per artefact
* Weapons can only get +skill that is relevant

Approximately 1 in 10 artifacts gets a +skill.

Also update Fencer's Gloves (and art-data.pl) to use these new
properties.


alexjurkiewicz:
I was thinking of this while deleting the spear of botono. I'd like to extend the system to generate skill bonuses for spellcasting and other skills and to weight the generated skill depending on the likely use of the armour (eg +/-spells on robes, +/- weapons on plate).

alexjurkiewicz:
todo: update shopping.cc artefact_value

MarvinPA:
This doesn't sound like an especially interesting property to me. It's very narrow, and I don't think it's even especially great on the Fencer's Gloves (where it's at least a unique, vaguely thematic unrand property).

alexjurkiewicz:
I think this prop is interesting on weapons: a good weapon with +skill in a
cross trained category would become more appealing, and -skill on a weapon
you like would create a choice about "overskilling".

More generally I like that this prop is sort of a vague meleedude analog to
spell enhancers. Slay is to Wiz as +Axe is to a fire enhancer!

alexjurkiewicz:
21:44 <+|amethyst> chequers:  I do think -skill on armour/jewellery probably should go
21:44 <+|amethyst> chequers:  since it's so likely to be irrelevant

cbuchananhowland:
I'm concerned that +slaying is more like Wiz & a fire enhancer at once, but so is +Axes. The differences are 1) +Axes doesn't affect aux attacks, and 2) the scaling is a bit different. I'm not sure if either of those is a compelling enough difference to add this on top of +slaying -- tho I suppose weapons can't get +slaying right now.

If players find the property on a weapon that's worth using and also haven't skilled to min delay already, it could affect their skilling choices. I'm having trouble deciding whether I think that it would significantly impact skilling choices and whether that's a good thing. 

I think I'm leaning slightly against, but wouldn't be against trying it out.

MarvinPA:
I agree with the concerns PleasingFungus and gammafunk mentioned in IRC, and still don't think it's interestingly distinct from a more narrow +slaying - even though it does have some slightly different effects, I don't think the extra complication makes them worthwhile.

alexjurkiewicz:
Sounds like this idea isn't well received in general, so I'll lcose this PR... for now...

PR #57, opened 2015-06-08
Optionally wait for full HP and MP before exploring

robobario:
A minor improvement to auto explore. The option defaults to false. (my flatmate wanted it to be like the ToME auto-explore)


neilmoore:
Hm, it feels like this should use rest_wait_percent rather than 100%.  See should_stop_resting in player.cc, for example.

robobario:
I've updated is_completely_rested to use should_stop_resting internally.

neilmoore:
It's in, thanks!  Turns out should_stop_resting isn't exactly what you need because it uses == where this should use >=, sorry about that.  I fixed it up (and removed the stray init.txt).

PR #60, opened 2015-06-11
Ecu altar fixes

alexjurkiewicz:


MarvinPA:
The logic-moving commit there has two different versions of _pray_ecumenical_altar (the old one as well as the updated one).

MarvinPA:
Fixed that and pushed, minus the Ru bonus.

PR #63, opened 2015-06-12
Fix Ru's apocalypse damage

amalloy:
It was supposed to be NdX, but was accidentally XdN.

Also tidied up a bit by removing some repetition, and moving related
variables closer together.


amalloy:
This is a non-trivial nerf to apocalypse, by the way. EV of 16d4 is 40, whereas EV on 4d16 is 34, plus more variance. May want to consider whether that is a desirable change - maybe the existing implementation is the right amount of damage even if it wasn't intentional.

PR #64, opened 2015-06-13
Ecumenical Altar tweaks

alexjurkiewicz:
1. Update feature description
2. Give benefit for Monks when they roll Gozag or Ru
3. Fix the "no god" bug

### DIFF
--- a/crawl-ref/source/dat/descript/features.txt
+++ b/crawl-ref/source/dat/descript/features.txt
@@ -88,8 +88,7 @@ A dried-up fountain that has seen better days.
 A faded altar to an unknown god
 
 An ancient altar, stripped of all identifying characteristics by the passage of
-aeons. Surely the altar's deity would greatly appreciate one who converts at so
-ancient a site of worship.
+aeons. You could pray here to convert to an unknown god.
 %%%%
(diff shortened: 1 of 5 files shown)


alexjurkiewicz:
I know Marvin doesn't like special treatment for Gozag/Ru (ref 43b0ec6) but I think it's good to for the altar to consistently reward Monks no matter what God is rolled.

MarvinPA:
The description has already been updated, and I've pushed the GOD_NO_GOD fix. I still strongly dislike the idea of having a bunch of extra special cases for Monks. In general pretty much anyone using the altar has good odds of just getting unlucky (and rolling Xom, or someone completely unsuited to the character), it seems perfectly fine for that to happen to apply a tiny bit more to Monks.

gammafunk:
I agree that adding a lot of special cases to fix undesirable results for some chars isn't good. Fixing the lack of extra piety for Monks for a couple gods but not fixing e.g. an FE getting Dith (or a Wn starting with fire spells or anyone else with access to fire spells before the altar) or other much more problematic results is inconsistent, and we don't want to go down the route of handling many special cases for bad combinations. Closing this PR.

PR #65, opened 2015-06-15
Ashenzari boosts UC when shield hand is bound.

alexjurkiewicz:
In some ways this is a problem -- Ash gives a skill boost to UCers who
can freely swap to other weapons. On the other hand, this gives UC
fighters a weapon boost from Ash.

### DIFF
--- a/crawl-ref/source/godpassive.cc
+++ b/crawl-ref/source/godpassive.cc
@@ -612,6 +612,7 @@ map<skill_type, int8_t> ash_get_boosted_skills(eq_type type)
     case (ET_SHIELD):
         if (bondage == 2)
             boost[SK_SHIELDS] = 1;
+            boost[SK_UNARMED_COMBAT] = 1;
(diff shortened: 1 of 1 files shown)


plathrop:
Yes please! Make Monk^Ash useful!

MarvinPA:
I don't see a good reason for UC fighters to be able to get a skill boost without having to be bound for it, really. Seems fine that worshipping Ash means you have to wield something, if you want a boost for it. Maybe it should be possible to curse your hands, preventing you from wielding anything at all (not really a serious suggestion, but if this is really a problem that needs fixing then that seems like a better solution to it than this).

cbuchananhowland:
I agree w/ MarvinPA.

gammafunk:
Agree that this is a bad idea. Giving a boost to a slot that you can still swap is basically against the concept of Ash. Closing this PR.

PR #66, opened 2015-06-15
Make the heading of Ash's skill-boost table stand out

amalloy:
A player noted that when it looks like all the other rows, it is hard to immediately tell what this table is all about. So, I made it red. Some ascii art lines could also work, or even just a single line like 

    -------------------

between heading and body rows.


MarvinPA:
Pushed but with white instead of red.

PR #68, opened 2015-06-19
Clean up and tweak Frances's spell list

drke2:
Summon Demon was listed twice, so split the duplicate SD's chance into
the other spells with a preference for SD itself and Haste (the other
spell with increased cast frequency).


drke2:
On second glance the SD casting rate was probably intended.

gammafunk:
Squashed the commits and merged, but forgot to issue close in the commit message.

PR #70, opened 2015-06-20
make monsters have better self-preservation around clouds

coledot:
Fixes bugs involving monster movement near clouds. Wandering monsters no longer path into clouds. Intelligent monsters want more HP before entering a damaging cloud; in other words, friendly monsters are less likely to do walk into a friendly cloud & die.

https://crawl.develz.org/mantis/view.php?id=2980
https://crawl.develz.org/mantis/view.php?id=9759


bhickey:
A few folks wanted to do some more testing on this to see if the AI behaves in the "right" way. Let's get this into trunk for some hands on testing. Pardon the merge latency delay.

PR #73, opened 2015-06-22
Add framework for brand-based artefact weapon lookup.

alexjurkiewicz:
Not ready for merge. Just looking for feedback about the idea and technical approach.


wheals:
I guess there are no obvious problems with this, though getting rid of buggy-vorpal would be nice there. But anyway, if it's not ready to merge, no reason to keep it open here.

PR #74, opened 2015-06-22
Tweak the Sack of Spiders

drke2:
Make it worse at lower evocations, better at mid to high evocations.

The issue with sack of spiders is that most of the Spider Nest
population deals a similar amount of damage (20~, speed 15), mainly
differing in special traits. This allows a character to utilize a large
portion of this item's potential with very modest (8~) evo, using
trapdoor spiders and redbacks to overwhelm things. This is further
encouraged by the fact that the higher-tier spiders that would actually
be significant upgrades (tarantella and wolf spider) take a very large
investment in evo to gain access to. Most characters don't see a lot of
variety with this item because of this.

This commit addresses this by adjusting the weight and level for
trapdoor and redback while increasing weight of the basic (10 dam, low
HD) spider. This is compensated by making the higher-tier spiders
accessible with a more reasonable level of evo, with everything except
ghost moth available at evo 15. This gives the item a similar feel to
rod of clouds wrt scaling, but gives more incentive to go past 15, as
further skill investment continues to improve the spider population.

### DIFF
--- a/crawl-ref/source/evoke.cc
+++ b/crawl-ref/source/evoke.cc
@@ -1017,16 +1017,16 @@ string manual_skill_names(bool short_text)
 
 static const pop_entry pop_spiders[] =
 { // Sack of Spiders
-  {  0,  10,   5,  FALL, MONS_GIANT_COCKROACH },
-  {  0,  10,   5,  FALL, MONS_SCORPION },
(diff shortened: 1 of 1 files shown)


gammafunk:
Merged in 0.17-a0-1483-g86df3dc

PR #75, opened 2015-06-23
Hammertime

alexjurkiewicz:
Club / Hammer supremacy!


MarvinPA:
This was discussed in IRC a bunch and had various problems pointed out, adding various mindelay complications and exceptions here isn't really a good idea.

PR #76, opened 2015-06-23
Allow undead player species to leave ghosts.

alexjurkiewicz:
It's interesting lore, but not well explained to players and just
results in a lower variety of player ghost types.


alexjurkiewicz:
ping

wheals:
This is a harmless flavour thing, and enough such things have been removed that I don't think it's worth getting rid of this one. I don't believe there are serious gameplay implications either way.

PR #81, opened 2015-07-05
Small adjustment to orb spider generation on the sack of spiders

drke2:
PR note: This is a change I wanted to make but forgot about, their
range should stop at 25 like redbacks since they show up too much
at 27, and in retrospect flat is better than semi.


gammafunk:
A made a commit in 0.17-a0-1489-g3ee15ac based on our irc discussion. Removes the breakpoint at 25 evo altogether and keeps the chances for orb spiders reasonable.

PR #82, opened 2015-07-11
Distinguish between 0-xp monsters and non-threatening ones

amalloy:
The most notable thing this fixes is that lurking horrors show up in the monster pane now, and are targetable with `x=`. As mentioned in the commit comment, someone should check whether the decisions I made about which mons_class_gives_xp checks to change and which to leave alone, because I may not have understood the intent in all cases. Similarly the decisions in mon-data.h about which to mark unthreatening may not be quite right.


neilmoore:
Mantis issue #9863

neilmoore:
Thinking about it some more, making ordinary ballistos not count as firewood is problematic I think: not only does it causes a ./crawl -test failure "ballistomycetes has speed 0", but it changes polymorphing, slow healing, autofight, and several other things.  Probably mostly academic, but as long as ballistos technically exist...

You can't tell from the monster class alone whether a ballisto is no-threat, because it might be active or inactive.  Perhaps something like:
    bool mons_is_threatening(monster *mons)
    {
        return mons_class_is_threatening(mons->type) || mons_is_active_ballisto(mons);
    }
Then use that in several places that use mons_class_is_threatening.

wheals:
Will this fix mantis issue [#9532](https://crawl.develz.org/mantis/view.php?id=9532)?

doy:
Rebased and merged separately with some more fixes in 6b0f55fb.

PR #85, opened 2015-07-16
Don't say that flying things 'stumble backwards' when trampled

amalloy:
Instead, they should be shoved backwards or something.

### DIFF
--- a/crawl-ref/source/melee_attack.cc
+++ b/crawl-ref/source/melee_attack.cc
@@ -3476,9 +3476,10 @@ bool melee_attack::do_knockback(bool trample)
 
     if (needs_message)
     {
+        const string verb = defender->airborne() ? "are shoved" : "stumble";
         mprf("%s %s backwards!",
              defender_name(false).c_str(),
-             defender->conj_verb("stumble").c_str());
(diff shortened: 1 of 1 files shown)


MarvinPA:
We use "stumble" in a bunch of other places, I'm not really sure if it's worth changing all of them while flying (like stumbling around while confused, no idea how you'd make that make sense while flying).

amalloy:
i don't think there are any other actions that cause stumbling to actually happen. There are two that say like "are you sure you want to stumble around while confused" or "you nearly stumble into lava". The second doesn't apply while flying anyway, and the first seems okay because it is about you being unable to coördinate your feet or whatever, which could apply while flying.

The only other instance of "stumble" is that when a high-powered Zin recitation confuses a monster it will print the message "The orc priest stumbles around in disarray", which seems about the same as stumbling around while confused.

All of these seem to be a different meaning of "stumble", ie "move around clumsily", than when being shoved by a huge dang elephant. Maybe my objection isn't really related to flying at all? I wouldn't mind just unconditionally changing this to "shoved backwards" regardless of flying state.

PR #94, opened 2015-07-21
use unwind_bool typedef

ldierk:
Replaces the two remaining uses of unwind_var\<bool\> with the appropriate typedef


gammafunk:
It's probably best if you don't merge master for a pull request branch of a simple change like this, since we'll want to rebase out that merge commit. You could force-push a rebased commit if you like, but it's probably not necessary to worry about keeping such a simple branch up to date (it will always rebase easily)

A question: what entry would you like for us to use in .mailmap? We tend to use names on irc or the shorter github account names, so is 'ldierk' ok?

PR #96, opened 2015-08-04
Makes move and travel honour exclusions on stairs.

4Hooves2Appendages:
Auto-travel / explore can venture into dangerous territory, passing
through excluded stair cases.

This change causes a prompt before moving or travelling onto an excluded staircase.

See image of problematic behaviour:
crawl error


MarvinPA:
You've got an unrelated commit in here somehow.

Also for reference, I'm relatively sure this was an intentional change for staircases that are excluded with a single-square exclusion, since most commonly the aim for doing so is to prevent actually using the staircase, rather than walking past it (and blocking walking past them can be annoying for staircases in corridors).

The example in the screenshot is a case where this doesn't work out though, yeah. Perhaps if the staircase exclusion has any adjacent exclusions at all, it should prevent travel past it?

4Hooves2Appendages:
Thanks for the comment.

I also thought that the behaviour would be intentional, but especially to someone inexperienced the current behaviour must actually seem like a bug.

I'm not sure how best to resolve it. Maybe single-excluding a staircase should ask whether to exclude the staircase on the other floor? Or maybe that should be a seperate option?

MarvinPA:
http://sprunge.us/dSFJ would be my proposed fix, it definitely resolves the issue in the screenshot but I don't know if it just ends up making things unnecessarily complicated (or if there are other feasible edge cases where it'd be problematic).

4Hooves2Appendages:
It would definitely fix the example above. It would leave single exclusions in corridors.

Do you think it's actually a big problem to prompt people before walking onto an excluded staircase? Aren't they mostly going to use the staircase? Also there is actually an option to exclude the other end of the stairs.

4Hooves2Appendages:
I have an idea, but it will take a bit of time.

PR #106, opened 2015-08-31
Use success% for Makhleb demon friendliness.

alexjurkiewicz:


bhickey:
Are you trying to kill players? Because this is how you kill players.

PR #108, opened 2015-09-07
Remove old messages in webtiles

kpdonn:
I've noticed that when playing on webtiles the game is very responsive
at the start but then gets slower and slower the longer it goes on. It
also snaps back to being responsive by just reloading the page. Using
Chrome's javascript profiler I was able to track the primary cause down
to the way messages are added to the page.

Basically every message just added a new div to the page after all the
previous ones. After playing for a while there can be thousands of divs
on the page which bogs down the performance.

This commit removes old messages that cannot be seen by the player every
time a new message is added. It also removes an unnecessary jQuery
scroll command and an unused messages array.

### DIFF
--- a/crawl-ref/source/webserver/game_data/static/messages.js
+++ b/crawl-ref/source/webserver/game_data/static/messages.js
@@ -4,7 +4,6 @@ function ($, comm, client, util, options) {
 
     var HISTORY_SIZE = 10;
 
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
Nice work tracking this down -- the slowdown is a really longstanding bug.

PR #109, opened 2015-09-10
Tiles lighting rework

BodrickLight:
Each tile gets a light colour calculated from the distance to the player; whether the tile is currently visible; whether the tile is lit by a halo or umbra; and a few other factors. This colour is then overlaid on the tile, and blended with adjacent tiles to create a smoother transition.

The amount of blending can be controlled with the tile_light_blur rcfile option. Setting this to 0 will disable the blending completely, while a value of 16 will blend the entire tile. The default value of 6 was chosen to give a good balance of being able to determine the base colour of the tile (important for being able to tell if a tile is in LOS or not), while retaining some smoothing between adjacent tiles.

This method could be expanded further to add coloured lighting based on branch, light spreading further than one tile etc.; but this is a simple implementation for now. It should be easy to change the cell colour calculation algorithm in tileview.cc::_get_colour without changing the rendering process if this is necessary in the future.

Some examples:
Dungeon:
image

Shoals:
image

Volcano + halo + Dithmenos umbra + orb glow:
image

Hybrid:
image


BodrickLight:
I wasn't sure whether to leave it in or not - the current version of qw.rc doesn't work in trunk, and I was using it for stress-testing my changes. I just took the latest version from https://github.com/elliptic/qw.

gammafunk:
Some comments after looking at the build of this branch on each dungeon branch. It seems to have some issues with visual artifacts when the blending affects lighter-colored wall tiles with relatively little "color transition" (sorry I don't know the right term).

I uploaded some examples of what I mean. Here are screenshots of ice cave in WebTiles and Tiles, respectively:

WebTiles icecv
Tiles icecv

You can see that the blending leaves notable artifacts on wall tiles with shading on adjacent edges, dividing the outer portion of the wall that's shaded into rectangles. You can also see this on lighter permarock tiles and other light wall tiles that are relatively uniform:

Zig
Wizlab

The blending looks really nice on wall tiles with a lot of texture, and seems to also work best with floor tiles having a lot of texture as well, since that texture seems to break up the straight lines of shading:

Dungeon

I'm not sure if there's an easy way to make the blending not so linear. For most wall tiles the blend does look fine, but there are some wall tiles like the ones I mention where I'd say the effect makes the wall visual significantly worse. Perhaps others don't feel it's as much a problem as I do, but I wanted to comment before doing further merge work.

BodrickLight:
That's very odd - I don't get the artifacts you're seeing on my machine. It looks like there's a 1px gap between the lighting rectangles, which is why they don't appear to blend particularly well. Your first screenshot (the webtiles one) seems correct - I presume the rest were all done with local tiles builds? It looks like it might be something to do with anisotropic filtering or antialiasing though - I'll have a fiddle around and see if I can figure out what's going on there.

For reference, ice caves look like this on my machine:
image

gammafunk:
The second screenshot of the ice cave was Tiles, but the rest were WebTiles. I would say that even your screenshot is not correct in the sense that the effect renders some of those wall tiles not very well. If you look at the wall tiles at (relative to player) 3 east, 1 south and then 1 east, 2 north the result isn't great. The hard blending lines don't work very well on these tiles. I'm not sure if excluding wall tiles from the effect might be the answer, or possibly it could be good if the blend was less linear in some way. Others may disagree with me about this, however.

But for tracking down any unintended effects in your code, let me know if you need another test. I'm testing on Linux (using Chrome in WebTiles).

BodrickLight:
Ok, I see what you're saying. I'll see if I can figure out a way to soften the blending effect somehow.

PleasingFungus:
Looks like this isn't going anywhere, sadly. Feel free to re-open if you get things working!

PR #110, opened 2015-09-11
Make Blink a single-level mutation

drke2:
Extra levels only serve to reduce the fail rate, leaving the HP cost
and range the same. It's also a fairly low-weight mutation, so extra
levels, when they do occur, tend to do so later in the game - when
the fail rate reduction is less significant due to the player's higher
XL and likelihood of having other sources of Blink. Therefore, getting
extra level(s) in Blink is one of the least interesting results of being
mutated.

Given that the mutation provides a simple ability with fixed properties,
it seems more logical and expected that it would be only one level.

Also clarify the message received upon losing the mutation.

PR: A fail rate tweak seemed necessary since the current values seem
balanced around having higher levels actually do something. At
mutlevel 1 the current fail rate feels too high at the XLs when the
ability is most likely to be used, especially taking into account the
additional cost of 5% MHP. The tweak is approximately equivalent to
the mutation becoming level 1.4. Having said that, I'd be fine with
not merging the 2nd commit if it's not desired.

### DIFF
--- a/crawl-ref/source/message.cc
+++ b/crawl-ref/source/message.cc
@@ -761,9 +761,8 @@ public:
     }
 
 #ifdef USE_TILE_WEB
-    void send(int old_msgs = 0)
+    void send()
     {
-        unsent += old_msgs;
         if (unsent == 0 || (send_ignore_one && unsent == 1)) return;
 
         if (client_rollback > 0)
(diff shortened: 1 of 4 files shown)


MarvinPA:
Looks good to me, but should probably have a TAG_MINOR_VERSION bump and some save compat code to set players with level 2 or 3 of the mutation to level 1 (similar to TAG_MINOR_DETERIORATION for example)?

drke2:
Done I believe, sorry about the extra commit(s).

PR #111, opened 2015-09-15
Mention branch rune on player entry.

alexjurkiewicz:


alexjurkiewicz:
Issues:

* Doesn't mention the demonic rune (should mention it on Pan entry, or for every pan level? Make sure not to mention for pan lord floors)
* Code is pretty split up. Probably best to have a single function take care of messaging

MarvinPA:
The "Abyssal Rune of Zot" should be capitalised as "abyssal rune of Zot" for consistency, and for the Pan runes I think it'd be better to say "This floor/level/realm/whatever" rather than "This branch".

alexjurkiewicz:
Good catches, fixed in this commit. (I'll squash before merge.)

alexjurkiewicz:
Reckon this is ready for merge now. You get notified about every rune on branch entry if you haven't picked it up yet.

PR #113, opened 2015-09-18
Ashenzari "curse hands"

alexjurkiewicz:
This is my second attempt at improving Ash for UC users (see #65).

Cursing your hands gives a UC boost but prevents (w)ielding items. Uncursing your hands removes the boost and restrictions. Cursing and uncursing uses a curse weapon or remove curse scroll respectively.

The code is not great quality right now (no idea what happens with felids/similar, the boost amount needs consideration, etc etc) but I want to see if it would be accepted before working further.

### DIFF
--- a/crawl-ref/source/godpassive.cc
+++ b/crawl-ref/source/godpassive.cc
@@ -213,6 +213,9 @@ void ash_check_bondage(bool msg)
         else
             s = ET_JEWELS;
 
+        // Ashenzari bound hands
+        if (i == EQ_WEAPON && player_hands_cursed())
+            cursed[ET_WEAPON] = 3;
         // transformed away slots are still considered to be possibly bound
         if (you_can_wear(i))
         {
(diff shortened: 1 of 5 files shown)


alexjurkiewicz:
Updated code to support a more natural interface by directly using scrolls, rather than additional abilities.

alexjurkiewicz:
Closing, no support for the idea in general:

00:28 <+Medar> I don't really like it either, fwiw
00:28 <+MarvinPA> s/nemelex/ash, but something like that yeah, cursing a thing that isn't even an item
00:28 <+Medar> non-trivial, non-obvious addition that I don't see the reason for
00:29 <+Medar> ash is already one of the most versatile gods, and can certainly work with uc already
00:29 <+Medar> and if not, does it have to?
00:29 <+reaverb> Ash not being able to give curse benefits to UC users doesn't strike me as significantly different as Troq
                 and Oka not being able to gift weapons to UC users.
00:29 <+MarvinPA> part of ash's thing is that you have to find items that you don't mind being stuck to you for a while,
                  that's never an issue with uc
00:29 <+MarvinPA> and yeah what medar said too
00:29 <+reaverb> well /useful/ weapons.

PR #117, opened 2015-09-24
Increase the stat cap to 144

alexjurkiewicz:
See https://crawl.develz.org/tavern/viewtopic.php?f=8&t=15144


neilmoore:
Your base stats are stored (both in-memory and marshalled) in int8_ts, so you can't raise the cap above 127 without changing at least some stuff in player.h and tags.cc (and it looks like tileweb.cc as well?)

alexjurkiewicz:
Sorry, I think 120 will be totally fine then. Want me to update this patch?

neilmoore:
Did some testing, and it looks like setting base stats to 99 then wearing a ring of int sets the base back to 72.  It looks like the functions `_normalize_stat`, `jiyva_stat_action`, and `_player_armour_adjusted_dodge_bonus` all have relevant instances of hard-coded "72".

PR #118, opened 2015-09-26
Allow selection of visible layers with '|'.

stahara:
* Toggle (m)onsters, (p)layer, (i)tems, and (c)louds visibility.
* Double-press '|' to invert all (behaves like current '|' command).

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -693,6 +693,14 @@ always_show_exclusions = true
         This option has no effect in tiles builds, where exclusions are
         always shown both in the viewport and on the level map.
 
+layers_toggle = monsters, player, items, clouds
+        The layers to hide when using the '|' command.
+
+layers_menu = false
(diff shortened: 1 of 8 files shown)


wheals:
I'm not convinced that there need to be options connected to this, when || to get the old behaviour is so easy.

stahara:
I didn't think it would get accepted without it, so it's there.  I can remove it if you like.  It would be much cleaner that way.

MarvinPA:
'|' is probably used rarely enough as-is that messing with its functionality is no problem, I agree t hat this would be better with no option.

evktalo:
I actually do use the command, and I welcome the new functionality. (Also agree about not needing the option.)

stahara:
I also added PR #145 as a squashed and rebased version of this series of commits, since a lot of what was added in the way of configuration (though not functionality) has been again removed.  That PR only modifies 3 files and is much simpler than this series.  You can use whichever you prefer. >8)

PR #120, opened 2015-09-29
Untangle the mutation-conflict code.  I think.

stahara:


crawl-ref/source/mutation.cc
```diff
@@ -97,62 +97,54 @@ static const body_facet_def _body_facets[] =
...
+enum mutation_resolution
+{
+    MUTR_REPLACE, // Remove the existing mutation and replace with the new one.
+    MUTR_COEXIST, // Allow both mutations to exist.
+    MUTR_MODIFY,  // Delete a level of the existing mutation.
+    MUTR_IGNORE,  // Discard the new mutation, leaving existing mutations unchanged.
+};
+
+struct mut_conflict {
+    mutation_type       muts[2];
```
PleasingFungus: Maybe make this a pair`<mutation_type, mutation_type`>?

crawl-ref/source/mutation.cc
```diff
@@ -97,62 +97,54 @@ static const body_facet_def _body_facets[] =
     { EQ_BOOTS, MUT_TALONS, 3 }
 };
 
+enum mutation_resolution
+{
+    MUTR_REPLACE, // Remove the existing mutation and replace with the new one.
+    MUTR_COEXIST, // Allow both mutations to exist.
+    MUTR_MODIFY,  // Delete a level of the existing mutation.
+    MUTR_IGNORE,  // Discard the new mutation, leaving existing mutations unchanged.
+};
```
PleasingFungus: This approach doesn't make sense. override isn't an arbitrary boolean toggle between two modes, it's a signal to attempt to add the mutation even when it'd be normally impossible. Splitting the 0/1/-1 values into two mutation_resolutions gives you the flexibility to say things like "replace existing mutations, UNLESS override is set, in which case you should just give up (MUTR_IGNORE) instead!" Not only do you not need that kind of flexibility for existing interactions, it doesn't make any sense.

I'd recommend just replacing the -1/0/1 values with an enum.

crawl-ref/source/mutation.cc
```diff
@@ -97,62 +97,54 @@ static const body_facet_def _body_facets[] =
...
  */
-static const int conflict[][3] =
-{
-    { MUT_REGENERATION,        MUT_SLOW_METABOLISM,        0},
-    { MUT_REGENERATION,        MUT_SLOW_HEALING,           0},
-    { MUT_ACUTE_VISION,        MUT_BLURRY_VISION,          0},
-    { MUT_FAST,                MUT_SLOW,                   0},
+static const mut_conflict mut_conflicts[] = {            // resolution:
+    // mutation_a               mutation_b                  if_forced     if_unforced
+    { {MUT_REGENERATION,        MUT_SLOW_METABOLISM},       MUTR_REPLACE, MUTR_COEXIST },
```
PleasingFungus: Why is this one special cased?

crawl-ref/source/mutation.cc
```diff
@@ -97,62 +97,54 @@ static const body_facet_def _body_facets[] =
...
- *      don't give the new mutation, and consider it a success. 
- *
- * It makes sense to have two entries for the same pair, one with value 0
- * and one with 1: that would replace all levels of the old mutation if
- * forced, or a single level if not forced. However, the 0 entry must
- * precede the 1 entry; so if you re-order this list, keep all the 0s
- * before all the 1s.
+ * Conflicting mutation pairs and how to resolve them based on the nature of the incoming
+ * mutation (forced or unforced). Entries are symmetric (so if A conflicts with B, B
+ * conflicts with A in the same way).
```
PleasingFungus: Worth commenting that the order is important, since that's the order in which conflicts are evaluated. (E.g., trying to add MUT_SLOW_HEALING while you already have both MUT_REGENERATION and MUT_NO_DEVICE_HEALING will have different effects depending on what order those two are listed in.)

crawl-ref/source/mutation.cc
```diff
@@ -906,13 +898,13 @@ static mutation_type _get_random_mutation(mutation_type mutclass)
  */
 int mut_check_conflict(mutation_type mut, bool innate_only)
 {
-    for (const int (&confl)[3] : conflict)
+    for (const mut_conflict conflict : mut_conflicts)
```
PleasingFungus: This should be "const mut_conflict &conflict"; otherwise you're making a copy of the mut_conflict, which you don't need to do.

crawl-ref/source/mutation.cc
```diff
@@ -935,53 +927,37 @@ static int _handle_conflicting_mutations(mutation_type mutation,
                                          const string &reason,
                                          bool temp = false)
 {
-    // If we have one of the pair, delete all levels of the other,
-    // and continue processing.
-    for (const int (&confl)[3] : conflict)
+    for (const mut_conflict conflict : mut_conflicts)
```
PleasingFungus: Likewise, this should be ...&conflict.

crawl-ref/source/mutation.cc
```diff
@@ -935,53 +927,37 @@ static int _handle_conflicting_mutations(mutation_type mutation,
...
         {
-            const mutation_type a = (mutation_type)confl[j];
-            const mutation_type b = (mutation_type)confl[1-j];
+            if (you.innate_mutation[conf_mut] >= you.mutation[conf_mut])
+                return -1;
 
-            if (mutation == a && you.mutation[b] > 0)
+            switch (override ? conflict.forced
+                    : temp   ? (conflict.unforced == MUTR_IGNORE ? MUTR_IGNORE : MUTR_COEXIST)
+                    :          conflict.unforced)
```
PleasingFungus: For future, this sort of code should be broken out, probably into its own function. Something like:

`static mutation_resolution _get_mutation_response(const mut_conflict &conflict,
                                                  bool force, bool temp)
{
    if (force)
        return conflict.forced;
    if (!temp)
        return conflict.unforced;
    if (conflict.unforced == MUTR_IGNORE)
        return MUTR_IGNORE;
    return MUTR_COEXIST;
}`

PleasingFungus:
Added various inline comments; not sure if you'll be notified otherwise, so here's a top-level comment.

stahara:
Thanks for the comments.  I'll look over them when I get off work.  I'll say now, however, that I'm not entirely sure about how to answer a few of these, since this was merely in response to a complaint that the existing code was very difficult to read.  This is simply a rewrite of what was already there.  I suppose one of the benefits of doing so is to expose some of those odd quirks that were previously difficult to recognize, but I don't think I'm the one to decide what should be done about them.

PleasingFungus:
Worth pulling some variant of this in at some point, but it looks like this one is stalled.

PR #123, opened 2015-10-01
Scale protection brand with weapon/shield skill.

alexjurkiewicz:
Protection brand provides triple the old bonus at skill 27, scaling
linearly to 0 at skill 0.

ref https://crawl.develz.org/tavern/viewtopic.php?f=8&t=17862&view=unread&sid=505ee590f9424aca856b1d0563bcec02#unread

### DIFF
--- a/crawl-ref/source/defines.h
+++ b/crawl-ref/source/defines.h
@@ -162,6 +162,8 @@ const int BASELINE_DELAY  = 10;
 const int GOURMAND_MAX            = 200 * BASELINE_DELAY;
 const int GOURMAND_NUTRITION_BASE = 10  * BASELINE_DELAY;
 
+const int UNPROTECTION_MAX = 50 * BASELINE_DELAY;
+
 const int CHUNK_BASE_NUTRITION    = 1000;
 
 const int ICEMAIL_MAX  = 10;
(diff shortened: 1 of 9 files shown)


gammafunk:
This definitely won't work as-is, since it's going to encourage a lot of weapon swap. If you find any weapon of the same type as the weapon skill you train, you'll want to carry that around, and then play the game of keeping it wielded for any turn you're forced to be in los of a ranged attacker, switching at the last minute. This is already a problem for the protection brand, but this commit makes this worse overall, since now it gives more ac for skill > 9 or so.

Furthermore it's a huge change to the strength of the brand yet you don't adjust the brand frequency.

alexjurkiewicz:
Good point. How about scaling up the protection bonus over x turns, like Gourmand?

alexjurkiewicz:
Added a `-AC` debuff which halves your armour for 50 turns after unequipping protection. Mark protection brand as dangerous (like distortion).

Note: doesn't work properly with shield protection brand yet. I'll fix that if people like the idea. I'm not sure if I do right now -- it might be too complex.

MarvinPA:
Closing this, since as gammafunk points out it doesn't really fix the issue, and working around that with unwield penalties seems over-complicated. Protection/evasion egos fundamentally just aren't super interesting, I think.

PR #125, opened 2015-10-01
Uniques always corpsify

alexjurkiewicz:
Or at least they don't coinflip() any more.

Also corpses from uniques always create hides when possible -- affects Purgy, Snorg and XTAHUA.


alexjurkiewicz:
Whoops, thanks @stahara

MarvinPA:
As discussed in IRC, this doesn't seem like it's worth a special case for.

PR #130, opened 2015-10-08
Drac breath slot

alexjurkiewicz:
Tested with dracs and non-dracs picking up ring of tele and then levelling up to 7

https://crawl.develz.org/tavern/viewtopic.php?f=8&p=245372#p245372

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -3443,6 +3443,8 @@ static int _find_ability_slot(const ability_def &abil)
     // Skip over a-e (invocations), a-g for Elyvilon
     int first_slot = letter_to_index('f');
 
+    ASSERT(first_slot < 52);
+
     switch (abil.ability)
     {
     case ABIL_ZIN_CURE_ALL_MUTATIONS:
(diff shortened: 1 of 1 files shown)


crawl-ref/source/ability.cc
```diff
@@ -3442,8 +3442,15 @@ static int _find_ability_slot(const ability_def &abil)
 
     // Skip over a-e (invocations), a-g for Elyvilon
```
wheals: Comment now outdated?

crawl-ref/source/ability.cc
```diff
@@ -3440,10 +3440,17 @@ static int _find_ability_slot(const ability_def &abil)
...
 
-    // Skip over a-e (invocations), a-g for Elyvilon
+    // a-e is for invocations
     int first_slot = letter_to_index('f');
-    if (you_worship(GOD_ELYVILON))
-        first_slot = letter_to_index('h');
+
+    // Reserve the first ability slot (f) for Draconian breath
+    if (species_is_draconian(you.species)
+            && you.experience_level < 7)
```
wheals: Maybe this should just check you.species == SP_BASE_DRACONIAN, so it wouldn't have to be modified if the XL when you grow up were changed?

PR #131, opened 2015-10-08
Update another xl_rank_name

drke2:
To avoid reusing "experienced" for a different level bracket than
it was previously.


wheals:
I'm not super happy with this name but I agree it's best not to reuse old adjectives in new capacities. As discussed on IRC we might replace these with numbers later which would of course solve any issues with the words.

PR #132, opened 2015-10-09
Fix for issues related to no-device-heal + rot.

smitnich:
Allow players to heal rot even with no device heal mutation. All of the other statuses are fixed by potions of curing even with no device heal, so fix this as well. Also let players use potions of heal wounds to heal rot, since they heal rot in the same way as curing.


wheals:
I believe wands/potions of heal wounds not healing rot for no device heal mutation is intentional, so as to not make them possibly useful in the middle of the game. Instead they should be useless all game, so the player doesn't have to keep them around for a possible use half a game later.

smitnich:
What about just letting curing healing rot then; all of the other effects of curing work with no device heal? Non Vine-stalkers can end up with no device heal 3, and be unable to heal rot without divine intervention; Vine-stalkers seem to have resist rot exactly to prevent this situation.

wheals:
That sounds fine to me.

PR #133, opened 2015-10-09
Tweak an arrival vault

drke2:
ldierk_water_garden placed toadstools, bushes, plants, and trees in four
corners.

A player unfamiliar with the nuances of stationary plant monsters may
choose to attack the plant or bush corner, and if this character has no
weapon or using untrained UC this will take hundreds of turns.

Change it to spawn trees in 3 corners with one toadstool exit, which the
dungeon builder flip/rotate will randomize.


drke2:
Forgot to remove SHUFFLE: 1t.

amalloy:
:+1:

PR #135, opened 2015-10-09
Add rc option to prompt before using escape hatches

amalloy:
I've had some close calls accidentally using a hatch that I thought was a stair.


alexjurkiewicz:
Enable by default imo

PR #142, opened 2015-10-11
Unrand attempt: the -6 whip "Skinflayer"

stahara:
An unrand whip with -6 enchantment, but which causes an increasing
amount of bonus damage with each successful strike.  The bonus expires
after about ten turns without another hit from the weapon.  Works only
on monsters with flesh.


gammafunk:
Hitting monsters is something you'll always want to do with a weapon, so it's basically the least interesting way to gain enchantment. This item is currently "become better than an average unbranded demon whip after a few hits, except for some enemies where it's always worse". What I'd suggest is to find a better mechanic to attach the "enchantment increases in combat" idea to. Possibilities include killing monsters (probably you could just start the enchantment higher, add more per kill), or having low HP; see the Sword of Power for the reverse of this. Attaching the enchantment to low HP would create a weird interaction with Death's Door. That may not have to be addressed if you go this route, but there should be a reasonable cap on the enchant the whip can get, regardless.

I'm closing this for now, but feel free to make a follow-up commit or pull request if you can rework the idea.

PR #143, opened 2015-10-12
Re-add Gozag duplication.

alexjurkiewicz:
This is now an ability that costs increasing gold with each use, and
duplicates a single item rather than a stack.

Based on current gold generation rates, typical three-runer can use this
ability a couple of times in a game:

    Runes | avg(goldfound) | avg(goldspent)
    3     | 33k            | 10k
    15    | 70k            | 13k


alexjurkiewicz:
See also #144

gammafunk:
This doesn't address one of the main concerns we had with Gozag duplication, which is that it rarely presents much of a decision. Getting new items from gold is also covered in a more interesting way by purchasing shops. Closing this PR.

PR #144, opened 2015-10-12
Gozag artefactise

alexjurkiewicz:
See also #143 -- this is an alternate gozag "capstone".

Add Gozag artefactisation. This is an ability that costs increasing gold with each use, and
works very similarly to wizmode &+.

Based on current gold generation rates, typical three-runer can use this
ability a couple of times in a game:

Runes | avg(goldfound) | avg(goldspent)
3     | 33k            | 10k
15    | 70k            | 13k

TODO:
* doesn't add plusses for non-weapons
* should we allow creating mega-artefacts by stacking new properties on
  top of an existing artefacts?


MarvinPA:
As discussed in IRC, this doesn't really fit Gozag's theme, and Gozag also already provides artefacts in a more reliable way through shops.

alexjurkiewicz:
Yep, agree. I think a better approach is probably just to make shops more reliable.

PR #145, opened 2015-10-14
Allow selection of visible layers with '|'.

stahara:
Toggle (m)onsters, (p)layer, (i)tems, and (c)louds visibility.

If you use only the '|' key, the functionality is exactly the same as
the original "show terrain".  However, you may now toggle certain layers
back on and that configuration will be remembered for the next time you
press '|'.


stahara:
This is a rebased and squashed version of #118.  That series of commits goes in a giant loop and much of what I added was removed in the final commit.  This one is the minimal diff, and is rebased on the current master for convenience.  Take whichever one you prefer.

PR #146, opened 2015-10-14
Add .clang-format.

alexjurkiewicz:
This doesn't match exactly with Crawl's current code-style, but it's a
good start.

### DIFF
--- a/crawl-ref/source/.clang-format
+++ b/crawl-ref/source/.clang-format
@@ -1,0 +1,8 @@
+---
+Language: Cpp
+
+BreakBeforeBinaryOperators: NonAssignment
+BreakBeforeBraces: Allman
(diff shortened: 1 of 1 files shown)


wheals:
We agreed on IRC a while back that this can't go much further until https://llvm.org/bugs/show_bug.cgi?id=19082 gets fixed. As the above commit shows, we did get something out of it.

alexjurkiewicz:
Another missing option: http://bugs.llvm.org/show_bug.cgi?id=23569

aidanholm:
Chained ternaries should be fixed with the merging of https://reviews.llvm.org/D50078, in the upcoming clang 11 release. That should be released around Sept 2020, so hopefully we'll be able to put clang-format to use then.

PR #149, opened 2015-10-15
Master

sesangsokuro:
ability name string to define keyword for translate
ability names in ability.cc to translate.h


neilmoore:
As noted, this is far from complete (the defines aren't even used); and I think it is better to use something like gettext rather than trying to roll our version thereof (and having a #define for every name in the code).

Of course, there are much bigger problems with translation when grammar is involved, but allowing names to be translated like this would be a good start.

PR #150, opened 2015-10-15
Add tracking for defensive actions including:

aolowin:
Armour types
  Dodged, Phased, or Deflected attacks
  Blocked and Reflected attacks
Add tracking of auxillary attacks.
Rework count_action() to have a standard way to store auxillary information instead
of doing manual bit shifting.

### DIFF
--- a/crawl-ref/source/actor.cc
+++ b/crawl-ref/source/actor.cc
@@ -370,11 +370,11 @@ int actor::apply_ac(int damage, int max_damage, ac_type ac_rule,
     saved = max(saved, min(gdr * max_damage / 100, ac / 2));
     if (for_real && (damage > 0) && (saved >= damage) && is_player())
     {
-        const item_def *body_armour = you.slot_item(EQ_BODY_ARMOUR);
+        const item_def *body_armour = slot_item(EQ_BODY_ARMOUR);
         if (body_armour)
             count_action(CACT_ARMOUR, body_armour->sub_type);
         else
-            count_action(CACT_ARMOUR, 0, 0); // unarmoured alttype
+            count_action(CACT_ARMOUR, 0, 0); // unarmoured auxtype
     }
 
     return max(damage - saved, 0);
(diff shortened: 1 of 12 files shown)


alexjurkiewicz:
Nice!

crawl-ref/source/actor.cc
```diff
@@ -368,6 +368,15 @@ int actor::apply_ac(int damage, int max_damage, ac_type ac_rule,
     }
 
     saved = max(saved, min(gdr * max_damage / 100, ac / 2));
+    if (for_real && (damage > 0) && (saved >= damage) && is_player())
+    {
+        const item_def *body_armour = you.slot_item(EQ_BODY_ARMOUR);
```
wheals: This is in an actor method, so just `slot_item(EQ_BODY_ARMOUR)` would work

crawl-ref/source/player.cc
```diff
@@ -5676,6 +5676,11 @@ void player::shield_block_succeeded(actor *foe)
 
     shield_blocks++;
     practise(EX_SHIELD_BLOCK);
+    const item_def *shield = you.slot_item(EQ_SHIELD);
```
wheals: Ditto

neilmoore:
This doesn't handle save compat of old action_counts.  That causes behaviour ranging from wrong labels to crashes when dumping a game that was transferred into this version, because the format of the 32-bit subtype number changed for unarmed, melee, and ranged attacks.

Also, if the subtype is negative, that overwrite the auxtype with -2 when combining the two.

Reverting for now so that servers don't have a hard-to-correct mix of crashy and non-crashy games.

neilmoore:
There doesn't seem to be a way to re-open a merged (as opposed to closed) pull request.  aolowin, could you resubmit this?  I would like to see it go in, it just needs a few fixes and there wasn't enough time to make them before the nightly server rebuilds.

PR #151, opened 2015-10-16
Make ring of shadows stack with Dith and Yred

pblur:
Currently, one of the more interesting stabby unrands in the game is incompatible with the stealth god (and with the undead miniony death god; but he's mostly a side point). This would make their effects stack manipulatively, so ring of shadows + dith max piety gives you 3* the stealth instead of 2*. That's good, but probably not too broken considering:
A) You get stabs on everything you want to already. Stealth apts are quite high, and your payoff is doubled, plus you have shadow step (ignoring yred).
B) Your primary threat is going to be noise, not your own unstealthiness.
C) You give up a ring slot for *invis, rN+ and SInv; not the worst ring by a long shot, but not something you'd generally be wearing late-game either.

I think this change would make the choice more interesting on some characters.

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -3239,11 +3239,10 @@ int check_stealth()
             umbra_mul = you.piety + MAX_PIETY;
             umbra_div = MAX_PIETY;
         }
-        if (player_equip_unrand(UNRAND_SHADOWS)
-            && 2 * umbra_mul < 3 * umbra_div)
+        if (player_equip_unrand(UNRAND_SHADOWS))
         {
-            umbra_mul = 3;
-            umbra_div = 2;
+            umbra_mul *= 3;
+            umbra_div *= 2;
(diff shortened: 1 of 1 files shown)


MarvinPA:
I don't think the ring of Shadows is by any means "incompatible" with Dith - it's still a (slightly) better ring of invisibility which is great if you're stabbing things. Current behaviour seems fine to me.

pblur:
Incompatible isn't quite the right word I guess. I mean that the flavor of the unrand does nothing, not that you'd never wear the ring. But if that would be too potent (and it would be strong)... sure.

gammafunk:
I don't think the issue is whether it's strong, it's more that it's a needless special case. The ring is still usable for those two gods, and for the other gods it supplies something special. Making the ring more complicated in order to make it more equally beneficial across gods is not a good tradeoff. Closing this PR.

PR #153, opened 2015-10-17
Make growth deterministic with no nearby enemies

smitnich:
Currently players can get any order of plant priority when using growth
by repeatedly activating and cancelling the ability for no cost, so
make growth always use the same order when no monsters are nearby.

### DIFF
--- a/crawl-ref/source/godwrath.cc
+++ b/crawl-ref/source/godwrath.cc
@@ -1495,7 +1495,7 @@ static bool _fedhas_summon_plants()
     {
         // Randomize the order so that plants don't always appear
         // in the same places when no enemies are nearby
-        shuffle_array(radius_points);
+        shuffle_array(radius_points[0]);
         unsigned target_count = random_range(2, 8);
         if (target_count < radius_points[0].size())
             prioritise_adjacent(you.pos(), radius_points[0]);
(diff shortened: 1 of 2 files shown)


neilmoore:
In the wrath code, the one-argument shuffle_array(radius_points[0]); should work just fine.  No harm either way (it expands to exactly the same call you are already making), just less typing.

crawl-ref/source/godwrath.cc
```diff
@@ -1493,6 +1493,9 @@ static bool _fedhas_summon_plants()
     // (assuming the player isn't already surrounded).
     else if (!radius_points[0].empty())
     {
+        // Randomize the order so that plants don't always appear
+        // in the same places when no enemies are nearby
+        shuffle_array(radius_points);
```
wheals: Is this correct, or should it be `radius_points[0]`?

neilmoore: Should still be radius_points[0], yes.

reaverb-Crawl:
I've implemented a different fix as of 0.17-a0-2209-g7c46689, not letting the player abort using the Fedhas growth ability.

A couple awkward problems with this solution that pushed me away from it:

A) This makes different directions on the game gird mechanically different. An example where this could be relevant is if you find a sleeping unique, retreat, and then want to create a plant out of their LOS to completely block them off.

B) The Fedhas growth targeter reveals the locations of adjacent invisible monsters. This is arguably a bug, but using the abort solution also solves this.

PR #157, opened 2015-10-19
 lillte simplify ability.cc 

anch0vy:
Add Initialize "result" and gather code of failure=0 ability


neilmoore:
It's in, thanks!  I used brace initialization to [simplify further](https://github.com/crawl/crawl/commit/c3f89d198632).

PR #158, opened 2015-10-19
 Clean up duplicate code at ability.cc 

anch0vy:
Remove duplicate your_talents call. And change condition check order.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -1131,23 +1131,18 @@ bool activate_ability()
         return false;
     }
 
-    vector<talent> talents = your_talents(false);
-    if (talents.empty())
+    bool confused = you.confused();
+    vector<talent> talents = your_talents(confused);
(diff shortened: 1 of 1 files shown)


neilmoore:
The double-call is intentional, so that if you are confused but don't have any abilities, it will tell you that you have no abilities rather than that you are too confused.  I would keep it as is, but I'm leaving this open so others can comment.

anch0vy:
I missed that case. sorry.
But I think It is okay to show "you are too confused" message before no ability message.
Because confused state is more important than no ability state. And beginner can know that ability can not use in confused even at no ability state.

neilmoore:
A reasonable argument; also, the exceptions (abilities that can be used while confused) are all god abilities, and you can never have a god and have *no* abilities (since Renounce Religion is always there, even if unavailable).

neilmoore:
It's in, thanks!

PR #159, opened 2015-10-19
Halve attribute values.

alexjurkiewicz:
Not ready for merge, ànd maybe not for 0.17, but thought I'd open this up for comments. The game works but significantly more balance work is required.
- If previously something used strength, now it uses strength \* 2, and
  similar for intelligence & dexterity.
- Reduces species starting stats (rounding down, for odd numbers)
  and job starting stats (rounding inconsistently to give +6 stats
  overall).
- Change +/-2 str/int/dex mutations to +/-1
- Reduce stat loss amounts from most sources (minimum of one, stat
  recovery time is unchanged)
- Reduce minimum stat to 2
- Slightly reduce -stat attributes on artifacts.

TODO:
- Stat drain / recovery numbers

### DIFF
--- a/crawl-ref/source/artefact.cc
+++ b/crawl-ref/source/artefact.cc
@@ -667,7 +667,7 @@ struct artefact_prop_data
 static int _gen_good_stat_artp() { return 1 + random2(3); }
 
 /// Generate 'bad' values for stat artps (e.g. ARTP_STRENGTH)
-static int _gen_bad_stat_artp() { return -2 - random2(4); }
+static int _gen_bad_stat_artp() { return -1 - random2(3); }
 
 /// Generate 'good' values for resist-ish artps (e.g. ARTP_FIRE)
 static int _gen_good_res_artp() { return 1; }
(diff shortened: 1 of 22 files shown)


neilmoore:
I am concerned about this being generally a character buff, since characters start at approximately the same effective numbers as before (subject to some rounding), but get effectively twice the increase on level-up.

On the other hand, decreasing starting stats to make up for that has its own problems: namely, that would put SpWz at 1 str and TrBe at 1 int, which are perilously close to 0.

alexjurkiewicz:
Well, there's a minimum number for starting stats. I reduced it from 3 to 2
in this patch. I also changed user level ups to every 4 levels, and
generally added 1 to species stat-up rates. All these numbers are
provisional, but I'm not intending this to be a buff.

Also, I reduced stat drain amounts, though that work probably needs extra
attention.

That said, it's probably impossible to make a change this general without
affecting balance in a multitude of situations...

alexjurkiewicz:
Reopening. If there is interest in this change, I can update it to suit current branch.

To expand on why I think this is a good change: it makes level up decision of s/i/d more meaningful. Because you get fewer attribute points, it's harder to get the exact distribution you want, which means choices should be tougher for the player.

alexjurkiewicz:
I still like this idea but think it's time to declare maintenance bankruptcy on this PR.

PR #164, opened 2015-10-23
Rework Gell's Gravitas for better synergy with Inner Flame

stahara:
- All monsters in range now collide with the target.
- Monsters are re-distributed as close to the target as possible first,
  then as close to their starting point as possible second.


wheals:
@sgrunt went with a different solution in 501c6074 and its two immediate parents.

PR #168, opened 2015-10-27
Remove amulet of resist mutation

pblur:
Remove amulet of resist mutation from randomly generated amulets. If I understand the codebase properly, this will still permit specifically placed rMut amulet in fleshworks wizlab, and shouldn't interfere with hat of the alchemist's rMut. Both of those sound like cool things to maintain even if ordinary rMut amulets are removed.

I expect we'd eventually need to rebalance malmutators, since those are in a kind of icky spot due to the OP nature of rMut currently.


pblur:
This is not nearly adequate to remove this... My bad.

PR #170, opened 2015-10-28
Add sturdy frame mutation (infinitevox).

alexjurkiewicz:
Reduce player base ER by 15/30/45%. Doesn't increase EV by that much
(+3EV in CPA), but allows significantly easier spellcasting.


alexjurkiewicz:
Other alternative I was thinking: -1/2/3 ER

alexjurkiewicz:
Updated to be -2/4/6 (minimum 0) ER

PR #174, opened 2015-11-14
Don't ask ghouls for potions of cMut (celerity)

amalloy:
Troves shouldn't ask for totally race-useless items, and that includes cMut for ghouls.


alexjurkiewicz:
What about other undead species?

amalloy:
Already handled by the code that was there before, @alexjurkiewicz. Mu don't get any potion tolls at all, and Vp can mutate so cMut is relevant.

alexjurkiewicz:
Undead mechanics strike again!

PR #175, opened 2015-11-16
Identify the Necronomicon when Kiku gives it to you

amalloy:


crawl-ref/source/godabil.cc
```diff
@@ -2026,6 +2026,7 @@ bool kiku_gift_necronomicon()
     {
         return false;
     }
+    set_ident_type(thing_created, true)
```
wheals: Should be `mitm[thing_created]`

amalloy:
Oops. I rolled back my commit to test the old version, and then accidentally committed a broken version of it. Updated it, @wheals.

PR #177, opened 2015-11-17
Add escape tunnels to evilmike_wizard_prison.

alexjurkiewicz:
Prevents the (very rare, but happened at least once) case where the
player could be inadvertently blinked into a silenced cell with no way
to escape.


alexjurkiewicz:
Could be better solved by making blinks from non-player-controlled sources (eg disto attacks) not blink the player to no_tele_into squares, if the player is currently on a no_tele_into square.

PR #181, opened 2015-11-23
Reduce the AC of rolling enemies.

alexjurkiewicz:
Should encourage players to make bad decisions around the
easy-to-mispredict motion of rolling boulder beetles.

### DIFF
--- a/crawl-ref/source/dat/descript/monsters.txt
+++ b/crawl-ref/source/dat/descript/monsters.txt
@@ -687,7 +687,7 @@ dragon's fearsome breath was lost in the process.
 boulder beetle
 
 A huge grey beetle with an almost impenetrable rocky carapace, capable of
-rolling at its foes at high speed.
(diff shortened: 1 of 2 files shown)


MarvinPA:
Boulder beetle movement isn't hard to predict, I don't really see how this is a good idea.

alexjurkiewicz:
If you're rolling in a cardinal direction it's pretty easy, but when the
movement is x tiles down for y tiles across the ratio isn't clear for
several turns. Also, the player is encouraged to make boulder beetles roll!

MarvinPA:
Well, I'm assuming you knew where you were when it started rolling so you can probably figure out the path regardless? And if it really is hard to figure out then it still seems like a bad idea to add a gameplay element relying on the fact that the existing mechanic is unclear. It also just seems like a new unclear mechanic itself. Also, ranged attacks exist!

alexjurkiewicz:
I think dealing with crawl's path quantization is very difficult, but agree with your other points so closing this.

PR #183, opened 2015-11-25
Remove blink frog healing from distortion attacks.

alexjurkiewicz:


wheals:
I'd be inclined to oppose: it's a cute flavour thing with fairly small consequences (disto versus them is still better than no brand on average). Blink frogs' desc could use a mention of the healing, though.

alexjurkiewicz:
Small consequence yes, but the consequence is prolonging a fight against
slippery enemies while using a brand that makes enemies even slipperyer!

wheals:
I'm not convinced enough to merge.

PR #184, opened 2015-11-25
Tweak panlord brand distribution.

alexjurkiewicz:
Make unusual (for the player) brands more common, and weaker brands less
common.

Not tied to my lists of "strong" and "weak" brands at all, if people want to dispute my categorisation :P I just like the idea of making certain brands more and less common.

### DIFF
--- a/crawl-ref/source/ghost.cc
+++ b/crawl-ref/source/ghost.cc
@@ -159,18 +159,21 @@ void ghost_demon::reset()
  */
 static brand_type _random_special_pan_lord_brand()
 {
-    return random_choose(SPWPN_FLAMING,
-                         SPWPN_FREEZING,
-                         SPWPN_ELECTROCUTION,
-                         SPWPN_VENOM,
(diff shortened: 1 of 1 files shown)


crawl-ref/source/ghost.cc
```diff
@@ -159,18 +159,21 @@ void ghost_demon::reset()
...
-                         SPWPN_DISTORTION,
-                         SPWPN_CHAOS);
+    return random_choose_weighted(10, SPWPN_FLAMING,
+                                  10, SPWPN_FREEZING,
+                                  10, SPWPN_ELECTROCUTION,
+                                  10, SPWPN_VENOM,
+                                  10, SPWPN_SPEED,
+                                  // Lower chance
+                                  5, SPWPN_DRAINING,
+                                  5, SPWPN_VORPAL,
```
wheals: I'm thinking maybe not have vorpal at all, since as far as the player can tell it's ~indistinguishable from having no brand.

PR #185, opened 2015-11-25
Simplify distortion_affects_defender.

alexjurkiewicz:
Split the proc chance out so it can be tweaked individually.


wheals:
You forgot to add `break`s (this is causing the crash).

PR #186, opened 2015-11-25
Tweak several branch end vaults

jmbjr:
Modifications based on observations during t0.17

Snake: Deadliest snake rune vault in t.
Reduced chance of guardian serpents and mana vipers.
Also slightly reduced chance of Vashnia appearing.

Spider: Current versions are pretty, but feel lacking.
Added Entropy Weaver and Spark Wasp as potential "Bosses"
Added Entropy Weaver as a Torpor Snail alternative.

Swamp:
Tweaked hydra and bog body weights
Replaced Spriggan Druid with Spriggan Berserker.
(a berserker feels like a more interesting choice than druid)

Vaults: Too many corridors to hide in
Widened hallways. Left loot and monsters the same.


crawl-ref/source/dat/des/branches/snake.des
```diff
@@ -1398,12 +1398,13 @@ MONS:   naga / nothing
...
         nothing w:20
-MONS:   guardian serpent w:20 / salamander mystic w:5 / naga ritualist w:5 / \
+MONS:   guardian serpent / salamander mystic w:5 / naga ritualist w:5 / \
         nothing
 MONS:   greater naga w:30 / naga warrior w:5 / naga sharpshooter w:5
-MONS:   anaconda / black mamba / shock serpent w:5 / mana viper w:15
+MONS:   anaconda / black mamba / shock serpent w:5 / mana viper
 
-KMONS:  Z = Vashnia / greater naga, greater naga
+KMONS:  Z = Vashnia / greater naga
```
wheals: This means if Vashnia was placed already it won't even put a greater naga; is that intentional?

jmbjr: Wheals: no that was not my intention. That was me spacing out when I made that edit. I've reverted that change and force-pushed an update

Thanks for the check!

PR #188, opened 2015-11-27
A trio of Pakellas Overflow Altar Vaults

jmbjr:
1)  Death Squad: high evo !apt band takes down filthy low evo !apt brute.
2/3 chance of a wand with 0 charges (occasionally will spawn a kobold with a wand with 1 charge)
3/10 chance of a troll corpse (early troll hide chance)
Only low-level wands will spawn (frost, flame, m. darts)
A dash of blood was sparingly used for thematic effect.
The walls vary in material, somewhat referencing the current blinking P altar.

2) Dig!
a wand of digging with a single charge with a transparent rock wall that hides a P altar.
I wasn't sure if this was against the rules for an overflow vault, but seems like the player has a very clear choice on whether to use the wand to get to the altar, or take it (and recharge it the hard way later).

3) Pakellas's Ersatz Instruments!
90% basic wands with 15 charges
10% any wand with normal charges
greed:30 and only 3 wands

the basic wands cost between 150-250. 
the good wands like hw and haste are >1k
cold/fire, etc are around 600

### DIFF
--- a/crawl-ref/source/dat/des/altar/overflow.des
+++ b/crawl-ref/source/dat/des/altar/overflow.des
@@ -1565,6 +1565,49 @@ MAP
 .............
 ENDMAP
 
+# high evo !apt band takes down filthy low evo !apt brute
+# 2/3  chance of a kobold or corpse with a wand
+# 3/10 chance for a troll corpse (i.e. chance for early troll hide)
+NAME:    johnstein_pakellas_death_squad
+DEPTH:   D:2-
+TAGS:    temple_overflow_pakellas uniq_altar_pakellas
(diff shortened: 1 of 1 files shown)


jmbjr:
forced a push with better rebased commits (one per vault now).

jmbjr:
updated with some adjusted shop weightings

jmbjr:
added greed back. 
see http://i.imgur.com/bh1BnH1.png for a sampling of prices (remember, the shop only generates 3 wands, this is just for reference).

updated first comment and fixed up some comments in the actual vault

PR #195, opened 2015-12-04
Rework malmutation

alexjurkiewicz:
Make lower-tier malmuters give temporary mutations instead of permanent ones.


crawl-ref/source/mutation.cc
```diff
@@ -788,30 +786,55 @@ static bool _accept_mutation(mutation_type mutat, bool ignore_weight = false)
...
 {
     const int tweight = lookup(total_weight, mt, 0);
     ASSERT(tweight);
 
     int cweight = random2(tweight);
+
+    // Perform an initial loop to see if we can increase an existing mutation
+    // Since tweight includes weights of mutations we won't use, this is
+    // probably (definitely) bad, biased math. XXX please fix me.
+    if (prefer_existing)
```
alexjurkiewicz: the logic of this block is totally off. I don't know how to make it smarter though...

alexjurkiewicz:
Note to self: the changes to enum.h probably need a save-compat fix.

MarvinPA:
Closing this, malmutation has had some small-scale changes, and I've also pushed a different implementation of the temporary mutation commit (making it try to apply multiple levels of fewer mutations).

PR #198, opened 2015-12-05
fix a novelty vault to work better with squarelos

jmbjr:
see !bug 10120
https://crawl.develz.org/mantis/view.php?id=10120

The old silent_spectre radius under circlelos covered a diameter of 25
squares.  In squarelos, the "diameter" is only 21.

This meant a few monsters were not properly silenced per the intention
of the vault.

One solution would have been to increase silence_radius to 12, but this
seems like a poor choice just to make a single novelty vault work
better.

Instead, the proposed fix makes a couple cells on the East side only 2
squares (just like some other existing cells).

Additionally, the silent_spectre is moved one space EAST to help cover
the new smaller cells.

Loot and number of mosnters remains unchanged. However, the ancient
lich, 6, is now BEHIND the diggable rock wall at the south and
completely outside of the silenced area.

If this is not desired, it could easily be moved to the 4 just above the
diggable walls.


jmbjr:
old circle-los: http://i.imgur.com/E07ydSo.png
new squarelos: http://i.imgur.com/14nqXM8.png
updated: http://i.imgur.com/YygIJd5.png

in-game webtiles shot: http://i.imgur.com/LzcNgir.png

note on the alich not being silenced, technically the player could still lure it back to the silenced part.

alexjurkiewicz:
Something else being talked about for this vault was getting rid of the hidden treasure room. Might be worth adding it to this PR as well.

jmbjr:
hm. getting rid of the treasure basically makes it a pretty big vault (+400 squares) with pretty much nothing of value in it. just walk around and stare at the silenced wizards. Maybe the loot could be downgraded a bit?  6 | and 4 * is pretty great loot. 

how bout turning the 4 | near the silent spectre into 1 |, 2*, and 1$.  then in the bottom hidden room, 1 |, 3 *, and the rest $.

If instead, the treasure is just removed, I wonder if the vault size should be reduced further so it doesn't take up such a huge space on the map.

PR #200, opened 2015-12-12
Lacertilians

ontoclasm:
A reptilian species with perma-faith and no wrath for switching gods.  So, anti-demigods, in a sense.  Generally weak otherwise.


jmbjr:
ontoclasm posted a patch file in ##crawl-dev at: https://dl.dropboxusercontent.com/u/1000017/crawl/lacertilians.patch 

I create the pull request via git am

I think the intent here is for sharing/review and if things look ok, I can host it as an experimental on cbro.

yrmvgh:
A god loving a race that is prone to swapping gods is a huge flavor glitch in my book. Those should be mutually exclusive.

I'd sooner expect to see a race with (No Wrath, Forlorn) or (Intrinsic Faith, Can't Switch Gods)

isloat:
This is an interesting idea, but I'm not sure how it'd work out in practice. Frequent god-switching and rapid faith gains means that La are encouraged to chew through gods that give permanent rewards before settling on one or two gods to clear an area with. There would be a lot of backtracking to the Temple / overflow altars. 

Setting up an experimental branch somewhere could offer some instructive feedback.

PleasingFungus:
Doesn't seem like this is going anywhere fast; we had an experimental branch a while back, but it's been so long that it was shut down! Please do feel free to reopen this when you decide to work on it again, just don't see any particular reason to leave it open in the meantime.

PR #203, opened 2015-12-21
Increase speed of longswords to make them less like a bad scimitar.

aolowin:
Except for early game there's very little reason to look at a long sword. With the speed improvement additive brands become a bit more interesting on them.


PleasingFungus:
It's fine for many weapon classes to only be relevant early-game.

PR #205, opened 2015-12-31
Acid crabs

ebering:


MarvinPA:
Closing this - the current sources of late corrosion (entropy weavers, caustic shrikes) are pretty decent threats, and their methods of applying corrosion are probably a better approach than clouds that'll also damage other monsters.

PR #211, opened 2016-01-08
Ignore splash damage for reach attacks.

dkorduban:
Currently if you attack a jelly with a polearm from distance, you still get a message "your hands burn". That doesn't make much sense for me. I mean, if I get burned from distance 2 attacking jelly, why can't jelly just spit acid from distance 2 actively? I guess the idea about this monster was that it was only close range.

Also, all other player auxiliary/monster flavour melee attacks require being adjacent except from this one piece.

### DIFF
--- a/crawl-ref/source/melee_attack.cc
+++ b/crawl-ref/source/melee_attack.cc
@@ -2105,6 +2105,10 @@ void melee_attack::attacker_sustain_passive_damage()
     if (attacker->res_acid() >= 3)
         return;
 
+    // Ignore splash damage for reach attacks.
+    if (!adjacent(attacker->position, defender->position))
+        return;
+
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Wrt flavor, one can imagine the impact of the hit spraying acid backwards (at the attacker); it's not an active action on the jelly's part, obviously.

No strong opinion on the mechanics.

crawl-ref/source/melee_attack.cc
```diff
@@ -2105,6 +2105,10 @@ void melee_attack::attacker_sustain_passive_damage()
     if (attacker->res_acid() >= 3)
         return;
 
+    // Ignore splash damage for reach attacks.
+    if (!adjacent(attacker->position, defender->position))
```
wheals: should call `pos()` on them, not access the private field `position`

PR #212, opened 2016-01-09
Origin/sif hunger

aolowin:
Give Sif a passive buff for spell casting.


PleasingFungus:
Is there any context/reasoning for this?

aolowin:
The main reasoning is for an early/mid-game boost. Miscast reduction doesn't happen til high piety and book giving is pretty slow. Amnesia isn't needed until late when you have a larger selection of potential spells to choose from. Late game spellcasting skill and higher int mean hunger isn't a big issue but it definitely can be earlier. Right now Sif feels like you get channeling and then nothing for a long time.

PleasingFungus:
To reduce the feeling of "you get channeling and then nothing for a long time", 5d26797 moves miscast protection to 2*. Thanks for the PR and the reasoning!

PR #213, opened 2016-01-09
The cloak of Preservation.

aolowin:


neilmoore:
A few things (ignoring balance):
1. there is an occasional message saying that the cloak looks more tattered, but that doesn't do anything.  The message seems misleading: it implies that the damage shaving is damaging the item.
2. why does the cloak allow itself to respawn when it crumbles after lifesaving?

aolowin:
1. is flavour so you're not as surprised when it crumbles to dust.
2. is my misreading the effect of UNIQ_NOT_EXISTS. I'll remove that line.

alexjurkiewicz:
The flavour message could perhaps be better implemented as item speech -- see how the majin-bo works.

wheals:
I don't feel very good about introducing something that's basically a non-guaranteed amulet of life saving. I think the unrand could stand out even if it were just damage shaving on a +0 (or maybe even lower enchantment) cloak,

MarvinPA:
I don't think damage shaving is really a good equipment property. It works fine on the Fortitude card as something you can have consumable access to, but on a cloak it seems like it'll just be always plain good, or just plain not very good if it has enough -AC that the damage shaving wouldn't compensate for it, except for against the relatively rare sources of AC-ignoring damage.

PleasingFungus:
Closing per above comments.

PR #217, opened 2016-01-13
Allow training Necromancy when carrying a pain-branded weapon

edsrzf:
Some (old) discussion here, which generally supported the idea: https://crawl.develz.org/tavern/viewtopic.php?t=14453

### DIFF
--- a/crawl-ref/source/itemprop.cc
+++ b/crawl-ref/source/itemprop.cc
@@ -2013,6 +2013,9 @@ bool item_skills(const item_def &item, set<skill_type> &skills)
         skills.insert(SK_EVOCATIONS);
     }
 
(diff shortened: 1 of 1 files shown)


plampila:
If this is merged (seems reasonable to me), I suppose magical staves should also allow training skills associated with them.

PleasingFungus:
Per https://crawl.develz.org/mantis/view.php?id=9341 and earlier discussions, the current behavior is intentional.

IIRC there were two primary issues. First, we don't want to mislead players into making what is typically a very bad decision, i.e. training a spellcasting skill without any spells that use it. (There's exactly one case I can think of when that'd be particularly reasonable.) Second, we want to minimize the number of special cases for skill training rules.

PR #218, opened 2016-01-13
fixed monsters not triggering traps after using stairs (#9654)

ahamidullah:


wheals:
This somehow has no change from master. I suggest you re-open a new pull request.

PR #223, opened 2016-01-16
Numerous tweaks to monster spawns and map layout

Floodkiller:
Randomize the placement of the death scarabs within the rune chamber to
prevent bunching/immediate LRDing of half the swarm.  Normalize placement
of harder monsters (moths of wrath/ghost moths/emperor scorpions) so that
there randomly it isn't likely to have 10 of them.  Adjust ratios of other
monster spawns as well.  Randomize entrance to inner loop to limit kill
corner/funneling in addition to increasing variability in layout.  Change
"rock statues" outside entrance to actual granite statues, and change
sarcophagus to clear_rock_wall so that the statues aren't called
"sacophogi" while keeping rune preview effect.  Change anti-apport wall
around sarcophagus into stone Tomb wall to better match aesthetic of inner
chamber.

### DIFF
--- a/crawl-ref/source/dat/des/branches/spider.des
+++ b/crawl-ref/source/dat/des/branches/spider.des
@@ -1665,69 +1665,6 @@ xxcc-.i.5...""...h...."".......-"c
   xc----@-"-----x------""-@----ccx
 ENDMAP
 
-#Tomb flavoring, death scarabs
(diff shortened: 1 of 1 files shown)


gammafunk:
Duplicate of #222, so I'm closing this one.

PR #225, opened 2016-01-17
make stash search show matching shop items, even when the shop name matches (#10162)

ahamidullah:


wheals:
Merged in b454403a3a34

PR #226, opened 2016-01-18
Trove commits

drke2:
Some small tweaks based on item rarity, no more than a 1 or -1 change. Some average out the same but with more dice for greater consistency.

Some universally useful consumables were added as options, plus !exp to give an incentive to dip into early zig levels. Silver rune was added because Vaults isn't the de facto easiest third rune anymore. (golden was tempting but seemed a tad sadistic)

Created a trove since there's no trove currently that places misc items. It places 6-7 misc items, 6-7 scrolls of recharging,  2-3 rods, and 2-3 wands, with a 1/3 chance of a magical staff. Vault weight was left at default, unsure if 5 or 15 would be more appropriate.

Squash excess if needed.

### DIFF
--- a/crawl-ref/source/dat/des/portals/trove.des
+++ b/crawl-ref/source/dat/des/portals/trove.des
@@ -886,6 +886,35 @@ wwwwwwwwwwwwwwwwwwwwwwwwwwwww
 wwwwwwwwwwwwwwwwwwwwwwwwwwwww
 ENDMAP
 
+# Misc items/evocations trove. 15 items.
+NAME:   trove_misc_throne_room
+COLOUR: x = blue
+COLOUR: ' = yellow
+SUBST:  d = dde
+KFEAT:  _ = altar_pakellas
+FTILE:  'T_ : floor_limestone
+ORIENT: encompass
+TAGS:   no_item_gen no_monster_gen allow_dup
+ITEM:   acquire misc / any rod w:4
+ITEM:   acquire wand w:15 / any misc w:3 / acquire magical staff w:2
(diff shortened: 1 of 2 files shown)


drke2:
from IRC:
[16:46] <chequers> even if silver is not the easiest third rune, it's still the most common, so making it a trove entry item feels like making the trove effortless entry for many games
[18:06] <+PleasingFungus> slimy is interesting because the most common 3-rune path doesn't involve it, so fulfilling that request means getting a different and/or extra rune

Best suggested options:
a) Remove silver, keep slimy 
b) Keep both
c) Remove both (+abyssal?) and add golden/demonic/hell runes instead (maybe with a lower chance of being chosen?)

Looking for more opinions on this.

MarvinPA:
I'd agree with chequers and PF in terms of keeping the slimy rune over the silver rune.

I'm not sure about some of the added tolls. Demon weapons and dragon armours work because if one doesn't randomly show up, there are places you can go to fairly reliably find enemies that might drop one - that's not really the case for lajatangs. Scrolls of recharging were explicitly removed in the past in favour of just asking for more highly-charged items in the past - I think just bumping up the charge requirement on the wands like you already have done is fine (and doing the same for weapons also looks good since I don't think those have even changed since enchant weapon scrolls were merged and made to always succeed). I think consumables like agility and resistance also used to be possible tolls but were removed in favour of having a shorter list of consumables, they might be fine to add again though I guess.

drke2:
OK, rebuilt the branch, only additions are HDA and the potions, and an additional +1 average weapon ench.

PR #228, opened 2016-01-21
Don't ask for armour from draconians or octopodes in troves

edsrzf:


drke2:
Restricting Dr seems iffy, as dragon armour serves as a sort of "this trove requests x scrolls of enchant armour", and they do want several of them per game, and more if using a shield. Op seems fine to me though.

PR #230, opened 2016-01-28
fixed monsters not triggering traps after using stairs (#9654) (resubmit)

ahamidullah:


wheals:
Thank you, and sorry about the Git issues! Since it wasn't so clear, your commit is in master in eb5f3690e09941ceda6e321a459bb1802e32c8c1.

PR #231, opened 2016-01-29
Enhance and expand upon the captain's cutlass

drke2:
The formula for disarming success has been changed and it is now based
on target HD. Disarming should be nominally easier to pull off.

Enchantment has been improved from +5 to +7. (was +10 pre-disarm)

There is now a chance when disarming to deal 18-35 (average 26.5)
non-elemental damage. This is akin to having a powerful flat damage
brand versus monsters wielding a weapon, but it will only trigger while
disarming.

TSO followers are still allowed to wield the weapon, but they won't use
the disarm feature.

PR notes: This seemed more interesting than just giving it +10 slay again.
The idea behind the attack is a disarmed monster is probably going to
be off-balance or trip/stumble, etc. so he's briefly vulnerable to your stabber
persona. Damage is a bit less than 2 elec brands, which seems to be about
right from testing it vs. orc and vaults.

The damage was made fairly substantial to make it rewarding, I don't have 
a strong feeling about randart power balance. It's certainly better than any 
past iteration of this specific unrand (vs monsters with weapons anyways) but
imo not as good as, for instance, arc blade.

Other goals were to have better disarm success formulas in general, by
incorporating HD and resolving inflated stab damage with one_chance_ins
instead. (it seems bad to have it work every time in a stab scenario against
uniques etc.)

### DIFF
--- a/crawl-ref/source/art-func.h
+++ b/crawl-ref/source/art-func.h
@@ -1274,25 +1274,44 @@ static void _OCTOPUS_KING_world_reacts(item_def *item)
 
 ///////////////////////////////////////////////////
 
+static void _CAPTAIN_equip(item_def *item, bool *show_msgs, bool unmeld)
+{
+    if (you_worship(GOD_SHINING_ONE))
+    {
+        _equip_mpr(show_msgs,
+                   "You feel dishonourable wielding this.");
+    }
+    else
+    {
+        _equip_mpr(show_msgs,
+                   "You feel a cutthroat vibe.");
+    }
+}
+
 static void _CAPTAIN_melee_effects(item_def* weapon, actor* attacker,
-                                   actor* defender, bool mondied, int dam)
+                                actor* defender, bool mondied, int dam)
 {
     // Player disarming sounds like a bad idea; monster-on-monster might
     // work but would be complicated.
-    if (!attacker->is_player() || !defender->is_monster() || mondied)
-        return;
-
-    if (x_chance_in_y(dam, 75))
+    if (coinflip()
+        && dam >= (3 + random2(defender->get_hit_dice()))
+        && !x_chance_in_y(defender->get_hit_dice(), random2(20) + dam*4)
+        && attacker->is_player()
+        && defender->is_monster()
+        && !mondied)
     {
         item_def *wpn = defender->as_monster()->disarm();
         if (wpn)
         {
-            mprf("You knock %s %s to the ground with your cutlass!",
-                 apostrophise(defender->name(DESC_THE)).c_str(),
-                 wpn->name(DESC_PLAIN).c_str());
+            mprf("The captain's cutlass flashes! You lacerate %s!!",
+                defender->name(DESC_THE).c_str());
+            mprf("%s %s falls to the floor!",
+                apostrophise(defender->name(DESC_THE)).c_str(),
+                wpn->name(DESC_PLAIN).c_str());
+            defender->hurt(attacker, 18 + random2(18));
+            did_god_conduct(DID_UNCHIVALRIC_ATTACK, 3);
         }
     }
-
 }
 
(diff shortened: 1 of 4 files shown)


crawl-ref/source/art-func.h
```diff
@@ -1274,25 +1274,55 @@ static void _OCTOPUS_KING_world_reacts(item_def *item)
 
 ///////////////////////////////////////////////////
 
+static void _CAPTAIN_equip(item_def *item, bool *show_msgs, bool unmeld)
+{
+    if (you_worship(GOD_SHINING_ONE))
+        {
```
wheals: style: braces go on the same column as the `if`

crawl-ref/source/art-func.h
```diff
@@ -1274,25 +1274,55 @@ static void _OCTOPUS_KING_world_reacts(item_def *item)
...
         {
-            mprf("You knock %s %s to the ground with your cutlass!",
-                 apostrophise(defender->name(DESC_THE)).c_str(),
-                 wpn->name(DESC_PLAIN).c_str());
+            mprf("You disarm %s, knocking %s onto the floor!",
+                defender->name(DESC_THE).c_str(),
+                wpn->name(DESC_A).c_str());
+            did_god_conduct(DID_UNCHIVALRIC_ATTACK, 3);
+            if ((!x_chance_in_y(defender->get_hit_dice(), random2(20) + dam*4)
+            && !stab) || (stab && !one_chance_in(3)))
```
wheals: `&&` should be aligned with the previous line, and the `||` etc. should be on its own line

crawl-ref/source/art-func.h
```diff
@@ -1274,25 +1274,55 @@ static void _OCTOPUS_KING_world_reacts(item_def *item)
...
-    if (!attacker->is_player() || !defender->is_monster() || mondied)
-        return;
-
-    if (x_chance_in_y(dam, 75))
+    if ((coinflip()
+        && dam >= (4 + random2(defender->get_hit_dice()))
+        && attacker->is_player()
+        && defender->is_monster()
+        && !mondied)
+        && (!stab || stab && one_chance_in(3)))
```
wheals: I'd expect it to be more likely if you're stabbing; am I missing something?

crawl-ref/source/art-func.h
```diff
@@ -1274,25 +1274,55 @@ static void _OCTOPUS_KING_world_reacts(item_def *item)
...
+                   "You feel a cutthroat vibe.");
+        }
+}
+
 static void _CAPTAIN_melee_effects(item_def* weapon, actor* attacker,
-                                   actor* defender, bool mondied, int dam)
+                                actor* defender, bool mondied, int dam)
 {
+    // Use a formula that isn't based on damage for tier 1 and 2 stabs
+    const bool stab = (defender->paralysed() || defender->asleep()
```
wheals: Could this use `find_stab_type(attacker, defender) >= STAB_HELD_IN_NET`?

MarvinPA:
This looks sort of overcomplicated, is the interaction with stabbing really needed? I'm not entirely convinced about giving it damage on disarming but it seems like if that's going to be the case it could just always trigger on disarming, and adjust the disarm chance as appropriate, rather than stack up more random chances.

drke2:
I felt like it was bad to give a basically guaranteed disarm on stabs given how the damage formula is balanced around normal hits with a rapier of speed that tend to be fairly low damage.

The second chance is basically along the same lines of rewarding higher damage dealt, but it could be removed I suppose.

drke2:
Merged the two checks together, making the first check a bit easier to compensate so that the damage can always trigger on disarming.

drke2:
squashed into one commit - adding TSO interaction doesn't seem worthy of a second commit.

PR #233, opened 2016-02-07
Clean up merfolk swimming checks (#10253)

qoala:


crawl-ref/source/transform.cc
```diff
@@ -1974,15 +1974,13 @@ bool transform(int pow, transformation_type which_trans, bool involuntary,
...
-                      && you.species == SP_MERFOLK)
-    {
+    // Land the player if we stopped flying.
+    if (was_flying && !you.airborne())
         move_player_to_grid(you.pos(), false);
-    }
+
+    // Update merfolk swimming for the form change.
+    if (you.species == SP_MERFOLK)
+        merfolk_check_swimming(false);
```
qoala: this will be redundant if we also called `move_player_to_grid` for landing, but as all the `merfolk_*_swimming` methods are idempotent, I think being more clear about what's needed outweighs the performance cost of calling this twice?

(same for `untransform` below)

PR #235, opened 2016-02-11
Panlord tweaks

alexjurkiewicz:
Mess around with AC/EV, rF/rC immunity, and rElec levels

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -601,74 +601,74 @@ static string _describe_demon(const string& name, bool flying)
     const uint32_t seed = hash32(&name[0], name.size());
     #define HRANDOM_ELEMENT(arr, id) arr[hash_rand(ARRAYSZ(arr), seed, id)]
 
-    const char* body_descs[] =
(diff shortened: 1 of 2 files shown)


drke2:
:+1:

MarvinPA:
I don't think elec vulnerability should be used here (or really anywhere especially).

alexjurkiewicz:
What's the diff for you between elec vuln and fire/cold vuln?

alexjurkiewicz:
I've significantly added to this commit since @drke2's thumbs up so it may no longer apply. Comments welcome.

Overall this is a pretty significant buff to pan lords, I think. I tried to keep their "personality" -- it's still possible to find the occasional pan lord with terrible defenses and summon butterflies, just less likely.

alexjurkiewicz:
As per IRC, I'm not sure if 9e24628 is a good idea. It might be a better idea to a) remove low level hexes from pan lords (this would remove almost all hexes though) or b) make monster hexes less resistable in general (so that reaching 0/1% success chance against the player isn't a thing that happens)

PR #237, opened 2016-02-19
A few build system improvements

hasufell:
Most of these fix build failures.

### DIFF
--- a/crawl-ref/source/Makefile
+++ b/crawl-ref/source/Makefile
@@ -605,7 +605,7 @@ endif
 #
 
 ifndef NO_PKGCONFIG
-ifeq ($(shell which pkg-config 2> /dev/null),)
+ifeq ($(shell which $(PKGCONFIG) 2> /dev/null),)
 NO_PKGCONFIG = YesPlease
 endif
 endif
--- a/crawl-ref/source/Makefile
+++ b/crawl-ref/source/Makefile
@@ -46,6 +46,7 @@
 #    USE_ICC       -- set to use Intel's compiler
 #    LTO           -- set for better optimization but slower compilation,
 #                     requires gcc4.6+
+#    NO_TRY_GOLD   -- if set don't try to detect a working gold linker
 #    NOASSERTS     -- set to disable assertion checks (ignored in debug mode)
 #    NOWIZARD      -- set to disable wizard mode.  Use if you have untrusted
 #                     remote players without DGL.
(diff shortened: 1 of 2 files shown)


neilmoore:
A few things I noticed.  Noting them here rather than on the individual commits because commit comments are easy for other devs to miss.  Most of these apply to 6d58ad3

  * You don't want to use the supplied EXTERNAL_FLAGS when building things with HOSTCC/HOSTCXX (e.g. the tile generator).  This is important because we support cross-compilation (that's how we make the official Windows builds), and the supplied flags might include options that are only applicable to the target architecture, not the host architecture.
  * It looks like the supplied external flags would get doubled up, because they're already added to CFLAGS later in the makefile (around lines 108 + 1176 for cflags, around line 1055 for ldflags)
  * In particular, this causes Android builds to ignore EXTERNAL_FLAGS_L because it clears LDFLAGS just a few lines down.  (I'm not sure why it does that in the first place: I have to imagine it has to do with the Android SDK providing some default flags that we want to ignore.)
  * In general, I think it's better for user-supplied flags to appear later in the command line rather than earlier, because that allows them to override things (like warning flags or optimisation flags) that are added to CFLAGS elsewhere in the Makefile.

hasufell:
> Most of these apply to 6d58ad3

Afais, your comments apply only to that commit. I am fine to remove it. Does the rest look ok?

neilmoore:
My only concern about the others is that the updated ncurses rules ignores NO_PKGCONFIG.  Not sure whether it would be better to set PKGCONFIG=false when NO_PKGCONFIG is set; or to wrap all this inside an ifndef NO_PKGCONFIG (and provide an alternative branch).

hasufell:
Well, NO_PKGCONFIG could just be refactored out, but I don't want to touch more Makefile code, so I'll go with another ifdef.

neilmoore:
There are at least some situations where people explicitly pass NO_PKGCONFIG=y as an option to make.  On the other hand, I guess they could do that now with PKGCONFIG=false

hasufell:
> There are at least some situations where people explicitly pass NO_PKGCONFIG=y as an option to make. On the other hand, I guess they could do that now with PKGCONFIG=false

Yes they could, but that also has a risk of breaking peoples build scripts, so I'll leave the amount of changes to a minimum.

neilmoore:
Many thanks!  I also factored out some duplicated logic for ncurses vs ncursesw ([0.18-a0-1358-gd0e7041](https://github.com/crawl/crawl/commit/d0e70415cf64)).

hasufell:
great, thanks

PR #243, opened 2016-03-03
Rework Gozag incite.

alexjurkiewicz:
Currently you can avoid most Gozag wrath by walking backwards whenever
an enemy is incited. This changes makes the buffs permanent and removes
the relatively weak effect of might.

Incited monsters get an 'incited by Gozag' enchantment indicator so the
player can tell what's up.

Now, spellcasters are hasted and everything else gets berserked.


MarvinPA:
Merged, thanks. Added an additional tweak to make haste undispellable too.

PR #244, opened 2016-03-08
Antispam option

giann:
# Added autofight_warning option.

_The goal of this option is to be the counterpart of `tile_key_repeat_delay=0` for console mode._

Since we don't know if the player keeps the `tab` key down in console mode, I keep track of time elapsed between two tab commands in milliseconds (using `std::chrono`).
If that delta is less than the autofight_warning option, we stop the user with the message "_I need to give more thought to this attack_".

The option has no effect if the `tab` command was issued by a macro.

### Note
Since this is my first patch, I look forwards to your comments !

### DIFF
--- a/crawl-ref/source/dgnevent.h
+++ b/crawl-ref/source/dgnevent.h
@@ -13,31 +13,33 @@
 // Keep event names in l_dgnevt.cc in sync.
 enum dgn_event_type
 {
-    DET_NONE            = 0x0000,
+    DET_NONE              = 0x0000,
 
-    DET_TURN_ELAPSED    = 0x0001,
-    DET_MONSTER_MOVED   = 0x0002,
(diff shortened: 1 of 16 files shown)


giann:
Deleted `handle_real_time_delta`. Everything is done in handle_real_time.
Time is kept in milliseconds with `you.real_time_ms` and converted down to seconds into `you.real_time`.
At save loading I set `you.real_time_ms` to `you.real_time * 1000`.

crawl-ref/source/dgnevent.h
```diff
@@ -13,31 +13,31 @@
 // Keep event names in l_dgnevt.cc in sync.
 enum dgn_event_type
 {
-    DET_NONE            = 0x0000,
```
amalloy: Why did you realign (misalign) all these flags without making any meaningful changes?

PR #247, opened 2016-03-11
have_passive: Jiyva, Vehumet, Nemelex, Qazlal, Ru, Pakellas

giann:
Didn't find how Ru's retribution is called for `passive_t::upgraded_aura_of_power`.


neilmoore:
Several of the piety ranks seem incorrect when compared with the existing code:

  * `resist_corrosion` uses `piety_breakpoint(2)` so should be rank 3.
  * I'm not sure where the rank 5 breakpoint for `jellies_army` came from, but that seems wrong: it means you won't get any jiyva jelly spawning or offlevel item eating until *****
  * `jiyva_eat_offlevel_items` should not be checking those three passives: Jiyva will eat offlevel items even if it doesn't actually give you a bonus.  Perhaps there should be a separate passive for this (rank -1).
  * Veh `mp_on_kill` should be rank -1, `spells_success` rank 3, `spells_range` rank 4.
  * Nem `cards_power` should be rank 0.
  * Qaz `resist_own_clouds` should be rank 0, `storm_shield` rank 1, `upgraded_storm_shield` rank 4, and `elemental_adaptation` rank 5
  * Ru `aura_of_power` should be rank 1; `upgraded_aura_of_power` seems to be unused, but would be rank 2.

In general: if the code being replaced says just `you.worship(...)`, that's rank -1; if it says `in_good_standing(god)`, that's rank 0; and if it says `in_good_standing(god, N)` or `you.worship && you.piety >= piety_breakpoint(N)`, that's rank N+1.

Also:

  * the `have_passives` for jiyva bonuses (in `jiyva_slurp_item_stack`) are in the wrong place: they should probably appear in `_sacrifice_one_item_noncount`, since that's the function that currently has the piety checks.
  * I'm not sure if `unlock_slime_vaults` should actually be a passive: it's more like an active ability that happens automatically.  But I guess that's debatable.

PR #248, opened 2016-03-11
Reduce max weapon enchantment to +5

aolowin:
After the change to enchant scrolls it became very easy to get a +9 weapon,
often before depths. This change makes melee a bit more difficult, gives
less incentive to hoard EW scrolls for your end weapon, and is a slight
buff to artefacts/unrands.


MarvinPA:
When this was discussed in IRC it seemed to be agreed that if there actually is a problem with it being too easy to enchant weapons to +9, this is probably the wrong way to go about fixing it.

If a change is needed, some of the alternate fixes suggested included just making EW rarer, or somehow requiring multiple scrolls to reach higher enchant levels (imitating the old failure chances somewhat).

PR #249, opened 2016-03-17
Ds mut nerfs

alexjurkiewicz:
Both these commits are pretty crazy on d:1-d:5 after my past changes, here's an attempt to make them more sensible.

Spiny commit is done, but I'm looking for comments on the powered by death rework. I think the numbers need further tweaking.


alexjurkiewicz:
ping

PR #251, opened 2016-04-02
fix for expose_to_element not being called properly (10293)

rawlins:
This started as an attempt to add expose_to_element back into beam.cc so that Qazlal characters properly adapt in response to spells; it spiraled out of control due to consequences for other aspects of exposure to elements.  The PR will have two main effects: (i) cause Qazlal elemental adaptations to trigger (hopefully properly) on all appropriate spells, and (ii) cause cold-blooded players to be slowed properly on all appropriate spells.  Neither of these were working correctly.

This also eliminates the call to melt player enchantments from expose_to_element: it is called from elsewhere and wasn't working properly in any case (it expects damage, not "strength").  However, this call may have been doing some token melting in special cases (token in the sense that it wouldn't have a real impact on duration).

A few other small changes: duration of adaptation no longer scales with strength (which was only doing something in weird special cases anyways).  one of the brands was being checked wrong.  Correctly trigger physical protection on earth spells (and others).

### DIFF
--- a/crawl-ref/source/attack.cc
+++ b/crawl-ref/source/attack.cc
@@ -1588,11 +1588,17 @@ bool attack::apply_damage_brand(const char *what)
         calc_elemental_brand_damage(BEAM_FIRE,
                                     defender->is_icy() ? "melt" : "burn",
                                     what);
+        defender->expose_to_element(BEAM_FIRE, 2);
+        if (defender->is_player())
+        {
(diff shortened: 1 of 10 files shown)


rawlins:
I found some more cases of melting that aren't handled properly, working on a fix.

rawlins:
This now gets everything I've thought to test.  For better or worse, moving maybe_melt_player_enchantments back to expose_player_to_element wouldn't work, because it expects damage, not "srength".

PR #252, opened 2016-04-05
Mummy unconfuse

Hurricos:
This branch implements a change to Mummies' Self-restoration ability.  It is now usable while confused and cures confusion.  Damn tarantellas.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -313,7 +313,7 @@ static const ability_def Ability_List[] =
     { ABIL_STOP_SINGING, "Stop Singing",
       0, 0, 0, 0, {}, abflag::NONE },
     { ABIL_MUMMY_RESTORATION, "Self-Restoration",
-      1, 0, 0, 0, {}, abflag::PERMANENT_MP },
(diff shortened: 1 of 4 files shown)


cbuchananhowland:
I'm reluctant to allow mummies to always have access to a cure for confusion; it removes one of their signature weaknesses. I feel strongly that we should move confusion to a place where it's reasonable for all characters to deal with monsters that cause confusion w/ !curing. Are we there yet? No, but this isn't the right solution.

Hurricos:
PleasingFungus found this solution acceptable, and it consumes MaxMP.  I
thought it was reasonable considering how many Mummies die in Spider to a,
what, 7HD enemy?

Shall I open another branch in which Tarantellas do not increase
you.duration[DUR_CONF] while confused, or just leave people to die?

MarvinPA:
I agree with Lasty, and maybe also with the -dev backlog suggestion that Mummies just lose the self-restoration ability now that stat loss recovers with XP. I guess undead generally should always lose stats instead of rotting HP in that case, which is probably okay.

Hurricos:
Fair enough.

... want me to implement it? ;)

MarvinPA:
Closing as per above discussion.

PR #254, opened 2016-04-14
Prevent elemental adaptation at max resists

rawlins:
Before, the resistance would apparently trigger and show the rF+ status
indicator, but have no effect -- despite the misleading status
indicator.  With this patch it will silently fail if the player is at
max resists for the relevant element at the time of exposure.  This
could be a very slight buff in that AC can always trigger, and it is now
more likely to be continuously on when receiving hybrid damage (e.g.
brands) when at max resists.  However, AC is the weakest of the resists
at the point in the game where this is typically relevant.

This doesn't do anything with equipment, so you could still get the
status indicator on at max resists by letting it trigger and then
putting on the relevant ring.

### DIFF
--- a/crawl-ref/source/godpassive.cc
+++ b/crawl-ref/source/godpassive.cc
@@ -1032,17 +1032,23 @@ void qazlal_element_adapt(beam_type flavour, int strength)
         case BEAM_LAVA:
         case BEAM_STICKY_FLAME:
         case BEAM_STEAM:
+            if (you.res_fire() >= 3)
+                return;
             what = BEAM_FIRE;
             dur = DUR_QAZLAL_FIRE_RES;
             descript = "fire";
             break;
         case BEAM_COLD:
         case BEAM_ICE:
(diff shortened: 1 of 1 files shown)


MarvinPA:
I think this is intended (and fine) behaviour, making it so that your adaption will still adjust on taking damage a bunch even if you have lots of resists already (most noticeable for rElec) and so aren't necessarily set up to make the most of it.

rawlins:
Thanks for the feedback, this new version will shut off other resists if a new one gets rolled but wouldn't trigger because of max resists.  It should have exactly the adaptation swapping behavior as master, but fix the interface issue.

MarvinPA:
Closing this - as discussed in IRC a while ago, it's fine to still give the status light at max resists since the % screen is still the place to look for full resist information (and it's consistent with, say, potions of resistance). Shutting off the ability when it would trigger at max resists feels potentially more complicated and confusing to me.

PR #255, opened 2016-04-14
Allow armored quarterstaff users of all races (fighter)

yrmvgh:
If a player wants to start with a quarterstaff and armor (plus a shield they can't wield), let them. They may be planning to transition into an enhancer staff later in their career, or perhaps they just want to lean on the awesome early game power of a quarterstaff while they train up shields and other skills.

A quarterstaff start is a great start, and for some races, an extraneous shield is preferable to being sent down the Gladiator-style light armor route.

(also, this simplifies the special case for formicid, so that a player who chooses DdFi today has the opportunity to notice the otherwise "hidden" quarterstaff option that works so well for a Formicid)

### DIFF
--- a/crawl-ref/source/ng-restr.cc
+++ b/crawl-ref/source/ng-restr.cc
@@ -101,12 +101,6 @@ char_choice_restriction weapon_restriction(weapon_type wpn,
         return CC_RESTRICTED;
     }
 
-    if (wpn == WPN_QUARTERSTAFF && ng.job != JOB_GLADIATOR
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
I suggest non-FO quarterstaff fighters should not start with a shield.

PleasingFungus:
My feeling is that the shield is pretty core to the fighter 'package'; not, of course, that all fighter starts are required to keep using shields, but that it's something that should always be a big part of the start. (It's also nice to have the quarterstaff option be more a differentiator for gladiators.)

Thank you for the submission, and apologies for the slow response!

PR #257, opened 2016-04-28
Let claws/hooves work through gloves/boots.

alexjurkiewicz:
It's a bit simulationist to have this restriction and is very difficult
to communicate to the player. There's not really any other circumstance
you wouldn't want to fill an equipment slot.

Gloves can be considered the finger(claw)-less/ Wolverine-compatible
variety, and I don't think boots would prevent kicking from working
anyway.

### DIFF
--- a/crawl-ref/source/attack.cc
+++ b/crawl-ref/source/attack.cc
@@ -1301,7 +1301,6 @@ int attack::calc_base_unarmed_damage()
 
     int damage = get_form()->get_base_unarmed_damage();
 
-    // Claw damage only applies for bare hands.
     if (you.has_usable_claws())
         damage += you.has_claws() * 2;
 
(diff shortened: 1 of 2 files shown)


wheals:
Is not changing talons intentional?

alexjurkiewicz:
Nope, oversight

alexjurkiewicz:
Fixed talons

MarvinPA:
This fact is listed in glove/boot descriptions now, and is also communicated for claws by the Wp: line changing when you equip gloves I think? No strong opinion on whether it should be changed or not (I'd weakly tend towards it being fine as-is now) but the descs would need to be updated again if this were merged.

alexjurkiewicz:
Happy to change descriptions if people like this change. I don't like how
it's currently the one reason not to wear an aux slot. It's a trade off,
but not particularly well communicates to the player.

PleasingFungus:
As MPA said, I think this is fine-ish now.

PR #258, opened 2016-04-30
https://crawl.develz.org/mantis/view.php?id=10384

yrmvgh:
https://crawl.develz.org/mantis/view.php?id=10384


MarvinPA:
Pushed this, could you include a commit message for any future PRs rather than just linking the Mantis issue (describing what the commit does, and you can reference the Mantis issue just by number).

PR #261, opened 2016-05-07
commenting convention: don't comment out code

yrmvgh:
PleasingFungus said something to this effect in IRC, and it seems a useful "rule" to have defined in the coding conventions.


PleasingFungus:
I'd put that at the end of the section, not the beginning.

PR #262, opened 2016-05-10
shaft-proof formicid

yrmvgh:
theme: mastery over hoies in the earth
physical: 6 strong limbs prevent you from falling in
gameplay improvement: shafts as a trap can make the piayer's own, already-risky shaft ability useless

inspired by tavern https://crawl.develz.org/tavern/viewtopic.php?f=8&t=19860

### DIFF
--- a/crawl-ref/source/traps.cc
+++ b/crawl-ref/source/traps.cc
@@ -933,6 +933,11 @@ void trap_def::trigger(actor& triggerer, bool flat_footed)
         // Known shafts don't trigger as traps.
         if (trig_knows)
             break;
+            
+        // FO are too tunnel-skilled and many-limbed to "fall" for shafts.
+        if (you.species == SP_FORMICID)
+            mpr("You detect a shaft and decide not to fall in.");
+            break;
 
         // A chance to escape.
         if (one_chance_in(4))


MarvinPA:
I don't see that shafts being more dangerous for Formicids is any kind of problem that needs to be solved, it seems perfectly reasonable for them to maybe put you into a situation where you self-shaft again to recover. Note that they're already immune to another kind of trap (teleportation) without needing any extra hidden special-casing.

yrmvgh:
It's not a "problem", game-wise. Shafts are rare enough that you can go an entire game without falling in one, Fo or otherwise. So changed or unchanged, it will make little difference.

It's simply a flavor glitch (They can sense monsters through earth and steel, they can dig rapid but they can't detect an ogre-sized fissure?), the solving of which would probably do more good than harm.

ada482:
While shafts are rare gameplay wise, they're far more lethal in the early game to formicids than other races. If a formicid falls three floors, and finds themselves in a situation where they need to immediately escape, it's either a fighting retreat, trying to dig through the walls next to the weakest threat, or shafting yourself potentially another two floors. Going from D:2 threats to D:5 threats to D:7 threats would be highly frustrating even for the best of players, especially when the formicid's greatest early game ability (sensing threats through walls) is severely hampered by not having cleared rooms to retreat to. And assuming that the formicid does have a somewhat lucky self-shaft after a poor shaft, they still need to potentially fight that threat on their rise back up to where they originally fell from. Their choices are so tightly limited after falling, and all of them contain the same risks for the same benefits. If having immunity to shafts is such a deal breaker in terms of unique power, then at least stasis should hard lock any falls to single floor increments, or the shaft-self should be replaced by a build shaft ability, providing the formicid with the ability to make a safe retreat position in what's essentially a mass out-of-depth problem.


Or we could just note that ants climb vertical surfaces in nature, and that it'd be absurd for them to suffer a pitfall, just as it'd be absurd for cats with scissor legs to be stealthy.

dpeg:
I like the commit. Agreed, the change is mostly about flavour but I think it strengthens the species' role a bit.

gammafunk:
I wouldn't call formicids being susceptible to shafts a "flavor glitch" at all. Them being able to dig holes does say to me that they should be immune to a trap holes appearing underneath them.

As MarvinPA points out, there's no need to compensate for any perceived susceptibility of Formicids to shaft traps in particular. They have two innate abilities for dealing with the limitation they're designed around, and those abilities are more than what most species have in the early game when they lack consumable escape items.

PleasingFungus:
I'm gonna agree with MarvinPA and gammafunk here. Thank you for the submission!

PR #263, opened 2016-05-12
nerf ossuary bog body

yrmvgh:
These things are more dangerous than an ogre, in a portal that players can typically count on being low in danger and poor in loot. Nasty surprise for the unspoiled, especially players with a few wins under their belt who know what to expect from low-level portals.

Still deals 25 damage!!!!!!!!!! but is a lot easier to kill... if it doesn't 1-hit the Spriggan/DE player.

### DIFF
--- a/crawl-ref/source/dat/des/portals/ossuary.des
+++ b/crawl-ref/source/dat/des/portals/ossuary.des
@@ -685,7 +685,7 @@ MONS:    generate_awake zombie place:D:6 /\
 MONS:    w:15 withered plant /\
          mummy \
          spells:throw_frost.18.wizard;slow.27.wizard \
-         name:magical n_suf /\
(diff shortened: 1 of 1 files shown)


cbuchananhowland:
IMO this really shouldn't be a bog body at all. If this has to exist, it should be a mummy with Throw Frost or something. That would fix the damage problem, and yes it's a problem.

yrmvgh:
That sounds good too. Anything but the current situation. This one special monster in one particular vault is the 3rd biggest killer in ossuary! (after mummy and zombie)

4Hooves2Appendages:
I lost a decent centaur to a couple of bog bodies. It was avoidable by just leaving, so I was obviously stupid. They should definitely stay! Mainly because they kill players.

PR #268, opened 2016-05-15
this might fix the dithmenos issue

yrmvgh:
but it begs the question, what about monsters that are **wielding** fiery weapons, should we hit them too?

https://crawl.develz.org/mantis/view.php?id=10417


neilmoore:
This does not work (it tries to look for the exact name "flaming", except that it actully makes the name "flaming").  We shouldn't go by the name, anyway, or there will be weird results when a random panlord happens to get a matching name.  Finally, the check should go into mons_is_fiery() rather than being used alongside it.

PR #273, opened 2016-05-23
Show average monster hitpoints in x-v.

alexjurkiewicz:
🙀  change 👹  slippery slope 🙈  avoid drama

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -3082,12 +3082,14 @@ static void _add_energy_to_string(int speed, int energy, string what,
  * @param value[in]         The current value represented by the bar.
  * @param max[in]           The max value that can be represented by the bar.
  * @param scale[in]         The value that each + and . represents.
+ * @param geometric[in]     The value to increment each +/.'s scale.
  * @param name              The name of the bar.
  * @param result[in,out]    The stringstream to append to.
  * @param base_value[in]    The 'base' value represented by the bar. If
  *                          INT_MAX, is ignored.
  */
 static void _print_bar(int value, int max, int scale,
+                       int geometric,
                        string name, ostringstream &result,
                        int base_value = INT_MAX)
 {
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
🙀

PleasingFungus:
bugged:
- derived undead (e.g. skeletons)
- mutant beasts
- probably ghost demons
- other stuff?
- #crawlcode

MarvinPA:
As discussed in -dev a while ago, this doesn't seem like a very good number to display since it doesn't really provide much useful information with player damage output not being displayed (and player damage output is probably not something that can be displayed in a good way).

If it's especially desirable to have something like this just to allow comparing health between monsters, this is probably a case that''d work just fine with adjectives or bars like AC/EV.

alexjurkiewicz:
I don't entirely disagree, Chris.

I think in crawl-dev there was general consensus to show more data as real
numbers. Showing HP seems like a good place to start: it might not be that
useful in isolation, but the data isn't available in crawl right now, it's
trivially available outside in IRC, and it's a good proof of concept for
showing more numbers.

MarvinPA:
I don't agree that it's a good place to start, or that it's something that needs any kind of proof of concept - it's not a new idea, other numbers have already been added in cases where it seems reasonable and useful (hex success chances, for example), and bars or adjectives have been used instead where that seems more appropriate (monster AC/EV). I think this falls into the latter camp, and something like monster melee base damage might easily fall into the former.

isloat:
I think a health bar (same as AC/EV) works better. Numbers are only meaningful if you have something to compare them against, and current Crawl doesn't show damage numbers. If a monster has 1770 average health, you'd only be able to grok what that means by comparing against all other monsters with more/less numeric displayed health. 

A health bar above AC/EV in xv is able to be compared against itself. Antaeus has +++++++++, a bat has +.

isloat:
Merged, improperly, so now I have to close this manually. Thank you!

PR #276, opened 2016-05-26
Prevent flight-expiry caused drowning/lava-death.

alexjurkiewicz:
Now, flight expiring when you're over unlandable terrain (eg, lava and
(for most species) deep water) will enable 'emergency flight' mode,
denoted with a red Fly status mark.

While in emergency flight, the player is drained each turn and has
vertigo (which persists for a few turns after landing).

Comments appreciated on coding style and penalties.


PleasingFungus:
Merged in somewhat modified form as b0f541b5c617.

PR #278, opened 2016-05-28
Make flight ignore grasping roots

RBrandon:
This removes the odd and spoilery secret tech of flying over deep water
to avoid getting rooted by a shambling mangrove. It also removes the
evasion-based chance to avoid getting rooted while flying as that is
also spoilery. This makes roots much more straightforward.

### DIFF
--- a/crawl-ref/source/mon-abil.cc
+++ b/crawl-ref/source/mon-abil.cc
@@ -1028,7 +1028,7 @@ void check_grasping_roots(actor* act, bool quiet)
         }
     }
 
-    if (!source || !feat_has_solid_floor(grd(act->pos())))
+    if (!source) 
     {
         if (act->is_player())
         {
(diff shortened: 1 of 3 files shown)


PleasingFungus:
Why not remove the evasion-based chance of evasion (change to some fixed random chance of evasion, maybe), and make grasping roots drop you into emergency flight if you're over deep water? A key part of the reason for grasping roots to exist is that dropping out of flight in swamp is bad in interesting, because of how the terrain restricts your movement; it'd be a shame to lose that.

RBrandon:
I'm not convinced that forcing emergency flight over deep water would be a significant improvement on the status quo. My impression on the purpose of mangroves was to force the player to choose to fight them (an above average threat) or not by slowing movement. I aimed to preserve that aspect rather than the effect on flight. Finally, I think that deep water flight canceling would require a few extra changes to work, none of which I am particularly inclined to put the work into :wink:

PleasingFungus:
OK, then.

PR #279, opened 2016-05-30
Display reflection chance for warlock's mirror

Sandman25DCSS:
Monster spells description displays chance for player to be banished, confused etc so player can make a decision about using MR items.
Warlock's mirror is functionally the same, player can make decision about using amulet of reflection.

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -16,6 +16,7 @@
 #include <string>
 
 #include "adjust.h"
+#include "art-enum.h"
 #include "artefact.h"
 #include "branch.h"
 #include "butcher.h"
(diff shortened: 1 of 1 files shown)


Sandman25DCSS:
http://postimg.org/image/wo0g9qbl7/

alexjurkiewicz:
Should be "The shield has **a**". Also, you can embed images directly into github comments (drag & drop)

Sandman25DCSS:
I am not native Engilsh speaker so I am sure the message can be improved a lot. Also I suspect it should start with "It" instead of "The shield".

Sandman25DCSS:
Thank you about the hint, I don't use GitHub much.

Sandman25DCSS:
description

PleasingFungus:
Rewritten and pushed as 50c6034, thanks! For future, a few small notes:

- It's better to avoid duplicating code - copy-pasting the omnireflect formula between beam.cc and describe.cc is inviting future trouble; either more work when someone tries to change it, or bugs when someone changes one but not the other. Better to make a shared function that both uses call. 
- Code and commit messages should be wrapped to 80 characters, when practical.
- Running util/checkwhite and util/unbrace will remove some common formatting issues.

Thanks for the patch!

Sandman25DCSS:
Thank you, I will keep that in mind.

PR #280, opened 2016-05-30
Merge Naga and Centaur bardings

alexjurkiewicz:
This has a few in-game effects:
* Both barding randarts can be worn by either player species. Centaur
  with lightning scales!
* Acquirement is slightly (30% vs 32.5%) less likely to generate a
  barding for Ce/Na.
* Placement weight should be the same, except that yaktaurs and nagas
  can both spawn with bardings now.

All naga & centaur bardings are converted into the new sub type
'barding' on load.

Current bugs:
* Tiles for naga/centaurs are incorrect in most cases


cbuchananhowland:
I didn't understand the motivation for this tavern thread -- this does create the odd result that the two randarts can be worn by the wrong species, which is bad IMO. It does not seem to solve any problems that I can see.

MarvinPA:
Agreed, I'm also not convinced and don't see any new arguments since the last time this was suggested.

alexjurkiewicz:
Not going to push this one :)

PR #284, opened 2016-05-31
Rework Xom's smoke trail.

alexjurkiewicz:
Turn it from a "bad" effect (which was really more a no-op) to a good
effect that leaves a random cloud trail behind you.

This is considered good because you can use it to cut off monster
pathing and so on, although cutting off your escape is a potential
problem.

### DIFF
--- a/crawl-ref/source/xom.cc
+++ b/crawl-ref/source/xom.cc
@@ -2672,13 +2672,20 @@ static void _xom_repel_stairs(bool unclimbable)
 static void _xom_moving_stairs(int) { _xom_repel_stairs(false); }
 static void _xom_unclimable_stairs(int) { _xom_repel_stairs(true); }
 
-static void _xom_colour_smoke_trail(int /*sever*/)
+static void _xom_cloud_trail(int /*sever*/)
 {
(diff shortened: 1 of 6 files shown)


crawl-ref/source/xom.cc
```diff
@@ -2672,13 +2672,20 @@ static void _xom_repel_stairs(bool unclimbable)
 static void _xom_moving_stairs(int) { _xom_repel_stairs(false); }
 static void _xom_unclimable_stairs(int) { _xom_repel_stairs(true); }
 
-static void _xom_colour_smoke_trail(int /*sever*/)
+static void _xom_cloud_trail(int /*sever*/)
 {
-    you.duration[DUR_COLOUR_SMOKE_TRAIL] = random_range(60, 120);
+    you.duration[DUR_CLOUD_TRAIL] = random_range(60, 120);
+    you.props[XOM_CLOUD_TRAIL_TYPE_KEY] =
+        random_choose_weighted(20, CLOUD_CHAOS,
```
wheals: I think it'd be nice to keep a chance of it being purely cosmetic (`CLOUD_MAGIC_TRAIL`), as now.

alexjurkiewicz:
Updated, you can squash on merge.

PR #285, opened 2016-05-31
Standardise Xom confusion application.

alexjurkiewicz:
Don't try to confuse monsters that will never give XP (oods,
battlespheres, etc) (minmay).

Change how Xom player confusion works -- be less likely to confuse
surrounding creatures as severity increases.

### DIFF
--- a/crawl-ref/source/xom.cc
+++ b/crawl-ref/source/xom.cc
@@ -1033,6 +1033,24 @@ static void _xom_do_potion(int /*sever*/)
     level_change(); // need this for !xp - see mantis #3245
 }
 
+static void _confuse_monster(monster* mons, int sever)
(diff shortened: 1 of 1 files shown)


wheals:
Please fix the compile failure.

Also no-XP is the wrong check; it should probably use `defender->holiness() & (MH_NONLIVING | MH_PLANT)` like [`SPWPN_CONFUSE` does](https://github.com/crawl/crawl/blob/master/crawl-ref/source/attack.cc#L880).

alexjurkiewicz:
Fixed, squash and merge again please.

RBrandon:
I always thought that confusing a fungus was a very Xom-like thing to do :smile:

PR #287, opened 2016-06-01
add a few more speech lines

yrmvgh:


PleasingFungus:
Merged, more or less. Thanks for the submission!

PR #289, opened 2016-06-03
Remove the distinction between 'axe' and 'polearm' baileys.

alexjurkiewicz:
This should have no effect on baileys beyond the obvious of:
- Axe / Polearm signalling outside the bailey goes away
- There's no split in bailey types, so weights shift around.

I performed an extremely dumb "merge" of entry monster/item types in some cases, if any of the merging was too dumb let me know.

### DIFF
--- a/crawl-ref/source/dat/des/portals/bailey.des
+++ b/crawl-ref/source/dat/des/portals/bailey.des
@@ -12,7 +12,7 @@
 ##############################################################################
 
 {{
-function bailey_portal(e, ptype)
+function bailey_portal(e)
   local timeout_turns = crawl.random_range(600, 800)
 
   local messager =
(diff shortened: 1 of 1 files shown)


isloat:
Why should the distinction between axe and polearm baileys be removed?

alexjurkiewicz:
Allows more equitable distribution of bailey map chances, which are (seemingly) intended to be pretty flat but were distorted by the bailey subtype distinction.

PleasingFungus:
Per discussion, I don't think the issues we solve here (vault weight skewing) are very major. The very minor niftiness of getting a warning of the bailey contents (via weapon type) outweigh them.

Hey, I closed it before the end of the year!

PR #290, opened 2016-06-03
Invisibility changes

Sandman25DCSS:
Please see https://crawl.develz.org/tavern/viewtopic.php?f=8&t=20149


PleasingFungus:
The general feeling on the devteam seems to be that this isn't necessary, and i personally like the items & effects in question, so I'm closing this. Sorry for the belated response!

PR #291, opened 2016-06-04
Describe possible number of charges for sack of spiders.

Sandman25DCSS:


Sandman25DCSS:
It does not make much sense to inscribe sack spider with "used N times" when unspoiled player does not know how many charges it can have.

PleasingFungus:
Sacks of Spiders no longer have charges.

PR #295, opened 2016-06-14
Add PID to mapstat log file output.

alexjurkiewicz:
This makes it easier to run several mapstats in parallel.


wheals:
Can you test this on Windows?

alexjurkiewicz:
Hmm, I don't have a Windows compile environment...

alexjurkiewicz:
Would be better to implement this by adding a `--mapstat-output-file` type option

PR #296, opened 2016-06-14
Reduce circular_temple_12 weight.

alexjurkiewicz:
Previously this temple placed in 12.5% of games, compared to 3.5% for
the next most common. It's good to have variable weight temple maps, but
with the current number of maps (70), you still get such a rarity
distribution without a single map dominating the top placement chance.

Fuller placement % analysis here:
https://docs.google.com/spreadsheets/d/1n48f-wt36iYd-ige6bcvD5JZEYofcj-oaunE7QCBp9s

### DIFF
--- a/crawl-ref/source/dat/des/branches/temple.des
+++ b/crawl-ref/source/dat/des/branches/temple.des
@@ -900,7 +900,7 @@ ENDMAP
 
 ##########################################################################
 # Circular temples
-#   Total weight: 90 = 40+10+10+10+5+5+5+5
+#   Total weight: 60 = 10+10+10+10+5+5+5+5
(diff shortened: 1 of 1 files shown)


wheals:
Do you know what sort of effect this has on the average number of gods in Temple? I figured it could cause a fairly big change.

alexjurkiewicz:
It looks like the average temple size goes up, although my quick math isn't good enough to figure out by how much. (Added a column to the google sheet)

wheals:
A more thorough revision was done in cc390b852004c86346b21af8bb9989e556c7ef73 (and there are no longer any huge outliers like that one was).

PR #299, opened 2016-06-16
Double starting skill for unarmed SK_WEAPON starts.

alexjurkiewicz:
It's not really a hard decision between starting with n weapon skill and
n unarmed combat skill, so double the skill for picking UC.


PleasingFungus:
Trolls don't exist?

PleasingFungus:
Closing, per discussion from several days ago ("I don't really want to encourage most races to use pure UC").

PR #300, opened 2016-06-17
Regen mut strength normalisation

alexjurkiewicz:


nikheizen:
It always bothered me that the regeneration mut scaled so much differently from other sources of regen (aside the spell). I very much like this PR and the messaging changes are a nice addition.

PleasingFungus:
Tweaked and pushed, thanks!

PR #301, opened 2016-06-18
WIP: add BOUND_SOUL tile glyph

jmbjr:
commit 708bd0eed added the Bind Souls spell to necromancers, however the only way to know if a monster has had its soul bound is to e'x'amine the monster. It would be preferable to have a Tiles glyph/icon to indicate the bound soul state (especially since the current Bind Souls spell has an infinite duration, similar to durable summoned monsters).

this PR adds a "Bound Soul" icon, but there's still some issues with getting the icons to appear in-game that I need some help fixing.

this is the icons.png file that is created during compilation. http://i.imgur.com/FT07gNc.png
New icon is at the top-right of the other status icons.


jmbjr:
I was using commit: 67f566f1b146 as a partial guide, so I'm sure there's some extra plumbing that I'm totally missing

crawl-ref/source/tilepick.cc
```diff
@@ -1755,6 +1755,8 @@ tileidx_t tileidx_monster(const monster_info& mons)
         ch |= TILE_FLAG_DRAIN;
     if (mons.is(MB_IDEALISED))
         ch |= TILE_FLAG_IDEALISED;
+    if (mons.is(MB_BOUND_SOUL))
+       ch |= TILE_FLAG_BOUND_SOUL;
```
jmbjr: looks like MB_BOUND_SOUL is set up correctly in mon-info.cc in: 708bd0eeda

jmbjr:
for clarity, everything compiles and the game plays just fine, but if I create a necromancer, &m, and wait around for it to bind some souls, I see the monsters get the bind_soul status (via 'x') but I don't see the icon. 

There are no errors that I can see

jmbjr:
I bet I need to shift things but I don't really fully understand how

PleasingFungus:
This doesn't actually compile, which makes me very curious about what's going on with your end. Possibly it's failing and you didn't notice, or you're compiling something else...?

I'll see about fixing it up.

PleasingFungus:
Fixed version pushed as 7a0dbba, thanks!

PR #302, opened 2016-06-21
Weight temple map selection by altar count.

alexjurkiewicz:
Before picking a temple map, decide how many altars we want. Crawl temple weights manage two things separately:
* Average temple size
* Temple map weighting

Temple size range weights are as per regret-index's desire in cc390b8.


alexjurkiewicz:
I'm not very familiar with map code, I think the code has several bugs right now, so this PR is open to get approval for the concept / implementation in general before I go back and finish the 80% of work squashing bugs :)

PleasingFungus:
I think this one is moot after the big r-i temple revision?

alexjurkiewicz:
Correct

PR #303, opened 2016-06-22
Remove see invisible from items

Sandman25DCSS:
Please see https://crawl.develz.org/tavern/viewtopic.php?f=8&t=20396 for reasoning and discussion

### DIFF
--- a/crawl-ref/source/artefact.cc
+++ b/crawl-ref/source/artefact.cc
@@ -364,7 +364,6 @@ static map<jewellery_type, vector<jewellery_fake_artp>> jewellery_artps = {
 
     { RING_MAGICAL_POWER, { { ARTP_MAGICAL_POWER, 9 } } },
     { RING_FLIGHT, { { ARTP_FLY, 1 } } },
-    { RING_SEE_INVISIBLE, { { ARTP_SEE_INVISIBLE, 1 } } },
     { RING_STEALTH, { { ARTP_STEALTH, 1 } } },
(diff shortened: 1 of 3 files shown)


PleasingFungus:
Your complaint is that it's too easy to swap to weapons or rings to spot invisible monsters, so you remove it from hats?

dpeg:
I agree with the basic reasoning: carrying a ring of SInv around and swapping it in occasionally is not very interesting. Since I also think that invisible monsters are fun to fight (if it does not happen too often), I support the change. (This is basically the same reasoning behind "rMut removal.)

I can see leaving SInv on hats. If we keep this, then we need to check whether invisible monsters are indeed rare enough. Might be necessary to reduce their generation a bit.

Sandman25DCSS:
My intention was to remove see invisible from all sources: hat, ring, randarts. Only unrandarts keep it.

Sandman25DCSS:
Teleporting away, changing hat to one with see invisible and returning back to kill that Unseen Horror is too similar to teleporting away from Mottled Dragon, changing cloak to one with preservation ego and returning back to kill the dragon.

PleasingFungus:
OK.

I believe this was closed by accident, but the consensus on the devteam is that this isn't a good change. sInv certainly doesn't 'trivialize' any otherwise non-trivial monster, and removing sInv makes interacting with invisible monsters much more annoying. There's probably room for some monster invisibility changes, but this is the wrong way to go about it.

Thanks for the submission!

PR #306, opened 2016-06-26
Forbid *Corrode on CPA artefacts

ChrisOelmueller:
Much like dwarven items, crystal plate mail is exceptionally resistant to corrosion, after all.


PleasingFungus:
I asked around, and dev consensus seems to be that the old flavor is no longer canon. Sorry!

PR #308, opened 2016-07-05
Xom improvements

zooko:


PleasingFungus:
**Design notes**: Do you have any reasoning for the warning?

The !xp effect seems extraordinarily strong; is there any reason you're making it more common than it is already? (Also, you realize that !xp is still useful even at XL 27, yes?)

The problem with Enslave as an autonomous god action is that it messes with the UI. If you want to keep attacking the enslaved enemy (e.g. if no other enemies are nearby), your autofight (tab) is suddenly nonfunctional, along with targeting for spells, wands, etc, *and* you have to say 'yes' to a warning... it's not the end of the world, but it's an annoyance that we generally don't want to inflict on players without a good reason.

**Technical notes**: you've got some commented-out code sitting around (in files.cc), and your second commit seems to have a lot of things going on at once. It might be nice to separate out the new actions, at least.

Is there a reason that enslave is its own action, instead of being one of the xom_good_enchant effects?

Does Xom have dialogue for the new actions?

PR #314, opened 2016-07-21
Fix for: 0010555: Big Wings Weirdness

Skilld18:
Updates the player tile to when the wings mutation is acquired to draw draconian wings


PleasingFungus:
Pushed as bf6fe6030a, thanks!

PR #315, opened 2016-07-28
Replace some magic numbers with new EQ_ enum values

amalloy:
Added enum values:

- `EQ_FIRST_EQUIP = EQ_WEAPON`
- `EQ_FIRST_JEWELLERY = EQ_LEFT_RING`
- `EQ_LAST_JEWELLERY = EQ_RING_AMULET`

Used these enum values to clarify the intent of various instances of iterating subsets of the available equip slots.

I'm not pushing this directly because it involved a number of fiddly little changes that (a) would be easy to get wrong, and (b) are hard to find places to test, so I think it could benefit from someone looking over it before merging in.

### DIFF
--- a/crawl-ref/source/dbg-asrt.cc
+++ b/crawl-ref/source/dbg-asrt.cc
@@ -353,7 +353,7 @@ static void _dump_player(FILE *file)
     fprintf(file, "\n");
 
     fprintf(file, "Equipment:\n");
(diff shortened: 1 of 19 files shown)


amalloy:
Applied as 36c65c9, with some fixes suggested by @PleasingFungus.

PR #319, opened 2016-08-07
Monster creation checklist.

isloat:
This is a /docs/develop/ file meant to assist any future development of
new monsters. There are multiple files that all need to be edited, and
there seems to be no central guideline for which file needs which
changes. Especially for new developers, this should help familiarize
them with the code base.

I invite more experienced developers to edit / expand this file and to
correct any mistakes, misunderstandings, or omissions.


PleasingFungus:
Squashed and pushed.

PR #321, opened 2016-08-08
Dream sheep

isloat:
Dream sheep are magical sheep that spawn in bands in late Lair, about
half as often as elephants. They can also spawn as part of a torpor
snail band, as well as accompanying cyclopes in Shoals.

Dream sheep are only slightly tougher than common sheep, but are more
evasive, have higher magic resistance, and have much higher HD.

Dream sheep gather their shared magical dust to overwhelm the player.
When successful, the target is slept for 5 to 20 turns. Dream dust
checks Clarity, but it does not check MR. (TODO: If this turns out to
be too strong, then a MR check is entirely reasonable to add.) The
strength of the dust effect depends on how many dream sheep are
visible to the player. The player receives 4 to 8 turns of sleep
immunity when a successful sleep is applied, unlike the standard 1
turn of immunity.

As an emergency-flagged spell, dream sheep will directly cast Sleep
instead of relying on dream dust.

On their own, dream sheep are not very dangerous for where they spawn,
even when the player is being put to sleep. When combined with other
monsters, their sleep ability makes slow but hard-hitting monsters
very dangerous. Dream sheep are only the third monster in the game
to have access to Sleep, after Aizul and satyrs.

This implements an idea of PleasingFungus's from a couple years ago
(sheepcursed masses) and, before refactoring and rewriting, used
DracoOmega's _starcursed_scream and _will_starcursed_scream code as
a base for the code for its sleep dust ability.

Sheep have been removed from the game. Sheep in vaults have been
mostly replaced with iguanas, and occasionally with dream sheep.
Cyclopes in Dungeon and Depths no longer spawn with ungulates.


gammafunk:
After testing these, they seem not too much threat as a pack even in the largest band size, but the sleep does give them some extra damage. With other dangerous monsters and bad positioning, the sleep effect is a lot more dangerous. Seems like an improvement over ordinary sheep, which currently exist only for flavor. What I'd propose is replacing ordinary sheep with these, preventing cyclops from generating with a flock in Dungeon, but generating said flock in Shoals.

For the dream effect, I'm not sure coordinating breath is a good idea. Having the player x over each sheep to see the cumulative timeout isn't good, and their ability isn't even themed as a breath. PleasingFungus mentioned that it might be better to use the sleep immunity duration instead.

The summon sheep spell seems like more of a flavorful joke than anything. You could achieve a similar effect by just increasing the band size, and It's pretty bad to so many identical summoners and have to guess which one to target when it creates summons.

PleasingFungus:
Wrt using sleep immunity, you'd probably want to set the immunity duration to the current breath timeout, and then change player::awaken() to only override the immunity duration if it's not yet set. Something like that.

isloat:
I have been away but I will be back at my desktop tomorrow evening and plan to start working on these suggested revisions.

@gammafunk, the summoning is mostly a leftover from the original idea (a monster branching off starcursed masses, which can replicate themselves to eventually ensure a paralysis given enough time) and I don't have any attachment to keeping it. If you think it complicates the monster then the spell is easily cut.

Dream sheep originally had an ENCH_DUSTED keeping track of their cooldown, before ##crawl-dev pointed out that it'd be simpler to just use the already-implemented ENCH_BREATH_WEAPON effect. Functionally it's the same either way. 

The sheep herd coordinates "breath" to ensure that they spend their time mostly productively instead of trying to fire off mini-sleeps that will either be ineffective (as the player is already asleep) or fail (because they can't combine sleep strength to overwhelm the player). If the sheep could fire off sleeps separately I'm concerned that dream sheep may be too overwhelming to deal with, unless resorting to using extensive sleep-immunity durations.

gammafunk:
PF pointed out that the summoning thing was to make individual sheep more dangerous. It might be possible to give them sleep or some other spell as an emergency spell to help with this.

Regarding the sleep immunity and making sheep not waste turns, you could simply have the sheep refuse to perform the sleep action when the immunity is active. This is something they could check instead of the LOS coordination in `_will_sheep_sleep()`. I think that the player having to 1) know that the sleep can only happen when all sleep don't have the breath timer and  2) check all sheep individually using x are bad things. The player's sleep immunity isn't displayed, which I suppose is one thing against that approach, but this works fine for things like paralysis and banishment in practice.

isloat:
I've implemented the suggested changes and updated the PR message to reflect the changes made.

ozyx:
I love the Phillip K. Dick quote... clever. :+1:

PleasingFungus:
Pushed, thanks! :godmode:

PR #322, opened 2016-08-08
Stronger messaging for Dg attribute boost.

isloat:
Distinguish Demigod stat increases at XL%3 from the standard stat
increase for every other race.

This re-opens #318 which was accidentally closed.


PleasingFungus:
Pushed, with a small wording tweak.

PR #323, opened 2016-08-08
Spell Library (second attempt)

shacklettbp:
(r)ead spellbooks to put them in your library where they can be read from the M screen. After (r)eading a spellbook it is destroyed like a scroll. My intent with this patch is to move towards a system where all memorization is done by memorizing spells from your library in the (M) screen, although currently the (M) screen still shows spells from books in your inventory or on the ground that you haven't read yet. You can get to the old book spell list interface through the inventory screen, but I disabled individual spell (M)emorization in that screen (if this isn't desired it is not necessary for the rest of the patch).

The next step would be an autoread feature so books are automatically read on
pickup, and the spells are added to your library, as well as changing starting
classes to start with spells in their library rather than in a book which just has to be read on the first turn.

### DIFF
--- a/crawl-ref/source/describe-spells.cc
+++ b/crawl-ref/source/describe-spells.cc
@@ -554,19 +554,12 @@ string describe_item_spells(const item_def &item)
 void list_spellset(const spellset &spells, const monster_info *mon_owner,
                    const item_def *source_item, formatted_string &initial_desc)
 {
-    const bool can_memorise = source_item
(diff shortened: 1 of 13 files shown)


isloat:
Is this still being worked on? IIRC back in the summer after you submitted this PR there was some discussion about taking a different approach to this. I haven't kept up with this, so let me know.

isloat:
Please resubmit if/when you have time to finish the feature.

PR #325, opened 2016-08-17
Fix for bug#10614

adibis:
I have tested compile. Have tested in wizmode by dropping all items and trying to read/quaff from floor. Have tried to do those actions when valid items were not on floor and it is working as expected.


PleasingFungus:
Rebased and pushed.

PR #326, opened 2016-08-22
Improve mottled draconians (Lightli)

alexjurkiewicz:
Let breath sticky flame have perfect accuracy (can be flavoured as you
just spraying napalm everywhere in front of you), and give +1
conjurations aptitude.


PleasingFungus:
Belatedly: buffing sticky flame seems fine, but the apt buff seems pretty arbitrary and uninteresting. (The spell Sticky Flame having Conjurations as a school does not mean that the Sticky Flame effect has any particular reason to be associated with conjurations, any more than any of the other breaths would.)

MarvinPA suggests buffing sticky flame range to 2-3 instead of giving it auto-hit, so as to avoid making it too similar to the spell.

PR #327, opened 2016-08-22
0010514: Item_slot not working properly

Sandman25DCSS:
https://crawl.develz.org/mantis/view.php?id=10514

### DIFF
--- a/crawl-ref/source/items.cc
+++ b/crawl-ref/source/items.cc
@@ -1904,7 +1904,7 @@ static bool _merge_stackable_item_into_inv(const item_def &it, int quant_got,
                                                     DESC_INVENTORY).c_str(),
                         quant_got);
         }
-
+        auto_assign_item_slot(you.inv[inv_slot]);
         return true;
     }
 


neilmoore:
This causes the function to set inv_slot to the wrong item slot: it would be the slot the item came from rather than the slot it moved to.  That could cause `_got_item` and `_check_note_item` (both called from `_put_item_in_inv`) to affect the wrong item, which might for example cull the wrong thing from your shopping list.

You'd need something like this, but tested:

    if (item_def *newit = auto_assign_item_slot(you.inv[inv_slot]))
        inv_slot = newit->link;

Also, I am mildly concerned that this would cause letters to adjust every time you pick up a stackable item, when really it should only happen when you thereby identify the stack.  If I had identify scrolls auto-assigned to letter `i`, then I manually `#adjust`ed the stack to `r`, I'm not sure I'd want the stack to move back to `i` every time I pick up another scroll.  But I suppose that is debatable.

Sandman25DCSS:
Thank you for taking a look at this, Neil. I see I am not qualified to fix the bug, sorry.

PleasingFungus:
Closing, then, but leaving the associated bug open.

PR #328, opened 2016-08-23
Don't let haste + finesse bring the player below 0.2 attack delay.

GenericPseudonym:
They're not intended to stack, but in the rare case where the player was
running into the limit at 0.2 attack delay (e.g. with a finessed
quickblade), adding haste on top could let them go down to 0.1.

### DIFF
--- a/crawl-ref/source/player-act.cc
+++ b/crawl-ref/source/player-act.cc
@@ -320,7 +320,8 @@ random_var player::attack_delay(const item_def *projectile, bool rescale) const
         // longer so when Haste speeds it up, only Finesse will apply.
         if (you.duration[DUR_HASTE] && rescale)
             attk_delay = rv::max(random_var(haste_mul(2)), div_rand_round(haste_mul(attk_delay), 2));
-        attk_delay = rv::max(random_var(2), div_rand_round(attk_delay, 2));
+        else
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Reproduced the bug in testing, but I'm not familiar with the code in question, so I'll leave this for someone else to review ( @wheals , maybe?)

neilmoore:
This commit doesn't work.  It makes finesse + haste give 4x attack speed, when it's supposed to give only 2x.  Without this **if** (which is supposed to cancel out haste, i.e. makes the attack take longer), we would give 3x attack speed.  And it doesn't even prevent getting speed 0.1 turns (you can test with a mindelay quick blade under haste and finesse).

The problem is that the minimum attack delay of 2 that we set here is later scaled by haste.  So the change should involve setting a different floor.

GenericPseudonym:
Apologies for that. I'm a total newbie and somehow overlooked that!

Unfortunately I haven't tested the commit -- compiling in MSVC 2010 runs into some errors.

neilmoore:
It looks like, when the div_rand_round in haste_mul and the one at the end of this method round in different directions, you can still end up with an occasional 0.1-delay attack.

I'll fix this in a different, and I hope simpler, way.

PR #331, opened 2016-08-26
Make Clarity resist sleep

isloat:
Affects Aizul, satyrs, dream sheep, sleeping needles, hibernation, one
Hell miscast effect, and Dithmenos wrath.

With the removal of "Clarity, the only way for players to gain Clarity
is via Ashenzari (where the buff wouldn't hurt), a mutation (rare!), or
the autumn katana (very rare!), so this seems fine to do.


isloat:
Hmm, the commit message should probably also mention orange crystal plate armour.

PR #332, opened 2016-08-29
Give Yiuf a better cloak

alexjurkiewicz:
Now players can trivially avoid Crazy Yiuf, give them (more of) a reason
to risk banishment.

### DIFF
--- a/crawl-ref/source/mon-gear.cc
+++ b/crawl-ref/source/mon-gear.cc
@@ -1966,8 +1966,11 @@ static void _give_armour(monster* mon, int level, bool spectral_orcs)
         }
         break;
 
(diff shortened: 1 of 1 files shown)


PleasingFungus:
This feels like loot inflation to me. In the early game, 40 XP, a cloak, and a chaos quarterstaff is already enough to tempt me to enter crazy yiuf's little corner of the dungeon pretty often!

MarvinPA:
Agreed, guaranteed plain aux armour is already pretty decent for early dangerous uniques (as with Robin's helmet for example).

alexjurkiewicz:
Fair enough. I feel like I only enter during bouts of suboptimal play.

PR #335, opened 2016-08-29
Sheep no spell-sleep

isloat:
Removes dream sheep emergency spell Sleep.

Moves dream dust from mon-abil.cc into mon-cast.cc.

Removes unused flag SPFLAG_BATTLE.


alexjurkiewicz:
Does this make dream sheep look like they have no spells in x-v? I'd
consider it a problem.

crawl-ref/source/mon-cast.cc
```diff
@@ -4937,6 +4938,80 @@ dice_def resonance_strike_base_damage(const monster &mons)
     return dice_def(3, mons.spell_hd(SPELL_RESONANCE_STRIKE));
 }
 
+static void _sheep_message(int num_sheep, int sleep_pow, actor *foe)
+{
+    string message = "";
```
amalloy: Can't this just be `string message;`?

crawl-ref/source/mon-cast.cc
```diff
@@ -4937,6 +4938,80 @@ dice_def resonance_strike_base_damage(const monster &mons)
...
+
+    // Determine messaging based on sleep strength.
+    if (sleep_pow >= 125)
+        message = "You are overwhelmed by glittering dream dust!";
+    else if (sleep_pow >= 75)
+        message = "The dream sheep are wreathed in dream dust.";
+    else if (sleep_pow >= 25)
+    {
+        message = make_stringf("The%s dream sheep shake%s wool and sparkle%s.",
+                               num_sheep == 1 ? "" : " herd of",
```
amalloy: Using "herd" here leaves you in an awkward position re singular vs plural verbs. Why not just leave it as "the dream sheep" whether singular or plural, as in the other cases?

isloat: Good idea, fixed.

crawl-ref/source/mon-cast.cc
isloat: Fixed.

crawl-ref/source/mon-cast.cc
```diff
@@ -4937,6 +4938,79 @@ dice_def resonance_strike_base_damage(const monster &mons)
...
+    {
+        mprf(MSGCH_MONSTER_SPELL, "%s%s", message.c_str(),
+             sleep_pow ? " You feel drowsy..." : "");
+    }
+}
+
+static void _dream_sheep_sleep(monster& mons, actor& foe)
+{
+    // Shepherd the dream sheep.
+    vector<monster*> dream_herd;
```
amalloy: Why collect a vector of all of them, if all you care about is how many there are? Just start an int at 0 and increment it for each sheep.

crawl-ref/source/mon-cast.cc
```diff
@@ -5156,6 +5230,11 @@ void mons_cast(monster* mons, bolt pbolt, spell_type spell_cast,
         _mons_cast_haunt(mons);
         return;
 
+    // SPELL_SLEEP_GAZE ;)
```
amalloy: What is this comment all about?

isloat: A joke about how it's placed near other gaze spells and works similarly.

crawl-ref/source/mon-cast.cc
isloat: That's a remnant from DracoOmega's _starcursed_scream code. I'll simplify that.

amalloy: Let me know when that's done and I'll merge it in.

isloat: Already squashed the fix into the commit before I saw this.

PR #336, opened 2016-08-31
Reduce sleep damage bonus

isloat:
This PR is just in case dream sheep become a problem. It seems that they've stabilized after the monspell changes and the sleep resistance duration got improved. Probably can be closed in a couple weeks if it's still unneeded.


isloat:
Sleep doesn't need a nerf.

PR #346, opened 2016-09-04
Follow symlinks when searching for fonts

abbradar:
In NixOS several fonts are symlinked into the fonts directory. This patch fixes their discovery.


neilmoore:
For the purposes of building redistributable binaries, it seems like it would be better to have the binary refer to the fonts directory rather than the exact location of the font file, since the latter is likely to change if OS font packages are renamed or merged.

When the original version of the command returned a font symlink instead of its target, what error did that cause?  I wonder if we can fix that and follow the symlink at runtime instead of build time.

abbradar:
It didn't find the font because `-type f` without `-L` is false when the file is a symlink. I think it can be fixed also with `-type f -or -type l`.

abbradar:
I've pushed the new patch and confirmed that it works.

neilmoore:
Thanks!

abbradar:
No problem, thanks for working on the great game ~_^.

PR #348, opened 2016-09-05
Templated random_choose_weighted

amalloy:
To fix https://crawl.develz.org/mantis/view.php?id=10651

Could merge myself of course, but would appreciate review by someone familiar with the perils of variadic templates.


amalloy:
Applied as 1485906a677b694affc0dc5204d97fc6e3a662b8.

PR #349, opened 2016-09-05
Tutorial text fix

catleeball:
The current tutorial text states "Incidentally, you can also cast spells by pressing
Ctrl+leftclick on the monster." This may be a little unclear, since ctrl+clicking won't prompt for which spell to use, but rather cast the spell last used. We may want to update this text.


catleeball:
This related to this bug by the way: https://crawl.develz.org/mantis/view.php?id=10654

PR #350, opened 2016-09-05
Ring of Woe. The companion to Robustness/Shaolin/Shielding.

aolowin:


MarvinPA:
I think the the gauntlets of War already have "plain good +slaying unrand" covered, really.

PleasingFungus:
Yeah, there's no particular need for 'symmetry' here. (Do we also need +8 str rings, +8 dex...?)

PR #351, opened 2016-09-06
Repair xom_monty_hall

isloat:
Restore an ungulate as the prize.

The jellies have been replaced with slime creatures, to maintain intent
of the dream sheep being the 'prize' (as in, the easy fight out of the
three doors). Since the vault is now more dangerous, it has been moved
deeper in the Dungeon.


PleasingFungus:
xom_monty_hall is an overflow altar, meaning that it should be between D:2-7.

PR #352, opened 2016-09-06
Auto-ID manuals the player steps on

isloat:
Having to pick up the manual to see what it'd train was annoying (especially with a full inventory) and
inconsistent with current book behavior.

Since you have to actually reach the tile that the manual resides on, the 'mystery' is still preserved.

### DIFF
--- a/crawl-ref/source/items.cc
+++ b/crawl-ref/source/items.cc
@@ -937,10 +937,10 @@ void item_check()
 /// If the given item is an unknown book, ID it; return whether we did.
 static bool _id_floor_book(item_def &book)
 {
-    if (book.base_type != OBJ_BOOKS || book.sub_type == BOOK_MANUAL)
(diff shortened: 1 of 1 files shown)


PleasingFungus:
What's the messaging like here? Do you get informed about whatever you just stepped on?

isloat:
(xv view) A paperback manual.
(after stepping) You see here a manual of Fighting.

MarvinPA:
As far as I can tell, this works for the first manual in a given game and then stops working for any subsequent manuals (they're considered fully identified already?).

isloat:
I've fixed the bug MarvinPA mentions and squashed it into the original commit. 

As far as I could tell, the bug happened because of line 943, where it checked if books have been fully identified. As far as I can tell there's no static list of manual types, so it saw that we'd identified one manual (out of one) and then returned out of the function. Adding an exception for manuals gets us intended behavior.

PR #353, opened 2016-09-06
Scales tweaks

alexjurkiewicz:
* Remove rough black scales
* Large bone plates give only +SH
* Rugged brown scales give only +hp%


alexjurkiewicz:
Maybe bone plates could give +reflect?

MarvinPA:
I agree with |amethyst's comment in IRC that making rugged brown scales just be +HP makes them overlap too much with robust, the other changes seem reasonable though. I'd maybe suggest removing iridescent over rough black scales (since the extra AC + a drawback makes it maybe a little further away from the AC + resist scales), but either's probably fine.

alexjurkiewicz:
Updated PR to not touch brown scales.

I'm not really a fan of scales that affect multiple things at each level --
it feels a bit fiddly for what should be a relatively simple part of a Ds's
kit. That's why I kept iridescent over black.

PR #355, opened 2016-09-08
Prompt for confirmation before eating mutagenic/forbidden/noxious food (10657)

suncrab:
Issue 10657 is about prompting before eating mutagenic chunks as a Zin worshiper, but it seems like the classes of food characterized by "is_bad_food" should prompt before eating regardless of god.

### DIFF
--- a/crawl-ref/source/food.cc
+++ b/crawl-ref/source/food.cc
@@ -34,6 +34,7 @@
 #include "notes.h"
 #include "options.h"
 #include "output.h"
+#include "prompt.h"
 #include "religion.h"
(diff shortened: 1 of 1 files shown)


PleasingFungus:
This might be better implemented as an auto-inscription, so that players can use rcfile settings to toggle the warnings off.

PleasingFungus:
My comment stands, but some prompt is better than no prompt :) thanks for contributing!

PleasingFungus:
Ah, apparently d09ab47ac29d7 already fixed this. Oops...

PR #358, opened 2016-09-13
Mennas weapon upgrade

alexjurkiewicz:
Always give mennas a holy base weapon type, which makes him the only monster who can possibly spawn with a trishula in-game.


PleasingFungus:
imo, choose one of these commits and stick with it. i like it as Monster Loot, don't really care about trishulas one way or the other.

isloat:
Merged in 2f7c216221523. Per PleasingFungus's suggestions I moved the special case back into data.

PR #359, opened 2016-09-13
Remove giant newts, replace with frilled lizards

isloat:
Not the most earth-shattering of changes, but giant newts and rats were nearly identical, differing only in these things: giant newts swim faster, have 0/15 defenses instead of 1/10, are cold-blooded (for the vanishingly small chance that it would matter), and are of a different genus. Rats serve well enough as the most basic monster, and giant newts were *very frequently* confused with giant geckos*, so it makes sense to target the poor newt.

*: I think Crawl should generally avoid generic adjectives in monster names, or at least avoid reusing adjectives from other monster names.

Frilled lizards have SPELL_SPRINT, very slightly more HP, and slightly different spawn ranges. They move in bursts, but they attack at speed 10, so they shouldn't be deadlier than newts were. Their movement should be noticeable, and teaches new players that certain monsters can have unique movement patterns.

They will need a new tile. I did a palette swap on giant newts (if only to indicate to trunk players that this is a different monster) but it's amateurish and temporary.

Enjoy these adorable videos of frilled lizards!
https://youtu.be/fKMtUOw3Uxw
https://youtu.be/utD66Z5Qb74
https://youtu.be/B41I0yMvEgY


isloat:
Some people have suggested to name them "frilled dragons" (keeping glyph on 'l', of course).

crawl-ref/docs/develop/levels/advanced.txt
```diff
@@ -389,7 +389,7 @@ This example is from sewer.des:
 
 {{
 function sewer_random_monster_list(e)
-  e.set_random_mon_list("bat w:20 / giant newt w:20 / small snake / \
+  e.set_random_mon_list("bat w:20 / giant cockroach w:20 / small snake / \
```
PleasingFungus: this looks like an oversight, not that it matters

crawl-ref/source/dat/des/altar/overflow.des
```diff
@@ -2085,11 +2085,11 @@ TAGS:   overflow_altar_trog
 DEPTH:  D:2-9, Orc, !Orc:$
 : if you.absdepth() < 6 then
 MONS:   orc ; hand axe . animal skin
-MONS:   giant gecko / rat / giant newt / quokka
+MONS:   giant gecko / rat / quokka
```
PleasingFungus: oversight?

crawl-ref/source/dat/des/portals/sewer.des
```diff
@@ -777,7 +777,7 @@ ENDMAP
 NAME:   kennysheep_door_sewer
 TAGS:   no_item_gen no_monster_gen no_pool_fixup
 ORIENT: encompass
-MONS:   kobold w:40 / bat / giant newt / giant cockroach / \
+MONS:   kobold w:40 / bat / giant cockroach / \
```
PleasingFungus: oversight?

crawl-ref/source/enum.h
```diff
@@ -2527,7 +2527,12 @@ enum monster_type                      // menv[].type
 #if TAG_MAJOR_VERSION > 34
     MONS_GIANT_LIZARD,          // genus
 #endif
+#if TAG_MAJOR_VERSION == 34
     MONS_GIANT_NEWT,
+#endif
+#if TAG_MAJOR_VERSION > 34
+    MONS_FRILLED_LIZARD,
+#endif
```
PleasingFungus: this can be simplified (put frilled lizards right after giant lizards, use an #else for giant newts)

PleasingFungus: also, it would probably be better to re-use the enum

PR #360, opened 2016-09-13
Punk rework

isloat:
Punk now applies SPWPN_ACID to its missiles, grants rCorrosion, has had its enchantment level lowered, and has been recolored. 

SPWPN_ACID now functions for weapon brands (but shouldn't ever spawn independently of Punk).


PleasingFungus:
the build breakage is real, not a glitch

isloat:
The problem was in makeitem.cc, when an assert was ran on line 351, it would crash the build because SPWPN_ACID now exists, but couldn't be applied to a weapon.

ACID has been moved to the ranged section, using SPWPN_REAPING as a justification for not special-casing. The FULLDEBUG=1 build passes locally, fingers crossed that Travis validates as well.

isloat:
Merged in b3e3175eb58c

PR #362, opened 2016-09-14
Fewer giants

isloat:
"Giant" as a prefix was used in too many creature names. This PR reduces its frequency, and in doing so, helps further distinguish some Lair creatures.

Giant frogs -> bullfrogs
Giant geckos -> leopard geckos
Giant leeches -> tyrant leeches
Giant cockroaches -> cave roaches
Giant eyeballs -> floating eyes
Giant spores -> ballistomycete spores
Greater naga -> naga exemplars

https://en.wikipedia.org/wiki/American_bullfrog
https://en.wikipedia.org/wiki/Leopard_gecko
http://www.livescience.com/10984-rex-leeches-enormous-teeth.html


MarvinPA:
The frog genus should probably just be MONS_FROG rather than retaining giant frogs? Also I wouldn't necessarily bother mentioning size in the description (as in the new tyrant leech description), "A big foo" is probably already overused in monster descriptions.

isloat:
For record/reference:

All commits except 'cave roaches', 'naga exemplars' and 'shrink monster descriptions' have been merged.

re: naga exemplars, I believe that "Greater" violates the species/class naming theme (cf naga warrior, naga mage, naga sharpshooter, naga ritualist). For any other species "Greater Foo" would sound awkward -- "greater deep elf", "greater ogre", "greater human". 

re: cave roaches, most of the other alternatives were bad. Could just go with simply "cockroach".

nikheizen:
> Greater naga -> naga exemplars
Instead what you should do is rename Hell Beasts to Lesser Hell Beasts.

isloat:
Closing, 'giant cockroaches' have no good alternative names, 'claymore' should probably have more discussion, and 'naga exemplar' isn't good enough.

PR #364, opened 2016-09-17
Long blade viability tweaks

drke2:
In practice, riposte offers similar utility to the other weapon effects (cleaving, reaching) with the caveat that the effectiveness of it is dependent on your EV value. So -1 dam was fairly appropriate, and brings the common 1h swords (falchion, long sword, scimitar) in line with hand axe and war axe, except there's another tier in the middle.

-2 dam feels like overkill for all but triple sword, and has the awkward result of double swords being strictly better than great swords, so great sword gets +1 dam and +1 delay.

Scimitar having the worst accuracy of all 1h swords seems odd, as it breaks the existing progression of higher damage -> lower accuracy that exists in most weapon types.

The high holy wrath weighting isn't really desirable aside from the flavor of TSO holy swords, and it's still common enough to represent the brand's association with the weapon type. Lowering it improves the quality of random long swords, scimitars, etc. that are used in the early game.


isloat:
This looks good, though I'm not sure I agree that triple sword need their delay lowered.

PleasingFungus:
a number of people have complained about double swords being strictly better than great swords, but i don't think that's actually a problem - we already have many other 'strictly better' weapon upgrades, even within swords themselves. 

no opinion otherwise.

drke2:
Updated PR to not touch triple swords as was discussed in -dev.

drke2:
Set GS at 15 dam/17 delay.

isloat:
I've merged bfa84d5. The brand redistribution was controversial enough in -dev that I don't want to touch it until additional consensus can be made.

MarvinPA:
Closing this since the great sword change has been merged, and as discussed in -dev it's not really desirable to buff long blades across the board again by adjusting the brand weighting.

PR #365, opened 2016-09-18
Make Zin hate mutagenic chunks properly.

m-aigner:
Commit message says it all. Fixes the following mantis issue: https://crawl.develz.org/mantis/view.php?id=10657 .


neilmoore:
I would suggest making that check use  'forbidden.*chunk', or better yet merging it with the previous check, 'forbidden.*(chunk|corpse)'.   Otherwise you are applying it to non-food items as well, which overrides the darkgrey colour for useless god-hated items (like potions of haste for mummies of cheibriados: currently darkgrey, red with your change to food_colouring.txt).

isloat:
Merged in d09ab47ac29d7, thank you!

PR #366, opened 2016-09-19
Don't display extra message for known scroll of acquirement

Sandman25DCSS:
It's weird to see exciting "This is a scroll of acquirement!" when you already know what you are reading.

Read which item? (? for menu, Esc to quit)
As you read the scroll of acquirement, it crumbles to dust.
This is a scroll of acquirement!
[a] Weapon [b] Armour        [c] Jewellery [d] Book [e] Staff  
[f] Wand   [g] Miscellaneous [h] Food      [i] Gold [j] Ammunition

### DIFF
--- a/crawl-ref/source/item_use.cc
+++ b/crawl-ref/source/item_use.cc
@@ -2746,7 +2746,10 @@ void read_scroll(item_def& scroll)
         break;
 
     case SCR_ACQUIREMENT:
-        mpr("This is a scroll of acquirement!");
+        if (!alreadyknown)
+        {
+            mpr("This is a scroll of acquirement!");
+        }
         // included in default force_more_message
         // Identify it early in case the player checks the '\' screen.
         set_ident_type(scroll, true);


PleasingFungus:
Shouldn't there still be some prompt? If not "this is a scroll of acquirement", then "Acquire:", something like that.

Sandman25DCSS:
Should I add "else" with "Acquire what?" prompt?

jdeeny:
I feel like "Acquire which type of item?" would be a nice prompt. I think it should happen even if the scroll is previously unknown, not inside an else clause. As in:

This is a scroll of acquirement!
Acquire which type of item?
[a] Weapon [b] Armour [c] Jewellery [d] Book [e] Staff
[f] Wand [g] Miscellaneous [h] Food [i] Gold [j] Ammunition

PleasingFungus:
@Sandman25DCSS Either that or @jdeeny's suggestion would be fine.

Sandman25DCSS:
mpr("Acquire which type of item?");

;)

PleasingFungus:
Apologies; my sep. 29th comment was wrong. This patch would've been fine. Very sorry for the delay!

Sandman25DCSS:
No problem. Thank you for merging it in!

PR #367, opened 2016-09-22
Exclude "MacOSX.sdk" from the SDK search for Mac OS X builds

aegolden:
In Sierra with Xcode 8 installed an SDK "MacOSX.sdk" with no version
number appears in the list of installed SDKs and this causes the
SDK version to come out empty.  We should exclude this special case
from the list of SDKs to consider so we find only the versioned
SDKs.  This change gets crawl building again on Mac OS X using, e.g.
"make APPLE_GCC=y NO_PKGCONFIG=y CONTRIB_SDL=y TILES=y".


PleasingFungus:
Pushed, thanks!

PR #373, opened 2016-09-27
Replace get_menu_colour_prefix_tags with menu_colour_item_name

jdeeny:
I felt like this name better described what the function is being used for.

### DIFF
--- a/crawl-ref/source/itemname.cc
+++ b/crawl-ref/source/itemname.cc
@@ -3967,12 +3967,11 @@ string item_prefix(const item_def &item, bool temp)
 /**
  * Return an item's name surrounded by colour tags, using menu colouring
  *
- * @param item_def The item being queried
- * @param desc The description level to use for the name string.
- * @return A string containing the item's name surrounded by colour tags.
+ * @param item The item being queried
+ * @param desc The description level to use for the name string
+ * @return A string containing the item's name surrounded by colour tags
  */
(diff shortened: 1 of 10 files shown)


crawl-ref/source/itemname.cc
```diff
@@ -3964,7 +3964,14 @@ string item_prefix(const item_def &item, bool temp)
...
-string get_menu_colour_prefix_tags(const item_def &item,
+/**
+ * Return an item's name surrounded by colour tags, using menu colouring
+ *
+ * @param item_def The item being queried
+ * @param desc The description level to use for the name string.
+ * @return A string containing the item's name surrounded by colour tags.
+ */
+string menu_colour_item_name(const item_def &item,
                                    description_level_type desc)
```
wheals: Second line is no longer aligned

crawl-ref/source/itemname.cc
```diff
@@ -3964,7 +3964,14 @@ string item_prefix(const item_def &item, bool temp)
     return result;
 }
 
-string get_menu_colour_prefix_tags(const item_def &item,
+/**
+ * Return an item's name surrounded by colour tags, using menu colouring
+ *
+ * @param item_def The item being queried
```
wheals: Should just be `item`

jdeeny: Do you mean the name of the parameter should be item?

PR #376, opened 2016-09-30
Changes to human aptitudes

3FVCE:
https://crawl.develz.org/tavern/viewtopic.php?f=8&t=21550&view=unread&sid=1ec1db73ca8a06314be6e3b54a21ebae#unread
Suggestion from this post to make humans +0 on all aptitudes


MarvinPA:
I don't think aesthetics alone is a good reason for this, and the current state indicates that the vague "baseline" for spellcasting/stealth/invocations does differ from other skills.

PR #379, opened 2016-10-05
Tell player if monster can never be recited to

m-aigner:
Fixes mantis issue 10416.

I can't help thinking that there's a better way to do this than declaring a vector I never use, but writing a whole new function to make the one-line change seems excessive when this works.

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -3623,18 +3623,24 @@ void get_monster_db_desc(const monster_info& mi, describe_info &inf,
 
     if (in_good_standing(GOD_ZIN, 0))
     {
-        const int check = mi.hd - zin_recite_power();
-        if (check >= 0)
-            inf.body << It << " is too strong to be recited to.";
-        else if (check >= -5)
-            inf.body << It << " may be too strong to be recited to.";
+        recite_counts retval;
+        monster *m = monster_at(mi.pos);
+        if (zin_check_recite_to_single_monster(m, retval) == RE_INELIGIBLE)
+            inf.body << "There is no point in reciting to " << it <<".";
         else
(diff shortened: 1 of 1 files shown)


MarvinPA:
This causes crashes when using monster lookup (?/M), since it tries to check the recite eligibility of a monster that doesn't exist. I guess it could just skip the recite line in that case, although it also seems feasible that this approach could lead to information leaks of some sort? Possibly it'd avoid that since it's not giving you the full breakdown of which type of recite would affect it (which could vary in monsters with multiple spellsets).

m-aigner:
There's nothing easier than making it skip the line in monster descriptions. (It would be enough to turn the "if (...)" I added into "if (m && ...)", as far as I can tell.) If this is what you prefer, I can make the change in a minute.

I suspect the other approach is better, though: I can see a player trying ?/M to check if they can recite to shadow dragons before entering vaults, or to death oozes before entering slime, and I think the game should tell them. If you agree that this is desired, I can do that too, though it requires more changes. (Specifically: the current zin_check_recite_single_monster takes a monster and can't really be made to take a monster_info, so I'd need to write a can_be_recited_to(monster_info) function, which from a brief look into the code might entail turning one or two methods of monster into methods of monster_info. It all looks perfectly possible, it's just slightly more complicated than the first option.)

MarvinPA:
Adding things to monster_info if necessary seems like probably the better way to do it, yeah (or at least, that's what I'd looked into in the past and put off repeatedly on because it seemed like too much effort!).

isloat:
I agree with MPA that adding this to monster_info is good. 

Can you fix the crash that he pointed out?

rawlins:
I merged this code (with some modifications) in e769cdc, thanks for putting together the PR!

PR #380, opened 2016-10-05
Add article to Sojobo's description

m-aigner:
Mantis #10712, though you could also close that with "Working as intended", as I'm not sure it reads better with an article.


PleasingFungus:
After reading this about ten times, I think I agree that it reads better without the article. Sorry for the delay!

PR #385, opened 2016-10-19
Put a symlink to init.txt in a sensible place for Mac OS X users

aegolden:
Mac OS X users will never expect to find init.txt inside the .app
bundle.  The standard place for such files on Mac OS X is in
~/Library/Application Support/<app name>, so on startup we should
put a symbolic link to whatever init.txt file we're going to use
in ~/Library/Application Support/Dungeon Crawl Stone Soup (or
wherever we're putting morgue and saves anyway).

Note that symlink will not overwrite an existing file.

### DIFF
--- a/crawl-ref/source/initfile.cc
+++ b/crawl-ref/source/initfile.cc
@@ -76,6 +76,9 @@
 #include <shlobj.h>
 #elif defined (__APPLE__)
 extern char **NXArgv;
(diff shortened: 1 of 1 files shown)


aegolden:
Force pushed replacement commit to address @neilmoore's observation that the original patch behaved incorrectly in the case where crawl_rc was already an absolute path.

PR #388, opened 2016-11-02
Maxwell's thermic engine, a flaming and freezing sword.

aolowin:


beeceropteryx:
hilarious, but if you do the math it's basically a branded greatsword that (a) has a low skill requirement and (b) lets you use a shield at the cost of hurting your resistances (so basically a wash).

PleasingFungus:
I'm not sure that this adds enough interesting decisions/gameplay to justify the added code (and added messages!). The theme is cool, but it's basically trading resists for damage in a very elaborate way?

aolowin:
The code can be refactored to share the same functions as the elemental staff so there's not much overhead there. For a different take it could toggle brands with an evocable effect - starts as flaming, evoke refrigeration, becomes freezing, evoke ring of flames, becomes flaming again. Loses a little of the Maxwell's demon flavour that way.

isloat:
I'm not a fan of this summoning demons; I think the flame/frost flavor was better.

aolowin:
It's still flame/frost, just with an added evocable effect.

isloat:
I've played around a bit with this locally and it plays much better than it looks on paper. The brand switching is cool and the demon summoning ties together the Maxwell's Demon reference you're making here. 

I'm more favorable towards this PR than I was previously. Any other devs have opinions?

isloat:
I personally really enjoy this idea, but other devs have expressed reservations.

1. Demonic horde of weak demons is probably not a very good ability. Try making it summon only sun demons / ice devils.
2. If this gets added, it's going to likely replace the Dagger of Chilly Death / Scimitar of Flaming Death. It might be simpler to, instead of using the mechanic you have here for damage type selection, to just merge these two unrands' effects into this item: randomly casts sticky flame or flash freeze on the victim.
3. Why isn't this a double sword! I get the demon blade connection to Maxwell's demon... but this is a great opportunity to have an unrand double sword.

isloat:
CanOfWorms has whipped up spiffy sprites for the Thermic Engine:
http://i.imgur.com/VrUNGQQ.png 
http://i.imgur.com/xCNTqkD.png

(This can be easily altered if it gets changed into a double sword; don't worry on that front.)

aolowin:
It's now a double sword with variable pluses. The demon summoning's been removed. I played around with the sticky flame/flash freeze idea but it just seemed like it added more message lines with little effect. My hope with the increasing pluses is to encourage staying in a fight longer since if you disengage they decay rapidly.

isloat:
Seems much simpler than before (a good thing, especially given the special code this requires), and taking 12 turns to fully power up is an interesting effect that isn't seen in many other parts of the game. 

I'd suggest making it power down more rapidly, though. 33% chance to lose a plus on each turn means it takes 36 turns of not fighting to fully power down, on average. That's a lot of time! (Well, assuming I'm interpreting this code correctly.)

PR #390, opened 2016-11-04
Add the orb status light whenever in Zot

alexjurkiewicz:


PleasingFungus:
Isn't this a bit annoying for autotravel?

MarvinPA:
The message colour entry there doesn't actually set a colour, it looks like (ideally the message would use MSGCH_ORB rather than having an entry in messages.txt I think, but that's not so easily supported in the context of branch entry messages).

Issue #389, opened 2016-11-03
WebTiles. Show unvisited areas on minimap.


AndSDev:
This path allow to show unvisited areas on minimap (WebTiles).

**Main idea:**
On the minimap simply define unvisited area - if black areas are not surrounded by walls. I propose highlight these border areas.

**Standart minimap:**
![image](https://cloud.githubusercontent.com/assets/13585370/19964416/a38fc596-a1d2-11e6-88a6-69c97c261529.png)
**New minimap:**
![image](https://cloud.githubusercontent.com/assets/13585370/19964269/e3ef297a-a1d1-11e6-982d-b0410cb4e7e5.png)

**Patch** (all changes in minimap.js):
```diff
diff --git a/crawl-ref/source/webserver/game_data/static/minimap.js b/crawl-ref/source/webserver/game_data/static/minimap.js
index 3313cca..eae6f4c 100644
--- a/crawl-ref/source/webserver/game_data/static/minimap.js
+++ b/crawl-ref/source/webserver/game_data/static/minimap.js
@@ -162,6 +162,25 @@ function ($, map_knowledge, dungeon_renderer, view_data,
     function update(x, y, cell)
     {
         cell = cell || map_knowledge.get(x, y);
+
+        if (cell.mf && cell.mf != enums.MF_WALL && cell.mf != enums.MF_MAP_WALL)
+        {
+            for (var i = -1; i <= 1; i++)
+            {
+                for (var j = -1; j <= 1; j++)
+                {
+                    if (i != 0 || j != 0)
+                    {
+                        var c = map_knowledge.get(x + i, y + j);
+                        if (!c.mf)
+                        {
+                            set(x + i, y + j, minimap_colours[enums.MF_MAP_FLOOR]);
+                        }
+                    }
+                }
+            }
+        }
+
         if (x == player.pos.x && y == player.pos.y)
             set(x, y, minimap_colours[enums.MF_PLAYER]);
         else
```

alexjurkiewicz:
I like it! Will really help manual explorers clear out the last of a level

isloat:
This should be a PR instead of an issue -- but great job on this patch! I like it, waiting to hear what other devs think.

AndSDev:
Added PR.

> Used MF_MAP_WALL color instead MF_MAP_FLOOR (more contrast).
> 
> Original minimap:
[...]

PR #392, opened 2016-11-10
WebTiles. Show unvisited areas on minimap. Fix #389.

AndSDev:
Used MF_MAP_WALL color instead MF_MAP_FLOOR (more contrast).

Original minimap:
image
MF_MAP_FLOOR color:
image
MF_MAP_WALL color:
image

PR #394, opened 2016-11-11
Tiles. Show unvisited areas on minimap.

AndSDev:
This path allow to show unvisited areas on minimap (in Tiles-version, as in #389 for WebTiles).

Standart minimap:
image
New minimap:
image

**P.S.** May be problems in the Labyrinth or in the Abyss (not tested).


isloat:
Are the issues that we discussed in previous #389 and #392 fixed now? 

(By the way, feel free to close your own PRs if you're making a new PR instead. I'm leaving #392 open for now, in case there's something you want to do with it further.)

PR #395, opened 2016-11-12
Removed healing as possible chaos effect

giann:
Fix issue https://crawl.develz.org/mantis/view.php?id=10762


MarvinPA:
I think a better fix would probably just to remove healing as a possible chaos effect, as is done with wands of random effects - it's not all that interesting, and the special-casing there is pretty weird for an already-complex effect.

giann:
I'll go ahead and do that

PR #396, opened 2016-11-15
Change slow regeneration to a two level mutation

alexjurkiewicz:
Remove the first level (slow regeneration near monsters) and rename to
inhibited regeneration.

Ghouls used to get slow regeneration 1, now they receive inhibited
regeneration 1, which is worse. :skull:


PleasingFungus:
This causes crashes if a deep dwarf (or anyone else with slow regen 3) loads the game and presses A. (You need save compat.)

alexjurkiewicz:
I've tested loading a DD and the conversation appears to happen properly, I think this is good to go!

isloat:
I like getting rid of a mutation that was almost always unnoticeable, but I'm slightly concerned about the impact on ghouls and this making Malmutation slightly scarier. Maybe this is okay? Ghouls are still a strong race, and Malmutation *should* be scary.

alexjurkiewicz:
Ghouls have single turn chunk eating as of several months ago, and they were already pretty strong, so I think this change is fine for their overall power level.

Re: mutation getting scarier, possibly. It would certainly be insta-cure-mut-quaff. But the chance of it happening must be quite low.

If it's a real concern, we could split the mutation into two separate single level mutations. "No Regeneration" could become a DD-specific mut with 0 random weight, while "Inhibited Regeneration" stays as-is.

isloat:
That does work to solve the problem but you'd have to make them mutually exclusive, wouldn't you?

alexjurkiewicz:
Sure, that's possible. Do you think such a solution is over engineered
though? I don't feel this change is going to significantly affect the
optimal curemut strategy

alexjurkiewicz:
Is anyone keen on merging this? I'll update the PR if so.

PleasingFungus:
I still like 2-level slow regen; someone was concerned about slow regen 3
(or the new 2) being way too nasty as a random mut, tho. Would it make
sense to have some new mut level between current slow 2/new 1 and current
3/new 2?

PR #397, opened 2016-11-15
Simplify base UC damage for forms

alexjurkiewicz:
Blade hands, statue form and dragon form all had base unarmed damage
that scaled with strength (and dexterity, for blade hands). Remove this
scaling and replace it with flat damage.

Naive damage comparison, using the average stats of winning transmuters
at various XLs:
* XL10 (17str, 20dex)
* XL15 (22str, 24dex)
* XL20 (25str, 25dex)
* XL27 (28str, 28dex)

Blade Hands (10% more at XL10, 20% less at XL27):
* XL10: Old: 20, New: 22
* XL15: Old: 23, New: 22
* XL20: Old: 25, New: 22
* XL27: Old: 27, New: 22

Statue Form (same at XL10, 20% less at XL27):
* XL10: Old: 12, New: 12
* XL15: Old: 13, New: 12
* XL20: Old: 14, New: 12
* XL27: Old: 15, New: 12

Dragon Form (40% more at XL10, same at XL27):
* XL10: Old: 23, New: 32
* XL15: Old: 27, New: 32
* XL20: Old: 29, New: 32
* XL27: Old: 31, New: 32

### DIFF
--- a/crawl-ref/source/dat/descript/spells.txt
+++ b/crawl-ref/source/dat/descript/spells.txt
@@ -128,7 +128,7 @@ deplete magical power.
 Blade Hands spell
 
 Causes long, scythe-shaped blades to grow from the caster's hands, increasing
-melee damage to a degree depending on strength and dexterity.
+melee damage significantly.
 
 While transformed, any equipped weapons are melded. The caster also becomes
 unable to evoke wands, and casting spells becomes somewhat more difficult.
@@ -495,8 +495,8 @@ Dragon Form spell
     desc = desc .. "."
 
     return desc
-}} The caster's melee damage will be substantially increased to a degree
-depending on strength, and they become much more robust but less evasive.
+}} The caster will find this form offers incredible melee damage and greatly
+increased robustness (though reduced evasiveness).
 
 While transformed, any equipped weapons and armour are melded.
 %%%%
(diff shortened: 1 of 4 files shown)


isloat:
Why is it beneficial to have flat damage over scaling damage?

alexjurkiewicz:
I think there's a benefit to having forms have consistent base damage calculations. Either all should use str/dev damage scaling or none should. Since other weapons (eg all other forms, uc, and all weapons) have flat base damage, it seemed least surprising to give all forms a flat base damage too.

isloat:
I don't play Transmuters much. I assume stats still would have a bonus effect on the damage done?

alexjurkiewicz:
Yes, all melee weapons have their base damage multiplied by strength,
including UC. It's just stats modify some form UC base dmg too.

isloat:
Sounds good then! Will wait on other devs seconding/thirding this before I push, as it's a somewhat significant change.

frozenca:
It seems really huge nerf to Chei + Statue Form.

MarvinPA:
Flat base damage for forms seems reasonable to me (or also good is damage scaling with spellpower, as with Hydra Form) - removing Blade Hands scaling with dexterity is particularly good, since now that strength/dex weighting has been removed it's the only type of 

Implementation-wise, this should just change the base_unarmed_damage entries in form-data.h and remove the overrides in transform.cc (and should also update the descriptions in spells.txt and status.txt, although to be fair I think the original PR predates those being added).

alexjurkiewicz:
Updated as per @MarvinPA.

I tried to keep the numbers somewhat similar to their current range so there won't be much difference for the average player. However if changes were desired I think:
* Dragon form should really do more damage than a max skilled troll (eg 36)
* Blade hands could do less damage in exchange for a level of robust

PR #399, opened 2016-11-18
Improve Xom form spell casting

alexjurkiewicz:
Firstly, remove Xom's ability to cast spells at 0 tension. The spell
list was summon butterflies + forms, none of which were very interesting
(or fun) to cast without tension.

Secondly, change Xom to also grant the player bonus skill in unarmed
combat when transforming them into a pure UC form (so all forms except
beastly appendage, statue and lichform). The bonus is equal to player's
XL.


MarvinPA:
I'm not a big fan of Xom granting UC skill - feels like it'd maybe be better just to have those forms be a bunch rarer.

alexjurkiewicz:
I feel like Xom's action list is getting smaller these days though :(

alexjurkiewicz:
Closing this, I don't think there was much support. I think without a change like this Xom shouldn't transform the player into pure UC forms, since 90% of players will just run away.

PR #400, opened 2016-11-18
Makefile cleanup mac osx

aegolden:
After the discussion on ##crawl-dev about the process for building the Mac OS X packages I spent some time studying the Makefile and found that some of the complexity around Mac OS X builds could be reduced.  In particular:

- We don't need to find/use the earliest available Mac OS X SDK when building.
- Some of the environment variables recommended by INSTALL.txt are not necessary.

### DIFF
--- a/crawl-ref/INSTALL.txt
+++ b/crawl-ref/INSTALL.txt
@@ -204,17 +204,13 @@ installed through the AppStore as a free download; the associated command line
 tools can then be installed by opening Xcode, opening the Preferences dialog,
 and clicking Install on the Command Line Tools line in the Download tab.
 
-Mac builds then use the Unix build process described above, but require you to
-add 'APPLE_GCC=y' to the 'make' command-line. In addition, to build the
-graphical version of Crawl, you must add 'NO_PKGCONFIG=y' and 'CONTRIB_SDL=y'.
-
-So, to build build Crawl on a Mac you could use the following example command
-lines.
+Mac builds then use the Unix build process described above. To build Crawl on
+a Mac you could use the following example command lines.
 
 For the console version:
-  make APPLE_GCC=y
+  make
 For the graphical version:
-  make APPLE_GCC=y NO_PKGCONFIG=y CONTRIB_SDL=y TILES=y
+  make TILES=y
 
 
 Building on Windows (MSYS2)
--- a/crawl-ref/source/Makefile
+++ b/crawl-ref/source/Makefile
@@ -407,79 +407,13 @@ INCLUDES_L += -Iutil -I.
 
 ifdef APPLE_GCC
 
-MARCH := $(uname_M)
-
-
-ifndef NO_AUTO_SDK
-
-# The SDK locations were moved in 10.8; this snippet tries to find them
-# there first, then reverts to the original location.
-XCODE_SDK_SUFFIX := Platforms/MacOSX.platform/Developer
-# Try to use xcode-select -p to find the current user-configured Xcode.app.
-XCODE_SELECT_PATH := $(shell xcode-select -p 2>/dev/null | sed 's/\/$$//')
-ifeq ($(shell test -e "$(XCODE_SELECT_PATH)/$(XCODE_SDK_SUFFIX)" || echo NOPE),)
-DEVELOPER_PATH := $(XCODE_SELECT_PATH)/$(XCODE_SDK_SUFFIX)
-else
-# Otherwise, try /Applications/Xcode.app.
-XCODE_DEFAULT_PATH := /Applications/Xcode.app/Contents/Developer
-ifeq ($(shell test -e "$(XCODE_DEFAULT_PATH)/$(XCODE_SDK_SUFFIX)" || echo NOPE),)
-DEVELOPER_PATH := $(XCODE_DEFAULT_PATH)/$(XCODE_SDK_SUFFIX)
-else
-# If all else fails, maybe it's in /Developer.
-DEVELOPER_PATH := /Developer
-endif
-endif
-
-# Find the oldest SDK available, in attempt to make this build as
-# backward-compatible as we possibly can.
-SDK_VER := $(shell ls $(DEVELOPER_PATH)/SDKs | grep -v "MacOSX.sdk" | sort -n | head -1 | perl -pe 's/^MacOSX//g;s/.sdk$$//g')
-
-ifeq ($(SDK_VER),10.4u)
-SDK_VER := 10.4
-endif
-
-ifndef SDK_VER
-$(error You do not seem to have any Mac OS X SDKs installed! This build is doomed to fail)
-endif
-
-endif
-
-ifndef SDK_VER
-ifeq ($(MARCH),ppc)
-SDK_VER := 10.4
-endif
-ifeq ($(MARCH),i386)
-SDK_VER := 10.4
-endif
-ifeq ($(MARCH),x86_64)
-ifdef TILES
-SDK_VER := 10.6
-else
-SDK_VER := 10.5
-endif
-endif
-endif
-
-# Mac OS X 10.4 adds a 'u' on the end of the SDK name. Everything
-# else is much easier to predict the name of.
-ifeq ($(SDK_VER),10.4)
-GCC_VER := 4.0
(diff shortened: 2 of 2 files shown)


aegolden:
@geekosaur says that even though Crawl will build with that latest (unversioned) SDK in Xcode 8.1, if you pass -mmacosx-version-min=10.9, for example, there will be runtime errors on earlier systems do to Apple bugs.  So possibly 1c70fb4 is no good.

I think the other commits are OK since the Makefile itself sets APPLE_GCC and NO_PKGCONFIG when it detects Darwin.

geekosaur:
@aegolden, the assumption made by crawl builds is incorrect for Xcode 8 because it does not version its SDK; this is largely Xcode's fault. And, along with Apple deciding that said unversioned SDK uses `clock_gettime` even when installed and building on 10.11, despite `clock_gettime` only being present in 10.12, it suggests that Apple has stopped supporting *any* backward compatibility. We can't do anything about that. We also haven't ever claimed you could use a newer Xcode to build for an older one, because Apple has always set the restrictions there (usually one past version is supported and others will fail, but Xcode 8 only supports the current OS version SDK even when running on an older OS).

Simply removing the SDK code means you are following Apple's lead and desupporting every Xcode version prior to 8.0, and therefore every version of Mac OS prior to 10.12 (because even Xcode 8 for 10.11 creates programs that will only run on 10.12).

`APPLE_GCC` has been defunct since 10.7, or more precisely Xcode 4, and should not be used at all. From that point on, Apple has not shipped gcc, it has shipped a rebadged clang. For example, on my 10.8 VM: 
```
Brandons-Mac-2:~ allbery$ gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 4.2 (clang-425.0.28) (based on LLVM 3.2svn)
Target: x86_64-apple-darwin12.6.0
Thread model: posix
Brandons-Mac-2:~ allbery$ xcodebuild -version
Xcode 4.6.3
Build version 4H1503
```
And it's actively harmful to use `APPLE_GCC=y` on Xcode 4 and later because it disables optimization to work around bugs in gcc 4.2 that are not present in any Apple-shipped version of clang.

geekosaur:
Additional note: with an unmodified OS X, crawl will not build on pre-10.7 because of the C++11 requirement. It is however still possible to create a C++11-capable ecosystem on older OS X versions, for example via https://trac.macports.org/wiki/LibcxxOnOlderSystems (Homebrew may have something similar). It might be nice to not rule that path out because we "know" only 10.7 and up support C++11.

aegolden:
> Simply removing the SDK code means you are following Apple's lead and desupporting every Xcode version prior to 8.0, and therefore every version of Mac OS prior to 10.12 (because even Xcode 8 for 10.11 creates programs that will only run on 10.12).

This isn't exactly true.  It's certainly possible to build 10.7 and 10.8 compatible apps using Xcode 8.  "Normal" mac app targets typically get a Deployment Target in their configuration that's allowed to be less than the SDK used for the build.  If there's some specific problem with Crawl (or an API Crawl uses, which Apple borked somehow) then maybe we can't target earlier versions while building with the latest SDK, but it should in principle be possible to do so.

The binary at http://aaronegolden.com/DCSS.zip was built with the 10.12 SDK but with -mmacosx-version-min=10.10, and a couple folks on ##crawl-dev were successfully running it the other night on 10.11.  I'm not sure how much wrangling it takes to trip the clock_gettime thing.

geekosaur:
On Sun, Nov 20, 2016 at 10:42 PM, Aaron Golden <notifications@github.com>
wrote:

> with -mmacosx-version-min=10.10


That doesn't do what you think. It builds with the current SDK, but changes
what the various compatibility macros think the system is; the result will
work if the resulting binary doesn't use load commands the older system
doesn't have (this is safe for 10.11 and possibly 10.10 but will fail on
earlier versions), and if there aren't any other changes not captured by
Apple's availability macros. It's likely to break with autoconf, for
example, since that doesn't go through the availability macros to see what
symbols are in what libraries. (clock_gettime might just manage to not be a
problem for crawl because it's not using autoconf. 10.9 and earlier will
almost certainly break SDL2.)

In general it is not as safe as forcing the appropriate SDK, which is why
most applications that support building for older OS X versions require /
expect you to use `MACOSX_DEPLOYMENT_TARGET instead of just
`-mmacosx-version-min`.

-- 
brandon s allbery kf8nh                               sine nomine associates
allbery.b@gmail.com                                  ballbery@sinenomine.net
unix, openafs, kerberos, infrastructure, xmonad        http://sinenomine.net

geekosaur:
Also, I'm, sorry to say, that particular confusion tends to indicate someone who hasn't learned (usually the hard way) just how much of a mess the Apple ecosystem is with respect to backward compatibility. The fact that it *almost* works doesn't help there.

alexjurkiewicz:
How old a version is it important to support? Based on the steam hardware
survey, 10.10 is about as old as you need to go:
http://store.steampowered.com/hwsurvey/

geekosaur:
On Sun, Nov 20, 2016 at 10:59 PM, Alex Jurkiewicz <notifications@github.com>
wrote:

> How old a version is it important to support? Based on the steam hardware
> survey, 10.10 is about as old as you need to go:
>

We used to support back to 10.4. That was raised to 10.7 when crawl went
C++11, because building and packaging a C++11 ecosystem to go with it is
prohibitively complex. Raising it past 10.9 would make a number of things
easier (notably, I could build and sign on the same system; also, would no
longer need the extra hackery for C++11 support on 10.7/10.8), but I am not
the one to make that call.

-- 
brandon s allbery kf8nh                               sine nomine associates
allbery.b@gmail.com                                  ballbery@sinenomine.net
unix, openafs, kerberos, infrastructure, xmonad        http://sinenomine.net

aegolden:
I know this an old post, but it appears to indicate that -mmacosx-version-min and MACOSX_DEPLOYMENT_TARGET are synonymous:  http://stackoverflow.com/questions/25352389/difference-between-macosx-deployment-target-and-mmacosx-version-min-compiler-op

Additionally, I've maintained a couple of apps on the app store for the last few years that have deployment targets of 10.7 and 10.8, and I've always built them with the latest SDK and haven't had any trouble.  Once I *did* accidentally use an API that was not available yet in 10.7 and I got a support request and was able to fix it… but in *general* it hasn't been a problem to build with the latest SDK and deploy to older versions.

geekosaur:
On Sun, Nov 20, 2016 at 11:07 PM, Aaron Golden <notifications@github.com>
wrote:

> I know this an old post, but it appears to indicate that
> -mmacosx-version-min and MACOSX_DEPLOYMENT_TARGET are synonymous:
> http://stackoverflow.com/questions/25352389/difference-
> between-macosx-deployment-target-and-mmacosx-version-min-compiler-op
>
Grrr, I have confused that env with `SDKROOT`. (and will probably have to
edit things and rebuild to correct that, and now I wonder if that is part
of why the last version was sometimes blowing up --- this also has
implications with respect to C++11 support. why does apple have to make
things so complicated...)


-- 
brandon s allbery kf8nh                               sine nomine associates
allbery.b@gmail.com                                  ballbery@sinenomine.net
unix, openafs, kerberos, infrastructure, xmonad        http://sinenomine.net

aegolden:
Here's a build done with the proposed changes, off of trunk, and with an extra change to *definitely* call clock_gettime in main.  If this one works on folks various older systems then I think Apple must have fixed the clock_gettime thing.  If it crashes on launch then… well clock_gettime is still an issue.

If you run it from the command line, like by cd'ing into the package contents and running the binary in the MacOS directory directly you'll see a log about clock get time on launch.

 http://aaronegolden.com/DCSS-DefinitelyCallsClockGetTime.zip

aegolden:
OK, so now I've confirmed that no calls to clock_gettime are actually compiled into crawl for Mac.  In fact the 0.19 and 0.19.1 Mac tiles and console binaries on http://crawl.develz.org right now were built using these changes.  If we don't hear from users saying that there are issues running on older macs then I'm thinking this is probably a pretty safe change after all.

PR #402, opened 2016-11-18
Deep elf knight spellbook unification

alexjurkiewicz:


MarvinPA:
This actually just removes all spells from deep elf knights (their entry in mon-data.h needs updating too).

isloat:
Closing since this is ancient, broken, and also since I plan on implementing another approach for deep elf knight spells.

PR #405, opened 2016-11-21
Buff !cancellation (Yermak)

alexjurkiewicz:
Instead of removing 1000-5000 contamination, remove 3000-5000.

For reference, yellow contam is 3333 and (light)red is 5000.

### DIFF
--- a/crawl-ref/source/potion.cc
+++ b/crawl-ref/source/potion.cc
@@ -454,7 +454,7 @@ public:
         debuff_player();
         mpr("You feel magically purged.");
         const int old_contam_level = get_contamination_level();
-        contaminate_player(-1 * (1000 + random2(4000)));
+        contaminate_player(-1 * (3000 + random2(2000)));
         if (old_contam_level && old_contam_level == get_contamination_level())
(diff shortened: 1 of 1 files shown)


MarvinPA:
Why?

alexjurkiewicz:
I think the current range is really broad, compared to most other potions in the game (eg hw & magic); and like Yermak mentioned in the tavern thread, it's not very exciting finding you need to quaff multiple potions to deal with a *Contam artifact or hell effects.

MarvinPA:
I could see reducing the range slightly (to 1500-4500 or something), but it seems fine as-is for the more common use cases of *Contam equipment and regular spellcasting/miscasts (as well as its use for just dealing with things like Mark, petrification etc). Hell effects/level 3 miscasts are pretty much a totally different thing with huge levels of contam that don't show up anywhere else, if there's any problem there it makes a lot more sense to me to reduce the amounts of contam they give than to change the potion.

alexjurkiewicz:
Fair enough, I'm happy to leave as-is

PR #409, opened 2016-11-23
Only count connected branches for Gozag service fee

alexjurkiewicz:


PleasingFungus:
Why.

alexjurkiewicz:
Abyss wandering inflates gozag's service fee by counting every piece of
gold generated. Same with Pan (although there at least every piece of gold
could be collected).

PleasingFungus:
Feels like most portal vaults should still count, though.

alexjurkiewicz:
I don't disagree, but it felt like special cases on special cases. If you'd like, I can restrict this to only ignore the Abyss?

PleasingFungus:
Sure, that works for me.

alexjurkiewicz:
Here we go

PR #416, opened 2016-11-27
Don't let the player wield useless items

alexjurkiewicz:
Sorry, speedrunners 😱

### DIFF
--- a/crawl-ref/source/item_use.cc
+++ b/crawl-ref/source/item_use.cc
@@ -505,6 +505,13 @@ bool wield_weapon(bool auto_wield, int slot, bool show_weff_messages,
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Goal is to prevent miskeying? Would it be better to add a prompt for wielding useless items instead?

alexjurkiewicz:
Yes, that's the goal, but the use-case of wielding useless items (bread swinging) is pretty niche, so it feels like an acceptable loss in order to simplify the interface ("why does the game even let me wield a scroll???").

(Bread swinging isn't completely removed with this PR ofc, you can still swing a rod, or many weapons when you have low skill.)

isloat:
This crashes if you bring up the wield screen, then enter '-' for none. 

The error is: "ERROR: range check error (-2 / 52)"

PR #419, opened 2016-11-30
Allow travel trails to be cleared while in the level map.

johnny-0:
For consistency and convenience, adjust the level map to handle CMD_CLEAR_MAP in the same manner as the main screen.

Related commit: 6207113de71659a0cbe17c8978a4c5037fd4aa84

#### Notes / concerns:
- There's now a call to ```crawl_view.set_player_at(you.pos())``` after clearing the map in the level map view.  I didn't notice this having an impact during testing.
- Is map_knowledge the best home for the relocated _do_clear_map() function?
- I added some documentation to clear_map() while dealing with clear_map_or_travel_trail(), mainly for context.  Although I believe it's accurate, it could use a once-over.

### DIFF
--- a/crawl-ref/source/main.cc
+++ b/crawl-ref/source/main.cc
@@ -1893,21 +1893,6 @@ static void _do_rest()
     _start_running(RDIR_REST, RMODE_REST_DURATION);
 }
 
-static void _do_clear_map()
-{
-    if (Options.show_travel_trail && env.travel_trail.size())
-    {
-        mpr("Clearing travel trail.");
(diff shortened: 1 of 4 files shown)


isloat:
Hey, thank you very much for your contribution to Crawl. I tested this out and everything seems to work fine. The concerns you brought up don't seem like a problem to me: map_knowledge is a good place to put it (maybe other devs might disagree?) and the documentation seems on point. 

Pushed in c8f91ca9c73902d6c7f25a75cddaa7dd092dc799, closing.

PR #423, opened 2016-12-30
Remove some spurious warnings on equip attempts.

colinmorris:
There were some cases where we'd give pointless warning prompts on equip operations that were doomed to fail anyways (e.g. because the corresponding slot was cursed). Defer warning about inscriptions until we're sure the equip operation is possible. 

Fixes:

- https://crawl.develz.org/mantis/view.php?id=10860
- https://crawl.develz.org/mantis/bug_view_page.php?bug_id=10826
- https://crawl.develz.org/mantis/view.php?id=4134

Also includes some minor refactoring, and additional comments and TODOs.


isloat:
This is a good UI fix, but the patch looks like it doesn't compile. If you haven't disappeared into the ether (I'm sorry we took so long to notice this PR!) then can you take a look at fixing this up?

neilmoore:
I think the build failure here was actually because of the bug fixed in 0.20-a0-398-gd1c1da2, nothing to do with the PR at all.

colinmorris:
After merging in master, it seems like Travis is happy.

glennmatthews:
Introduces https://crawl.develz.org/mantis/view.php?id=10940

PR #426, opened 2017-01-06
Add Cyno species.

Floodkiller:
Add Cynos, a species with a +2 apt in all skills, but a max skill of 14.


jiocy:
Assorted thoughts after splatting a few Cyno Wanderers:

Firstly, I tried out the race despite hating the idea. Reading it is the opposite of reading something exciting like VS or Ds. However, I gave it a shake and found the race actually rather fun, because of the way all options are open to you at all times.

- Maybe fighting and spellcasting should be exceptions, and able to go up to 27. 
 -     They could be immune to the cap and skill normally, or special, going up automatically based on something else, such as your XL, the number of runes you've collected, or the average value of your other skills.
- On the topic of runes: what if instead of a cap of 14, the cap was instead 10, or 12, plus 0.5 for each rune you've collected?
- On the topic of the cap: 14 is a fine enough number, goes great with lajatangs, but I'm curious why you chose it. With a cap of 15, shields can be a less awkward option. And shields seem like a good choice for Cyno, because onehanders suffer less from the skillcap than twohanders.
- Something cool could/should happen at XL14. 😁 Or whatever the final skillcap ends up being. A free dose of beneficial mutation? 
- The race could use a few twists, maybe. Are you planning to add any?
 -     Being able to swap weapons faster would suit their jack of all trades tendencies

Floodkiller:
Thanks for the feedback!  I've made some changes to the species that have yet to be pushed to CBRO which may address many of the points of your feedback (you can read through the commits or the Tavern thread if you want to see more information).  The main change, however, is that the hard skill cap has been replaced with a 'soft' skill cap.  This starts out with fast training (at a +3 aptitude) and still allows training for skills up to 27, at the cost of the skill aptitudes slowly becoming worse as the skills get higher (training any skill higher than 21 will be at a -4 aptitude).  I've also added a flavor mutation to the species, and may plan on slightly bumping aptitudes up or down a number from the flat average to spice things up.

Please feel free to provide feedback on these changes when they hit CBRO (or you can compile directly from my branch if you don't wish to wait), regardless if it is positive or negative.  The more feedback, the better!

Floodkiller:
Closing pull request to rebase.

PR #427, opened 2017-01-06
Fix incorrect messages on branch transition

colinmorris:
Sometimes the game would say "Welcome back to x!" even though it was your first time in x. e.g. https://crawl.develz.org/mantis/view.php?id=10868 
This fixes that, and simplifies some of the floor transition logic.

(There's definitely plenty more room for refactoring `floor_transition`. There are a few existing "I don't know why this is here" comments in there which I am totally on the same page with. But this is a start.)

### DIFF
--- a/crawl-ref/source/stairs.cc
+++ b/crawl-ref/source/stairs.cc
@@ -322,7 +322,7 @@ static bool _check_fall_down_stairs(const dungeon_feature_type ftype, bool going
             ouch(1, KILLED_BY_FALLING_DOWN_STAIRS);
 
         // Note that if going downstairs, this only does damage.
-        // Tt doesn't cancel the level transition.
+        // It doesn't cancel the level transition.
         if (going_up)
         {
             you.turn_is_over = true;
--- a/crawl-ref/source/stairs.cc
+++ b/crawl-ref/source/stairs.cc
@@ -635,14 +635,9 @@ void floor_transition(dungeon_feature_type how,
         how = branches[old_level.branch].entry_stairs;
     }
 
-    // Check for falling down the stairs or portal. (Why can't you fall in the abyss?)
-    if (!going_up && !shaft
-        && how != DNGN_ENTER_ABYSS
-        && how != DNGN_ABYSSAL_STAIR
-        && how != DNGN_EXIT_ABYSS)
-    {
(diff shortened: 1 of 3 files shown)


crawl-ref/source/stairs.cc
```diff
@@ -627,23 +635,13 @@ void floor_transition(dungeon_feature_type how,
...
-        && player_in_connected_branch()
-        && old_level.branch != you.where_are_you)
-    {
-        mprf("Welcome %sto %s!",
-             you.chapter == CHAPTER_POCKET_ABYSS ? "" : "back ",
-             branches[you.where_are_you].longname);
-    }
-
-    // Falling down the stairs or portal.
+    // Check for falling down the stairs or portal. (Why can't you fall in the abyss?)
```
colinmorris: Oh yeah, I found this curious. Anyone know why we check for abyss here? The only thing I could think of was that maybe it would be weird (and cruel) to make the player trip down the stairs if they were banished. But if that's the concern, I think it'd be better to just check `forced`.

crawl-ref/source/stairs.cc
```diff
@@ -326,8 +321,8 @@ static bool _fall_down_stairs(const dungeon_feature_type ftype, bool going_up)
         else
             ouch(1, KILLED_BY_FALLING_DOWN_STAIRS);
 
-        // Note that this only does damage if going downstairs;
-        // it doesn't cancel the level transition.
+        // Note that if going downstairs, this only does damage.
+        // Tt doesn't cancel the level transition.
```
PleasingFungus: Typo!

crawl-ref/source/stairs.cc
PleasingFungus: 3f120941 / https://crawl.develz.org/mantis/view.php?id=6975 suggests that's the reason; 0a4efdf8 is also involved, but that predates useful commit messages.

crawl-ref/source/stairs.cc
colinmorris: Whoops, fixed.

crawl-ref/source/stairs.cc
colinmorris: Cool, I changed it to just check forced. Did a quick test and everything seemed ok.

PR #428, opened 2017-01-07
Tweak OOD spawn, and kick trivial monsters from hell

frozenca:
1. Current OOD spawn adds +1-5 to dungeon depth in
monster generation. This can give D:1 701 turn centaur, or
D:2 585 turn killer bee band, which are both quite harsh.
So make this weaker for D:1~4 by tweaking the cap of fuzz value.

2. Probability of current super-OOD spawn is
(1400-floor*117)-3000 turn : 0.02%
3000-9000 turn : linearly scaled from 0.02% to 100%

The 0.02% part is practically meaningless, and in fact,
it has been just "death trap for unluckies".
Remove this term, and remove (1400-floor*117) condition accordingly.

---------------------------------------------

Allowing necrophages, mummies, phantoms, crimson imps in
one of the latest dungeons is bad idea.

Gehenna: Increase Brimstone Fiends/barlugs/hellions,
add orbs of fire

Cocytus: Increase blizzard demons, introduce
Ice Fiends to regular set

Tartarus: Introduce Tzitzimimeh to regular set,
and add death scarabs/curse toes

Dis: Add Hell Sentinels/war gargoyles/iron dragons


isloat:
I appreciate the effort you've put into this. Thanks for taking the time to submit a patch to Crawl. That said: there's already an outstanding branch (https://github.com/crawl/crawl/compare/bosch) that addresses many problems with the Hells, and has a solution for monster spawn tables which is similar to your proposal to eliminate a lot of the trash from the spawn tables. Development has been stalled on this for a bit but should be restarting progress soon. For several reasons I'd prefer to merge the Hell improvements alongside the other changes that the devs have planned for postgame changes, instead of on its own. 

Further, I strongly disagree with putting orbs of fire into Gehenna and think they should remain a Zot exclusive, regardless of how thematically delicious it is to place a pure fire monster into the fire-flavored Hell. Orbs of fire are very strongly identified with Zot and placing them elsewhere significantly weakens their connection to this special branch and weakens the identity of the branch itself as well. 

I'm also not really a fan of what you've proposed here for OOD generation tweaks. While I do agree that having such a tiny chance to occur on the early floors isn't great, I think it does help to keep the more experienced players on their toes while navigating the early game, and when a serious OOD does happen it can dramatically change the tone of a given playthrough. I believe Lasty may have a different solution to harshness of OODs but that may be a while off.

frozenca:
Hell rework: Dev's changes seem much better than mine.

OOD for D:1~4: Facing centaur in D:1 at 701 turn and killer bee bands in D:2 at 585 turn looks just unfair. The criterias of '700', '1400-117*depth' had been set before the introduce of autoexplore so it maybe used to plausible number at that time but maybe not now, so if the game should punish players who don't do speedrunning or manually explore in D:1~4, then keeping consistently the same fuzz value five for every single depth would be bad, because spawn depth+5 is not a big deal after the early part whereas calling D:6 monsters in D:1 could be really deadly.

Super-OOD: 0.02% base prob for interval [1400-117*N, 3000] turns is just a bombardment in really unlucky game. If we really should punish players with super-OOD who don't drag THAT much then the base prob should be significantly higher than as it is (though I think there is really no fun to do so) or completely removing the 0.02% prob of [1400-117*N, 3000] interval otherwise.

isloat:
I don't have any strong opinions on OODs, but this PR isn't going anywhere. If you have an idea for how to approach OODs differently, such as you mentioned in above comment, then submit a new PR please. Getting rid of super-OODs sounds okay to me, but there may be a reason they're there in the first place? I don't have enough history working on Crawl to know. 

The monster generation tables do need work, and the dev team has been poking at it for some time now. There's a planned cleanup for Dungeon (and Vaults) that's been suggested for some time now. Whenever one of us has the free time and the inclination to take a look at it is probably when it'll happen. Perils of a volunteer project...

Thanks for your work.

PR #429, opened 2017-01-12
Archaeologist background 

PabloMansanet:
- Starts with terrible skills  (only 1 SP in Stealth) and no weapon. Stats are 3 str, 3 dex, 6 int. Needs to improvise quickly to survive.
- Starts fully clothed with +1 gloves, robe, boots and hat, as well as a **dusty tome** and **ancient crate**.
- At level 3, the tome reveals itself to be a manual of some skill (less SP than normal manuals).
- When the manual is finished, the crate opens revealing an artifact dependant on that skill.

Designed to force you into offbeat builds, and force you into bad (read: fun) decisions. Discussed [here](https://crawl.develz.org/tavern/viewtopic.php?f=8&t=22779).

I'm opening a PR in case there is enough interest to make a experimental.


isloat:
This is a fresh and interesting idea and I'm very willing to give this a shot.

I'm concerned that it breaks too significantly from the rules and standards set by the other backgrounds, and I think other devs might share similar reservations here. That said, Crawl is getting old and sometimes an unexpected shot in the arm can help inject vitality and creativity into the game.

I do like that you've set it up in such a way that it prevents startscumming. And much like Demonspawn (the most popular race by a significant margin), I suspect this would be a popular background because of the "gambling" factor combined with the potential to roll something good. Unlike some other backgrounds, Archaeologist has a high replay factor due to getting different manuals/artefacts. 

One suggestion for improvement I would make is to put weightings on the artefacts - have most artefacts share a common weight, but select several artefacts and make them spawn more rarely. Like dpeg says: 

>"It actually helps if some choices are rarer than others. You might be tempted to make everything equally likely, so as to maximise cross-game variance, but the fact is that eventually the player will see things repeat. Having "common" and "rare" levels helps greatly to make even the 1000th game feel different from the usual, if you run across something you hardly ever do."

isloat:
What happens if you get a tome for a skill that you can't train? Manual of Crossbows when you haven't found a crossbow yet.

PabloMansanet:
Hey! Manuals should allow you to train a skill you don't have the items for :) (although I believe my Archaeologist implementation is bugged in that regard. Haven't had a chance to revise it yet).

MarvinPA:
I'd be very much against a background that hands out unrands, and even if it just gives out randarts or the other proposed items it seems like it'd have the same problems that the recently-discussed background with a spellbook of only L6+ spells would have (except to an even worse extent, probably).

PabloMansanet:
@MarvinPA What were the problems with the high-level spell background? I wasn't around for that discussion.

isloat:
I think this is a neat idea SteelNeuron, but I think it breaks too significantly from the guidelines and precedent we have for Crawl backgrounds. Backgrounds are starting kits that you can reliably rely on (with the one exception being Wanderer, where it is explicitly random), and the starting kits are *starting* kits -- you need to rely on whatever you find from floorgod to really progress further into the game.

Sacrificing that reliable start in exchange for future power is something that's much more the domain of species than it is for backgrounds. I think there's too many concerns about how a background like this would interact with species that have frontloaded (or backloaded) strength. 

Still, it's a fresh and intriguing idea; just, not right for Crawl. I encourage you to come up with more!

PR #430, opened 2017-01-14
Implement the Ieoh Jian Council

PabloMansanet:
This is the rebased PR for the Ieoh Jian Council, should be much less of a mess!


isloat:
This god has been implemented. Congratulations!

PR #431, opened 2017-01-17
Substantially buff the necklace of bloodlust & rework firestarter

alexjurkiewicz:
It's rare for players to voluntarily use *Rage items, so let's make the
item much better.


MarvinPA:
Everything other than the Wrath of Trog deliberately uses ANGRY: 5, to make "*Rage" have a consistent meaning (for a while in the past, lots of unrands had different values which is why it's implemented as-is, might be better to make it just be a bool property and special-case the Wrath of Trog now).

alexjurkiewicz:
I forgot `*rage` was normalised. Updated (and fixed firestarter to actually provide immunity)

PR #433, opened 2017-01-24
Render useless items in dark grey in Known Items menu

glennmatthews:
In the "Known Items" menu, render useless items in dark grey instead of the default light grey used for other non-overridden items. Useful when configuring manual autopickup overrides for a randomly chosen character as it reminds me to e.g., not enable javelin pickup if I'm playing a kobold.

### DIFF
--- a/crawl-ref/source/dat/database/help.txt
+++ b/crawl-ref/source/dat/database/help.txt
@@ -33,11 +33,12 @@ known-menu
 Pressing an item type's letter key selects that item.
 
 <h>Selecting item types</h>
-Items types displayed in grey use the default autopickup options set in your
-configuration file. Selecting a type of item overrides those options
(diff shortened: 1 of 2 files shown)


isloat:
There is a problem with this: now all items cycle through three different stages when toggled in this screen.

For a kobold: javelins go from dark grey -, to white +, to white -, then back to dark grey -
arrows go from grey -, to white +, to white -, then back to grey -

If you can get that fixed (there should be two stages only for each item, white +, and grey or dark grey - depending on restriction) I'll push this since this is a convenient UI fix for characters with item restrictions.

Thank you for your contribution!

isloat:
|amethyst has informed me that I was mistaken and that three-state toggling is the intended default behavior. There's actually no problem with this patch. Oops. Sorry!

glennmatthews:
Thanks!

PR #435, opened 2017-01-25
Fix for pain bond and aux attacks (10472)

rawlins:
The basic problem was that radiate_pain_bond was called recursively from
monster::hurt (which radiate_pain_bond calls), and could loop.  When the
loop reached the original target, if the embedded call to hurt killed
the monster, it was guaranteed to clean up at that point, ignoring the
`clean` argument to the original call.  In a number of cases that relied
on a <1hp monster not being cleaned up, this led to odd messages or
crashes.

This commit fixes the problem by refactoring the recursion into a single
function (radiate_pain_bond) that tracks the original target.  It defers
cleanup for that target to the outer call of monster::hurt, and so will
respect the cleanup_dead argument for that function.  As a side effect,
pain bond damage is calculated slightly more correctly for passive
damage effects.  I've tested this as carefully as I can and am fairly
confident that it prevents the DEAD_MONSTER messages as well as the
crash from constriction, but there were a lot of hard-to-replicate
special cases that I haven't tested directly.

### DIFF
--- a/crawl-ref/source/mon-util.cc
+++ b/crawl-ref/source/mon-util.cc
@@ -5566,7 +5566,7 @@ int max_mons_charge(monster_type m)
 }
 
 // Deal out damage to nearby pain-bonded monsters based on the distance between them.
-void radiate_pain_bond(const monster& mon, int damage)
+void radiate_pain_bond(const monster& mon, int damage, const monster* original_target)
 {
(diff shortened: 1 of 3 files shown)


isloat:
Pushed in f70bbbcf307fc583d5194bdf2c106e342bb257ec. Thank you very much for figuring out this bug and fixing it!

PR #437, opened 2017-01-26
Many monster equipment changes

Doesnty:
Monsters have more equipment slots: two more weapon slots, a cloak slot, a headgear slot, a gloves slot, and a boots slot. Their jewellery slot has been broken into an amulet and two ring slots. They know to put things in the new slots.

Monsters get slaying bonuses to all attacks, not just weapon attacks.

Shields of protection now work for monsters.

Robin now wears her helmet in the headgear slot instead of the shield slot. Because she didn't get any AC from her helmet before, she loses a point of base AC to compensate.

Nikola now wears his gloves in the gloves slot instead of the shield slot.

Crazy Yiuf, Maurice, Nikola, and draconians now wear their cloaks in their cloak slot instead of as body armor.

Gastronok wears his hat on his head instead of as body armor.

All monsters that wear bardings do so in their boots slot instead of as body armor.

Joseph now spawns wearing a stupid-looking hat. (This is the most important change.)

Nagaraja can wield four weapons, and spawn with four light one-handed weapons. If this is not desirable, it can easily be reverted by removing their M_FOUR_WEAPONS flag, reverting their attacks to normal, and reverting their weapon table in mon-gear.cc to use gnoll weapons. Entropy weavers could concievably four-weapon as well but I opted not to give them the flag.

Unfortunately, I am not an artist, and as such did not rig the nagaraja tile to display four weapons. They will probably need a new pose, as their upper right-hand weapon would leak into the tile above them, and their hands are posed awkwardly for right-facing off-hand weapons.

Monsters now have five attack slots. Only nagaraja use the new slot (for constriction+four weapons).

The helmet slot has been renamed to headgear, to minimize confusion with the helmet/hat dichotomy.


alexjurkiewicz:
Wow, nice work!

isloat:
This is a change that increases complexity by a lot but increases bigly parity between monsters and players, *and* opens up more design space; I'm interested. What do other devs think?

Doesnty:
neever mind then

might turn some of the smaller parts of this into their own, more broken-up PRs

PR #439, opened 2017-01-30
FreeBSD Compilation Changes

Koshroy:
This PR makes a couple changes to successfully build DCSS on a FreeBSD environment using `clang`. A summary of the changes:

- The `Makefile` includes `sqlite` from `/usr/local/include` instead of `/usr/include`.
- Perl scripts that look for `perl` at `/usr/bin/perl` now look at `/usr/bin/env perl` to find
  Perl instead. Instead of passing the `-w` flag to `perl`, `use warnings` is used instead.
- `crash.cc` cannot seem to find `strsignal` so instead we use `sys_signame`.

This has been tested with both `TILES=y` and without. The Doxygen files still hardcode the path for Perl to `/usr/bin/perl` but I believe coming up with a way to point Doxygen to the correct Perl is a deeper discussion.


jiocy:
Cool as heck.

Koshroy:
Any feedback on the PR? I can flatten the commits if you all would like, or if there's a cherry-picking process you all go through feel free.

jiocy:
I think they take awhile to notice things, you might want to try hitting their IRC channel, I got some good advice there. It's ##crawl-dev on Freenode.

neilmoore:
A few comments:

- This will need to be squashed into one commit, but we can take care of that.

Koshroy:
* Removed erroneous spaces from the Perl scripts.
* Reorganized the headers in `crash.cc`.
* Added proper spacing to code added in `crash.cc`.

Koshroy:
I encountered an issue where I couldn't call `strsignal` despite including the headers suggested in the manual, so I went with `sys_signame` instead.

neilmoore:
Many thanks!  I made some fixes for our coding style, and also to use our TARGET_OS_FREEBSD rather than the underlying __FreeBSD__, in [0.20-a0-636-g32200f4](https://github.com/crawl/crawl/commit/32200f4c3c99)

PR #440, opened 2017-02-05
reduce intensity of conjure flame verb

jiocy:
"Roar" in Crawl is usually a loud noise, typically one issued by a beast or a Trog-worshipper.

Conjure flame, however, is quite a quiet spell and cautious (but inexperienced) players might avoid casting it for fear of a loud "roaring" noise if they fail to check the full spell details in the z ? ! menu. The description of the sound could stand to be toned down a little.

Perhaps "The fire ignites dangerously||menacingly!" would retain some of the original phrase's feeling of power and violence, at the risk of throwing around unnecessary adverbs!

Uncertain whether "The fire roars!" was being called by other sources than Conjure Flame, sources which perhaps **are** loud enough to earn a "roar".

### DIFF
--- a/crawl-ref/source/spl-clouds.cc
+++ b/crawl-ref/source/spl-clouds.cc
@@ -90,7 +90,7 @@ spret_type conjure_flame(const actor *agent, int pow, const coord_def& where,
         // Reinforce the cloud - but not too much.
         // It must be a fire cloud from a previous test.
         if (you.see_cell(where))
-            mpr("The fire roars with new energy!");
+            mpr("The fire blazes with new energy!");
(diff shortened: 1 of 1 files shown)


MarvinPA:
A "roaring fire" is a common phrase just used to mean a fire that's burning strongly, and doesn't describe the sound it makes in the sense of an actual creature roaring. I don't think this particularly needs changing, but maybe just "blazes" instead of "roars" would be okay.

isloat:
Fires roaring is idiomatic speech, though, and we have plenty of players where English isn't their native language. Our noise mechanics are already subtle, I think it's good to avoid (mildly) confusing the player on that count.

PR #441, opened 2017-02-06
Gloves of Boxing

Doesnty:
Boxing brand for gloves that gives +4 slaying to unarmed attacks. While it was brought up that this creates no-brainers for UC characters, it can also create interesting choices for characters who find it before investing heavily in a melee weapon. I also think the argument that it creates no-brainers is silly, as gloves of archery do the same thing for anyone using ranged combat and most of the ego gloves do in general as they're quite uncommon; many characters only get one pair anyway.

UC is generally considered poor for characters other than trolls (who can't use them), ghouls (who finally have reasonable gloves to use), and transmuters (who meld them), so this shouldn't be a huge balance upset.


yrmvgh:
Might be nice (and less narrow) for it to apply to aux attacks. Rings of slaying already do...

PR #442, opened 2017-02-07
Remove archery brand.

Doesnty:
Got told gloves of archery shouldn't be a thing, so here's a patch that disables them.


isloat:
It's been a while since you posted this PR, so I thought I'd update: while gloves of archery are probably going to be going away, I and a couple of other devs want to see an additional ego for glove slot before we do away with them. So I'm gonna put this PR on hold until we can come up with a sufficiently good idea for a new ego and implement it, then merge this. Thanks for the patch!

PR #443, opened 2017-02-07
Identify wands when the player steps on them.

Doesnty:
Wands can be trivially type-identified by picking them up, and it's rather annoying when encountering a wand with a full inventory to figure out the safest item to drop, pick up the wand, then immediately discard it and pick up the item again because it's a trash wand (e.g. random effects or flame outside early-game). Spellbooks are in the same boat and identify just by standing over them; this extends that behavior to wands.

### DIFF
--- a/crawl-ref/source/items.cc
+++ b/crawl-ref/source/items.cc
@@ -934,29 +934,39 @@ void item_check()
     }
 }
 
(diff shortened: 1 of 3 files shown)


giann:
At that point why bothering with identification at all ? We could just know what they are when they come into view.

PR #445, opened 2017-02-08
Unrand vaults

alexjurkiewicz:
Unrand vaults for mask of the dragon and shield of ignorance


isloat:
Pushed in f0b4e372af6c and f59e5ab63d09, after some surgical complications. Thank you!

PR #447, opened 2017-02-08
Piercing Shot

Doesnty:
Simple level 5 charm that makes all thrown/fired missiles penetrate for a while, at a cost of 2 MP per projectile. Should be fun since penetration was fun but is incredibly rare to get to use now that it no longer shows up as an ammo brand.


MarvinPA:
A better way to make the penetration brand more common (if that's desirable) would probably be just to make it generate more often on launchers. I don't think there's anything particularly distinguishing about penetration that would merit it being a spell, considering that weapon branding spells have been being steadily removed.

Doesnty:
What problems does this spell share with the older brands? It's not a no-brainer to cast this at the start of a fight since it guzzles MP rapidly (5 on cast+2 per shot) for no extra damage against most single targets. It's not a no-brainer to learn it since it's level 5. It's not stepping on any god's shoes since the only way you're getting a penetration launcher is a super lucky randart crossbow roll. I'm also not convinced it is brokenly good like Warp Weapon was. It's arguably stepping on Throwing's shoes, but Throwing has other factors keeping it distinct already (reliance on Shoals or an ammo god, blowguns, being 0-handed).

I think making penetration crossbows more common would be much less interesting than having a spell that allows for crossbows to get penetration. Penetration is a strong effect, and if it was readily available as a normal brand it would blow the others out of the water. This method makes it less free than just making vanilla penetration crossbows drop, as you have to invest experience in Charms and MP to utilize it.

isloat:
I find Doesnty's arguments to be compelling, especially about the MP cost being a significant dampener on how often it can be used. I also agree about making the brand more common being less interesting than having a spell that provides a special interaction.

Previous brand spells were cast-and-forget, this has to be actively maintained with ongoing MP costs. I think it's a whole different beast.

isloat:
already explained in ##-dev, leaving here for posterity (plus a little cleanup):

Doesnty, I've been mulling over your Piercing Shot PR for a while now and I finally figured out what bothers me about it and why I think it won't be going in (aside from MPA's objections, that is):

It's essentially a bolt spell with very weird behavior, that uses ammo that isn't meaningfully constrained (unlike early game with arrows/S2S or rocks/Sandblast), and the "spell power" (ie damage and accuracy) relies on non-spellcasting skills. 

On its face it's a cool upgrade for ranged characters, but once you dig into it, it starts falling apart. 

This was initially a not-bad take on a branding spell: high initial investment + per-use MP burn is a good constraint, vs previous cast-and-forget behavior. So that makes me slightly sad that I have to close this. But don't let this discourage you!

PR #449, opened 2017-02-09
Several is_useless_item() fixes

glennmatthews:
1. `is_useless_item()` checks for potion and ring of flight did not check for innate racial permaflight ability, only whether you were permaflying at the moment. (Hence, toggling permaflight on and off would change the perceived uselessness of these items.) Added missing `racial_permanent_flight()` check to these two cases.
2. `is_useless_item()` checks for scroll of fog and wand of clouds considered Still Winds to be a permanent condition rather than a temporary one. Added missing `temp` check to these two cases.


isloat:
Seems solid. I'm going to give this a testing tomorrow before I merge, but cursorily going over the code I doubt there's any real problems with this. Thanks for putting in the work to contribute to Crawl!

PR #450, opened 2017-02-09
Enhanced sound support

Autofire:
This fork features some enhancements for the sound system. Although my plans initially were more ambitious, this will do for now.

- Adds `sound_file_path` option, making the current sound option **much** easier to use.
- Moved all the sound code to its own set of files for future expansion (I have some plans). Also, makes it painless to fiddle around with `PLAY_SOUND_COMMAND`, which previously was in AppHdr.h despite being used in just one function.
- (**UPDATE**) Adds multiple cues that can be used in place of typical regular expressions. Some examples are `PICKUP`, `MEMORISE_SPELL`, and `CHANGE QUIVER`. See sound.h for a full list. I intend of expanding this.
- Very incomplete support for a `hold_sound` option, which is _supposed_ to pause the game while the sound is playing. This pause is intended to be sequential, delaying the rest of the message until the sound finishes. I have two purposes for this: giving battles a real-time feel, and to highlight those epic victories. This feature doesn't work as intended, so I dropped the in-game documentation for now.
- (**UPDATE**) Adds `one_SDL_sound_channel`. Originally, sound behavior depended on whether you were playing with SDL or using the `SOUND_PLAY_COMMAND` define; if playing with SDL/Tiles, only one sound would play at once, whereas `SOUND_PLAY_COMMAND` allowed as many sounds as were necessary. By default, `one_SDL_sound_channel` is set to false, making it behave as `SOUND_PLAY_COMMAND`. However, players can restore the old behavior by setting this option to true.
- (**UPDATE**) The version screen now tells the user how sounds will be played. (i.e. "Windows Multimedia API," "External command," or "SDL_mixer")

Anyway, I know this works as intended on Linux and Cygwin, and I did run `crawl -test` to double check stuff. **UPDATE: I have now verified that it works with Tiles!** However, I can't test with Windows, Mac, etc. I may or may not have broken the sound stuff. At worst, some more files need to be included in sound.cc.

Finally, the game works as normal when compiled without sound support. I only tested this in Linux, but I'm certain that it should be the same everywhere else.

Happy crawling!
~Autofire


isloat:
Hey, great job on this! There's been attempts at sound patches before, but this seems more thorough than usual.

Unfortunately since I'm deaf I cannot confirm for myself that this works properly, so I'll have to hand it off to another developer. Good luck!

Autofire:
I made a very small tweak, commenting out `SOUND_PLAY_COMMAND`. This was how I originally found it, and I forgot to change it back to this. I meant to do this sooner...but I forgot. :sweat: 

I didn't know that there had been sound patched before, @isloat . Maybe I should have tried salvaging one of those... _shrugs_.

Autofire:
Sadly my latest version has issues with the mute button, with it also being mapped to other things. I can't figure out an unused key combination, and I can't easily revert my change at the moment. If this pull request gets accepted, please find a good keybind.

neilmoore:
In trunk, with a few fixes ([0.20-a0-778-g4270d38](https://github.com/crawl/crawl/commit/4270d38ebf63)).  Thanks!

PR #451, opened 2017-02-10
Have ash id stacks when standing on them (10580)

rawlins:
Previously, `god_id_item` was called only from a distance, and then only on the
top item of the stack, so items not at the top of the stack would not be id'd
when you move onto their square.  This fix ids the whole stack when you stand
on it if appropriate.  This may overcall `god_id_item` but I haven't spotted
any detrimental effects.  Addresses 10580, 10919.


isloat:
Thank you!

PR #452, opened 2017-02-12
cause Amulet of Faith to interact with Gozag

yrmvgh:
Gozag isn't just a beautiful and witty shopkeeper, Gozag is also a god!
That's why a background spent meditating in a monastery waives the fee,
and that's why a player of great faith could stand to be blessed by Gozag.

Wearing =faith increases Gozag's corpse-to-gold conversion rate by 25%.
Removing =faith increases the price of summoning merchants by 200.

### DIFF
--- a/crawl-ref/source/invent.cc
+++ b/crawl-ref/source/invent.cc
@@ -1591,7 +1591,6 @@ bool needs_handle_warning(const item_def &item, operation_types oper,
     if (oper == OPER_REMOVE
         && item.is_type(OBJ_JEWELLERY, AMU_FAITH)
         && !(you_worship(GOD_RU) && you.piety >= piety_breakpoint(5))
-        && !you_worship(GOD_GOZAG)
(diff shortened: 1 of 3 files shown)


MarvinPA:
Faith was very deliberately disabled under Gozag (95b2ebd6e474c1b9fc7abbb2067aab1d975cc92e) and this doesn't fix any of its problems (in fact it looks like it adds back the exact same bad special case that previously existed).

yrmvgh:
Oh, really? Blast! Well, thanks for taking a look.

isloat:
I doubt it, sorry. :( 

Xom's interaction with Faith works because it increases the amount of Xom situations, which is perfectly in keeping with Xom and the point of the amulet (more god actions!). If you take off faith Xom just simply becomes less interested in you.

With this Gozag special case, taking the amulet off permanently alters the game. That's not good.

PR #454, opened 2017-02-14
Add a noise meter to the HUD

rawlins:
Noise is one of the more complicated systems in dcss, yet it's barely
represented in the UI.  This is partly because there's just so much
information to represent.  This commit attempts to partially solve this
problem by providing a very simple UI element that summarizes noise
heard by the player during the last turn in the form of a meter.  This
meter peaks at the loudest noise audible at the player's position during
that turn, and will switch between several different colors (details
vary slightly for different interfaces), described below.  I chose to
use sound that has propagated to the player, so more distant events will
be attenuated by distance and dungeon features -- the idea is to give a
measure as to how likely things are to move towards the player's
position.

While experienced players may have internalized a lot of this, most of
this information is not at all apparent to less experienced players, and
I think this UI element is potentially extremely useful (as opposed to
clutter).

1. lightgray: the noise level at the player's position is within a range
that makes it unlikely to propagate outside the player's los (though
this can happen depending on terrain).  Or more simply: this noise is
probably basically ok.  Most low-medium melee damage is in this
category, as are most low-damage spells.

2. yellow: the noise heard at the player's position is likely to
propagate outside the player's los, up to twice the los radius.  Or more
simply: this noise is bad.  Shouting, distant fireballs, high damage
melee attacks, high damage spells (e.g. many bolt spells), door creaks
are in this category.  longbows and bigger crossbows are on the border
of this category.

3. red: the noise is like to be heard at a substantial distance, on an
open level more than 2x the los radius.  Or more simply: this noise is
very bad.  This category includes howler monkeys, lightning bolt, nearby
fireballs, qazlal at pieties above ****.  (Though thunder clouds may
randomly produce red noise at nearly any piety.)

4. lightmagenta: these sounds are off the meter, the loudest sounds you
can make.  Includes: shatter, gong, ood collision, orb apport/pickup

Potential issues/complications:

* The above categories could be tweaked or added to.  How they use
the meter could also be tweaked.  (Loudness maps linearly onto the
meter so about half of it is occupied by red.)
* This implementation uses a heuristic for branch noise adjustment
that is kind of hacky.  A better (but much more complicated) solution
might be to sample noise at various distances directly from the noise
grid.  Right now I just use a constant addition or subtraction to model
the effects of branch noise; in e.g. shoals or crypt this will push
some of the category boundaries for colors around.
* An alternative approach would be to not use attenuated noise at all,
but find the loudest noise (before propagation) in the player's los.
* This implementation doesn't save the noise state, so it isn't displayed
after loading.
* In wizmode there is also a numeric version of the meter displayed,
but I decided this is too uninterpretable to be useful to players.
(But, the heuristic for normal noise branches is: in a completely open
area a sound of loudness l will propagate approximately l/.85 spaces.)
* exposing this to the player may reveal some cases where it is not
entirely intuitive that no sound is made.  For example, berserking is
silent (but unstealthy) except when opening doors.

### DIFF
--- a/crawl-ref/docs/crawl_manual.rst
+++ b/crawl-ref/docs/crawl_manual.rst
@@ -234,9 +234,15 @@ Place
   branch. The starting branch is called Dungeon, so that the place information
   will read "Dungeon:1" for a new character.
 
-Gold
-  Displays the number of gold pieces you have found. Gold is found scattered
-  around the dungeon, and is primarily used to buy items from shops.
+Noise
+  This is a colored bar indicating the loudness of noise that you heard on your
+  last turn. The color provides a rough guide to how far away the noise it
+  indicates might be audible.  If the bar is gray, the sound is less likely to
+  be audible outside of your line of sight (at least in an open area); if it is
+  yellow, the sound is likely to be audible outside of your line of sight, and
+  if it is red the sound will be audible at a substantial distance.  If the bar
+  turns magenta, you have made one of the loudest noises in the dungeon.  Terrain
+  may affect how noise spreads beyond this indicator.
 
 Time
   This indicates the amount of time that has passed since entering the dungeon,
--- a/crawl-ref/source/duration-data.h
+++ b/crawl-ref/source/duration-data.h
@@ -583,7 +583,7 @@ static const duration_def duration_data[] =
         {{"", trog_remove_trogs_hand},
           {"You feel the effects of Trog's Hand fading.", 1}}, 6},
     { DUR_GOZAG_GOLD_AURA, 0, "", "gold aura", "", "", D_NO_FLAGS,
-        {{ "", []() { you.props[GOZAG_GOLD_AURA_KEY] = 0; }}}},
+        {{ "", []() { you.props[GOZAG_GOLD_AURA_KEY] = 0; you.redraw_title = true;}}}},
     { DUR_COLLAPSE, 0, "", "", "collapse", "", D_NO_FLAGS },
     { DUR_BRAINLESS, 0, "", "", "brainless", "", D_NO_FLAGS },
     { DUR_CLUMSY, 0, "", "", "clumsy", "", D_NO_FLAGS },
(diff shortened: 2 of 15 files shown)


rawlins:
Here are two screenshots.  Keep in mind that this meter is more about the change than the static value, so looks better than this in action :-)

screenshot 2017-02-13 20 03 21
screenshot 2017-02-13 20 03 49

rawlins:
Also, I made a short screencast of me putting the noise meter through its paces in local tiles, here:

https://www.dropbox.com/s/xn1s2co3um3617q/noisemeter_screencast.mov?dl=0

(Note that this was done in wizmode, so the number you see next to "Noise:" isn't normally displayed.)

rawlins:
Also: noisemeter_gong
(Idea due to alexjurkiewicz)

giann:
This is awesome

rawlins:
I've updated this based on feedback in IRC to take up less space.  Here are screenshots of the HUD for the half-width version in 80x24 console, local tiles, and webtiles; in the local tiles one I suppressed the number so the spacing without it is easy to see.  As discussed in IRC, it may make sense to eliminate gold from the display except for gozag, but I haven't done this yet.

screenshot 2017-02-13 23 27 17
screenshot 2017-02-14 10 54 10
screenshot 2017-02-14 12 14 09

@giann Thanks!!

frozenca:
Pretty good!

But, I think the best way would be eventually DoomRL's sound system, that is, the player can "hear" sound effects and the size and direction match with in-game noise values. Devs, how do you think about officially introducing CWZ sound patch?

rawlins:
@frozenca there's definitely lots of room for more sophisticated presentations of noise, another thing that's been tossed around in ##crawl-dev a bit is the idea of using color overlays on tiles to indicate loud noise sources.  I don't know much about the CWZ sound system I believe there's also someone working on better sound support for at least local tiles right now.

This approach has the advantage of being something that works in console over ssh though :-)  I think it's consistent with the other ideas.

isloat:
This looks great. Let us know when you hit a final version on this.

rawlins:
Current state of things:

screenshot 2017-02-16 00 26 04

rawlins:
One more console screenshot; this also shows the way gold displays in this PR for gozagites (It's very similar in tiles/webtiles).

screenshot 2017-02-16 23 54 52

PR #456, opened 2017-02-14
Remove alarm traps in ossuary_tomb_3

alexjurkiewicz:
They were after the player had dealt with (nearly) all monsters in the
portal, and the portal is small enough there's no real need to move them
earlier.


isloat:
I don't see alarm traps at the entrance in this commit, just their removal at the end.

PR #457, opened 2017-02-14
Fix issue 10940 introduced by #423

glennmatthews:
In `wear_armour`, don't call get_armour_slot until we know it's an armour item.

Fixes crash issue 10940 introduced by #423.


glennmatthews:
Thanks! I don't believe I have access to update the status of https://crawl.develz.org/mantis/view.php?id=10940 - can someone do so?

PR #459, opened 2017-02-16
Barbs affect more movement methods. More restrictions on hopping.

glennmatthews:
The following movement actions now warn the player if barbed, and trigger pain if
followed through:
- Formicid Shaft Self ability
- Barachian Hop ability (issue [10889](https://crawl.develz.org/mantis/view.php?id=10889))
- Ru Power Leap ability
- Uskayaw Line Pass ability

Other god movement abilities (Shadow Step, Bend Space) are more
explicitly "Blink-like" and so still do not trigger barbs.

Additionally, some more restrictions on Hop to make it behave less like cBlink:
- Can't hop if exhausted
- Can't hop if stuck in liquefied ground
- Can't hop if stuck in grasping roots
- Can't hop if netted

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -1791,24 +1791,21 @@ static spret_type _do_ability(const ability_def& abil, bool fail)
 
     case ABIL_SHAFT_SELF:
         fail_check();
-        if (you.can_do_shaft_ability(false))
+        if (you.can_do_shaft_ability(false)
+            && check_move_barbed("dig")
+            && yesno("Are you sure you want to shaft yourself?", true, 'n'))
         {
-            if (yesno("Are you sure you want to shaft yourself?", true, 'n'))
-                start_delay<ShaftSelfDelay>(1);
-            else
-                return SPRET_ABORT;
+            start_delay<ShaftSelfDelay>(1);
         }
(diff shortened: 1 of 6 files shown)


isloat:
After a period of discussion in ##crawl-dev, the consensus is that the increase in arbitrariness isn't really worth the tradeoff in reduction of abstraction / increase of "flavor". 

Hop already has its own -Hop status and tying it to Exh is too much. The other conditions are rare and specific, and while it notionally makes sense to stop the frog from hopping when its movement is restricted, it also increases cognitive load on the player to remember all these situations where it would apply.

As for the barbs half of this PR, it's not viewed favorably for similar reasons described above, and one additional one: Crawl has really no good way to distinguish between "physical" and "magical" movement abilities. So, again, it comes down to the player having to remember multiple, various special cases instead of a single one (moving the character directly with the movement keys triggers Barbs pain).

glennmatthews:
OK, fair enough. Thanks for the consideration! If this is rejected, feel free to close 10889 as well.

isloat:
For what it's worth, this generated discussion about splitting up our various cooldowns instead of loading it all onto the Exh status. And thank you for submitting a patch for Crawl, it's appreciated!

PR #461, opened 2017-02-17
Split seven enums out of enum.h

alexjurkiewicz:
Long term goal is to entirely remove enum.h. This will help speed up
compilation when making changes to enums, since you won't have to
recompile literally every file.

Currently, crawl compiles fine as non-debug console build on OSX. Other
build targets not yet tested.

Also uses '#pragma once' instead of the more traditional '#ifndef
MY_DEFINE_H' approach for the new files.


alexjurkiewicz:
I'd like feedback on how to proceed before I do, but I'm happy to contribute somewhere between 50-99% of the work separating out enums.

Current feedback from pleasingfungus: should put the files not in a subdir (eg `enum/ability-type.h` -> `ability-type.h`)

PleasingFungus:
Pushed, thanks!

PR #463, opened 2017-02-19
Fix a few typos in docs and source

lischenko:
Duplicated "to", for example:
"Select a spell to read its description or *to to* memorise it"


MarvinPA:
Merged, thanks!

PR #464, opened 2017-02-19
Improved Linux integration

Mailaender:
This adds install rules and metadata files for all Linux distributions. The concept of the appdata.xml file which might be new to some folks is explained at https://en.opensuse.org/openSUSE:AppStore.


Mailaender:
See https://build.opensuse.org/request/show/459057 for an integration test.

neilmoore:
Looks good, just a few comments.
1. The launcher refers to "Exec=crawl-sdl", but I don't think we use that binary name by default.  'make install' just calls the binary 'crawl', and the Debian build for example uses 'crawl-tiles'.
2. It would be nice to have a launcher for console builds too.

Mailaender:
I addressed the name clash which every distribution solved by rename workarounds. Standardized this to the `crawl-tiles` name which your Debian files use. https://build.opensuse.org/request/show/459433

Mailaender:
You can change this appdata.xml file afterwards. It is meant to be maintained by the original developer team. To me the man page is enough for the console build for the moment. Also I am inspired by http://www.playdeb.net/game/Stone%20Soup which uses the tiles variant to tease people with screenshots.

moozaad:
Just reviewing the opensuse package for 0.22 master trunk. Has this been superseded? If not, is it still worth standardising for 0.22 release?

moozaad:
It's still used in opensuse to allow both clients installed at the same time and to have nice user friendly desktop menu entries.. https://build.opensuse.org/package/show/games/crawl You can see the 4 patches/PR there. As a packager, it's definitely nice to have!

moozaad:
I'm working on the 0.24 build for opensuse. This PR is still valid from what I can see.

rawlins:
I'm going to close this since it seems to me that the key ideas/work here has been incorporated into #2188. Very sorry that this has languished so long, and I really hope we can get some version of the changes proposed here merged soon.

PR #465, opened 2017-02-19
New arrival vault

chereshnev:
New arrival vault called  "antagonism"  was added. Comments, criticisms and recommendations are welcome! Vault was designed from scratch.

### DIFF
--- a/crawl-ref/source/dat/des/arrival/small.des
+++ b/crawl-ref/source/dat/des/arrival/small.des
@@ -3179,39 +3179,21 @@ xtttttttttttttttttttttx
 ENDMAP
 
 ###############################################################################
-# This arrival vault represents a confrontation between the faithful zealots
-# of The Shining One and Yredelemnul.
  
-NAME:    chereshnev_arrival_zealots
-TAGS:    arrival no_monster_gen no_item_gen no_pool_fixup generate awake \
-         no_trap_gen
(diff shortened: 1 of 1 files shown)


MarvinPA:
I don't think arrival vaults should immediately spam the player with a bunch of unimportant combat messages (there are a couple of existing ones using krakens that are very similar to this in fact, and they should probably be changed not to do so). "Assorted monsters behind glass" is also extremely overused in arrival vaults, would probably be best to avoid more of that really.

chereshnev:
I reworked the vault. Although the monsters behind the glass is still here. If you have specific requirements, I am ready to listen. Otherwise, I will remove the branch and my new vault.

alexjurkiewicz:
I think there are a lot of "monsters behind glass" arrival vaults, but there's always room for more with lower weight, right? IMO, if you give the two factions a bit of space to roam around in (say six squares), and have only ~2 monsters per side, it would be quite nice.

chereshnev:
> I think there are a lot of "monsters behind glass" arrival vaults, but there's always room for more with lower weight, right? IMO, if you give the two factions a bit of space to roam around in (say six squares), and have only ~2 monsters per side, it would be quite nice.

> 

Thank you very much for your feedback! But I'm afraid the developers have already said their word. Maybe they're right. But I don't know this because it was my first vault.

PR #467, opened 2017-02-19
Add deep troll leather armour and troll iron armour.

Doesnty:
It bugged me how every troll drops the same green hide with the same stats, so I made blue and gray variants for the late-game trolls to drop.

Deep troll leather armor gives 7 AC with 7 ER, the same as swamp dragon armour. It offers regeneration instead of rPois. They're kinda abundant in Depths and Vaults so I didn't want it to be excellent, just decent.

Troll iron armor gives 10 AC with 18 ER, the same as plate mail. It offers regeneration instead of possibly having an ego or enchantment built-in. Given that any race can wear it, it may need to be toned down, but you usually have to kill an iron troll to get your hands on it so I don't think it's a huge game-changer. It's also mostly worse than storm dragon armour. In theory it should provide rF+ and rC+ but that would make it too desirable, I think.
I'm not too attached to the name, but "iron troll leather armour" felt wrong.


Doesnty:
<MarvinPA> i don't really think yet more base types of armour are a good thing to add
<Brannock> why not?
<gammafunk> why would it be?
<MarvinPA> because there are already a lot that are insufficiently well-differentiated? and these are literally just "existing armour type, but better"
<Brannock> I mean, I want to know what MPA's reasoning is, not arguing "hey why not toss it in"
<Brannock> I agree about "is X but better"
<Doesnty> i'd be fine with just making different appearances for troll leather
<Doesnty> that's the main thing that bugs me
<Doesnty> that they all drop green hides
<gammafunk> they all have the same inner skin
<Doesnty> this just seemed like the most logical way since iron trolls also have massive AC
<gammafunk> that's the part that turns into the armour
<Brannock> lol
<Brannock> I'm okay with cosmetically different troll armour
<MarvinPA> you peel off the outer skin to get to the delicious armour beneath it
<gammafunk> but yeah I think as far as new armour types go
<gammafunk> you are guilty until proven innocent
<MarvinPA> (but yes cosmetic differences sound fine to me)
<Brannock> Doesnty, can you resubmit (or a separate PR) with the cosmetics only?
<Doesnty> yeah
<Brannock> I'd support that one

PR #472, opened 2017-02-24
Allow swimming in deep water

alexjurkiewicz:
All species can now move through deep water with a 2.5x penalty. The
penalty for moving through shallow water is also fixed at 1.5x rather
than the previous random multiplier of 1.3x - 2.0x.

When swimming in deep water, non-native swimmers cannot attack -- it's
not indended for deep water swimming to be part of tactical combat.

Crawl has historically had big problems with players possibly getting
stuck due to deep water -- see lots of island vaults needing to plant
escape hatches and the like to prevent people trapping themselves.

There are also issues with monster gear near deep water -- if you want
to grab a monster's items / corpse you generally just have to step back
once before landing the killing blow so they die over land rather than
deep water.

### DIFF
--- a/crawl-ref/source/main.cc
+++ b/crawl-ref/source/main.cc
@@ -3260,6 +3260,7 @@ static void _move_player(coord_def move)
                                                                : "walk";
 
     monster* targ_monst = monster_at(targ);
+
     if (fedhas_passthrough(targ_monst) && !you.is_stationary())
     {
         // Moving on a plant takes 1.5 x normal move delay. We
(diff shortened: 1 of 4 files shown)


alexjurkiewicz:
The build failure is a Travis timeout

alexjurkiewicz:
Note to self:

```02:25 < Sequell> (1/1) hellmonk said (5h 16m 4s ago): just so you're aware, your allow swimming pr
                 prevents casting spells or attacking if you're flying over deep water with a non-swimming
                 species
```

alexjurkiewicz:
I will fix this if there's interest in the PR

isloat:
After group discussion in -dev this morning, we mostly concluded that adding another level of bad-swimming to the game wasn't desirable. We agreed that reducing instances of the player getting stuck was a good thing, but it was brought up that these island vaults already have escape hatches (as you mentioned originally) and that we don't have ways out of teleport closets aside from hatches (and Passwall). 

Deep swimming doesn't really add anything interesting to the game, though, while opening up situations for players to get themselves killed by miskeying into deep water. Could add a prompt for REALLY move into the deep water?, but again that would interfere with flow of play. 

Personally, I'm not sure what to do about the whole "move enemies out of deep water before killing if you want their gear" problem. I don't think this PR would have cleanly addressed it, though. 

As of 242eb585b wading now has a fixed speed penalty.

PR #473, opened 2017-02-24
Load saves where desolation is incorrectly global

rawlins:
Some games transferred from pre-0.19 to post-0.19a will then save with
desolation's place info incorrectly numbered as the global place slot,
triggering an assertion failure on load.  This patch loads these files
without crashing; it limits the time-frame of this fix with a new minor
tag.  No one who has tried has been able to duplicate the save file
corruption, though we have several saves that result from it, so the
root cause may return or still be present.

Fixes some saves found attached to 10696 and 10788.

### DIFF
--- a/crawl-ref/source/tags.cc
+++ b/crawl-ref/source/tags.cc
@@ -3923,13 +3923,12 @@ static void tag_read_you_dungeon(reader &th)
         {
             // This is to fix some crashing saves that didn't import
             // correctly, where the desolation slot occasionally gets marked
-            // as global on conversion from pre-0.19 to post-0.19a. Is setting
-            // the place info for branch `i` really a good idea? There's
-            // actually no guarantee that the save order will match the
-            // current branch order; this is really more of a best guess.
-            // HOWEVER, in all the instances I have, this works, and the place
-            // order hasn't changed since the iterator was introduced. (advil)
-            place_info.branch = static_cast<branch_type>(i);
+            // as global on conversion from pre-0.19 to post-0.19a.   This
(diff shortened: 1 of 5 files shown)


crawl-ref/source/tags.cc
```diff
@@ -3917,10 +3919,23 @@ static void tag_read_you_dungeon(reader &th)
...
+        if (place_info.is_global() && th.getMinorVersion() <= TAG_MINOR_DESOLATION_GLOBAL)
+        {
+            // This is to fix some crashing saves that didn't import
+            // correctly, where the desolation slot occasionally gets marked
+            // as global on conversion from pre-0.19 to post-0.19a. Is setting
+            // the place info for branch `i` really a good idea? There's
+            // actually no guarantee that the save order will match the
+            // current branch order; this is really more of a best guess.
+            // HOWEVER, in all the instances I have, this works, and the place
+            // order hasn't changed since the iterator was introduced. (advil)
```
amalloy: If the only known occurrence of this bug is that Desolation is incorrectly marked, and we're not sure `i` is right in all cases, what about asserting that `i == BRANCH_DESOLATION` before setting `place_info.branch` to be `BRANCH_DESOLATION`?

rawlins: That sounds reasonable, I'll update it tonight.  The minor tag idea (which was PF's suggestion) also helps with this, because I know that the save order hasn't changed in the past.

rawlins: This suggestion led to me figuring out what was causing the corruption in the first place, so, thanks!

PR #478, opened 2017-02-26
make Ba description less negative

yrmvgh:


isloat:
This reads a little awkward because of the final sentence -- you've paired the slow movement and skill mastery in the same sentence, but they aren't really related to each other.

PleasingFungus:
I don't think the description is overly negative right now.

PleasingFungus:
Thank you for going to the trouble of writing this up, regardless!

PR #481, opened 2017-02-28
Add the skeleton species

PabloMansanet:
Skeletons are low HP, high mitigation undead species with a knack for
combat and offensive magic (conjuration) but without the necessary subtlety
for the more advanced schools of magic, or to excel at stealth,
invocations or evocations.

The skeleton has an always active ability that allows it to enter a
trance until all living creatures in LOS are dead. This provides it with
damage shaving and a skill boost, but will mesmerize it against the
closest living being until everything dies, so there is a constant risk
involved in using it.


isloat:
This species's raison d'etre is inherently flawed.

Reaping is an activated ability that locks the player into mesmerisation until no enemies remain in LOS. In exchange the player gets buffs to defense and offense. You characterize this as a "constant risk", but what I immediately recognize is that Reaping will only ever be used when the player can tightly control the situation: ie, luring enemies back, setting up killholes, or specifically luring enemies to areas where the player can take advantage of doors and corners to be able to reliably break the mez if necessary. No player will Reap when they think it might get them killed, unless they're truly out of options.

So essentially Reaping means this species is even **more** encouraged to lure away enemies and whittle at a crowd than the average Crawl species. Because of the strength of Reaping, and even (and especially) if Reaping was made more powerful, players want to use it in optimal conditions and to minimize the risk of getting caught out in a bad way. This is further exacerbated by the average length of any given Crawl playthrough. In a much briefer game, I could see a situation where some players would have no problem using Reaping to chew through crowds, at the risk of losing their character. When the stakes are "Lose the past two or three hours of playtime" then players are going to become increasingly conservative with this ability over the course of a playthrough. Would a Skeleton player use this in V:$? Depths? 

And in the case where the skeleton player does just deal with luring everything to controllable situations, this strategy is already very strong (and has been chipped at occasionally over the years of Crawl development) -- does it really need to be even stronger with the benefits from Reaping? 

But surely Berserk has the same problems? It does, but it's much less severe. They aren't locked into berserking permanently as long as monsters are visible. The post-berserk player is still able to move away to escape from bad situations. Because of that there's significantly less incentive to lure everything back to an absolutely safe spot. What about the Obsidian Axe? I think, personally, there's a difference between it being a very risky trait of an otherwise-good item and being the core mechanic of a species.

There are a couple of ways I can think of to address these issues (none of which are really in keeping with established precedent for how other abilities in Crawl work): disallow Reaping when near walls or requiring multiple visible enemies before Reaping can be triggered... but of course there are still ways to work around these (luring weak enemies coming to mind, something that's gotten other ideas quashed in the past). 

The rest of the species doesn't have anything standing out aside from undead characteristics and a boilerplate set of aptitudes, unfortunately. Thanks for the PR, but this won't be merged.

PR #482, opened 2017-02-28
Lava orc / Djinni cleanup

rawlins:
Lava orcs and Djinni haven't been startable species in years, but their code lingers -- and there is a lot of it.  (|amethyst calculated that this PR removes 1378 net lines of code.)  This PR:

- on load, converts Lava Orcs to Hill Orcs, and Djinni to Vine Stalkers (both after a special full "yes" confirmation).  They both get to fly for a little while, and Djinni get some extra nutrition, but that's it.
- eliminates most code involved in both species.  What's left is the bare minimum to support save compatibility.

Hopefully this won't break anything else, but it's a pretty big change so it's entirely possible I missed something.  It doesn't mess with anything in `rltiles` because I wasn't quite sure what could be changed safely.

Here are some save games (generated in the last point in history to support starting a character for each of these species) for anyone who wants to test the conversions:

[lavaorc-djinn-saves.zip](https://github.com/crawl/crawl/files/807371/lavaorc-djinn-saves.zip)

(Because one of the LO saves has its stats maxed via &A, you get a slightly weird order of the training screen vs the load confirmation, but this is harmless as far as I can tell.)

Credit to |amethyst for the basic idea this PR implements.

### DIFF
--- a/crawl-ref/source/aptitudes.h
+++ b/crawl-ref/source/aptitudes.h
@@ -1332,6 +1332,7 @@ static const species_skill_aptitude species_skill_aptitudes[] =
     APT(SP_OCTOPODE,        SK_EVOCATIONS,      1),
 
 #if TAG_MAJOR_VERSION == 34
+    // Need to keep this as long as the species are still around
     // SP_DJINNI
     APT(SP_DJINNI,          SK_FIGHTING,       -1),
     APT(SP_DJINNI,          SK_SHORT_BLADES,   -2),
--- a/crawl-ref/source/aptitudes.h
+++ b/crawl-ref/source/aptitudes.h
@@ -1332,7 +1332,9 @@ static const species_skill_aptitude species_skill_aptitudes[] =
     APT(SP_OCTOPODE,        SK_EVOCATIONS,      1),
 
 #if TAG_MAJOR_VERSION == 34
-    // Need to keep this as long as the species are still around
+    // It seems that we need to keep this as long as SP_DJINNI/SP_LAVAORC
+    // (which  are needed for loading old Djinni saves) is still around.
+
     // SP_DJINNI
     APT(SP_DJINNI,          SK_FIGHTING,       -1),
     APT(SP_DJINNI,          SK_SHORT_BLADES,   -2),
--- a/crawl-ref/source/cloud.cc
+++ b/crawl-ref/source/cloud.cc
@@ -906,9 +906,7 @@ bool actor_cloud_immune(const actor &act, const cloud_struct &cloud)
     }
 
     // Qazlalites get immunity to clouds.
-    if (player && have_passive(passive_t::cloud_immunity))
-        return true;
-    return false;
+    return player && have_passive(passive_t::cloud_immunity);
 }
 
 // Returns a numeric resistance value for the actor's resistance to
--- a/crawl-ref/source/cloud.cc
+++ b/crawl-ref/source/cloud.cc
@@ -908,16 +908,6 @@ bool actor_cloud_immune(const actor &act, const cloud_struct &cloud)
     // Qazlalites get immunity to clouds.
     if (player && have_passive(passive_t::cloud_immunity))
         return true;
-#if TAG_MAJOR_VERSION == 34
-
-    if (player && you.species == SP_DJINNI
-        && (cloud.type == CLOUD_FIRE
-            || cloud.type == CLOUD_FOREST_FIRE))
-    {
-        return true;
-    }
-#endif
-
     return false;
 }
 
--- a/crawl-ref/source/files.cc
+++ b/crawl-ref/source/files.cc
@@ -79,6 +79,7 @@
 #include "unwind.h"
 #include "version.h"
 #include "view.h"
+#include "wiz-you.h"
 #include "xom.h"
 
 #ifdef __ANDROID__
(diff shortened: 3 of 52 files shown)


crawl-ref/source/aptitudes.h
```diff
@@ -1332,6 +1332,7 @@ static const species_skill_aptitude species_skill_aptitudes[] =
     APT(SP_OCTOPODE,        SK_EVOCATIONS,      1),
 
 #if TAG_MAJOR_VERSION == 34
+    // Need to keep this as long as the species are still around
```
amalloy: What does "still around" mean, here? As I understand it you have removed the species; do you mean until there are no save files anywhere with Djinni? If so, I think be more explicit in the comment.

crawl-ref/source/cloud.cc
```diff
@@ -908,16 +908,6 @@ bool actor_cloud_immune(const actor &act, const cloud_struct &cloud)
     // Qazlalites get immunity to clouds.
     if (player && have_passive(passive_t::cloud_immunity))
```
amalloy: Since you're removing the last interesting branch here, I'd just write this as:

    return (player && ...);

But that's a style choice; feel free to disagree.

crawl-ref/source/files.cc
```diff
@@ -2085,6 +2091,53 @@ bool get_save_version(reader &file, int &major, int &minor)
...
+    {
+        if (!yes_or_no("This <red>Lava Orc</red> save game cannot be loaded as-is. If you "
+                       "load it now, your character will be converted to a Hill Orc. Continue?"))
+        {
+            you.save->abort(); // don't even rewrite the header
+            delete you.save;
+            you.save = 0;
+            end(0, false, "Please load the save in an earlier version if you want to keep it as a Lava Orc.\n");
+        }
+        wizard_change_species_to(SP_HILL_ORC);
```
amalloy: Should we really be doing this with wizmode functions? It's probably okay I guess, but seems a little weird. Also, isn't it possible to compile without wizmode support? In those cases this would break, right? I think we could take this one function out of wiz-you and move it into someplace else, since we now need it even for non-wizard games.

crawl-ref/source/files.cc
```diff
@@ -2085,6 +2091,53 @@ bool get_save_version(reader &file, int &major, int &minor)
...
+    // At this point the character has been loaded but not resaved, but the grid, lua, stashes, etc have not been.
+#if TAG_MAJOR_VERSION == 34
+    if (you.species == SP_LAVA_ORC)
+    {
+        if (!yes_or_no("This <red>Lava Orc</red> save game cannot be loaded as-is. If you "
+                       "load it now, your character will be converted to a Hill Orc. Continue?"))
+        {
+            you.save->abort(); // don't even rewrite the header
+            delete you.save;
+            you.save = 0;
```
amalloy: We use nullptr for nulls, not 0. But can you remove the delete and the set to 0? It seems like you can just call abort and wait for normal cleanup to destroy the package*.

crawl-ref/source/files.cc
```diff
@@ -2085,6 +2091,53 @@ bool get_save_version(reader &file, int &major, int &minor)
...
+        return true;
+    }
+    if (you.species == SP_DJINNI)
+    {
+        if (!yes_or_no("This <red>Djinni</red> save game cannot be loaded as-is. If you "
+                       "load it now, your character will be converted to a Vine Stalker. Continue?"))
+        {
+            you.save->abort(); // don't even rewrite the header
+            delete you.save;
+            you.save = 0;
```
amalloy: Same comments here as above of course.

crawl-ref/source/player.h
```diff
@@ -200,6 +200,7 @@ class player : public actor
     int lives;
     int deaths;
 #if TAG_MAJOR_VERSION == 34
+    // Need to keep these fields for save compat reasons.
```
amalloy: I don't think we do. You can remove them from player.h, and then change the code in tags.cc to read the bytes and throw them away, rather than reading the bytes and storing them in the player data.

crawl-ref/source/player.cc
```diff
@@ -3785,18 +3696,6 @@ void dec_mp(int mp_loss, bool silent)
 
 void drain_mp(int loss)
```
amalloy: Maybe we can remove this function now, and just call dec_mp directly?

crawl-ref/source/files.cc
rawlins: this is totally right, shouldn't even build without WIZARD defined; not sure why travis didn't catch this.  I'm refactoring it out of wizmode (which is cleaner anyways).

crawl-ref/source/files.cc
rawlins: Leaving this for future investigation since it is used in a bunch of other places in files.cc.

crawl-ref/source/files.cc
rawlins: (It seems that travis doesn't actually test NOWIZARD=y)

rawlins:
>  For one, Dj still have their fakemuts as mentioned; it should be impossible to become a Dj at all, so those fakemuts can just be removed. Another I saw is that a lava orc attempting to walk into lava gets a different message than a hill orc does.

I should have gotten these wrapped in to the rebase I just did.  I decided not to add any checks yet to the change species code for the reasons we discussed, and also I'd need to check how that interacts with chardump loading.

rawlins:
> For one, Dj still have their fakemuts as mentioned; it should be impossible to become a Dj at all, so those fakemuts can just be removed. Another I saw is that a lava orc attempting to walk into lava gets a different message than a hill orc does.

I think I got these as part of the rebase I just did.  I decided not to add any checks to species-changing yet for the reasons we discussed, plus I'd need to check how it interacts with the chardump loading code, but this could go in later.  A good place for them would be in `wiz-you.cc:wizard_change_species`.

PR #484, opened 2017-03-03
make pikel less awkward

yrmvgh:
maybe pikel's fine as-is, maybe not. maybe taking out the creepy stuff would be an improvement?

*shrugs*


frozenca:
I think that keeping Pikel as-is would fit its slavery theme.

isloat:
I don't think rape/sexual slavery is appropriate for Crawl. Pikel still has plenty of slaver lines, and the new lines are solid.

PR #485, opened 2017-03-03
Goldified Spellbooks

Doesnty:
This set of commits changes spellbooks to function like gold, runes and the orb. Instead of sitting in inventory, picking up a book causes it to cease to exist and add spells to the player's library, which they can memorize from at any time.

-Learnable spells in the book are listed upon picking it up.
-Books with no usable or new spells are tagged useless. This was meant to make shopping painless, but had the consequence of making book acquirement much different; it will consistently not give useless books but is more prone to empty acquirements. To counter this, I added code to make it attempt to give a manual if it would give a useless spellbook.
-Trog loses book burning and gains absolutely nothing in return.
-Sif Muna is not directly changed. They should probably beam spells straight to the library instead of dropping books at your feet now. Book gifts run off of acquirement and thus are more useful. Sif also stops giving books once the player has every spell instead of drowning them in a torrent of redundant books.
-Vehumet is completely unchanged. Offered gifts are not added to the library and are withdrawn as normal.
-Background descriptions could use touching up as the player doesn't see their starting book anymore.
-On loading an old save, all books in the player's inventory have their contents added to the library and then disintegrate.


yrmvgh:
I, for one, am impressed! I'll pull this into my fork and let you know if I notice any problems. Thanks for doing this!

yrmvgh:
I have been informed that this is likely to get rebased or refactored multiple times and that might cause a mess in my repo. I still want this feature pretty bad, so I'll be watching it closely!

yrmvgh:
I tried this in yiufcrawl and it's working without problems so far! Here's a screenshot of an existing character from a previous version converting a pile of books from the floor to the library:
http://i.imgur.com/wMMVj7L.png

Do you have any plans for maybe a sort/filter feature in the player library?

isloat:
Do you have any further changes or fixes you'd like to make to this? There was some discussion a couple weeks ago on c-r-d and in ##-dev about this. We're generally favorable, but there's a couple things that need to be ironed out first.

1. Trog. Have him burn spellbooks on sight if you're following the god. Retain book-burning behavior for that one Trog vault. 
2. Spell list sorting. With a large library of spells the memorisable spell list becomes very unwieldly. Additional sorting options would significantly help this. It would be great to have a toggle between book sort, spell school sort, and spell level sort.

MarvinPA:
Book burning on-sight doesn't work, as has been previously discussed - there are many identifiable vaults with books that are guaranteed, or with high chances of books. The only way that'd function as a conduct is to burn all books that generate, and I don't think that's a very interesting conduct.

Doesnty:
i'm apparently persona non grata now so someone else gets to finish this i guess

bye

PR #487, opened 2017-03-06
Add Cyno species.

Floodkiller:
Add Cyno species, which are humanoids with the heads of dogs.

Cynos are unique in their skill training process: instead of having a
static aptitude, it dynamically scales based on the current skill level of
the skill they are attempting to train.  They start out with an effective
aptitude of +4, but lose -4 from that effective aptitude every 6 skill
levels (minimum aptitude of -8).  This makes it more efficient for Cynos
to spread out their skills over attempting to focus on raising a single
skill to a high level.

Cynos also have see invisible, and can detect any items within 7 tiles
with their strong nose.  Other statistics include a stat line of 7 Str/8
Int/9 Dex + sid/5 levels, -2 XP, +3 MR per level, and +0% HP/+0 MP.


isloat:
I expressed this earlier in ##-dev but while I like the overall species I think they need one more "flavor" thing to push them over the top. It's not urgent or necessary that this be added before any potential merge, though.

Is "Cyno" the name we want to use for this? Changing species abbreviation is messy as hell for scoring tabulation, so a new name (if any) should be decided on before merging.

MarvinPA:
I don't see that any of the problems with the aptitudes scaling at hard breakpoints have been fixed, so I'd definitely be opposed to merging them without addressing that (and as discussed in -dev previously, I also don't think that scaling them gradually would necessarily be good, and it may be a fundamentally flawed mechanic).

crawl-ref/source/enum.h
```diff
@@ -191,4 +191,4 @@ enum
 enum
 {
     NUM_MAX_DOLLS = 10,
-};
+};
```
isloat: There should be an end of file newline after line 194.

crawl-ref/source/spl-goditem.cc
```diff
@@ -486,18 +487,23 @@ void debuff_monster(monster &mon)
...
     {
+        //Check which god may be providing detect_items and set map_radius
         if (have_passive(passive_t::detect_items))
         {
             map_radius = min(you.piety / 20 - 1, LOS_DEFAULT_RANGE);
             if (map_radius <= 0)
                 return 0;
         }
-        else // MUT_JELLY_GROWTH
+        else if (you.mutation[MUT_JELLY_GROWTH]) // MUT_JELLY_GROWTH
```
isloat: Good fixes, but should be in its own commit IMO.

crawl-ref/source/spl-goditem.cc
```diff
@@ -486,18 +487,23 @@ void debuff_monster(monster &mon)
 int detect_items(int pow)
 {
     int items_found = 0;
-    int map_radius;
+    int map_radius = 0;
```
isloat: Good fixes, but should be in its own commit IMO.

Floodkiller:
It would probably be fast enough to add something like Fangs 1 or Shaggy Fur 1.  However, I assume the feedback was asking for a unique flavor addition.  I will continue brainstorming to see if I can come up with something.

The origin of the Cyno name was deriving it from 'cynocephalus', from Greek mythology: http://www.theoi.com/Phylos/Kunokephaloi.html  
I'm not attached to the name at all, so feel free to bikeshed.
_______________
Although I did not resolve the issue with the aptitudes changing at hard breakpoints, I will explain the reasons why I did submitted it anyway (as well as comparison examples):

1.  Some breakpoints are bad because they encourage no brainier decision making: it makes no sense to train a weapon skill past the required amount for min delay, because the benefits that more weapon skill provides after that point have lost the most important reason to train the skill.  Even though Cyno has breakpoints that adjust the cost per skill level at specific points, it never makes the option literally worse, just more expensive.  When it occurs (at a whole number with a message warning), the player will likely make a cost/value calculation to determine if it is worth continuing training now or later, and this evaluation can always change depending on the player's current strategic situation in the game.  Even though the breakpoint may make this more pronounced than normal, this is what the player does on any other species, at any point a whole skill number is reached.  I don't see this system as one that creates explicit no brainier choices.

2. Some breakpoints are bad because they are hidden from the player: the relationship between dodging and dexterity changes calculations based on dexterity's value, which is hidden from the player.  For the Cyno skill system, the player is shown their effective aptitude in the skill menu, as well as being able to turn on the skill cost option to see how much the skill will cost.  Additionally, the player is notified whenever the effective aptitude changes.  Finally, the skill level system progresses on experience gain, which means the player is given ample warning and time to avoid training a skill after a change in cost has occured, if they choose to do so.  Even though the step rate and size is not displayed directly to the player, there is enough information available that the only time being unspoiled on that information may affect gameplay is potions of experience.  Even so, the player is more concerned about their resulting skill levels over correct aptitude display, and the skill levels are rendered correctly.

3. Some breakpoints are bad because they require you to hit them to have a positive effect: see weapon skill effect on min-delay or spellcasting effect on spell levels available.  This is not applicable to the Cyno skill system.

Additionally, there was enough pushback when discussing alternative ways forward that I decided, in combination with the above reasoning, to push the species in their current state for a more direct devteam consensus.  If the species is rejected for merging for the breakpoints, I will proceed to investigate alternative implementations at that time.

crawl-ref/source/spl-goditem.cc
Floodkiller: Submitted as pull request #488.

crawl-ref/source/spl-goditem.cc
Floodkiller: Submitted as pull request #488.

crawl-ref/source/enum.h
Floodkiller: Pushed commit to re-add end of file newline.

Floodkiller:
Changes to god-passive.cc should also have been in their own pull request instead of this one; they were leftover fixes from a previous iteration of Cynos.  Submitted as #489.

isloat:
Fwiw I agree with Floodkiller's arguments about why 'hard breakpoints' aren't a problem in this case. I earlier proposed tying the skill drops to XL but that would have led to several very bad playerfeel situations (scrambling to get the right skills boosted before you level up) while putting even heavier importance on the early game than Crawl already has in the first place.

6-12-18-27 also doesn't directly interact with any particular mindelay breakpoints that I can think of off the top of my head. I think whips hit mindelay at 12 skill and maybe a couple other weapons? So Cynos aren't particularly encouraged to "game" the system any more beyond the breakpoints.

yrmvgh:
Yes, I tried this race a few times and found that hitting the breakpoints gave a nice sense of "mini achievement" or completion, similar to the good feel you get when a new demonspawn mutation reveals itself.

I think a "flavor additon" like the now-angelic frog race could help cynos, because the experimental branch on cbro has had pretty low activity and the race may not be grabbing people's attention or imaginations.

That said, Crawl is perfectly happy to keep unpopular species around if they have good gameplay, and I think Cyno has good gameplay. It can illustrate to players the value of "putting a few points in Evo", which is useful because players that start out by following the standard `MiBe.Axe` advice (which is good advice) often wind up aggressively pursuing mindelay on an Executioner's Axe at the expense of their defenses and ranged/wand options.

I'm skeptical of the "breakpoints are automatically bad by fiat" argument, and hope more people will try it out!

Floodkiller:
There was some criticism in regards to the current skill system being too lenient, so I made a comparison chart for brainstorming a system that might better capture the aspect of "quick to level early, but discouraged from raising to the higher levels that most other characters would aim for".  It is located here for further reference: https://docs.google.com/spreadsheets/d/1UpAgbzL8OENPac_ubosSXxroNSN8llOtZWkuAUtyXX8/edit?usp=sharing

Colors are based on the aptitudes on the right side of the sheet; if a cell matches a color, it means it is less than or equal to the aptitude with the matching color (matching with the highest aptitude possible).  I can add any other suggestions to the chart for visual comparison if none of the current options are appealing.

Floodkiller:
There was a bikeshedding discussion from renaming Cyno last night, so here's the suggestions I remembered and wrote down:
-Keep Cyno/Cy
-Bultungin/Bu
-Gnoll/Gn (would require making sure it doesn't interfere with winning Gnome games)
-Mountain/Hill/xxx Gnoll
-High Elf (I don't really like this name as the player base already has too much attachment/expectations for it)
-Werewolf
-Pupper

If there are any other suggestions, of if anyone wants to say their preference, feel free.  I am fine with pretty much everything except for High Elf (but I'll concede if that's what everyone else wants).

isloat:
After looking into it I'm partial to Bultungin, and Crawl does have a tradition of trying to draw from local mythologies and legends.

sooheon:
Bultungin is pretty cool, I wonder if we can add any of the elements from that mythology to spice up this species more? There's were-hyena transformation, vampirism, child-eating, and a general aura of savagery which is not portrayed by the current iteration of "pupper learns tricks quickly". Maybe something like vampires' bat transformation, dropping into hyena form for faster movement, a vampiric bite attack, and sense items/enemies (smell)?

isloat:
I much prefer Bultungin over Gnoll, and in general I'd like to see Crawl continue on pulling from lesser-known mythologies and traditions (even if we occasionally step on cultural translation toes here and there). On the other hand, "Gnoll" and their short attention spans does tie into the player not seeing any gnolls in the Dungeon after a certain point. 

The rest of the species looks good. The singular objection has been the implementation of skill scaling, and I genuinely cannot see any possible way to abuse the breakpoints nor why they're truly a problem. The entire purpose of the species is rapid early development and very significantly slower mid/late development: super-low granularity, as used in Floodkiller's implementation, gives players a "stopping point" and conveys dilettantism well.

I think directly adjusting aptitude levels was more clear, but if you're sticking with the custom XP table, can you add some comments explaining how exactly the values were decided on/generated? Looking at the values and comparing them to the spreadsheet I can see that some of them are the +4 and +0 XP values at 5/10, but I don't understand how 13693 or the later values were arrived upon, Comments would make this clearer for any devs in the future, especially 1+ year from now on when a lot of this discussion will have become dusty and deep in the archives.

MarvinPA:
The entire point of changing to the current system of skill scaling was so that you wouldn't be penalised for not excessively micromanaging your skills, since gradual skill levels matter instead of only integer breakpoints. I don't think that problem is a good thing to bring back in species form.

isloat:
You mentioned in an earlier comment that gradual scaling wouldn't be a good idea. Would a custom XP table still have the same problems of gradual scaling that you were thinking of?

There's still some micromanaging in the current system, in that players will want to occasionally check their skills menu and see whether a skill's relative XP cost is getting out of control, and then evaluate whether other skills are more worthwhile/efficient to be training instead. With this species, this XP curve would be steeper, with a longer and flatter 'head' to the curve.

Floodkiller:
@isloat: Each value on the table is showing the total experience required to reach that whole skill number.  This value is generated by taking the previous row's total experience value and adding however much experience would be required at that row's given aptitude (based on the column's header description).  I use total experience values on the main chart as that is how skill_exp_needed formats the experience to skill level table.  Using your example of 13693 for skill level 15, it takes the previous skill level's total exp (14, 10723) and adds the amount of experience required to move from 14 to 15 under an aptitude of -6 (2970).  For another example from the same column, 1433 for skill level 9 takes the previous experience total for skill level 8 (983) and adds the amount of experience to required to move from 8 to 9 under an aptitude of +0 (450), as per the column description (+4 until skill level 7, decay or reduce effective aptitude by 2 each level until at the floor of -6).  Thus, even though the total experience required amounts may be different than the static aptitudes, the cost still matches up to the corresponding aptitude and keeps the information presented to the player correct (983 to 1433 at Cyno effective +0 is the same experience cost as 1800 to 2250 at a static +0).  This can also be seen on the Differences Between Levels sheet (although I have not updated it to contain the newer columns).  Without doing it this way and just using the static aptitude's total exp cost, changing the aptitude would result in an experience deficit on the first level of the new effective aptitude that needs to be made up, as well as cause confusion to the player.

I use a separate skill table for Cynos in skill_exp_needed and represent the aptitude in the manual bonus column as I felt it was the least messy, resource intensive way to code it.  One alternative would have been to actually alter the aptitude and raise exp earned in that skill to match the static aptitude's value (avoiding the skill level from dropping on change), but screwing with the exp earned is rather messy in the background and screws up Wizard Mode for Cynos.  Another solution would be to call skill_exp_needed inside of itself for Cyno to dynamically calculate the proper total exp needed for each skill level at runtime, but this effectively creates the same amount of int arrays in memory (if not more) than just having a separate table with hard coded numbers for Cyno.

isloat:
I've merged this PR, with some changes and further commits. Floodkiller has done an exemplary job putting together a species that's unique and plays well within Crawl's design constraints, as well as subjecting this to testing on an experimental branch and incorporating a lot of feedback. I want to present a closing argument because the skill scaling has been brought up a few times and I don't mean to dismiss it out of hand by merging this species without comment. 

I pulled this locally, compiled, and then played around with this species for a couple hours. I saw that, in practice, the "breakpoints" were not a relevant concern to how I played. The relative skill cost was what I was concerned about, exactly the same as when I play a species that uses the default skill experience table. Because the species informs the player when the skill requirements become steeper, players won't be caught off guard that they've been pouring XP into a hard-to-train skill. 

For both standard species and for Bultungin, I found myself checking the skill menu occasionally to see whether the relative skill cost had gotten out of control and whether I could use my XP more efficiently by focusing on other skills instead. In fact the interface for Bultungin was *better* in this regard, since the game warns me in the exact situation where I'd consider switching skills! I did not feel as if I was pressured into micromanaging my skills, nor did I feel that I was being punished for not cleaving exactly to the breakpoints. While I wasn't particularly fixated on how close I was (or wasn't) to a "breakpoint", I definitely felt the impact of the custom skill curve. I think this is an ideal situation for the design goals of the species.

The distinct difference is that Bultungin's experience curve is much *curvier* than the relatively level slope of the standard skill experience curve. In practice the two curves function the same: at any given level, the amount of XP needed to advance to the next level is going to be significantly higher than the amount of XP needed to advance to this level. In Bultungins' case, "significantly higher" is "very, very much higher" sometime after about level 10-11.

That the curve starts becoming so steep for Bultungin at a level significantly before where I would normally be comfortable abandoning my "main" skills and focusing on other skills, differentiates noticeably the arc of their gameplay compared to other races. I don't think I can overstate how steep this curve is, and how much it affects their midgame and lategame. The game is still beatable, but it requires a distinctly different approach to floorgod and combat than it would for, e.g., Minotaur or Merfolk. This is a wonderful characteristic for a species to have, to be markedly different than other species and to require different decisions and approaches throughout the game.

I think the original characterisation of "breakpoints" (in part stemming from previous iterations on the species) was misleading and unfortunately colored subsequent discussion on the species.

(As a note: Bultungin probably should be excluded from the Ashenzari Loremaster banner for tournaments.)

PR #490, opened 2017-03-10
Allow (un)wielding items while berserk

Doesnty:
"You are too angry to draw your sword" is a ridiculous state to be in. Warnings surrounding intentional berserk stay in place as swapping to a proper weapon is a waste of precious berserk turns.

Also enables the same for ranged weapon-having monsters who get unintentionally berserked, in the interest of player-monster symmetry. Given that launchers are used by almost nothing in the places where moths of wrath spawn, I don't think this is a huge balance change.

### DIFF
--- a/crawl-ref/source/item-use.cc
+++ b/crawl-ref/source/item-use.cc
@@ -494,17 +494,6 @@ bool wield_weapon(bool auto_wield, int slot, bool show_weff_messages,
         return true;
     }
 
-    // Switching weapons while berserk is usually a waste of time.
-    if (you.berserk())
-    {
-        string prompt = "Really switch weapons while berserk?";
-        if (!yesno(prompt.c_str(), false, 'n'))
-        {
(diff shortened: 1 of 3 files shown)


isloat:
When testing this, I don't receive any warnings for swapping weapons while berserk.

Doesnty:
I mean that the "are you sure you want to drink this potion of berserk rage while wielding a longbow" messages would stay in place, but there should probably be warnings around that too, thanks. Will do.

MarvinPA:
It doesn't make any sense to me to prompt on every single swap attempt if the point of the change is to let you swap to your melee weapon if berserked while wielding a launcher. Just a prompt if you try to switch to a launcher/non-weapon seems reasonable (and it should definitely avoid prompting twice).

isloat:
This looks like it only warns for launchers, not nonweapons like MPA mentioned (that is, potions and other combat-unusable items).

PR #491, opened 2017-03-11
Blinker bad-mutation

yrmvgh:
all teleports (trap, scroll, teleportitis, monster "teleport other") become uncontrolled blinks

Like "mp-linked wands", this badmut has potential for providing a small upside... if you are wearing *tele items or suffering from teleportitis.


yrmvgh:
Could make it also affect controlled blinks (whether at the base level, or by making it a 2 level mut)

isloat:
Do you think this would be an interesting bad mutation? Would this outweigh the crippling of teleportation? It's essentially -Tele. Of course, teleportation can place you nearby on a bad roll even without this mutation.

The messages for gaining/losing Blinker 1 are too obscure. "You are afraid to teleport" conveys something more like "Something very very bad will happen if you teleport" instead of "You will teleport only a short distance".

yrmvgh:
I think it would be interesting, and I think there aren't enough mutations in the pool at present. This one won't just randomly lead to a death like Berserkitis does, and "uncontrolled" blink is still very much manipulateable. If you're in a dire enough emergency, there's always ?blink, which can be followed with a clever ?tele from the good spot you just chose.

I thought the messages were in line with other mutation messages, but I'm happy to make them mork explicit too. One moment. :)

isloat:
This mutation doesn't work.

1. If you read a ?teleportation, you instantly blink, instead of after a delay as with current teleportation behavior.
2. Rings of teleportation still teleport you regularly instead of blinking you.

yrmvgh:
I'm not entirely sure what you mean by 2, but let me know whether I successfully did what you asked! :)

For 1, I think keeping the delay but also killing the range is "essentially -Tele" as you cautioned against earlier, but skipping the delay can make the mut's downside occasionally interesting, and abled to be played around.

isloat:
Re: 2), because you originally applied the behavior change to *only* reading the scroll of teleportation, other sources of teleportation weren't affected. Because you moved the affected behavior into _teleport_player now it works properly. 

Still on the fence about this mutation in general. But more badmuts are good! Will think on this for a bit.

MarvinPA:
I'd agree that this shouldn't change the effect to be instant instead of delayed (but also that not changing that probably makes it too severe, I'm not really convinced that it's a great concept generally).

Also you have missing braces that cause all non-blink teleports to silently fail.

isloat:
What if, instead of using the explicit blink behavior, it just greatly reduced your teleport range? 

Could also pair it with a "good" mutation that gave the player a higher minimum range for teleportation.

yrmvgh:
isloat, I do like that idea, I'm just less enthusiastic about it, because it's your idea at this point, not mine. :) I probably can and will code that at some point, but I'm curious - why, on the crawl/crawl repo, do people not add commits to one another's pull requests?

Being able to do that is one of the wonderful things about collaboration on github, but it seems to be Not Done, ever. I wouldn't be annoyed in the slightest if people just made the change, instead of asking, "Hey, cool idea, but can you scrap it and code this other related idea instead?" (I know that's going to sound snarky, but I really don't mean any snark or hostility in it, I swear!)

isloat:
>why, on the crawl/crawl repo, do people not add commits to one another's pull requests?

I'm actually not sure. I think I looked into this a few months ago and it was something like, the pull requester had to explicitly give permission for this when making the PR? Will make a note to look into this further.

MarvinPA:
Closing, since as discussed above it's probably not a good idea as-is.

PR #496, opened 2017-03-16
Filter distant duplicates in stash search (Siegurt)

rawlins:
This is a basic implementation of an idea from Siegurt [on tavern](https://crawl.develz.org/tavern/viewtopic.php?f=8&t=23274&p=309432#p309432) to show only the
closest copies of items when searching stashes. It does this only for floor items that aren't missiles. This implementation always does the filtering -- at present, there is no way to disable it.

I'm not sure if this is the ideal way to do it, and this version is probably missing some stuff, so feedback is welcome.

Here are some screenshots to give a sense for what this looks like, taken from an actual player save (XL23) I had around for debugging something else (though I had at some point done &|).  

edit: see below for updated screenshots

### DIFF
--- a/crawl-ref/source/stash.cc
+++ b/crawl-ref/source/stash.cc
@@ -457,6 +457,7 @@ vector<stash_search_result> Stash::matches_search(
         {
             stash_search_result res;
             res.match = s;
+            res.primary_sort = item.name(DESC_QUALNAME);
             res.item = item;
             results.push_back(res);
         }
@@ -469,6 +470,7 @@ vector<stash_search_result> Stash::matches_search(
         {
             stash_search_result res;
             res.match = fdesc;
+            res.primary_sort = fdesc;
             results.push_back(res);
         }
     }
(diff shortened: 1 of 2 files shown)


MarvinPA:
This should probably only apply to non-consumables, rather than just non-missiles. Could also fold it into the "show/hide useless" toggle and default that to "hide", maybe?

I didn't test to check but at a glance it looks like carrying an unidentified "glowing dagger" might prevent you from finding other glowing daggers and just list them all as duplicates? For any unidentified glowing/enchanted/ego-but-pluses-unknown item, probably all copies should be listed (or at least they shouldn't count as duplicates of things that you're carrying).

rawlins:
Screenshots from after rebase, which hopefully incorporates all the feedback so far.

search for "sword":

screenshot 2017-03-17 10 18 55

search for "bolt":

screenshot 2017-03-17 10 18 23

search for "bolt" with = toggled:

screenshot 2017-03-17 10 18 55

PR #497, opened 2017-03-17
re-remove area_hot

yrmvgh:
at least I'm pretty sure it was removed previously


yrmvgh:
ok actually it was never removed before, my bad. gonna close this when travis-ci finishes, I always like to know whether it built or not.

PR #498, opened 2017-03-22
don't recommend LTO for cross-compiling to windows

yrmvgh:
~~according to geekosaur, LTO doesn't do much even when it works correctly.~~ what geeko said is actually more complicated than that. bytecode something or other.

and according to me, cross compiling from Arch to Windows with LTO=y causes teleport scroll crashes!

### DIFF
--- a/crawl-ref/docs/develop/release/guide.txt
+++ b/crawl-ref/docs/develop/release/guide.txt
@@ -116,7 +116,7 @@ for those the list begins with step 3.
    the installer and stand-alone zips respectively. Unless you're on msys, you
    need to specify CROSSHOST as well:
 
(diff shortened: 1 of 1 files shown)


yrmvgh:
[cross compile LTO windows crash.zip](https://github.com/crawl/crawl/files/860579/cross.compile.LTO.windows.crash.zip)

rawlins:
I don't know anything about cross-compiling or this option, but a crash that specific is probably revealing an otherwise hard-to-spot bug somewhere with teleporting.  Too bad it's windows.

rawlins:
I'm not sure any more why this is lingering so long but in a recent discussion (crawl-dev-20180330.lg) gammafunk mentioned that he doesn't use LTO for cross-compiling the official builds either, so I'm going to merge this.

PR #499, opened 2017-03-22
Small fix

chereshnev:
Invert conditional in xom.cc


chereshnev:
I think it was wrong to use the master branch.

PR #501, opened 2017-03-24
Invert conditional

Mnoleg:
Invert conditional (xom.cc).


neilmoore:
It needs curly braces, too.  Our convention is to use braces if either the condition or the body is longer than a single line.

Mnoleg:
Ok, did it

Mnoleg:
Thanks!

PR #502, opened 2017-03-28
Rework WJC martial attacks

PabloMansanet:
This revision of WJC is a result of a conversation between me and Lasty,
in which we managed to converge into a set of changes that have
potential to solve the issues with both the current and past
implementations of the god. It stems from the agreement that WJC is not
to be a combo god, or to dictate a specific flowchart of abilities to be
used one after the other, but a set of mobility (and mobility control)
tools that allow the player to be at the ideal position to take enemies
out safely, even in the open.

Fundamental changes:
* Lunge now deals 100% weapon damage and has no extra effects. This
makes sense in context of a reduction in damage to wall jump and
whirlwind, so lunge is now the only martial attack that can match
tabbing in terms of damage. This leaves Lunge a niche for dealing with
polearms, ranged monsters, blinkers, etcetera while not superseding tab
or allowing degenerate behaviour like swapping with minions to be
optimal.
* Whirlwind and wall jump now deal 75% normal damage. This makes them
desirable to use against groups but not against a single target, unless the
player is also interested in the mobility itself.
* Whirlwind now causes Dizzy instead of Slow. Dizzy causes monsters to
walk slowly, but does not affect other actions. This is in line with the
goal of making all of the god's tools be based on positioning, not
combat modifiers.
* ~Wall Jump can again be used anywhere, but now comes at a cost of 1
piety per jump. This cost is represented in three places: In the ^^
window, in the power description when obtained, and reflected in a
message after every wall jump. This alleviates the interface problem of
wall jump being available sometimes but not others, and prevents is use
as a level traversal tool. Due to its reduced damage, it will
necessarily be used as a mobility skill first.~
EDIT: As per the second commit, wall jump has been reworked again. Take a look at that commit for info: a730685b0bc7ae3ced4b87fe43e65d15085929df
* Serpent's Lash does not increase the chance to cause status effects.
Instead, its effect on martial strike damage has increased. This pushes
it further into an offense tool, and particularly rewards lunge,
allowing the player to be explosive when necessary.

These changes reduce the power level of the god severely, but also
define its role more clearly and move it away from the mass damage field
occupied by Uskayaw. I believe that the strength of wall jump is still
enough to offset the lost power, but in case a buff is necessary, we
also discussed the possibility of the god passively trading a fraction
of your AC for damage bonus, in order to both punish high mitigation
builds and increase offense across all builds, in a step towards the
melee glass cannon idea (This is not part of the PR)

### DIFF
--- a/crawl-ref/source/dat/descript/ability.txt
+++ b/crawl-ref/source/dat/descript/ability.txt
@@ -863,7 +863,7 @@ Serpent's Lash ability
 
 Allows you to move at supernatural speeds for a short distance, causing the
 next two movement actions taken to be instant. Martial attacks performed under
-the effect of Serpent's Lash are more effective.
+the effect of Serpent's Lash deal increased damage.
 %%%%
 Heaven On Earth ability
 
--- a/crawl-ref/source/dat/descript/gods.txt
+++ b/crawl-ref/source/dat/descript/gods.txt
@@ -315,17 +315,18 @@ below. They will also be taught to move at unfathomable speeds for short
 distances, and become able to summon a storm of golden clouds to increase their
 precision and damage for as long as they keep using martial attacks.
 
-  .......       ...<lightred>@</lightred>... (*) <lightred>Whirlwind</lightred>: Move between two tiles adjacent to
+  .......       ....... (*) <lightred>Lunge</lightred>: Move directly towards an enemy to strike.
+  ..<lightred>@</lightred>.O..  -->  ...<lightred>@</lightred>O.. 
+  .......       .......
+
+  .......       ...<lightred>@</lightred>... (**) <lightred>Whirlwind</lightred>: Move between two tiles adjacent to
   ...<lightred>@</lightred>O..  -->  ....O.. an enemy to deal a weak, dizzying strike.
   .......       .......
 
-  ..x....       ..x.... (**) <lightred>Wall Jump</lightred>: Move into a wall to jump away from 
-  ..x<lightred>@</lightred>O..  -->  ..x.O<lightred>@</lightred>. it in exchange for some piety, distracting and
-  ..x....       ..x.... dealing some damage to nearby foes.
+  ..x....       ..x.... (***) <lightred>Wall Jump</lightred>: Move into a wall to jump away. It  
+  ..x<lightred>@</lightred>O..  -->  ..x.O<lightred>@</lightred>. isn't faster than walking, but will distract and deal 
+  ..x....       ..x.... damage to nearby enemies.
 
-  .......       ....... (***) <lightred>Lunge</lightred>: Move directly towards an enemy to strike.
-  ..<lightred>@</lightred>.O..  -->  ...<lightred>@</lightred>O.. 
-  .......       .......
 %%%%
 Kikubaaqudgha powers
 
--- a/crawl-ref/source/dat/descript/gods.txt
+++ b/crawl-ref/source/dat/descript/gods.txt
@@ -316,15 +316,15 @@ distances, and become able to summon a storm of golden clouds to increase their
 precision and damage for as long as they keep using martial attacks.
 
   .......       ...<lightred>@</lightred>... (*) <lightred>Whirlwind</lightred>: Move between two tiles adjacent to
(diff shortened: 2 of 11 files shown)


PabloMansanet:
Distraction is a stronger effect at the moment, and it gives some variety too, I think having two separate effects can give the movement choices depth, but I'm open about this.

1 piety per jump seems fine after some testing. It generally makes sure that you use WJ to gain distance or to get out of a cornered situation, which is a more interesting use that walking jumping an enemy for damage.

alexjurkiewicz:
I'm concerned that people will accidentally walljump far more than they
mean to. Usually when I'm moving around a level I end up multiple times
running into walls while I mash keys...

PabloMansanet:
Alex, I proposed a solution for this:

The first time you move into a wall, you get the message "You put your feet/claws/tentacles on the wall" which doesn't consume any time. The second time you perform the actual jump.

This is a bit more deliberate than the current format while still requiring one less keypress than an ability, and not requiring menu navigation.

I'm still unsure if it's necessary, but it's worth knowing that in case the accidental jumps are a real problem for many people, there's a straight forward interface solution to them.

isloat:
This latest commit fixes a lot of what I personally found problematic about Wall Jump. Very good steps are being made here.

MarvinPA:
This sounds like not at all an improvement to me, it's now just making it extremely overcomplicated in an attempt to make it be usable everywhere, and still not solving the problem of it being a boring reliable escape.

isloat:
@PabloMansanet Testing this locally, the Slow (Dizzy? ingame it displays as Slow) status doesn't expire after it is applied. This needs to be fixed.

@MarvinPA Testing shows that the new iteration is *far* less reliable as an escape than the original wall jump (prior to trunk changes). It's still potentially an escape, of course, but I can create space only part of the time, and usually only against one (sometime two) adjacent monster. With more monsters, or if any monsters are nonadjacent, they easily close distance in the two player turns it takes to wall jump. The situations where I'm able to create distance involve both 1. a successfully applied distraction 2. jumping directly away from the monster, usually into open space.

Of course, since nonadjacent monsters aren't distracted, they get two free 'turns' to act against the player when they wall jump. This can be very dangerous (orc priests, et al) and it also gives nonranged monsters opportunities to move towards the player and frequently into a position obstructing the player from advancing further safely. Because of the double-turn, subsequent wall jumps aren't capable of creating any further advantage.

I do think this introduces a new concern that players will want to drag monsters around until they're positioned onto the right terrain layout. This isn't ideal, but energy randomization addresses that somewhat, and players *already* can just walk directly away from monsters. In theory, the player could also repeatedly wall jump until they get a successful distraction, but if the monster is worth running away from, then it's not going to be worth exposing yourself to multiple attacks from that monster. 

I do not agree that the implementation is "extremely overcomplicated" and characterizing it that way seems like an exaggeration -- especially compared to some of the gnarlier parts of Crawl's codebase.

PabloMansanet:
@isloat Oops, I'm on it!

MarvinPA:
@Brannock: I'm not referring to the implementation or code at all, I'm referring to the design. Messing with monster movement energy at all in the first place isn't a great start, but only doing it if they're adjacent to some set of squares that's different to the set of squares you're attacking just seems ridiculous to me. Especially compared with the extremely simple and functional approach that already exists without all these attempts to hack around the various problems of enabling wall jump everywhere.

isloat:
@MarvinPA Ah, I wasn't clear on you meaning the design, thanks for clarifying.

I don't personally see a problem with conditionally altering monster energy, but I do recognize that the target-locked version of Wall Jump is much simpler in how it behaves and interacts with the rest of the game.

PabloMansanet:
I don't think it's particularly difficult for the player to understand that enemies close to your trajectory will be distracted. Once they have understood that, it's not that much of a stretch for them to realize that monsters aren't acting at 100% when you wall jump around them, and there is a message to show this "XXX is briefly distracted by your jump"

isloat:
@PabloMansanet Do you mind making a version of this without the changes to Wall Jump? WJ is incredibly controversial for whatever reason, and I and Lasty (to my knowledge) liked the other changes you've made. We can continue discussion on how to handle Wall Jump.

PR #503, opened 2017-03-29
Rework monk start to be more generalist

sooheon:
Some context may be found here: https://crawl.develz.org/tavern/viewtopic.php?f=8&t=23161

And the commit msg for the change: 

```
Distribute the weapon skill points to fighting (4), dodging (4), and
stealth (4). This results in Monks starting with the highest fighting and
dodging in the game, and tied for the highest stealth, but no weapon or
related skills.

Previous rationale for moving to a weapon choice start was that having such a
high skill investment into UC disallowed experimentation. But having weapon
choice results in Monks playing far too much like weaker Gladiators. This change
mirrors the Monks' saved-up piety: starting with high generalist fighting skills
is essentially saved up xp to be used toward a weapon of choice, once one is
taken from the floor or an enemy. Though many melee characters start by focusing
weapon skill anyways, Monks should see a better return on this strategy given
that their supporting skills begin at a higher point.

Flavor wise, this fits the idea of a monk who has trained their body for health
and discipline, but has not studied a particular weapon. Imagine an ascetic
stepping foot into the wrong cave, and being thrown into an adventure. Gameplay
wise, it gives a truly different starting experience from Fi/Gl/Hu in the
starting levels, and it jives corresponds well to the piety gimmick, in that it
is a boost to power *once a condition has been met* (finding an altar, finding a
weapon).

In the majority of cases, this should be a buff. Monks go from 11 skill points
total to 12, and skills are at higher levels, which are more valuable. As soon
as a monk picks up any weapon, they should be in a better place than old monks.
Further, this is a buff in the case that one quickly finds a good weapon which
is different from the type chosen at chargen (0 wasted xp). In the first ~30
turns, this is a nerf to survivability, requiring maximal use of the newly
buffed dodge and stealth.
```

Any feedback is much appreciated.

### DIFF
--- a/crawl-ref/source/job-data.h
+++ b/crawl-ref/source/job-data.h
@@ -188,9 +188,8 @@ static const map<job_type, job_def> job_data =
     { SP_DEEP_DWARF, SP_HILL_ORC, SP_TROLL, SP_CENTAUR, SP_MERFOLK,
       SP_GARGOYLE, SP_DEMONSPAWN, },
     { "robe" },
-    WCHOICE_PLAIN,
-    { { SK_FIGHTING, 3 }, { SK_WEAPON, 3 }, { SK_DODGING, 3 },
-      { SK_STEALTH, 2 }, },
+    WCHOICE_NONE,
+    { { SK_FIGHTING, 4 }, { SK_DODGING, 4 }, { SK_STEALTH, 4 }, },
 } },
 
 { JOB_NECROMANCER, {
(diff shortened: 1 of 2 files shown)


isloat:
I like detaching Monks from weapons very much. Their current weapon choices do nothing to meaningfully differentiate the background, and ends up infringing upon other Warrior background starts. Turning them into "you start with 'nothing'" is distinct from all other backgrounds, and mirrors Wanderers (who start with 'whatever'). Excellent idea and contribution, thank you.

PR #505, opened 2017-03-30
allow monsters to ignore AC if player is incap

yrmvgh:
...acitated. If the player is asleep or paralyzed, there's no reason for a monster to swing wildly at the player's armor, they can simply attack the player's face meat, or other gaps in the armour.

This may help correct AC being significantly more reliable than the other two defenses. Incapacitation will be more universally dangerous. This will hopefully result in fewer laments of "I forgot that (dream sheep, grinder, etc) are actually dangerous when you play a mage", because incap players won't sometimes happen to shrug off attacks during incapacitation and forget to take incap seriously when they play a squishy character.

One way to walk back the intensity of this change would be to allow defense bonuses (such as AC/EV rings, god bonuses, scale mutations, spells, or even the player's size category) to apply during incapacitation, under the justification that they naturally protect the character from harm regardless of the character's current state of consciousness - there's no "gap" to exploit and slice into. So only "base" ev/ac/sh would be set to 0 during incap.

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -5956,6 +5956,9 @@ int player::base_ac(int scale) const
 
 int player::armour_class(bool /*calc_unid*/) const
 {
+    if (you.incapacitated())
+       return 0;
+ 
     const int scale = 100;
     int AC = base_ac(scale);
 


isloat:
>One way to walk back the intensity of this change would be to allow defense bonuses (such as AC/EV rings, god bonuses, scale mutations, spells, or even the player's size category) to apply during incapacitation, under the justification that they naturally protect the character from harm regardless of the character's current state of consciousness

If you can implement this I'd like to see it.

MarvinPA:
Making some types of AC/EV work differently to others sounds not at all good to me - that behaviour is deliberately avoided for SH, and was recently removed for EV (see 2783ae5f). Even without that change I definitely don't see this being an improvement.

isloat:
I didn't know about the precedent in 2783ae5 or about the SH avoidance. I agree.

yrmvgh:
@isloat @MarvinPA thanks for the thoughts! I find that opening a PR is better for discussion than talking in ##crawl-dev, because it circumvents the timezone issue and puts your idea in a place where everyone will actually see it (unlike the wiki). It still bugs me that AC works fine during paralysis, given that you're defenseless and open to be "critically hit", but if this is the wrong way to go about fixing the AC superiority to other defenses, that's cool too. Back to the drawing board!

PR #506, opened 2017-04-02
Remove contamination from Haste.

Floodkiller:
Removes the contamination from gaining and extending the Haste effect, as
well as the per-turn contamination gain (and prevention of contamination
dissipation) while Haste is ongoing.

While Haste was a spell (and to a lesser degree, as a wand), it made sense
to limit the amount of times Haste was allowed to be used per tactical
encounter, as it could effectively be used as many times as desired
throughout the game.  However, now that the only source of Haste are
potions and unreliable god effects where the player cannot necessarily
guarantee they will receive Haste (Potion Petition, calling a potion shop,
decks of escape, Xom), tactical limitation no longer seems necessary.

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -3330,7 +3330,6 @@ void bolt::affect_player_enchantment(bool resistible)
     case BEAM_HASTE:
         haste_player(40 + random2(ench_power));
         did_god_conduct(DID_HASTY, 10, blame_player);
-        contaminate_player(750 + random2(500), blame_player);
         obvious_effect = true;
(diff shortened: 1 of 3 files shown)


Floodkiller:
It was also brought up that the original reason that contamination was added to Haste was to prevent combining it with Invis.  I would argue that the same reasoning in the commit message still applies to this: with a limited amount of Haste available per game, the player does not have the ability to abuse this combination to any significant degree before running out of potions of haste (unlike before where they could cast Haste at any point in combination with Invis, or use a wand of hasting and continuously recharge it).

PR #509, opened 2017-04-06
make hydras less baroque (hellcrawl)

yrmvgh:


Hellmonk:
prob would want to clean this up if there's actual interest in merging this to big boy crawl

yrmvgh:
meh

PR #511, opened 2017-04-10
Make Cutlasses Long Blades

mprobinson:
All other short blades do pure piercing damage, but cutlasses also
do slicing damage, which causes surprising head growth when
attacking hydras, contradicting the hydra description. Change to
Long Blades, which avoids this, and adds thematically appropriate
riposte attacks.


MarvinPA:
Making it do the same type of damage as other short blades and not chop hydra heads seems much better to me, since it's supposed to just be a renamed rapier in the first place - there are also a huge number of long blade unrands already and not very many interesting short blades.

PR #512, opened 2017-04-11
De-Eurocentricism

alexjurkiewicz:
Rename some items brought up in https://crawl.develz.org/tavern/viewtopic.php?f=8&t=23390


neilmoore:
"Forget-Me-Not" is already a randart name, so it might be best to avoid that (or remove it from dat/database/rand_arm.txt )

MarvinPA:
I don't have a problem with the idea generally but these specific changes don't feel like improvements. For the ring of Shaolin it'd probably be fine to just stick with something simple, as with the ring of Robustness and brooch of Shielding. I don't understand the theme change of the First Bow giving Int/MP at all (also I know there's been some discussion in favour of keeping the current name of Krishna in the past, can't look up the commits right now but that might be worth a look). I also don't see any reason for the longbow unrands to have matching pluses - if anything it'd probably be better to reduce Sharnga's pluses a bunch compared to the others due to its brand.

The lajatang change definitely seems fine - Agnes's new theme and description is also a fairly bad offender there, in that it also specifically refers to lajatangs as exotic.

alexjurkiewicz:
Thanks both. Updated based on your feedback @MarvinPA (which invalidates @neilmoore's)

isloat:
Merged.

PR #513, opened 2017-04-12
Replace spell success breakpoints with minmay's polynomial

rawlins:
Lots of consensus that something like this is a good idea -- replacing
these pre-dcss breakpoints has been under discussion since at least
2014. This version uses minmay's polynomial solution explained at:
https://crawl.develz.org/tavern/viewtopic.php?f=8&t=23414

The argument that swayed me to make a branch: "One great effect of
getting rid of chance_breaks is that spells transition from 100% to 99%
to 98% etc. etc. earlier, instead of staying at 100% for ages and then
quickly falling. A 99% fail spell is still useless but it makes it much
easier for unspoiled players to tell how close they are to getting it
castable, compared to seeing both level 5 and level 9 spells as 100%
fail."

I've done some quick checking but not careful checking, and have mixed
feelings about introducing doubles where there were only wholesome ints,
but it's just one line, what could go wrong.

### DIFF
--- a/crawl-ref/source/spl-cast.cc
+++ b/crawl-ref/source/spl-cast.cc
@@ -326,13 +326,16 @@ int raw_spell_fail(spell_type spell)
 
     // Don't cap power for failure rate purposes.
     // scale by 6, which I guess was chosen because it seems to work.
-    chance -= calc_spell_power(spell, false, true, false, 6); // realistic range: -6 to -366 (before scale -1 to -61)
+    // realistic range for spellpower: -6 to -366 (before scale -1 to -61)
+    chance -= calc_spell_power(spell, false, true, false, 6);
     chance -= (you.intel() * 2); // realistic range: -2 to -70
 
     const int armour_shield_penalty = player_armour_shield_spell_penalty();
     dprf("Armour+Shield spell failure penalty: %d", armour_shield_penalty);
-    chance += armour_shield_penalty; // realistic range: 0 to 500 in extreme cases.
-                                     // A midlevel melee character in plate might have 40 or 50, and a caster in a robe would usually have 0.
+    chance += armour_shield_penalty; // range: 0 to 500 in extreme cases.
+                                     // A midlevel melee character in plate
+                                     // might have 40 or 50, and a caster in a
+                                     // robe would usually have 0.
 
     static const int difficulty_by_level[] =
     {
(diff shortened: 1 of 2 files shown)


neilmoore:
Minmay already suggested avoiding floating-point by scaling the numbers by a large value and dividing at the end.  I'd suggest 262144, since that's a power of two and reasonably close to 1/(the first exponent).  That would give you something like:

    (x*x*x + 426*x*x + 82670*x + 6983254) / 262144

Also, you can use [Horner's rule](https://en.wikipedia.org/wiki/Horner%27s_method) to compute this with only 2 multiplications instead of 5 (or 3 multiplications instead of 6, for the version with a coefficient on the x^3 term):

   (((x + 426)*x + 82670)*x + 6983254) / 262144

rawlins:
This version is done-ish.  Right now the PR does two things:

- change `chance_breaks` into a polynomial; I used |amethyst's version in the end.
- add more resolution to the spellpower calculation.

**remaining decision point**: There is *still a breakpoint*, at around 45 -- where the old version switched to linear, `f(x)=x`, and I have replicated this behavior.  It's actually at 43 here because that's the relevant fixed point for this polynomial.  The polynomial continues slightly to the right of this line, so if the polynomial is used all the way up, this pushes fail rates above 43 down but not majorly so, which gives the appearance of training doing something earlier.  I think this could actually be desirable on simplicity grounds, and won't realistically impact casting safety for most cases (though with sif protection it may lead to earlier casting of certain spells).  Here's a wolfram alpha graph of the situation:

[plot of y=(((x + 426)*x + 82670)*x + 6983254) / 262144 and y=x and x=100 and y=100 for x=-183 to 126](https://www.wolframalpha.com/input/?i=plot+of+y%3D(((x+%2B+426)*x+%2B+82670)*x+%2B+6983254)+%2F+262144+and+y%3Dx+and+x%3D100+and+y%3D100+for+x%3D-183+to+126)

It's also worth noting that this set of changes doesn't entirely reduce the steppiness of fail rates, because of resolution of the actual die rolls.  That is, the true fail rate is slightly coarser than integer percentages between about 15% and 85%.  To deal with this, you'd need to add resolution to raw_spell_fail itself, and then change the rolls (not to mention _get_true_fail_rate). I'm not going to do that myself right now.

PR #515, opened 2017-04-15
New set of hopefully more neutral titles for Beogh

Doesnty:
Evangelist and Apostle as titles for an evil god of orcs are kind of uncomfortable, so I put together a new set from scratch; open to better ideas on these.


isloat:
I think these are too much on the generic side, sorry, and breaks from Beogh's established theme, flavor, and mechanics as an evangelist. Thank you for submitting a PR, though!

PR #518, opened 2017-04-17
Add new L7 Charms spell, Overwhelming Strike

Floodkiller:
Adds a new Level 7 Charms spell: Overwhelming Strike.  This spell, when
cast, gives the player the Strike status for a short period of time.  If
the player successfully hits an enemy while under the Strike status, a
powerful strike will be performed on the enemy based on spellpower and
the player's current remaining MP at the time of the strike.  After the
strike is performed, the player's MP is drained to 0 and the Strike status
is removed.

Overwhelming Strike can be found in the book of Enchantments.


Floodkiller:
Overwhelming Strike is one of many pull requests I plan on submitting as part of an overall Charms reform attempt I am working on.  This reform attempts to restructure Charms from a generic 'buff spells' school towards a more focused 'short-term tactical buffs/melee enhancements with advantages and disadvantages'.  This reform will also be looking at reworking/moving/removing existing Charms spells that do not currently fit this mold.

Overwhelming Strike has the advantage of moderate to incredibly large amounts of damage within a single turn to a single target.  However, it has multiple disadvantages that discourage reckless use.

First, it requires the user to successfully hit the target with a melee weapon.  This makes the spell less useful against evasive/blocking targets, and harder to land within the short spell duration with slower and less accurate weapons.  The range also requires the user to have some way to close the distance to the enemy before the spell wears off (or already be near/in melee range with the enemy).  Tying the spell to a status effect that triggers on melee instead of a range 1 spell also lets the user know that MP spent casting the spell does not contribute towards the damage portion of the spell.

Second, draining MP to 0 after use shuts down other MP using options (spells, god abilities, etc.) for the rest of the encounter, which is significant for anyone who has invested enough experience into spells to make Overwhelming Strike both castable in combat and useful (due to how heavily spellcasting affects the damage output).

Third, the spell is MP inefficient in two ways.  The first way is that it has a worse MP to damage ratio than most other damage dealing spells, even at max spellpower (which is harder to reach than elemental based spells due to only being able to boost it with generic spell enhancers and Int).  The second way is that it wastes much more MP on 'overkill' damage than any other single target spell, to a degree that it is comparable to firing 3-4 iron shots into a corpse after killing it 1-2.

If the current formula is considered too powerful, it would be recommended to either change the MP to dice ratio (currently 1:1) or the spellpower to sides ratio (currently 10:1 using div_rand_round) to what is considered reasonable.

rawlins:
I'm going to go ahead and close this PR (doing a pass on old PRs/issues), given the lack of response from anyone else after such a long time. I'm not entirely sure what the lack of response means, but I'd guess it's mostly something in the neighborhood of: "this proposal isn't obviously out, but I don't myself want to merge it".

I can tell you why I haven't merged it, which doesn't really speak for anyone else: I don't think that adding more confusing touch-type spells is great from an interface or design perspective (and I'm already on the fence about removing confusing touch). Also, while I understand the motivation behind the MP-drain mechanic, I don't think it really fits as a spell effect/consequence.

PR #519, opened 2017-04-17
break up a swastika or two

yrmvgh:
roderic_ancient_swastika swastika could, on an unlucky shuffle, wind up looking highly similar to the symbol now associated with the NSDAP of the 1920s.

roderic_greek_sauvastika could as well, when flipped by the dungeon builder.

Folks are more than welcome to make further edits to the PR; I have ticked the checkbox that allows it.
image


gammafunk:
The relevant vaults were removed in 0.21-a0-193-g085b6a0, so this commit is no longer necessary, thanks.

PR #521, opened 2017-04-19
Rework Zin's mutation protection.

alexjurkiewicz:
Instead of granting passive mutation resistance and a capstone curemut
ability, Zin will periodically remove mutations from worshippers.

With the recent !mutation change, the only way Zinites can remove
mutations is Zin's capstone ability. Any mutations gained after using
this ability must be kept "for life" (or by changing to another god,
quaffing !mutation and then changing back to Zin, which is fiddly).

A lot of the justification for reworking !mutation also applies to
changing how Zin works. Zin is currently a way to "opt out" of the
mutation game (however imperfectly this works in practice, as above),
which isn't that interesting a benefit of worshipping them. Zinites
simply ignore tactical positioning against malmutators. Replacing rMut
with slower mutation removal means that Zin reduces, but doesn't
completely eliminate the cost of positioning mistakes against
malmutators.

The mutation removal frequency is meant to be significantly higher
than weapon gifts from Trog/Okawaru, but not as frequent as mutation
gifts from Jiyva (the timeout is the same, but counts down 25% slower
Jiyva skips the gift timeout slowdown).

Mutation cleansing starts at 0*, to prevent any possibility for players
to want to keep their piety below a certain threshold.

(The mutation cleansing speed and minimum piety threshold may need to
be tweaked in future.)


isloat:
This is a very good change but will probably need to wait for the 0.21 dev cycle.

crawl-ref/source/religion.cc
```diff
@@ -991,7 +989,7 @@ static bool _give_nemelex_gift(bool forced = false)
             // included in default force_more_message
             canned_msg(MSG_SOMETHING_APPEARS);
 
-            _inc_gift_timeout(5 + random2avg(9, 2));
+            _inc_gift_timeout(15 + random2avg(9, 2));
```
rawlins: why does this change the nemelex gift timeout?

crawl-ref/source/religion.cc
```diff
@@ -1167,6 +1165,20 @@ static bool _give_pakellas_gift()
     return false;
 }
 
+static bool _give_zin_gift()
+{
+    if (!you.how_mutated()) {
```
rawlins: bracketing style :)

crawl-ref/source/religion.cc
```diff
@@ -1167,6 +1165,20 @@ static bool _give_pakellas_gift()
...
+        return false;
+    }
+    bool success = delete_mutation(RANDOM_MUTATION, "Zin's grace", true,
+                              true, true);
+    if (success) {
```
rawlins: bracketing style

crawl-ref/source/ability-type.h
```diff
@@ -82,7 +82,9 @@ enum ability_type
     ABIL_ZIN_VITALISATION,
     ABIL_ZIN_IMPRISON,
     ABIL_ZIN_SANCTUARY,
+#if TAG_MAJOR_VERSION == 34
     ABIL_ZIN_CURE_ALL_MUTATIONS,
```
rawlins: I'm probably missing something, but couldn't this ability just be removed entirely?  On that note, does this patch need to do something in tags.cc to fix up the abilities list?

crawl-ref/source/dat/descript/de/gods.txt
```diff
@@ -121,16 +121,6 @@ verzehren, die von ihren Anhängern woanders im Verlies liegen gelassen wurden.
 Besonders strenggläubige Anhänger werden genauso formlos wie ihr Gott, weil
 ihre Eigenschaften und Mutationen sich unter Jiyvas Einfluss verändern.
 %%%%
-Zin powers
```
rawlins: I don't know much about how crawl translations work but it seems a shame to delete all these because one word is wrong, maybe being the translational equivalent of protected isn't so far off the new version?

alexjurkiewicz: I believe the idea is translators would look at a different of the file, see the entry is deleted and then know the entry needs an update -- it's stable version focused IOW. But I can readd if you prefer

crawl-ref/source/ability-type.h
alexjurkiewicz: I assumed removing this would have saved compare issues. Action counts in morgue for example.

rawlins: yeah good point. I think you can remove everything but the entry in ability_type, based on 4eb006deb9de as an example template I found; just a matter of testing save importing with a character that has full piety.

alexjurkiewicz:
Will be merged as part of #522

PR #526, opened 2017-04-26
Gnoll clean-up/adjustments

Floodkiller:
Contains a new ability for Gnolls, adjustments to the Gnoll XP table, adjustment to Gnoll stat gain, and some code cleanup.


isloat:
some talk from ##-dev, for context, when this patch was initially presented for feedback:

<Brannock> I like shifting some of their power with aptitudes into an ability, as well as making the curve harsher
<Brannock> looking at that new XP table I think it'll help solve the issue of people trying to ignore the species's constraints by focusing on a minimum few skills
<Brannock> because that curve becomes really whoa
<alexjurkiewicz> yeah. How much XP do you need to reach level 14 in a skill now, compared to current gnoll and compared to 0 apt?
<alexjurkiewicz> because to me that's the sort of important cut off, weapon mindelay with most classes
<Brannock> 14? 6000 baseline, old gnoll: 10723, new gnoll: 15076

<Brannock> 15076 is still doable if you're very dedicated, which I think is fine (gnoll shouldn't lock you out *totally*), but it begins to heavily suggest "level something else"

<Floodkiller> many of the complaints about gnolls were about how bland they were, with the shifting aptitudes not being interesting enough to stand on its own
<kramin> I don't really understand that complaint :P
<kramin> I like the aptitude gimmick
<Pinkbeast> Is making the doggo apt lock even harder really the answer?
<kramin> yes imo
<kramin> needs to actually be hard to win the game without hybridising with gnolls
<Floodkiller> at least 2 three rune games went with a Trog-esque skill build, without Trog; another good half of the games hyper focused a couple skills and barely spread out unless required

some objections from minmay:

<minmay> My point is that there are a ton of ways to excel in a specialty without having much in the way of skill levels: good short blades, demon weapons, some spells, several gods
<minmay> If the goal is for "hybridizing" to be better than not hybridizing, that goal has already been met - in fact, it's met by the majority of species in the game
<minmay> But if you want specializing with gnolls to be way worse than doing it with other races I think you need more than an apt gimmick because specialization isn't that strongly based on skills<minmay> there are multiple gods that let you specialize in using allies to kill things, and be REALLY good at it, without any skills at double-digit levels
<minmay> but even without gods, weapons don't have skill requirements and spells have wizardry
<minmay> I think a lot of the complaints about blandness, incidentally, stem from a perception of highly specialized characters as exciting, and more optimally built characters as boring

--------------------

on a related (but not very relevant to this PR) topic I think the whole skill and aptitude system Crawl has is fundamentally flawed and a lot of what minmay objects to is rooted more in that system than the specific implementation of Gnoll; and the team has previously focused on (and eliminated) a few species that differentiated themselves solely through their aptitudes.

However, I think Gnolls take a different approach to this problem, and give a new twist on the system. I think strengthening the potency of the XP curve (which has a significant dampening impact on their mid/lategame), while also giving them an additional tool (Bark) helps to make them stand out as a species while moving them further away from "Human, but frontloaded".

MarvinPA:
Giving them fear on demand to make their already easy early game (after the point of +5 apts at the very start) even easier doesn't sound like an improvement in terms of balance to me (or particularly interesting as just a consumable turned into a species ability). I also don't see any good reason it should be usable while confused - currently that's consistently only allowed for divine self-healing, and it could probably be further restricted just to divine abilities that specifically cure confusion, even.

I'd definitely agree with minmay that fundamentally, messing with their aptitudes in this way won't work and strongly encourages them to focus on things with very high reward for low investment (like summons etc) - a bigger problem is certainly the way the system works overall and the fact that those things with such high reward for low investment exist in the first place, but while they do, it seems like a bad idea to build a species that's guaranteed to focus on them even more than others.

alexjurkiewicz:
You could give the ability at XL10 or something

Floodkiller:
@MarvinPA: Considering Cause Fear also exists in a starting spell book (Arcane Marksman), Gnolls already have access to early fear on demand if they desire (and the spell is better than the ability  because it is only noise 3 instead of noise 25, accessible at XL 4, and not tied to a breath timer).

Likewise, Spit Poison (Naga), Conjure Flame/Mephitic Cloud/scroll of fog (Draconian), Regen (Troll/VS) may be uninteresting on their own, but usually combine with other features on the species to make for an interesting mechanic.  Fearsome Bark was designed as a compliment to lower MR, as I don't think Gnolls currently have anything that actually justifies keeping it other than flavor with short attention span.  This is also why it was designed to be used while confused, which will likely happen more with Gnolls than any other species.  If the player gets into a bad fight (whether confused or affected by any other hex), they can attempt to scare the current encounter away and recover at the cost of potentially attracting others.

Another factor into why it is allowed to be used while confused is that shouting (making a loud noise) is already allowed under confusion, and Fearsome Bark is not tied to a directional or concentration-based aspect in flavor or mechanics that would justify being inhibited by confusion, except for internal mechanical consistency with abilities as a whole.  In fact, I would argue that the mechanical consistency aspect of disallowing spells/abilities during confusion is an issue, as it effectively makes Confusion a pseudo-paralysis that must be immediately cured and the situation escaped unless the player isn't in any actual danger.  I think many spells and abilities (not all, specifically concentration/multi turn spells and abilities) should be allowed during confusion, but at the cost of confusion affecting the success rates and directional/locational aspects (locking scrolls out due to inability to read during confusion is still fine in my opinion).
---------------
The XP adjustment aspect of this pull request was concentrated on fixing the trend I spotted in winning games (forcefully ignoring skill diversity and focusing a couple skills) and slightly increase the XP required to build up to the soft cap (placed around 12) before feature freeze/release; it is not an attempt at redesigning the system from scratch.  If the overall opinion is to hold back Gnoll from release due to disagreement with the current implementation, then I will continue developing and testing alternative systems until I find something that I feel is acceptable to submit.

isloat:
Responding to a few separate thoughts: the new XP table, Fearsome Bark, and variety of ROI in Crawl skilling.

* Whether or not the current implementation of Gnoll is desirable to implement in Crawl is, I think, independent of whether or not these changes *improve* the current implementation of Gnoll. The new XP table does a better job of pushing players towards broadening their builds and playing around with the impact of Gnoll on skill building. I think that is a definite improvement over the current situation, and I'm inclined to push this (not right away) so that Gnolls are in a better state until/if they're improved (if at all) instead of leaving them to sit unimproved. 

  I've taken the time to playtest the new implementation and I've found it much more difficult to "brute force" the species' constraints on skill building, to the point where if I kept on trying to brute force a focused build I would find myself significantly outclassed by the Dungeon's power curve. We have many, many other species that this applies to! (Good luck trying to play a dedicated Troll stabber.) Gnoll stands out here quite remarkably compared to other species: while others have a few certain builds that are generally unfeasible to focus on, for Gnoll it's _all_ possible builds that are very XP-inefficient for focusing on. 

* The thing that stands out to me about Fearsome Bark and why I think it's a good ability (even if it "merely" duplicates ?fear but with greater noise) is because it interacts with the rest of the speciation differences for Gnoll. Generalist builds tend to have a lot of trouble with crowds, compared to dedicated and focused builds that can pick off monsters one by one. Fearsome Bark gives Gnoll a way to clear out these crowds, at the very real risk of drawing _more_ enemies that they'll be even less able to handle. 

  I think this compares well to the Naga poison spit - on any other species baseline poison spit would be a problem (kiting) but on Naga it gives them a way to soften up approaching enemies and interacts with their general slowness and durability. 

  And, in general, I think it's better for Gnoll to have more going on, rather than fewer, especially since more than a few people (who are both favorable and unfavorable towards the species) have said that the XP table is insufficient on its own to differentiate them as a species. Compare Barachim where adding the LOS interaction rounded them out, and Vine Stalker which had multiple unique (at the time) mechanics, and have needed fairly little adjustments since then.

  I have no real opinion on allowing or forbidding it to be used while confused. Floodkiller's argument in favor of it seems sensible to me. Consistency around confusion (while being amusingly antithetical to the idea of 'confusion', but I digress) could be improved, though. 

* Entirely divergent from this conversation on Gnoll, I want to precaution against squashing (too much) the stuff that have high returns for low investments. Going too far down that path can end up reducing variety rather than increasing it. While I don't think Crawl is too vulnerable to this, I've seen this happen in other games, where the developers have tried to normalize and smooth out everything and it ended up making the game become stagnant and staid.

isloat:
Partially merged. As per discussion in ##crawl-dev Fearsome Bark needs either further constraints on it, or Gnoll just need a signature weakness.

PR #527, opened 2017-04-26
(Re)Add the Reaver background.

aolowin:
Reavers are focused on magical  offense and physical defense and
start with a buckler and a new book - the book of Brimstone. The
spell set is loosely based on the Fire/Earth reaver from the wiki.
The extra starting strength should allow them an easier transition
to heavier armour than a conjurer or earth elementalist and the
buckler is unique to the casting backgrounds.


MarvinPA:
This doesn't seem sufficiently differentiated from many of the pre-existing magical backgrounds (Cj, EE, FE) - the buckler and small amount of extra strength don't really merit it being a separate thing.

gammafunk:
I agree that simply giving the player a buckler and an assortment of spells that are nearly all available in the starting books of other classes doesn't make for a compelling new background. There's a fairly high bar for new backgrounds, since they have to feel like they play very differently from all the other ones available (and not simply "like this other background, but stronger"). Closing this.

PR #529, opened 2017-05-01
SDL Tiles: Don't disable screen saver

dessaya:
I don't know if I'm the only one, but I find it very annoying that Crawl locks the screen saver. [Apparently](https://forums.libsdl.org/viewtopic.php?t=4672&sid=867766d41acba06758c5315edc5c8641) it's the default behaviour set by SDL.

I tried setting `SDL_HINT_VIDEO_ALLOW_SCREENSAVER` but it didn't work. Calling `SDL_EnableScreenSaver()` did.


rawlins:
Thanks!

PR #532, opened 2017-05-12
Replace Gnoll aptitude mechanic with locked stats

Floodkiller:
This commit removes the Gnoll aptitude system in order to replace it with
a different main feature: locked, low stats.

This version of Gnoll has +3 in every aptitude at all skill levels, unlike
the previous system which reduced the effective aptitude as more skill
levels were gained.  In order to compensate, Gnolls have a base stat line
of 5 Str/5 Int/5 Dex. Additionally, these stats cannot be changed by any
source, temporary or permanent. For example, Gnolls cannot gain stats
from backgrounds, potions of might/brilliance/agility, Cheibriados, or
changing forms. Likewise, they cannot lose stats from stat drain or have
their stats shifted by Jiyva. This is an alternative approach to
encouraging adaptability and flexibility through being a generalist over a
specialist.

Balancing mechanics from previous iterations of Gnoll (like lower MR
growth) have been temporarily reverted to start balance testing from a
clean slate. They may be re-introduced at a later time if necessary or
interesting.


nikheizen:
My take on this is that it is a poor attempt at making gnolls better generalists, but a good effort to reinforce the "short attention span" flavour point.

Your stats are shitty, but you have a +3 apt in every skill. Find a Triple Crossbow? Get 20 crossbows by your first rune while still having good defenses from skills. A good speed weapon dropped while you are already invested in another melee weapon, no biggy! Fedhas altar? Fuck other skills when your oklobs and ballistomycetes are this strong.

The low spell power high skilling will also effect to concentrate the power of spells which don't need much spell power to be strong, like blink, PoG, regen, etc. This is probably not a good thing.

Anyway, I'm not against this change, but I don't think it will help with the "generalist" goal, the soft-cap skill system, while fucking terrible to play, was better at enforcing this.

Mechanically, it's better, and they will be a stronger race, but I don't think it achieves the goal you've outlined.

Zibokwe:
I like the concept now, it'll be interesting to see how it plays out, have to play more carefully but you can always adapt in nearly any situation.

Floodkiller:
No longer applicable to current Gnoll version, closing.

PR #534, opened 2017-05-14
Remove flight requirements from Hell rune floors

Floodkiller:
Adds solid ground/shallow water pathways in Hell rune floors (mainly
Cocytus) in order to avoid requiring the player to obtain a source of
flight/translocations to access the rune. Only rune access is kept in
mind; the level may be much longer to travel through without flight, and
non-rune items may still remain out of reach.

Specific breakdown of changed vaults:
coc_dpeg- Added a shallow water pathway from the middle to the closest
outside island, and added a circular shallow water pathway linking all the
outside islands.
coc_old- Added a short shallow water pathway across the moat to each
entrance to the middle.
coc_mu- Created a shallow water path that takes the player to all
landmasses on the longer route (flight still required for shorter route).
Added a curved shallow water path at the end to Antaeus's island.
coc_grunt- Added short shallow water paths in between each island.
coc_grunt_cove- Added short shallow water paths connecting the outside
islands, with the last island on each side connecting to the center land
mass.
geh_mu- Added a solid land bridge to access the fortress, and replaced the
lava tile in front of the rune with a door.


gammafunk:
This was merged in d008f67 in trunk, and cherry-picked to 0.20 as well, thanks!

PR #535, opened 2017-05-14
Make cooldowns for DDoor and Berserk accessible through status()

Furbeh:
Purpose for this is to allow custom rc to detect the cooldowns for Death's Door and Berserk through the use of status()


rawlins:
This shows them in % as well, but I think that's fine; I'm going to merge.

rawlins:
Thanks for the patch!

PR #539, opened 2017-05-19
Reduce advpng compression webtiles sprite sheets

alexjurkiewicz:
It'd be great for this to get into 0.20 btw!

Here's a quick chart of compression level vs size vs time, using
player.png webtiles sprite sheet (three runs per compression level,
picking the best run for time):

Compression Level   Size (kb)   Time (s)
                1         966        1.7
                2         893        3
                3         890       21.7
                4         885       61

While there's an argument to be made we should use the best possible
compression level since all players benefit, going from level 2 to 4
results in 20x slower compression for 1% smaller file, which seems
particularly not worth it.


isloat:
This seems good although I'm going to let other more technically-minded developers decide on this. Great catch, Alex!

PR #541, opened 2017-05-21
Improve "no ability" message [skip ci]

hawthornbunny:


nikheizen:
A terribly bland message which is no more informative than the previous one.

amalloy:
I think this is neither better nor worse than what we currently have. If you'd like the wording to be changed, it would help to provide some explanation of what you think is wrong with the current wording and why your proposed new wording is better.

MarvinPA:
I think the current message here is fine, yeah.

hawthornbunny:
Sorry, I completely missed the original replies to this! I know it's too late now, but my reasoning (which I gave in the crawl devchat back at the time) is this: the current message is wrong. It says that the player's character isn't good enough to have a special ability, but that's not true; they simply don't _have_ a special ability. Saying "You're not good enough" implies that the player can gain special abilities through achievement or skill, which is untrue, making the message a confusing oddity. Crawl is typically more accurate than this.

hawthornbunny:
And yes, the message is bland, but that's the point. It's informational. That's like saying "There are no items here" isn't exciting enough.

PR #544, opened 2017-05-31
Unify helmet & hat mutation restrictions

alexjurkiewicz:
Helmets can now be worn with any mutation except horns 3 or antennae 3.

Previously, helmets were the only equipment item that was broken by
getting level 1 in a facet mutation. This lead to players (especially
those worshipping Jiyva/Xom) having to juggle both a helmet and a hat as
possible items to wear, since they would often get a facet mutation
(horns 1, antennae 1, or beak 1) that would unequip their helmet.

Given helmets are only marginally better than hats (+1 AC), just give
them the same equipment limitations as hats. Note: helmets are still
equippable by fewer species, due to size limitations.

TODO: does _slot_conflict need to be updated?

### DIFF
--- a/crawl-ref/source/item-use.cc
+++ b/crawl-ref/source/item-use.cc
@@ -922,45 +922,6 @@ bool can_wear_armour(const item_def &item, bool verbose, bool ignore_temporary)
                 mpr("You can't wear any headgear with your large antennae!");
             return false;
         }
-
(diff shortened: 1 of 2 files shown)


alexjurkiewicz:
Might need to add helmet preventing headbutt aux attack (& retaliation)

alexjurkiewicz:
Closing, no interest

PR #546, opened 2017-06-10
Fix viable species/background option

cut1less:
Accidentally broken in the commits that added group headers to the species menu.

This should close [bug 11092](https://crawl.develz.org/mantis/view.php?id=11092) on Mantis.


rawlins:
Thanks for figuring this out, looks good!  For future reference your commit here (in the PR thread) would work better as the commit description.  (And then github will automatically pull it to the PR description for a single-commit PR anyways.)

PR #547, opened 2017-06-10
Add alternative vi bindings for the Workman keyboard layout

billohreally:
I'm pretty fond of the Workman keyboard layout so I created comfortable alternative vi bindings for it and thought I'd contribute them. The layout is pictured below if needed.

workman_keyboard_layout

### DIFF
--- a/crawl-ref/settings/init.txt
+++ b/crawl-ref/settings/init.txt
@@ -60,6 +60,9 @@
 # Alternative vi bindings for Colemak users.
 # include = colemak_command_keys.txt
 
+# Alternative vi bindings for Workman users.
+# include = workman_command_keys.txt
+
 # Override the vi movement keys with a non-command.
 # include = no_vi_command_keys.txt
 
--- a/crawl-ref/settings/workman_command_keys.txt
+++ b/crawl-ref/settings/workman_command_keys.txt
@@ -1,0 +1,81 @@
+# keymap with working vi keys for workman keyboards
+
+bindkey = [y] CMD_MOVE_LEFT
+bindkey = [e] CMD_MOVE_DOWN
+bindkey = [n] CMD_MOVE_UP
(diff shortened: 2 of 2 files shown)


amalloy:
Shouldn't there be bindings for `CMD_ATTACK_LEFT` and friends?

billohreally:
I'm not sure what CMD_ATTACK_LEFT and co. do so I'm reticent to say if they're necessary or not. 

However, I can confirm that the equivalent file for the dvorak keyboard layout currently in the repo (and the one that I used as a source for this one) does not have mappings for CMD_ATTACK_LEFT and co.

Furthermore, I've played quite a bit with it and I've had no usability issues with it, however my playtime was skewed towards melee non-magic classes.

Feel free to close this, no problem. If someone is using the Workman layout in the first place then they're likely to be more than competent enough to configure it themselves! ;)

rawlins:
`CMD_ATTACK_LEFT` etc attack in place without moving, by convention they are ctrl+movement keys.  I think without explicitly having them they'll remain default bound to ctrl + regular vikeys, though I'm not sure of that.  Probably the Dvorak bindings should have them too.

This seems reasonable to merge with that change, I looked at it a few times and never did anything because I don't know much about this layout.  But I think that applies to everyone on the devteam.

rawlins:
Not entirely sure why this was closed, but for posterity, I started to add CMT_ATTACK_etc for Dvorak keybindings, and the problem with adding them seems to be that they collide too much with other ctrl-bindings.  In principle those could be changed, but it's not so obvious how to do it (ctrl-f and ctrl-s being the worst ones).  To change the Dvorak ctrl-bindings I think I'd really need feedback from someone who actually uses it about what they'd prefer for the conflicting commands; Dvorak users still have access to the swing commands with ctrl+arrow keys.  

It looks like a similar problem would arise for this key binding.

PR #548, opened 2017-06-11
Transporter vault: Trog's Sanctum

EMTedronai:
Trog's most devout followers finally succeeded in creating a secret haven
devoid of all influences of forbidden magic. Will the hero dare to
desecrate this sacred place of silent rage?

The vault makes sure that the player is aware of the silence by placing
the entry transporter inside the silence field. Because of the silence it
is very easy to pull all monsters individually and fight them one by one.
In the end it comes to the player being able to fight the (potentially
berserking) boss monster while being silenced. Mainly magic focussed
characters should probably refrain from entering ;-)

Unfortunately for the player, the imprisoned magic users are not very
grateful for their rescue once the field of silence is being lifted :-/


EMTedronai:
Note that the 'boxiness' of the vault is so that the silence does not penetrate into other parts of the dungeon.

PR #549, opened 2017-06-11
Grate trap vault: Okawaru's Leap of Faith

EMTedronai:
A small idea which I found thematically appealing: Okawaru wants new followers to prove the worthiness by taking a plunge into the unknown. 

Ideally I would have wanted a permanent shaft trap instead of the escape hatch. But you have to work with what you've got...

### DIFF
--- a/crawl-ref/source/dat/des/traps/grate.des
+++ b/crawl-ref/source/dat/des/traps/grate.des
@@ -153,3 +153,30 @@ xxxxxxx
 xxxxxxx
 xxxxxxx
 ENDMAP
+
+# Okawaru wants new followers to prove their worthiness by taking a plunge
+# into the unknown.
+NAME:   emtedronai_okawaru_leap_of_faith
+DEPTH:  D:2-13
+TAGS:   luniq_grate_trap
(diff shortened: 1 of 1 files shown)


MarvinPA:
This doesn't seem like a good vault design to me, it's heavily skewed towards just punishing unspoiled players for not looking up the vault definition (even more so than most other runed door vaults).

EMTedronai:
Honestly, I do not feel too strongly about this vault myself. It was a "spur of the moment" idea that I found cute and quickly implemented.

If you think that it does not fit crawl particularly well, feel free to just close the request. This was just on the off chance that somebody actually found it fitting.

gammafunk:
It is a cute idea, and we do have other another pressure plate trap vault that's non-runed and pretty spoilery (the iron grate one), but it's probably not good to make more of these with or without runed doors.

These vaults can simply always be recognized by spoiled players try to function as something like a trap feature. Traps work better when they can't be seen ahead of time, so the spider basket one works a lot better than something like this or the iron grate vault. Closing this PR.

PR #551, opened 2017-06-21
Add an autumn forest themed bazaar vault

kenranunderscore:
This adds a medium-sized bazaar vault that places the player in an autumn forest (or, in 1 out of 4 cases, a rather dead one). Upon entering, the player has to follow a path that suddenly widens and leads to a clearing in the forest where a few secretive shopkeepers have settled down and are offering their wares. After a successful shopping trip, the player can resume their voyage to follow the path out of the forest and the bazaar.

This vault contains 3 to 4 normal shops (one is abandoned half the time) and one that is always abandoned.

Note: I don't know what type of bazaar vault is more useful at the moment (common or more rare ones), so I've taken 5 as a somewhat rare `WEIGHT`. I have in the meantime made antique shops more likely to appear to justify this. I guess it could arguably also be made very rare (if this pull request makes it in, anyway) as it differs from most/all of the other bazaar layouts in that it has no regular or symmetrical shape.

Here you can see a couple of pictures showcasing the vault and its two variants - dead trees/no dead trees, aka autumn/winter - in tiles and console mode (floor and shop halo can vary ingame as they are generated automatically):

bkfc_rare_entering
bkfc_rare_shops_exit
bkfc_green_full
bkfc_green_console
bkfc_rare_console

### DIFF
--- a/crawl-ref/source/dat/des/portals/bazaar.des
+++ b/crawl-ref/source/dat/des/portals/bazaar.des
@@ -1784,6 +1784,62 @@ x.......x     x.......x
 xxxxxxxxx     xxxxxxxxx
 ENDMAP
 
+# 3 to 4 shops, with an average of 3.5
+NAME:       bazaar_kenran_forest_clearing
+TAGS:       no_rotate no_vmirror
+ORIENT:     encompass
+WEIGHT:     5
+KITEM:      h = human skeleton
+KFEAT:      $ = any shop
+KFEAT:      A = abandoned_shop
+: if crawl.one_chance_in(4) then
+KFEAT:      F = V
+TILE:       t = dngn_tree_dead w:10 / dngn_tree_red w:3 / \
+                dngn_tree_lightred w:2
+COLOUR:     . = brown, t = brown w:10 / red w:3 / lightred w:2
+: else
+TILE:       t = dngn_tree w:3 / dngn_tree_yellow w:2 / \
+                dngn_tree_lightred w:2 / dngn_tree_red w:3
+KFEAT:      F = T
+COLOUR:     . = green, t = green w:3 / lightgreen w:2 / \
+                lightred w:3 / red w:2 / brown w:1
+: end
+NSUBST:     $ = 1=$A
+:           bazaar_setup(_G)
+MAP
+ttttttttttttttttttttttttt
+tttttttttttttttttttttt>tt
+ttttttttttttttttttttt.ttt
+ttttttttttttttttttttt.ttt
+ttttttttttttttGttt.t..ttt
+ttttttttttttttt.ttt..tttt
+tttt$.ttttttttt.....ttttt
+tt.t...ttttttt..t.ttttttt
+tt...ttttttt..ttttttt..tt
+tttt....ttt...tt..tt....t
+tttttt.t.t....t..t...t.$t
+ttttt..t..t.F..tttttttttt
+tttttt........t..Attttttt
+tttttttt.tt..t.t..ttttttt
+tttttt....t..t..htt.$tttt
+tttttt.t...t..t.tt...tttt
+ttt......t.....t....ttttt
+ttG.t..tt....tt...ttttttt
+t......tttt..tttttttttttt
+t$..t.tttttt....ttttttttt
+ttt....ttt...ttGttttttttt
+ttttttttt...ttttttttttttt
+ttttttttt..tttttttttttttt
+tttttttttt...tttttttttttt
+ttttttttttt..tttttttttttt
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
Nice work, on the floor tiles especially!

kenranunderscore:
@alexjurkiewicz Thank you, but I cannot take (full) credit for this ;) There are certain predefined bazaar "layouts" that are applied automatically -- halos around the shops for instance. It took me some fiddling to make it look good though. This seems to be why most (all?) shops in bazaars are usually in the open.

isloat:
This looks great!

crawl-ref/source/dat/des/portals/bazaar.des
```diff
@@ -1548,6 +1548,65 @@ xT.xxxxxxxxxxx.Tx
 xxxxxxxxxxxxxxxxx
 ENDMAP
 
+# 3 to 4 shops, with an average of 3.5
+NAME:       bazaar_kenran_forest_clearing
+TAGS:       no_rotate no_vmirror
+ORIENT:     encompass
+WEIGHT:     5
+KITEM:      h = human skeleton
```
amalloy: Right now the placement of this skeleton is fixed: what about putting a few Hs scattered around the map, with `NSUBST: H = 1=h / *=t.` to make one of them the skeleton and replace the rest with trees or floor randomly.

crawl-ref/source/dat/des/portals/bazaar.des
```diff
@@ -1548,6 +1548,65 @@ xT.xxxxxxxxxxx.Tx
...
+: end
+NSUBST:     $ = 1=$A
+SUBST:      $ = #:20 $
+:           bazaar_setup(_G)
+MAP
+ttttttttttttttttttttttttt
+tttttttttttttttttttttt>tt
+ttttttttttttttttttttt.ttt
+ttttttttttttttttttttt.ttt
+ttttttttttttttGttt.t..ttt
```
amalloy: I don't like the location of this G: it means you never get to see the two trees NE and SW of it, which leaves the map feeling a little unexplored. More generally I don't like transparent-but-impassable obstacles that are at a 90-degree corner between two walls.

This would be fine if it were moved south a tile, or east, or just put somewhere else where it's not in a corner.

crawl-ref/source/dat/des/portals/bazaar.des
```diff
@@ -1548,6 +1548,65 @@ xT.xxxxxxxxxxx.Tx
 xxxxxxxxxxxxxxxxx
 ENDMAP
 
+# 3 to 4 shops, with an average of 3.5
+NAME:       bazaar_kenran_forest_clearing
+TAGS:       no_rotate no_vmirror
```
amalloy: Why `no_rotate no_vmirror`? It seems like it ought to be just fine rotated or mirrored, right? What am I missing?

crawl-ref/source/dat/des/portals/bazaar.des
```diff
@@ -1548,6 +1548,65 @@ xT.xxxxxxxxxxx.Tx
...
+ttttttttt..tttttttttttttt
+tttttttttt...tttttttttttt
+ttttttttttt..tttttttttttt
+ttttttttttt.t.ttttttttttt
+tttttttttt..ttttttttttttt
+ttttt.tt...tttttttttttttt
+ttt<.....tttttttttttttttt
+ttttttt.ttttttttttttttttt
+ttttttttttttttttttttttttt
+ENDMAP
```
amalloy: Since you're going for a sorta chaotic random look, you could consider replacing some of the `t`s on the edge of the map with floor tiles at random, so that it doesn't always have the same shape every time. That's not an important feature or requirement, just something that seemed to fit with your current design.

crawl-ref/source/dat/des/portals/bazaar.des
kenranunderscore: I guess that was just my personal preference of how it looks. It's definitely fine to rotate/mirror and I'll get to that in my next commit.

crawl-ref/source/dat/des/portals/bazaar.des
kenranunderscore: This skeleton should be the owner of the fixed abandoned shop two tiles away. I think I like your idea better though. I'll play around with this a little.

crawl-ref/source/dat/des/portals/bazaar.des
kenranunderscore: That idea is superb. I'll have to test a bit to get a feel which trees look best when floor and vice versa.

PR #554, opened 2017-06-23
8 new vaults

nikheizen:
4 simple arrival vaults: three following simple aesthetic patterns, and one mimicking an abandoned settlement.
2 simple altar vaults based on letters of the alphabet.
**nzn_motion_control:** a dangerous late-dungeon/lair vault which plays off of the companionship between catoblepae and manticores. Sometimes featuring merfolk avatars too!
**nzn_random_ring:** a randomized vault consisting of 3 rings around a central altar+stairs down which can appear in many places.

### DIFF
--- a/crawl-ref/source/dat/des/arrival/simple.des
+++ b/crawl-ref/source/dat/des/arrival/simple.des
@@ -2029,3 +2029,71 @@ x.G.G.G.x
 x.......x
 xxxx+xxxx
 ENDMAP
+
+NAME: nzn_arrival_crosshair
+TAGS: arrival
+ORIENT: float
+MAP
+     x+x
+  cccx.xccc
+ cc..x.x..cc
+ c.c.x.x.c.c
+ c..c...c..c
+xxxx.....xxxx
++.....{.....+
+xxxx.....xxxx
+ c..c...c..c
+ c.c.x.x.c.c
+ cc..x.x..cc
+  cccx.xccc
+     x+x
+ENDMAP
+
+
+NAME: nzn_arrival_water_spiral
+ORIENT: float
+TAGS: arrival no_monster_gen no_pool_fixup
+SHUFFLE: {}
+MAP
+      @
+    wwwWwwwwww
+   wwwwwWwwwwwww
+  wwwwwwwWwwwwwww
+ wwwGwwwwwWwwGwww
+ wwwww.wwwwWwwwww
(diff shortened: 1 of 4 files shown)


crawl-ref/source/dat/des/arrival/simple.des
```diff
@@ -2029,3 +2029,71 @@ x.G.G.G.x
...
+x.................x
++.................+
+x.................x
+xxx.............xxx
+  xxx.........xxx
+    xxx.....xxx
+  xxx.........xxx
+xxx.............xxx
+x.......WWW.......x
+G.....WWwwwWW.....G
```
amalloy: These Gs confused me at first, because a granite statue was a plausible feature for you to include here, so I didn't look for a SUBST replacing them. Might be better to use a glyph that doesn't normally represent a feature, like E.

crawl-ref/source/dat/des/arrival/simple.des
```diff
@@ -2029,3 +2029,71 @@ x.G.G.G.x
...
+ wwwGwwWwwwwwGwww
+  wwwwwwWwwwwwwww
+   wwwwwwWwwwwww
+    wwwwwwWwwww
+           @
+ENDMAP
+
+NAME: nzn_arrival_depleted_hourglass
+ORIENT: north
+TAGS: arrival no_monster_gen no_pool_fixup no_rotate
```
amalloy: You probably wnat `no_vmirror` as well, so the "depleted" hourglass doesn't have all the sand on the top.

nikheizen:
Thanks for looking at them, amalloy.
I fixed the first three arrival vaults up a bit, and tweaked the last one to make it a bit cleaner.
Also added a couple new simple altar vaults and another float(inspired by one of evilmike's arrival vaults) to the PR.

Not sure if bunching them all up like this is preferred/not, so let me know.

gammafunk:
All pushed through two rebased commits except for the two letter random altar vaults. There are follow-up commits to tweak issues with most of these vaults, so feel free to take a look at those and let me know if you see any issues.

For those random altar vaults, it's preferable that they have some game-play concept to justify their existence. Nearly all the vaults of this type feature monsters or some special layout.

You could rework these vaults to be simple one-altar overflows, but even then drawing letters is not so interesting. Overflow altar vaults are probably the most useful type to create overall if you do want to try to make new ones.

PR #555, opened 2017-06-23
Add a lighthouse-themed bazaar vault

kenranunderscore:
This is an(other) attempt at straying away a bit from the classical bazaar vault looks by using some more irregular shapes. The vault `bazaar_kenran_lighthouse` features a circular hut built on an island containing 4 random shops. Its aim is to make the player think of a lighthouse in the sea.

The shoreline, as well as the transitions from shallow to deep water, and from deep water to open sea, respectively, vary slightly in each generated map.

At the moment, the centre of the lighthouse contains a trapped ophan (as to provide a source of light). Of course this is rather gimmicky, but I found it fun. It can easily be replaced by another shop or decorative interior though, should monster generation in this case be frowned upon.

Here are screenshots to show how it looks in tiles/console mode:
console
enter
full
interior

### DIFF
--- a/crawl-ref/source/dat/des/portals/bazaar.des
+++ b/crawl-ref/source/dat/des/portals/bazaar.des
@@ -1880,7 +1880,7 @@ KFEAT:  s = w
 KFEAT:  m = W
 KFEAT:  x = open_sea
 FTILE:  _A>1+ = floor_pebble, - = floor_rough, .< = floor_sand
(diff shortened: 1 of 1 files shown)


kenranunderscore:
I'll close this until (perhaps) the other PR has gone through. Then I can simplify the syntax and also remove the ophan in the middle.

PR #561, opened 2017-06-25
mpr문 번역 완료

sdynet:
mpr문 번역 완료


sdynet:
Leave a comment

nikheizen:
You probably don't want to include all the changes in the source code from EN->KR in this. (assuming the point is the demigod portfolio).

PR #562, opened 2017-06-25
Fix some older bazaar vaults

kenranunderscore:
Some older bazaar vaults did not work/look as originally intended due to tile randomisation. The problems were mostly due to vaults using `LFLOORCOL` in some way, which was then overwritten by the colour randomisation in bazaar_setup. Some of this was only noticeable in console mode, since the randomisation also sets `LFLOORTILE`.

To fix this and thus restore the intended looks of these vaults, several steps were taken:
- Extend `bazaar_setup` in way that makes it possible to control whether colour randomisation takes place or not.
- In vaults that used RNG to set a floor colour and then checked for it, the checks now use `dgn.get_floor_colour()` to get the correct colour (example vault: `bazaar_lake`). This has to take place after the call to bazaar_setup.
- Vaults that used to set a fixed floor colour now use the version of `bazaar_setup` that does not randomise the colours (example vault: `bazaar_minmay_generic_d`).

*Note*: The generation of shop halos is the only thing that needs to be done in the branch epilogue; everything else can be done before that. The code now reflects this, and vaults that do not want halos "delete" the epilogue (pass an empty string to `dgn.set_branch_epilogue`).
Otherwise, if the player had entered another bazaar before, the epilogue could have been set already (hence place halos).

I've tried to update some of the comments/documentation in `bazaar.des` as well.


kenranunderscore:
FYI: the travis build error only occurred in one job and is due to packages not getting installed properly, so shouldn't be a "real" error.

rawlins:
This looks basically good to me, not sure why it's lingering, any vault people want to weigh in? If I'm  the one to merge this I'd probably change it so that passing `nil` to set_branch_epilogue deletes the branch, rather than empty string. (Since the `ASSERT` is probably there as a sanity check against other ways for the epilogue to be empty.)

ebering:
Digging into it, Bazaar setup is the only codepath currently using an epilogue. I'm tempted to remove it, but I don't want to remove that functionality from dlua.

That assert has one purpose: to explicitly forbid using `dgn_set_branch_epilogue` to clear an epilogue. 

I'm preparing to land this so that it takes nil as well, but passing an empty string will have the same effect. (As it should, that function expects the name of an epilogue callback and "" is not a function name in Lua.)

PR #563, opened 2017-06-27
Rebalance potion of mutation

alexjurkiewicz:
Previous behaviour:
Remove 2-6 mutations, add 1-3 mutations, add one good mutation.

New behaviour:
Remove 1-2 mutations, add 2-3 mutations, add one bad mutation.

Mutation is a complex area of game mechanics. Some general points
describing the context and justification for this change:
* Generally, positive mutations are considered "nice to have", most
  negative mutations are ignorable, and a few are considered "must fix"
  (subdued magic, berserkitis, etc).
* Mutations come from a few major sources: !mutation, the malmutation
  monster spell, and mutagenic chunks.
* In the current system, removing a specific bad mutation is generally
  quite easy. It is also easy to return to a nearly-unmutated state. You
  can do both of these things by quaffing (plentiful) potions of
  mutation.

On malmutation:
* In the previous mutation system (with !curemut), malmutation made
  players pay a limited resource if they were mutated. There was tension
  between curing some ignorable mutations now vs risking not having
  curemut in the future for a potentially worse malmutation.
* This wasn't great, because the choice was generally a no-brainer (only
  cure must fix mutations), and the effect of malmutators was zero until
  the player ran out of !curemut. Then, they became potentially
  game-ending threats. This binary behaviour wasn't especially
  enjoyable.
* In the current system, malmutation isn't a significant threat.
  Mutation potions are plentiful and can replace must-fix bad mutations
  with ignorable mutations reliably.
* This also isn't especially enjoyable, you can simply ignore
  malmutation as a serious threat.

On mutation roulette:
* In the current system, players are incentivised to quaff potions of
  mutation surplus to their anti-malmutation needs. They're likely to
  get a good set of mutations and can simple quaff again if not.
* The !mutation design allows players to eat mutagenic chunks (even
  scumming for OOD monsters) and gamble on getting good mutations. If
  they don't, a potion of mutation will reset them to a baseline state,
  since it removes more mutations than it adds.

On a desired behaviour of mutation potions, mutation chunks, and
malmutation:
* Mutation should be a system players cannot completely ignore. This is
  why rMut amulets were removed in the past.
* Malmutation should be a credible threat at all points in the game,
  regardless of how many potions of mutations you should have.
* Receiving bad mutations from any source should have a significant cost
  to remove.

On the approach this change is part of:
* There should be a concept (not neccessarily represented explicitly in
  game code) of "mutation level".
* As the player is mutated, their mutation level generally increases.
* There's no generally-available way to decrease mutation level.
* A higher mutation level implies more mutations.
* A higher mutation level implies more bad mutations.
* It's possible for players to shuffle their mutations, but not without
  increasing their mutation level.
* This means the player can remove must-fix mutations, but only by
  increasing their mutation level.

This change tries to implement such a system by making the potion of
mutation add mutations on average, and not guarantee any good mutations.
Players can use the potion to remove existing bad mutations (although
they'll remove fewer mutations on average than before), but it's very
unlikely for them to reduce their overall mutation level.

Potential future tweaks may be required. In particular, it might be a
good idea to also add a good mutation.

### DIFF
--- a/crawl-ref/source/potion.cc
+++ b/crawl-ref/source/potion.cc
@@ -761,9 +761,9 @@ public:
     }
 };
 
-const int MIN_REMOVED = 2;
-const int MAX_REMOVED = 6;
-const int MIN_ADDED = 1;
+const int MIN_REMOVED = 1;
+const int MAX_REMOVED = 2;
+const int MIN_ADDED = 2;
 const int MAX_ADDED = 3;
 
 class PotionMutation : public PotionEffect
@@ -798,8 +798,8 @@ public:
         // Add mutations.
         for (int i = 0; i < add_mutations; i++)
             mutated |= mutate(RANDOM_MUTATION, "potion of mutation", false);
-        // Always one good mutation.
-        mutated |= mutate(RANDOM_GOOD_MUTATION, "potion of mutation", false);
+        // Always one bad mutation.
+        mutated |= mutate(RANDOM_BAD_MUTATION, "potion of mutation", false);
 
         learned_something_new(HINT_YOU_MUTATED);
         return mutated;
--- a/crawl-ref/source/potion.cc
+++ b/crawl-ref/source/potion.cc
@@ -761,11 +761,6 @@ public:
     }
 };
 
-const int MIN_REMOVED = 1;
-const int MAX_REMOVED = 2;
-const int MIN_ADDED = 2;
-const int MAX_ADDED = 3;
-
 class PotionMutation : public PotionEffect
 {
 private:
@@ -789,19 +784,14 @@ public:
     {
         mpr("You feel extremely strange.");
         bool mutated = false;
-        int remove_mutations = random_range(MIN_REMOVED, MAX_REMOVED);
-        int add_mutations = random_range(MIN_ADDED, MAX_ADDED);
 
-        // Remove mutations.
-        for (int i = 0; i < remove_mutations; i++)
-            mutated |= delete_mutation(RANDOM_MUTATION, "potion of mutation", false);
-        // Add mutations.
-        for (int i = 0; i < add_mutations; i++)
-            mutated |= mutate(RANDOM_MUTATION, "potion of mutation", false);
-        // Always one bad mutation.
+        mutated |= delete_mutation(RANDOM_MUTATION, "potion of mutation", false);
+        mutated |= mutate(RANDOM_MUTATION, "potion of mutation", false);
+        mutated |= mutate(RANDOM_GOOD_MUTATION, "potion of mutation", false);
         mutated |= mutate(RANDOM_BAD_MUTATION, "potion of mutation", false);
 
-        learned_something_new(HINT_YOU_MUTATED);
+        if (mutated)
+            learned_something_new(HINT_YOU_MUTATED);
         return mutated;
     }
 


isloat:
I think you make an interesting argument here. I agree that the current form of !mut might be too aggressive about bringing the player back to a certain baseline mutation level with the large amount of mutations removed. If nothing else, possibly 2-6 removed mutations could be reduced to 2-4. 

You say that it is "very unlikely" to reduce overall mutation level. Removing 1-2 mutations, adding 2-3 mutations, then adding a bad mutation seems like it would always increase overall mutation. 

One thing I want to address is "fun factor", which obviously shouldn't be a conclusive reason to change or not change a particular thing, but I think is important to consider. Because this version has a guaranteed bad mutation, for players !mutation goes from "hey I'll try this" to "only ever use this when necessary". 

Last thing: If surplus !mutation is a problem, why not reduce generation of !mut? This way, variety between games is increased -- e.g., finding Cigotuvi's lab and having a pile of potions to play around with.

MarvinPA:
I'd definitely be in favour of reducing the number of mutations removed, as well as making the potion rarer (and removing mutagenic chunks, probably). My current local implementation from a while ago went with 2-4 mutations removed and swapping the rarity of degeneration and mutation.

alexjurkiewicz:
Re: likelihood of reducing mutation level. Due to how crawl mutation system
works it's not completely uncommon for the potion as tested to reduce the
number of mutations a player has every now and then. At least in my testing.

I think fun is an important consideration. This potion wouldn't be used
except to cure bad mutations. I do like how players in current trunk are
incentivised to enter the mutation game. Maybe benemut could be
reintroduced? Or maybe tweaking the behaviour further can find a sweet spot.

alexjurkiewicz:
Thanks for the feedback. Updated the potion again.

Now: remove 1, add 1 good, add 1 bad.

A few good objections were brought up to the first iteration.

Potion of mutation has two uses:
1) Curing must-fix bad mutations
2) Get a few good mutations in a low risk way

The first rebalance left #1 in place (although nerfed a little), but #2
was broken. The potion's upside was drastically reduced, thus players
wouldn't quaff the potion unless they were attempting to fix an existing
bad mutation. In other words, it took the mutation system back to the
old !curemut days: players would try to avoid mutation and didn't have a
way to interact positively with the system.

This new iteration attempts to keep the potion as plausibly beneficial,
while still capable of dealing with must-fix bad mutations. Since only
one mutation is explicitly removed (more could be removed by the
addition of an opposing mutation sometimes), the potion becomes less
effective at removing mutations from players with a high mutation level.

Removing only one mutation might be too little. Possible future
iterations:
Remove 2, add 1 good, add 1 random.
Remove 2, add 3 random.
Remove 3, add 3 random.

alexjurkiewicz:
BTW, my least favourite part of the current potion is how many mutations it removes. There's no real sense of permanence to the mutations a character picks up. They might get horns 2 from one quaff and then immediately lose them on the second. I think mutations would feel more "impactful" if they appeared more permanent to players.

alexjurkiewicz:
Any thoughts on this?

rawlins:
Right now I'd favor a more conservative change like 2-6 => 2-4.

By the way, when writing the mutation test suite I included something that simulates mutation potions, this might be useful when trying to get a sense for how these proposals will tend to work out.  You can find it in mutation.lua, the functions test_potion and simulate_potion, there are various easily tweakable parameters (change silent to false at the beginning, comment out the other tests, and do something like `crawl -test mutation 2>&1 > /dev/null | less` to see the output).

alexjurkiewicz:
Closing in favour of #579. I opened a new PR to keep this discussion intact but separate.

PR #564, opened 2017-06-27
Stop diagonal running when encountering doors

kenranunderscore:
Consider the following scenario:

```
  ######+##
   @
```

When running (Shift-moving) east, the player will stop before the door in the wall. When moving through the Vaults, I encountered the following situation and noticed a different behaviour:

```
  @
  #
   #
   +#
     #
      #
```

When running SE, the player will just pass the door without stopping. This happens for any of the 4 diagonal directions.

This is due to the methods checking only the tiles to the left and the right relative to the movement direction. When moving horizontally or vertically, this does suffice. In the above example (and analogously for other diagonal movement) however, "left" means the directional vector `(1,0)`, and "right" is `(0,1)`. So when trying to find out whether the movement should stop, only the wall tiles are actually checked and the door isn't, even though it's passed by the player.

To fix this (in case it needs fixing), I've added another check that runs after the usual "left" and "right" ones and that is only executed if the player is moving diagonally. It is testing the "far-left" and "far-right" directions specifically, that is, perpendicular to the movement direction.
This check is only handling the exact case of "passing a door"; that is, only doors are tested for. It can easily be extended to include other features the player should stop at though.

The reason I didn't incorporate those "far-left" and "far-right" tiles into the `run_check` array is because then the diagonal movement would stop way more often and it could be considered annoying. So I've tried to restrict the actual change in behaviour to this one case only.

### DIFF
--- a/crawl-ref/source/externs.h
+++ b/crawl-ref/source/externs.h
@@ -698,6 +698,7 @@ public:
     bool notified_hp_full;
     coord_def pos;
     int travel_speed;
+    int direction;
 
     FixedVector<run_check_dir,3> run_check; // array of grids to check
 
@@ -735,6 +736,7 @@ public:
 private:
     void set_run_check(int index, int compass_dir);
     bool run_should_stop() const;
+    bool diag_run_passes_door() const;
 };
 
 enum mon_spell_slot_flag
--- a/crawl-ref/source/travel.cc
+++ b/crawl-ref/source/travel.cc
@@ -4026,6 +4026,7 @@ void runrest::initialise(int dir, int mode)
     // Note HP and MP for reference.
     hp = you.hp;
     mp = you.magic_points;
+    direction = dir;
     notified_hp_full = false;
     notified_mp_full = false;
     init_travel_speed();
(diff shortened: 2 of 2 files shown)


rawlins:
Thanks, looks good!  Diagonal shift-movement could maybe use further tweaking beyond this, but I agree that the door case is the most important immediate change and that adding further cases could result in more annoyance than improvement.  I'll merge this as-is and leave more possibilities for the future.

One case I'm still wondering about looks in miniature like this:

`....
@.#.
....
....`

If the player moves shift-diagonally down-right, they won't stop at the piece of wall at all, which seems counter-intuitive; if they started one space up they'd actually be stopped twice.  Another case I'm wondering about is this kind of setup:

`@#....
..#...
...#..
....#.
......`

Analogous to non-diagonal movement you might expect to stop at each of these openings, but running won't stop until the end -- but this seems like it could be annoying in practice (maybe this is a case you had in mind).

Not sure exactly what the right tweak is to get these to work as appropriate, but if you're interested in figuring this out I'd merge another PR :-)

PR #565, opened 2017-07-02
Change oka and trog acquirement texts to be more informative

suomenusko:
I think this should be pretty self explanatory. I have always found it mildly annoying when you get a gift in the middle of being shot at by a pack of yaktaurs, only to wonder where it went afterwards since it was just bolts. 

This changes the gift message text for Oka and Trog to be more specific as to what it gave, specifically ammo, weapon, or armour

### DIFF
--- a/crawl-ref/source/dat/defaults/messages.txt
+++ b/crawl-ref/source/dat/defaults/messages.txt
@@ -159,7 +159,7 @@ force_more_message += The gate opens wide
 force_more_message += With a loud hiss the gate opens wide
 
 # God gifts
-force_more_message += grants you a gift
+force_more_message += grants you (a gift|ammunition|armour|a weapon)
 force_more_message += offers you knowledge of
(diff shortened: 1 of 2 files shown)


rawlins:
This looks pretty reasonable to me; could you also update the default force_more_message to include these new messages?  (In `crawl-ref/source/dat/defaults/messages.txt`.)  Or possibly change your messages to "grants you a gift of armour" etc, which I think will be covered by the current defaults.  (Either option seems fine.)

rawlins:
Thanks for the PR! I merged this in [https://github.com/crawl/crawl/commit/06f091eb7428](https://github.com/crawl/crawl/commit/06f091eb7428) (github web interface didn't pick up that that commit was this PR, so I'll have to manually close this, but your authorship information is still there).  Two really minor things for future reference: I fixed a minor bracketing inconsistency with the house style, and I used your first note in the thread here as the commit message (since it's the commit message that is really preserved for posterity).

PR #567, opened 2017-07-03
Upgrade mask of the dragon guarded vault

alexjurkiewicz:
This vault isn't very lethal currently. It's essentially just a slightly
higher density section of Zot, and the lack of speed>10 monsters means
the open layout isn't a huge problem. So, make two changes:

1) Tweak the monster sets.
Old:
6 draconian, 6 dragon, 6(!) ghost moth
New:
9 dragon, 3 ghost moth, 3 moth of wrath

2) Replace the runed doors with transporters. Now, the player has to
step in, walk around the central stone feature, pick up the mask, then
get to the exit transporter. There is some randomness built in to which
direction the transporters operate in and where they are placed
internally.

Some players might get lucky and find the mask a few steps from where
they arrive, but it's not guaranteed, and Zot's -tele status means a
quick escape isn't that possible anyway.

### DIFF
--- a/crawl-ref/source/dat/des/variable/mini_monsters.des
+++ b/crawl-ref/source/dat/des/variable/mini_monsters.des
@@ -6413,7 +6413,6 @@ ITEM:    hat randart no_pickup
 : else
 ITEM:    mask_of_the_dragon no_pickup
 : end
-MONS:    storm dragon, shadow dragon, golden dragon, ghost moth, moth of wrath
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
Fixed some issues identified by @gammafunk.

Screenshots showing what the vault looks like now:
screen shot 2017-07-05 at 3 49 50 pm
screen shot 2017-07-05 at 3 50 09 pm

PR #568, opened 2017-07-05
Rename Demigods to Golems

alexjurkiewicz:
This species is characterised by inability to worship gods. The lore for
why Demigods cannot is along the lines of how they are too proud to do
so. This has never been greatly convincing, so rename them to the
occasionally suggested Golem.

Golems are flesh and blood creatures created from clay. They are just
like humans (they even bleed!) except their lack of humanity precludes
worship of the human gods.

There was an older "Golem" monster, so this PR has a few sticky points
that need extra reviewer attention. Specifically:
* monster-type.h
* mon-data.h
* mon-util.cc

Currently Golems cannot be placed by the level gen so they have been
removed from two vaults (arrival gimmick and pan gimmick vault). This
should be fixed.


alexjurkiewicz:
Also needs an update to Frederick's description.

MarvinPA:
This doesn't seem like any kind of improvement to the lore - what's "humanity" even supposed to mean in this context? Why can gargoyles worship gods despite being more construct-like than golems?

alexjurkiewicz:
Doesn't seem like any love for this PR

PR #569, opened 2017-07-13
Add new vaults and a serial

alexjurkiewicz:
* Four new serial corridors vaults
* A new serial layout, "rivers". This is a river-focused serial, similar
  to "bayou" (and the old, removed "ponds"), but without any plants and
  with more river-like (rather than pond-like) water design, along with
  a few waterlogged rooms. Includes ten vaults.
* Six new minivaults.

---
This change is [Reviewable](https://reviewable.io/reviews/crawl/crawl/569)

### DIFF
--- a/crawl-ref/source/dat/des/serial/corridors.des
+++ b/crawl-ref/source/dat/des/serial/corridors.des
@@ -159,3 +159,59 @@ xxxxxxxx...xxxxxxxx
       xG...Gx
       xx@@@xx
 ENDMAP
+
+NAME:   chequers_hall_statues_y
+DEPTH:  D, Depths, Elf, Crypt, Snake
+WEIGHT: 2
+TAGS:   transparent extra allow_dup serial_corexii_hall decor
+SUBST:  T = T:3 V:3 G:1 .:3
+MAP
+         x
+         x
+xx@@@xxxxxxxxx@@@xx
+xG....GxxxxxG....Gx
+xx.....xxxxx.....xx
+ xx.....xxx.....xx
+  xx.....T.....xx
+   xG.........Gx
+   xxx.......xxx
+     xx.....xx
+      xx...xx
+      xG...Gx
+      xx...xx
+      xx...xx
+      xx...xx
+      xG...Gx
+      xx@@@xx
+ENDMAP
+
+NAME:   chequers_hall_statues_wall_a
+DEPTH:  D, Depths, Elf, Crypt, Snake
+WEIGHT: 2
+TAGS:   transparent extra allow_dup serial_corexii_hall decor
+SUBST:  T = T:3 V:3 G:1 .:3
+MAP
+xxxxxxxxxxxxxxxxxxx
+xGxxxGxxxTxxxGxxxGx
+@.................@
+@.................@
+ENDMAP
+
+NAME:   chequers_hall_statues_corner_a
+DEPTH:  D, Depths, Elf, Crypt, Snake
+WEIGHT: 2
+TAGS:   transparent extra allow_dup serial_corexii_hall decor
+SUBST:  T = T:3 V:3 G:1 .:3
+MAP
+xxxxxxxxxxx
+xTxxxGxxxGx
+xx........@
+xx........@
+xx..
+xG..
+xx..
+xx..
+xx..
+xG..
+xx@@
+ENDMAP
--- a/crawl-ref/source/dat/des/serial/rivers.des
+++ b/crawl-ref/source/dat/des/serial/rivers.des
@@ -1,0 +1,202 @@
+# A serial layout focusing on little streams appearing through the
+# dungeon. Similar to bayou, but with less vegetation.
+
+NAME:   serial_rivers
+TAGS:   allow_dup extra luniq_serial decor nolayout_encompass
+DEPTH:  D, Depths, Lair
+CHANCE: 1%
+{{
+  local c = 4 + crawl.random2(3)
+  hook("post_place", function()
+    dgn.place_maps{tag="serial_rivers", count=c}
+  end)
+}}
+
+NAME:   chequers_serial_rivers_micro_a
+TAGS:   serial_rivers allow_dup extra decor no_pool_fixup transparent no_monster_gen no_pool_fixup
+DEPTH:  D, Depths, Lair
+MAP
+     @
+xxxx...xxxx
+xwwwWWWwwxx
+xxxx...xxxx
+ENDMAP
+
+NAME:   chequers_serial_rivers_micro_b
+TAGS:   serial_rivers allow_dup extra decor no_pool_fixup transparent no_monster_gen no_pool_fixup
+DEPTH:  D, Depths, Lair
+MAP
+     @
+xxxx...x
+xwwwwWWx
+xxxx...x
+ENDMAP
+
+NAME:   chequers_serial_rivers_micro_c
+TAGS:   serial_rivers allow_dup extra decor no_pool_fixup transparent no_monster_gen no_pool_fixup
+DEPTH:  D, Depths, Lair
+SUBST:  W = W:8 .
+MAP
+ W.
+.W...
+.WWW.
+.WWW.
+.WWW.
+...W.
+  .W
+ENDMAP
+
+NAME:   chequers_serial_rivers_micro_d
+TAGS:   serial_rivers extra decor no_pool_fixup transparent no_monster_gen no_pool_fixup
+DEPTH:  D, Depths, Lair
+MAP
+  @
+ www
+ w.w
+ w.w
+ w@w
+ENDMAP
+
+NAME:   chequers_serial_rivers_micro_e
+TAGS:   serial_rivers allow_dup extra decor no_pool_fixup transparent no_monster_gen no_pool_fixup
+DEPTH:  D, Depths, Lair
+SUBST:  . = W., W = W:8 .
+MAP
+wwwwwww
+ .....
+ .....
+  .@.
+ENDMAP
+
+NAME:   chequers_serial_rivers_room_a
+TAGS:   serial_rivers extra decor no_pool_fixup transparent no_monster_gen no_pool_fixup
+DEPTH:  D, Depths, Lair
+MAP
+xxxxxx+
+xw....x
+xw....x
+xw....x
+xwwwwwx
++xxxxxx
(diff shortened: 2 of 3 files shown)


alexjurkiewicz:
A few vault screenshots: 
screen shot 2017-08-02 at 11 42 42 am
screen shot 2017-08-02 at 11 43 12 am
screen shot 2017-08-02 at 11 44 14 am
screen shot 2017-08-02 at 11 47 08 am
screen shot 2017-08-02 at 11 48 41 am
screen shot 2017-08-02 at 11 49 10 am
screen shot 2017-08-02 at 11 49 40 am
screen shot 2017-08-02 at 11 49 51 am
screen shot 2017-08-02 at 11 50 07 am
screen shot 2017-08-02 at 11 50 28 am
screen shot 2017-08-02 at 11 50 34 am
screen shot 2017-08-02 at 11 50 51 am
screen shot 2017-08-02 at 11 50 54 am
screen shot 2017-08-02 at 11 51 08 am
screen shot 2017-08-02 at 11 51 14 am
screen shot 2017-08-02 at 11 51 18 am
screen shot 2017-08-02 at 11 52 28 am

crawl-ref/source/dat/des/serial/rivers.des
```diff
@@ -0,0 +1,168 @@
...
+xwwwwwWWWWWWWWWWWwwwwwx
+xwwwwwWWWWWWWWWWWwwwwwx
+xxxx..............@xxxx
+       .......... 
+ENDMAP
+
+NAME:   chequers_serial_rivers_large_a
+TAGS:   serial_rivers allow_dup extra decor no_pool_fixup transparent
+DEPTH:  D, Depths, Lair
+SUBST:  W = .:4 w:4 W, - = W .
```
gammafunk: You're going to have to do a more careful SUBST here, since you will (rarely) create little shallow water teleport closets where deep water surrounds a central shallow water or floor tile. It would probably be a good idea to have a central one-tile region of shallow water with a guaranteed shallow-water connection to the `-` outer region. This will still allow plenty of squares to have randomized water but will guarantee the player cannot be trapped.

crawl-ref/source/dat/des/serial/rivers.des
```diff
@@ -0,0 +1,168 @@
...
+ .  www
+w.
+ww@
+w .
+ENDMAP
+
+NAME:   chequers_serial_rivers_marsh
+TAGS:   serial_rivers allow_dup extra decor transparent
+DEPTH:  D, Depths, Lair
+SUBST:  . = w. t:2
```
gammafunk: Most of this deep water is being replaced with shallow water; not sure if this can generate areas that strand the player or not due to how pools are fixing things up, but you probably want to mostly use shallow water here. If you want some deep water, you have to be careful about stranding the player, so place it wisely.

crawl-ref/source/dat/des/serial/corridors.des
```diff
@@ -159,3 +159,73 @@ xxxxxxxx...xxxxxxxx
...
+xx..
+xG..
+xx..
+xx..
+xx..
+xG..
+xx@@
+ENDMAP
+
+NAME:   chequers_hall_statues_open_a
```
alexjurkiewicz: Fair enough, I'll drop it entirely.

crawl-ref/source/dat/des/serial/rivers.des
```diff
@@ -0,0 +1,168 @@
...
+TAGS:   serial_rivers allow_dup extra decor transparent
+DEPTH:  D, Depths, Lair
+KPROP:  w = no_tele_into
+MAP
+@     w
+.ww   w
+ .ww. Www.
+.  Ww. ..wW@
+ .      ww
+  .   w
```
alexjurkiewicz: I'll connect the closet to the rest of the water

crawl-ref/source/dat/des/serial/rivers.des
alexjurkiewicz: I'll subst with W only

crawl-ref/source/dat/des/serial/rivers.des
alexjurkiewicz: good idea, I'll add a core of shallow water

crawl-ref/source/dat/des/serial/corridors.des
```diff
@@ -159,3 +159,59 @@ xxxxxxxx...xxxxxxxx
...
+DEPTH:  D, Depths, Elf, Crypt, Snake
+WEIGHT: 2
+TAGS:   transparent extra allow_dup serial_corexii_hall decor
+SUBST:  T = T:3 V:3 G:1 .:3
+MAP
+xxxxxxxxxxx
+xGxxxGxxxGx
+xx........@
+xx........@
+xx..
```
amalloy: Again, the other serial_corexii_hall vaults have enclosed corridors, so having this one open seems weird. If you're doing this on purpose, why?

crawl-ref/source/dat/des/variable/mini_features.des
```diff
@@ -2758,6 +2758,101 @@ MAP
 ..............
 ENDMAP
 
+NAME: chequers_decorated_corner
+TAGS: decor extra
+DEPTH: D, Depths
+MAP
```
amalloy: Now here's one where I thought I might see some SUBSTs for the `T`!

crawl-ref/source/dat/des/variable/mini_features.des
```diff
@@ -2758,6 +2758,101 @@ MAP
...
+x.Gxxxxxxx
+x..x
+xG.x
+x@@x
+ENDMAP
+
+NAME: chequers_jeans
+TAGS: decor extra
+DEPTH: D, Depths
+SHUFFLE: G$
```
amalloy: I don't see a `$` to shuffle with here. Do you mean this?

    SUBST: G = G$

crawl-ref/source/dat/des/variable/mini_features.des
```diff
@@ -2758,6 +2758,101 @@ MAP
...
+.t.t.t.t.t.t.t.t.t.t.t.t.
+xxxxxxxxxxxxxxxxxxxxxxxxx
+ENDMAP
+
+NAME:   chequers_arboreum
+TAGS:   decor extra no_item_gen no_monster_gen
+PLACE:  D:1
+DEPTH:  D
+SUBST:  g = t.
+COLOUR: . = green, g = green
```
amalloy: So here we're turning each `g` into either a tree with a normal floor colour/tile, or a floor tile that's green/grass? If `.` is supposed to be the same as `g` after the transform, why not just use `.` instead of `g` to begin with? That is, replace all the `g`s in the vault with `.`, and then

    SUBST: . = t.
    COLOUR: . = green
    FTILE: . = floor_grass

crawl-ref/source/dat/des/serial/corridors.des
gammafunk: It's a decor vault that just adds an open L-shaped wall. The corridor notion is about connecting two sides of the vaults to the rest of the level. The decor vaults in this serial don't have to be fully enclosed corridors; they're trying to add some additional embellishment to structure normally generated by the layout (which is a pretty open layout when you look at levels it generates).

crawl-ref/source/dat/des/variable/mini_features.des
gammafunk: Alex was probably just trying to have a 50% chance to place nothing instead of a statue, in which case maybe
`SUBST: G = G.`
would be more clear. If the `$` was trying to place something else, then this glyph needs to be defined.

crawl-ref/source/dat/des/variable/mini_features.des
gammafunk: I agree that you don't need an intermediate glyph. Also, he will need to have `FTILE` set on both `t` and `.`, contrary to your code above, since otherwise trees will have a normal dungeon floor tile, which would look odd. Alex can simplify his original `FTILE` statement though:
```
FTILE: t. = floor_grass
```
instead of putting two statements setting the same tile.

amalloy: Okay. I wasn't sure if setting the tile for a tree would cause it to no longer look like a tree.

crawl-ref/source/dat/des/serial/corridors.des
```diff
@@ -159,3 +159,59 @@ xxxxxxxx...xxxxxxxx
...
+DEPTH:  D, Depths, Elf, Crypt, Snake
+WEIGHT: 2
+TAGS:   transparent extra allow_dup serial_corexii_hall decor
+SUBST:  T = T:3 V:3 G:1 .:3
+MAP
+xxxxxxxxxxxxxxxxxxx
+xGxxxGxxxGxxxGxxxGx
+@.................@
+@.................@
+ENDMAP
```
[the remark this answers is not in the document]
alexjurkiewicz: It usually gets used as the wall of an open space, you can see an example in the last image. This is secretly my favourite new vault for this serial!

crawl-ref/source/dat/des/variable/mini_features.des
alexjurkiewicz: oops. the intention was to place a statue, gold or nothing. So this needs work

crawl-ref/source/dat/des/serial/corridors.des
```diff
@@ -159,3 +159,59 @@ xxxxxxxx...xxxxxxxx
       xG...Gx
       xx@@@xx
 ENDMAP
+
+NAME:   chequers_hall_statues_y
+DEPTH:  D, Depths, Elf, Crypt, Snake
+WEIGHT: 2
+TAGS:   transparent extra allow_dup serial_corexii_hall decor
+SUBST:  T = T:3 V:3 G:1 .:3
```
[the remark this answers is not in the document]
alexjurkiewicz: Good catch. I intended to include a T in all of my added vaults to keep with the theme. I've added them to all these `corridors.des` vaults in a (soon to be pushed) commit

alexjurkiewicz:
Thanks for the feedback, sorry for the slow response. Fixed all issues

alexjurkiewicz:
I've updated this PR with a couple more serial vaults. I don't think there is any outstanding feedback.

PR #570, opened 2017-07-13
Don't let ghosts get racial fast speed or chaos brand

Doesnty:
just a suggestion


alexjurkiewicz:
+1 for speed normalisation. Not sure if chaos brand is any worse than, say, distortion

Ash-R:
I think chaos is ok given the collective chance of confusion, distortion, petrify and paralysis is ~5% and there's also a chance you'll get a good result like haste or invis. Inspect description should say which weapon they are holding instead of "weapon of <brand>" since you can't determine weapon type via the player doll either.

PR #572, opened 2017-07-15
Fix various typos in jewellery hints (fixes #571)

### LINKED ISSUE #571: Broken hint text (tiles, jewellery)
**PLATFORM:** local tiles

**VERSION:** 0.20.1 (appears to still be present on HEAD, see below)

**DESCRIPTION:** HINT_SEEN_JEWELLERY's output is "You have picked up a a piece of jewellery, either a ring or an amulet. You can click on it to put it on, and click again to remove it. By" (note: magenta color stops here) "right-clicking> on it, you can view its properties." (note: the following is in red) "</magenta>".

Here's a [screenshot](http://i.imgur.com/7D7vJhX.png).
This is most likely related to mistyped content in [hints.cc](https://github.com/crawl/crawl/blob/master/crawl-ref/source/hints.cc#L1329).

There appear to be mistakes in the console version of the prompt as well.

I'll fix (and open a PR) tomorrow if no one gets around to it by then (it's quite late here).

5paceToast:
Fixes:
- Repeated "a"
- Missing `</w`, which broke coloring.

### DIFF
--- a/crawl-ref/source/hints.cc
+++ b/crawl-ref/source/hints.cc
@@ -1327,7 +1327,7 @@ void learned_something_new(hints_event_type seen_what, coord_def gc)
         break;
 
     case HINT_SEEN_JEWELLERY:
-        text << "You have picked up a a piece of jewellery, either a ring"
+        text << "You have picked up a piece of jewellery, either a ring"
              << "<console> ('<w>"
              << stringize_glyph(get_item_symbol(SHOW_ITEM_RING))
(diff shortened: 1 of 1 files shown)


5paceToast:
Checked console and tiles variant, and they look more or less as I'd expect now (though things like `right-clicking` being white, but `right clicking` not so is a bit surprising, it could just as easily be intentional).

rawlins:
Thanks for the fixes!

PR #575, opened 2017-07-18
Gozag: don't offer useless haste/berserk in petition

alexjurkiewicz:
This change was inspired by the request here:
https://crawl.develz.org/tavern/viewtopic.php?f=8&t=23992

Since even gods cannot override racial restrictions against berserk
(undead, formicid) or haste (formicid), don't let Gozag offer these
potions to ineligible species. Previously, the effects would be added to
offered potions but have no effect (except increasing the potion cost).

This makes offered potions slightly better on average for affected
species.

### DIFF
--- a/crawl-ref/source/god-abil.cc
+++ b/crawl-ref/source/god-abil.cc
@@ -4021,6 +4021,15 @@ static void _gozag_add_potions(CrawlVector &vec, potion_type *which)
 {
     for (; *which != NUM_POTIONS; which++)
     {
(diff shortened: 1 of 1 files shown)


DoctoreMad:
I'm only seeing changes to haste/berserk in "File Changes" list.

alexjurkiewicz:
Yeah, this PR changed to only affect haste/berserk, the commit has an updated description

DoctoreMad:
Oh, I saw the original desc and thought you had missed something. On your way then.

PR #576, opened 2017-07-18
Reduce Deep Dwarf HP modifier from +20% to 0%

alexjurkiewicz:
Inspired by this thread:
https://crawl.develz.org/tavern/viewtopic.php?f=8&t=23991

Deep Dwarf has been an extremely strong species for a long time now.
Their power is derived from effectively unlimited heal wounds, as well
as strong skill aptitudes for playing a primarily melee fighter. On top
of these, DD get the equal-second best health aptitude in game, beaten
only by Troll & Ogre.

Beyond the obvious benefits of a high health pool, DDs benefit from it
by having greater freedom to heal only when their missing health is
greater than the max potential heal. Reducing DD's health aptitude will
make optimal healing riskier for a longer period of the game.

---
This change is [Reviewable](https://reviewable.io/reviews/crawl/crawl/576)

### DIFF
--- a/crawl-ref/source/species-data.h
+++ b/crawl-ref/source/species-data.h
@@ -102,7 +102,7 @@ static const map<species_type, species_def> species_data =
     "DD",
     "Deep Dwarf", "Dwarven", "Dwarf",
     SPF_NONE,
-    -1, 2, 0, 6,
+    -1, 0, 0, 6,
     MONS_DEEP_DWARF,
     HT_LAND, US_ALIVE, SIZE_MEDIUM,
     11, 8, 8, // 27


rawlins:
In absence of other dev input after such a long time I'm going to close this. Here's my own take, which doesn't speak for anyone else (and isn't something I have extremely strong feelings about, but I'm doing a pass on old issues/PRs):

1. The "no-brainer" argument from the linked tavern thread isn't something that applies to species choice. We do not try to balance species selection to be equal in difficulty.
2. If the problem is that DD are unfun (something I don't actually agree with) then it's really not clear to me that this change will improve things -- just make them a bit more tedious. But perhaps I'm missing something.
3. My personal biggest game design complaint about DD is that, within things you could do as DD, there's an imbalance between options that provide a ready healing source (makhleb, trog, vampiric weapons if you find them) and ones that don't. As far as I can see, this change would make that problem worse, not better.

PR #581, opened 2017-07-27
Rework strength damage modifier

alexjurkiewicz:
chart
[Spreadsheet with full numbers](https://docs.google.com/spreadsheets/d/1-ZZlhxy0iUUthRvxJHaZ9ZkqmBexBE36h0JWvYmEIUU/edit?usp=sharing)

Previously, the strength damage modifier had three forms depending on
your strength, from 0-9, 10, and 11+. The formulas were almost but not
quite the same, damage growth was fastest in the 0-9 range with +7.6%
per str compared to +5% per str in the 11+ range.

(Note: these +% figures are the max increase, since the algorithm
includes random2 the average increase is about half the quoted numbers.)

The way this formula changes based on strength isn't great because either:
1) If it should affect attribute increase choices, it's a non-continuous
breakpoint and therefore spoilery.
2) It it's too subtle to change attribute increase choices it's probably
pointless.

Change the damage modifier calculation to use a single continuous
formula for all strength values. The new formula grows at 5.8% with a
base (at 0 str) of 35%.

Replacing the formula neccessarily changes damage modifier values at
different strengths. The modifier is a little lower at typical strength
values, but gets higher at very low (<7) and very high (>28) strengths.
The biggest loss is around 11-14str, where players will lose about +10%
modifier.

```
Comparison of damage modifier at various strengths:
STR    Old    New
-------------------------
1       25%    35%
5       60%    65%
10     100%    95%
15     130%   120%
20     150%   150%
30     210%   210%
40     260%   270%
```

---
This change is [Reviewable](https://reviewable.io/reviews/crawl/crawl/581)

### DIFF
--- a/crawl-ref/source/attack.cc
+++ b/crawl-ref/source/attack.cc
@@ -1112,7 +1112,7 @@ int attack::player_stat_modify_damage(int damage)
     // Each point of strength over 10 increases this by 0.025 (2.5%),
     // strength below 10 reduces the multiplied by the same amount.
     // Minimum multiplier is 0.01 (1%) (reached at -30 str).
-    damage *= max(1, 75 + 2.5 * you.strength());
+    damage *= max(1.0, 75 + 2.5 * you.strength());
     damage /= 100;
 
     return damage;
--- a/crawl-ref/source/attack.cc
+++ b/crawl-ref/source/attack.cc
@@ -1108,8 +1108,11 @@ string attack::defender_name(bool allow_reflexive)
 
 int attack::player_stat_modify_damage(int damage)
 {
-    // 11 strength is ~1.0x damage multiplier.
-    damage *= random2(36 + div_rand_round(you.strength() * 57, 10));
+    // At 10 strength, damage is multiplied by 1.0
+    // Each point of strength over 10 increases this by 0.025 (2.5%),
+    // strength below 10 reduces the multiplied by the same amount.
+    // Minimum multiplier is 0.01 (1%) (reached at -30 str).
+    damage *= max(1, 75 + 2.5 * you.strength());
     damage /= 100;
 
     return damage;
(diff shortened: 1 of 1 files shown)


Ash-R:
Great work! I'd love to see some more needlessly complicated calculations/formulas similar to this tidied up likewise :3

gammafunk:
Your spreadsheet calculations are not correct for the proposed modifier. In the spreadsheet formula you try to show the max result of a random2() roll, but random2 returns [0,N), so that second formula should be

`=34+A2*5.7`

This is done correctly in your formula for the old modifier

alexjurkiewicz:
Whoops. Since I wanted what the spreadsheet showed, I modified the code (s/35/36)

rawlins:
I'm inclined to merge this in some form, possibly with a slight constant increase to smooth out the nerf to mid-level strengths. Anyone else care?

rawlins:
once randomness is factored in, the original proposal in this PR has some problems (and therefore so does my followup) -- in the pre-PR version, the base damage modifier is increased by a (multiplied) random value between 0 and str when above 10, and decreased by a (multiplied) random value when below 10. It is never selected purely randomly itself, and so for example is never 0. The PR proposal would allow a 0 draw at any str value. To be equivalent in structure, my proposal would still need the breakpoint at 10, and look something like this (where the modifier is scaled by 100):

    mod = 100    (if str = 10)
    mod = 100 - random2(50 - 5 * str)    (if str < 10)
    mod = 100 + random2(5 * str - 50)    (if str > 10)

I wonder if this randomness could be eliminated entirely? One possibility is simply a flat, non-random:

    mod = 75 + 2.5 * str

Which is the mean of the above random formula.

stenellad:
The change seems to have broken Travis

rawlins:
I think the current travis state is just the halloween test issue. I'm excited about this change but it'll need some careful fsimming before merging, I can take a look this weekend.

PR #582, opened 2017-07-28
Add a note when an unrand is first seen.

aolowin:
After they were changed to being identified the notes were bypassed.


amalloy:
Merged #593, which accomplishes the same idea.

PR #583, opened 2017-08-02
Increase Bai Suzhen's defenses

alexjurkiewicz:
A lot of her scariness comes from abilities that get scarier the longer
you are in combat (hydra summoning, thunder clouds) but she doesn't have
any particular staying power. Increase her health by ~1/3rd, at the cost
of a little primary attack damage.


alexjurkiewicz:
05:46 <+gammafunk> alexjurkiewicz: not going to merge
                   the bai suzhe PR, disagree that
                   bai suzhen has a problem more than
                   other uniques, or at least one
                   that would be solved by her having
                   more HP
05:47 <+gammafunk> if she has a problem it's fairly
                   small and it would be more her
                   abilities being dangerous
05:48 <+gammafunk> she has a massive amount of HP and
                   quite high AC as well, even if she
                   maybe doesn't use body armor
05:49 <+gammafunk> moving her placement shallower
                   would make her more dangerous, if
                   that's necessary, but we've sort
                   of historically said that anything
                   like a 1% kill ratio is a rough
                   metric of "good enough"

PR #585, opened 2017-08-05
Skill UI updates

rawlins:
This PR is for tracking some skill UI changes I'm working on as they progress, feedback is welcome.

**Initial commit:**

This adds information in the following format to weapon/shield info
about when the player will reach mindelay/no penalty (where the first
two lines were already there):

    This weapon's minimum attack delay (0.7) is reached at skill level 26.
         (Your skill: 20.8)
    At max training you would reach 26.0 in about 1.0 XLs.
    At current training (50%) you will reach 26.0 in about 2.3 XLs.

The basic idea is that this information allows more informed training
allocations at any given point.  I decided to use XLs as a measure
because I think players generally seem to have sense of XL progression,
and for most of the game it's something they're very attuned to.  Past
XL 27 it will say "the equivalent of N XLs", similar to how felid life
gain works.  Also, I couldn't think of any other way to quantify this.
Using XLs does to some degree reveal the amount of deep weirdness and
non-linearity in skill point / XL gain, but so be it.

In most cases this is an exact calculation, but there are a few things I
wasn't able to perfectly account for.  Crosstraining works, but it won't
take into account training multiple cross-training skills at the same
time.  Ash bonuses are approximate, because they give a skill point
bonus that is dependent on skill level (level, not points), and this is
somewhat hard to calculate.  It also uses only current ash piety.  It
ignores draining, ash penance, heroism, and any other skill effects I
forgot about.

This commit comes with a bunch of refactoring of skill / skill point
calculations in order to avoid code duplication and make some
calculations slightly clearer.  This commit is (hopefully) the first
step in a bunch of changes to the skill UI, including skill targets.
For example, the code that produces the weapon info is set up to support
arbitrary queries for skill targets.


Ash-R:
I think all the information about XL is a) bloat and b) confusing in the proposed form (max training?). The only thing that should be added is a line that informs of your current attack delay. e.g.
This weapon's minimum attack delay (0.7) is reached at skill level 26.
Your current attack delay (0.9) is achieved by current skill level 20.8.

rawlins:
> I think all the information about XL is a) bloat

Well, I don't really agree (or I wouldn't have started working on this in the first place).   Could you explain _why_ you think it is bloat?

There may be an argument to be made that it is more confusing than it is ehlpful, I'm still thinking about that -- but largely because xp gain itself is extremely confusing.  (max training = 100% on that skill.)

Sandman25DCSS:
IMHO primary use of this info is to decide either to go for 2h weapon or 1h+shield so information like
"At max training you will reach 15.0 in about 4.0 XLs" and
"At max training you will reach 24.0 in about 8.0 XLs" can be very useful.
Is it possible to support magic skills also?
It can be interesting to compare upgrading from glaive to bardiche and getting Shadow Creatures, for example. Though in this case we would need to know spell failure also and it can make things really complicated I suppose.

crawl-ref/source/skills.cc
```diff
@@ -1160,7 +1161,7 @@ pair<int,int> skill_level_to_diffs(skill_type skill, double amount, int scaled_t
     {
         // each loop is the max skill points that can be gained at the
         // current cost level, up to `target`.
-        int next_level = reduced ? skill_cost_needed(you_skill_cost_level) :
+        const int next_level = reduced ? skill_cost_needed(you_skill_cost_level) :
```
amalloy: What about something like:

    const int target_level  = you_skill_cost_level + (reduced ? 0 : 1);
    const int next_level = skill_cost_needed(target_level);

Factors out the repeated `skill_cost_needed` and `you_skill_cost_level`, making it clearer what the impact of `reduced` is. Not sure about the name for my suggested intermediate variable, since I'm not totally sure what you're computing here.

crawl-ref/source/skills.cc
```diff
@@ -1169,34 +1170,35 @@ pair<int,int> skill_level_to_diffs(skill_type skill, double amount, int scaled_t
...
+        delta.skill_points = min<int>(abs((int)(target - you_skill)),
                                  max_skp);
-        int delta_xp = (delta_skp * cost + 9) / 10;
+        delta.experience = (delta.skill_points * cost + 9) / 10;
 
         if (reduced)
         {
-            delta_skp = -min<int>(delta_skp, you_skill);
-            delta_xp = -min<int>(delta_xp, you_xp);
+            delta.skill_points = -min<int>(delta.skill_points, you_skill);
```
amalloy: `-min(x,y)` is a little weird. I think this is the best way to compute the thing you're trying to compute, but it could use a comment explaining why this is a good calculation to make.

crawl-ref/source/skills.cc
```diff
@@ -1145,7 +1146,7 @@ pair<int,int> skill_level_to_diffs(skill_type skill, double amount, int scaled_t
         you_skill += ash_skill_point_boost(skill, you.skills[skill] * 10);
     }
     if (target == you_skill)
-        return make_pair(0, 0);
+        return skill_diff();
 
     // XXX: reducing with ash bonuses in play could lead to weird results.
     const bool reduced = target < you_skill;
```
amalloy: `reduced` seems to have significant impact on a number of decisions throughout this function, but it's not clear from its name what is being reduced. From looking at how it's defined I still don't really know why `target < you_skill` is meaningful here, or why the changes it makes later make sense. Are we trying to compute...how much experience I would have to lose in order to no longer be at mindelay, or something weird like that? A more descriptive name and/or a lengthier comment explaining what it means would be welcome.

crawl-ref/source/skills.cc
```diff
@@ -1169,34 +1170,35 @@ pair<int,int> skill_level_to_diffs(skill_type skill, double amount, int scaled_t
...
 
-        int cost = calc_skill_cost(you_skill_cost_level);
+        const int cost = calc_skill_cost(you_skill_cost_level);
         // Maximum number of skill points to transfer in one go.
         // It's max_xp*10/cost rounded up.
-        int max_skp = (max_xp * 10 + cost - 1) / cost;
-        max_skp = max(max_skp, 1);
-        int delta_skp = min<int>(abs((int)(target - you_skill)),
+        const int max_skp = max((max_xp * 10 + cost - 1) / cost, 1);
+        skill_diff delta;
```
amalloy: I don't feel strongly about this, but what about using the `skill_diff(int, int)` constructor you defined? You're initializing to zero and then unconditionally overwriting, which seems a bit weird.

On the other hand I can see how this style might be more readable than what I suggest. Use your judgment on that, I just wanted to make sure you considered both ways.

rawlins: I consider that but I don't think that works for this use because the xp value is actually dependent on the skill point value, though maybe c++ constructors could handle that.

crawl-ref/source/skills.cc
rawlins: I've cleaned this up in the next commit, but one thing that may not have been entirely clear is that this function is also used when actually setting skill levels -- it's a refactored version of the old code that set skill levels by side effect, in order to do just the calculations with no side effects.  So when `reduced` (now `decrease_skill`) is true, it's computing how much xp/skp to remove to reach the target skill level.  The function I factored it out of is `set_skill_level`.

crawl-ref/source/skills.cc
```diff
@@ -1068,14 +1090,31 @@ static int _train(skill_type exsk, int &max_exp, bool simu)
...
+ * @param scaled_training how to scale the training values (for estimating what happens when other skills are being trained).
+ * @param crosstraining whether to factor in crosstraining.
+ * @param religion whether to factor in god bonuses (specifically, ash).
+ * @param manuals whether to factor in manuals.  Does not use exact skill points from manual -- only uses training rate bonus.
+ *
+ * @return a pair consisting of the difference in skill points, and the difference in xp.
+ */
+skill_diff skill_level_to_diffs(skill_type skill, double amount,
+                            int scaled_training,
+                            bool crosstraining, bool religion, bool manuals)
```
amalloy: This is a lot of bools, which will make it difficult to know what is happening at a call site: `skill_level_to_diffs(sk, 5, 10, true, false, true)` is pretty inscrutable. Is there any way to combine them, or use an argument whose value has more meaning? A `mode` enum, or fields in a bitmask or something?

PR #586, opened 2017-08-06
Increase priority of checking for hated spells when memorising

jmbjr:
see **https://crawl.develz.org/mantis/view.php?id=11169**

This update always makes hated or loathed spells show up red in the
memorisation menu and always provides a warning message if the user
tries to memorise these spells.

Additionally, for Trog, all spells are red in the memorisation menu.

There's a couple minor fixups with using a color enum instead of the color directly, and mixing up the warning priorities to something that seems to flow well. (i.e. check if Trog hates spellcasting before checking if Trog hates that particular spell to avoid a redundant warning)

Finally, added 3 docstrings in religion.cc for the functions I needed to use for the warnings.

Latest update 8/8/2017 implements amalloy's suggestions.
- Make some docstrings more consistent
- Move some warnings into a new function
- Fix some formatting 

Previously, the memorisation screen would prioritise "not enough spell
levels" over "divinely forbidden or hated". So if you worshiped Yred and
didn't have enough spell levels for Statue Form, it would show up as
gray, until you finally had enough levels, at which point it would
finally show up as red.

Similarly, after selecting the spell for memorisation, there were no
additional warning messages reminding the user that their god didn't
like the spell.

### DIFF
--- a/crawl-ref/source/religion.cc
+++ b/crawl-ref/source/religion.cc
@@ -3905,11 +3905,26 @@ bool god_likes_spell(spell_type spell, god_type god)
     }
 }
 
+/**
+ * Does your god hate spellcasting?
+ *
+ * @param god           The god to check against
+ * @return              Whether the god hates spellcasting
+ */
 bool god_hates_spellcasting(god_type god)
 {
     return god == GOD_TROG;
 }
 
+/**
+ * Will your god put you under penance if you actually cast spell?
+ *
+ * @param spell         The spell to check against
+ * @param god           The god to check against
+ * @param fake_spell    true if the spell is evoked or from an innate or divine ability
+ *                      false if it is a spell being cast normally.
+ * @return              true if the god hates the spell
+ */
 bool god_hates_spell(spell_type spell, god_type god, bool fake_spell)
 {
     if (god_hates_spellcasting(god))
@@ -3938,6 +3953,13 @@ bool god_hates_spell(spell_type spell, god_type god, bool fake_spell)
     return false;
 }
 
+/**
+ * Will your god excommunicate you if you actually cast spell?
+ *
+ * @param spell         The spell to check against
+ * @param god           The god to check against
+ * @return              Whether the god loathes the spell
+ */
 bool god_loathes_spell(spell_type spell, god_type god)
 {
     if (spell == SPELL_NECROMUTATION && is_good_god(god))
@@ -3945,6 +3967,27 @@ bool god_loathes_spell(spell_type spell, god_type god)
     if (spell == SPELL_STATUE_FORM && god == GOD_YREDELEMNUL)
         return true;
     return false;
+}
+
+/**
+ * Checks to see if your god hates or loathes this spell,
+ * or just hates spellcasting in general, and returns a warning string
+ *
+ * @param spell         The spell to check against
+ * @param god           The god to check against
+ * @return              Warning string if god has strong opinions on spell
+ *                      Empty string if god doesn't care about spell
+ */
+string god_spell_warn_string(spell_type spell, god_type god)
+{
+    if (god_loathes_spell(spell, god))
+        return "Your god extremely despises this spell!";
+    else if (god_hates_spellcasting(god))
+        return "Your god hates spellcasting!";
+    else if (god_hates_spell(spell, god))
+        return "Your god hates this spell!";
+    else
+        return "";
 }
 
 bool god_hates_ability(ability_type ability, god_type god)
--- a/crawl-ref/source/spl-book.cc
+++ b/crawl-ref/source/spl-book.cc
@@ -689,10 +689,6 @@ static spell_type _choose_mem_spell(spell_list &spells,
         int colour = LIGHTGRAY;
         if (vehumet_is_offering(spell))
             colour = LIGHTBLUE;
-        // Grey out spells for which you lack experience or spell levels.
-        else if (spell_difficulty(spell) > you.experience_level
-                 || player_spell_levels() < spell_levels_required(spell))
-            colour = DARKGRAY;
         else
             colour = spell_highlight_by_utility(spell);
 
@@ -874,6 +870,11 @@ static bool _learn_spell_checks(spell_type specspell, bool wizard = false)
 */
 bool learn_spell(spell_type specspell, bool wizard)
 {
+    string mem_spell_warning_string = god_spell_warn_string(specspell, you.religion);
+
+    if (mem_spell_warning_string != "")
+        mprf(MSGCH_WARN, "%s", mem_spell_warning_string.c_str());
+
     if (!_learn_spell_checks(specspell, wizard))
         return false;
 
--- a/crawl-ref/source/spl-util.cc
+++ b/crawl-ref/source/spl-util.cc
@@ -1068,7 +1068,8 @@ bool spell_is_form(spell_type spell)
  *                   (status effects, mana, gods, items, etc.)
  * @param prevent    Whether to only check for effects which prevent casting,
  *                   rather than just ones that make it unproductive.
- * @param fake_spell Is the spell evoked, or from an innate or divine ability?
+ * @param fake_spell true if the spell is evoked or from an innate or divine ability
+ *                   false if it is a spell being cast normally.
  * @return           Whether the given spell has no chance of being useful.
  */
 bool spell_is_useless(spell_type spell, bool temp, bool prevent,
@@ -1086,7 +1087,8 @@ bool spell_is_useless(spell_type spell, bool temp, bool prevent,
  *                   (status effects, mana, gods, items, etc.)
  * @param prevent    Whether to only check for effects which prevent casting,
  *                   rather than just ones that make it unproductive.
- * @param fake_spell Is the spell evoked, or from an innate or divine ability?
+ * @param fake_spell true if the spell is evoked or from an innate or divine ability
+ *                   false if it is a spell being cast normally.
  * @return           The reason a spell is useless to the player, if it is;
  *                   "" otherwise. The string should be a full clause, but
  *                   begin with a lowercase letter so callers can put it in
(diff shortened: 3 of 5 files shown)


jmbjr:
I'm not convinced I'm handling All the Situations since I'm still trying to acclimate to the codebase, but this feels like an incremental improvement.

It did seem a bit odd to warn the user that "Casting this spell..." will be bad when they are trying to memorize. They may think they accidentally hit z, but saying "Learning this spell is OK, but if you cast it your god won't like it" was a bit verbose.

The reason I added the warning in addition to the RED in the memorisation menu was for newbies who might not know what the Red meant.

crawl-ref/source/spl-util.cc
```diff
@@ -1344,8 +1344,12 @@ int spell_highlight_by_utility(spell_type spell, int default_colour,
     {
         return COL_FORBIDDEN;
     }
+    // Grey out spells for which you lack experience or spell levels.
+    if (spell_difficulty(spell) > you.experience_level
+             || player_spell_levels() < spell_levels_required(spell))
+        default_colour = DARKGRAY;
```
rawlins: I would use one of the semantic color constants in `spl-util.h` (or add a new one) rather than an actual color term.  `COL_INAPPLICABLE` might be appropriate, and is `DARKGRAY` anyways.

Also, this line should be in `{}` because the condition is two lines, I'm not sure why `unbrace` didn't catch it.

crawl-ref/source/spl-book.cc
```diff
@@ -874,6 +870,11 @@ static bool _learn_spell_checks(spell_type specspell, bool wizard = false)
 */
 bool learn_spell(spell_type specspell, bool wizard)
 {
+    if (god_loathes_spell(specspell, you.religion))
+        mprf(MSGCH_WARN,"Casting this spell will cause instant excommunication!");
+    else if (god_hates_spell(specspell, you.religion))
```
rawlins: I don't know this part of the code very well and so wasn't entirely sure on first glance that `god_loathes_spell` and `god_hates_spell` are the right ones (plus they aren't documented).  It might be worth adding a docstring to these two functions (which is a reasonable thing to do in a commit like this), and cross-checking with `spl-cast.cc:_spellcasting_god_conduct` if you haven't done so.  One more thing to double check is if this messaging makes sense for Trog, you might want something more general for that case.

Also, I might use different wording on the warnings in case punishments vary a bit by god ("Your god despises this spell" or something) but that's a matter of taste I think.

jmbjr: Good ideas.  

I added docstrings for the two functions, and one for god_hates_spellcasting.

I took a look at spl-cast.cc, but didn't see anything that jumped out at me, but I'm not sure my eyes are trained enough yet to catch the things I should be looking for.

I liked your suggestion on the warnings.  They've been made a bit more generic/vague, but I think it still gets the point across

crawl-ref/source/spl-util.cc
jmbjr: Thanks for the reminder on unbrace. I hadn't actually set up all the fancy git stuff from docs/develop/git, so it missed it.  fixed now.

and I handled the COL_INAPPLICABLE as well. Looks better now.

crawl-ref/source/religion.cc
```diff
@@ -3905,11 +3905,26 @@ bool god_likes_spell(spell_type spell, god_type god)
...
 {
     return god == GOD_TROG;
 }
 
+/**
+ * Will your god put you under penance if you actually cast spell?
+ *
+ * @param spell         The spell to check against
+ * @param god           The god to check against
+ * @param fake_spell    Whether the spell is evoked, or from an innate or
```
amalloy: "Whether X is Y, or Z" sounds like you are saying Y and Z are the two options. But what you mean here is "Is it either of Y or Z?" with the implication being that it could instead be Q (i.e., an actual spell you are actually casting with the `z` command).

Clearer would be to be explicit about the two options, and say which of them is `true` and which is `false`. For example,

> `true` if the spell is evoked or from an innate or divine ability; `false` if it is a spell being cast normally.

crawl-ref/source/religion.cc
```diff
@@ -3938,6 +3953,13 @@ bool god_hates_spell(spell_type spell, god_type god, bool fake_spell)
     return false;
 }
 
+/**
+ * Will your god excommunicate you if you actually cast spell?
```
amalloy: Thank goodness for this docstring. Reading just the diff my first thought was, "Really, we have a distinction between loathes and hates? I wonder what the difference is."

crawl-ref/source/spl-book.cc
```diff
@@ -874,6 +870,14 @@ static bool _learn_spell_checks(spell_type specspell, bool wizard = false)
 */
 bool learn_spell(spell_type specspell, bool wizard)
 {
+    if (god_loathes_spell(specspell, you.religion))
```
amalloy: Consider extracting these six lines into their own function that returns a string. Then when you call it here you won't have to duplicate the `mprf(MSGCH_WARN, ...)` prelude, and you can avoid interleaving side effects with computation.

crawl-ref/source/spl-util.cc
```diff
@@ -1344,8 +1344,14 @@ int spell_highlight_by_utility(spell_type spell, int default_colour,
     {
         return COL_FORBIDDEN;
     }
+    // Grey out spells for which you lack experience or spell levels.
+    if (spell_difficulty(spell) > you.experience_level
+             || player_spell_levels() < spell_levels_required(spell))
+    {
+        default_colour = COL_INAPPLICABLE ;
```
amalloy: Just `return COL_INAPPLICABLE;` here, and likewise the stanza below it.

crawl-ref/source/spl-util.cc
```diff
@@ -1344,8 +1344,14 @@ int spell_highlight_by_utility(spell_type spell, int default_colour,
     {
         return COL_FORBIDDEN;
     }
+    // Grey out spells for which you lack experience or spell levels.
+    if (spell_difficulty(spell) > you.experience_level
+             || player_spell_levels() < spell_levels_required(spell))
```
amalloy: The indentation on this hanging `||` doesn't match the one in the previous stanza. I think the previous one reflects "crawl style", such as it is, but if you find evidence that your style is more common in the crawl codebase then you should fix the one above it to match this one.

crawl-ref/source/religion.cc
jmbjr: good idea. I just copied it from another docstring in spl-cast.cc. 

Looks like spl-cast.cc, religion.cc, and spl-util.cc all have different wording for this param. Maybe I'll just change them all to the same thing (and see if I can find anymore examples lurking about)

crawl-ref/source/spl-book.cc
jmbjr: good idea. thanks

crawl-ref/source/spl-util.cc
jmbjr: another copy/paste thing after I floated these lines to the new function. I'll tweak it

jmbjr:
Latest update 8/8/2017 implements amalloy's suggestions.

Make some docstrings more consistent
Move some warnings into a new function
Fix some formatting

crawl-ref/source/religion.cc
```diff
@@ -3969,6 +3969,27 @@ bool god_loathes_spell(spell_type spell, god_type god)
...
+ * Checks to see if your god hates or loathes this spell,
+ * or just hates spellcasting in general, and returns a warning string
+ *
+ * @param spell         The spell to check against
+ * @param god           The god to check against
+ * @return              Warning string if god has strong opinions on spell
+ *                      Empty string if god doesn't care about spell
+ */
+string god_spell_warn_string(spell_type spell, god_type god)
+{
```
amalloy: You're ignoring the `god` argument and using `you.religion` instead.

crawl-ref/source/spl-book.cc
```diff
@@ -870,13 +870,10 @@ static bool _learn_spell_checks(spell_type specspell, bool wizard = false)
...
 {
-    if (god_loathes_spell(specspell, you.religion))
-        mprf(MSGCH_WARN,"Your god extremely loathes this spell!");
-    else if (god_hates_spellcasting(you.religion))
-        mprf(MSGCH_WARN,"Your god hates spellcasting!");
-    else if (god_hates_spell(specspell, you.religion))
-        mprf(MSGCH_WARN,"Your god hates this spell!");
+    string mem_spell_warning_string = god_spell_warn_string(specspell, you.religion);
 
+    if (mem_spell_warning_string != "")
```
amalloy: Should use `.empty()` rather than comparing to `""`.

crawl-ref/source/spl-book.cc
```diff
@@ -870,13 +870,10 @@ static bool _learn_spell_checks(spell_type specspell, bool wizard = false)
 */
 bool learn_spell(spell_type specspell, bool wizard)
 {
-    if (god_loathes_spell(specspell, you.religion))
-        mprf(MSGCH_WARN,"Your god extremely loathes this spell!");
-    else if (god_hates_spellcasting(you.religion))
-        mprf(MSGCH_WARN,"Your god hates spellcasting!");
-    else if (god_hates_spell(specspell, you.religion))
-        mprf(MSGCH_WARN,"Your god hates this spell!");
+    string mem_spell_warning_string = god_spell_warn_string(specspell, you.religion);
```
amalloy: I'd be inclined to just call it `warning`. That it's a string is implied by its type, and that it's related to memorising spells is implied from context. If you think it's important to emphasise that it's about memorising spells, I think you should not abbreviate to `mem`. Call it `memorisation_warning` or something.

amalloy:
Oh I forgot to say, thanks for fixing up all the `fake_spell` docstrings. Those kind of changes are never much fun, but I love to see the codebase getting cleaned up.

crawl-ref/source/religion.cc
jmbjr: doh!

jmbjr:
I squashed and tweaked the commit. It seemed like everything could fit into a single commit. Also wanted to make sure damerell got credit.

amalloy, I also fixed the issue of not actually using the god variable I passed to the new function.

Everything else is the same, except for wording inside the commit message and commit title.

amalloy:
Applied as 1756668458. Thanks @jmbjr!

PR #588, opened 2017-08-09
Adding Vehumet zealot background

fuchsiaruler:
Destroyers are mages that start without book and receive new spells from Vehumet instead.

As discussed in [that thread](https://crawl.develz.org/tavern/viewtopic.php?f=8&t=24085).


MarvinPA:
There's nothing that really makes this an interesting zealot class, as per the removals of many similar starting gods in the past.

njvack:
I agree that generally, fewer zealot starts are better than more. And if this were a "book start, except with Vehumet" kind of deal, I'd be against it. However, this is more of a "bookless conjurer" start which seems new to me. Instead of having a planned build you can rely on based on your starting book (and which will generally give a well-defined "good" path through Lair), you rely on the whims of Vehumet and the floor.

Maybe it doesn't play as differently as I'm thinking it would, but at least on paper it sounds like D:1 play would be different than any of the conjuration-focused book starts.

amalloy:
I thought a book-less conjurer start could potentially be interesting, so I tried a practice game. I have to say, it feels terribly weak. I'm XL5 on D:3 and still haven't been offered a single level 2 spell. [Apparently](https://github.com/crawl/crawl/blob/1756668/crawl-ref/source/religion.cc#L1665-L1666) the first gift (which I got at XL2) is always a level 1 spell and the second gift (which I got at XL3) can be either level 1 or level 2. There are some usable level 1 spells, but they're not a huge improvement over magic dart, so really you'd like to get something at least level 2 to work with, preferably 3. Waiting until XL5 or XL6 before that happens is miserable; you would do much better to pick a book start to survive the early game with a Vehumet altar to give you some mid-game direction.

I don't think this background actually produces an interestingly different style of play; it's just worse than starting with a book.

edgefigaro:
I too tried it out, 4 games HuCj, dropping book and taking vehu immediately, default setting of 15 piety. Its really bad. 

First game i got a flame tongue on D1, made it to D3 before getting poisoned by an adder and dieing.

Second game was bad, I made it to D3 XL 5 but never got a spell gift. I suspect this character is vulnerable to piety swings. I had no answers to worms or an orc in scale mail. I died to a an orc priest when my retreat was blocked. I probably could have been able to live with a hex wand and a venom dagger that dropped, but I'm not exactly trying to play flawlessly and make full use of a favorable floorgod. 

Third Game I upped the piety at the start from 15 to 35 (starting Be piety) and character plays more stably. Freeze is immediately offered for a second spell and a versatile kit. I get offered throw frost and hit ** early on D2 which I immediately take. Ignite poison is offered my first kill into D3. I'm XL 4 and only have 3 spell levels, and I pass.

D4 has dowan and duvessa which i don't want to fight. I try but my freeze isn't strong enough at the moment. A bullfrog skeleton I don't have good answers to at the moment, but its also not a threat and i chop it down with a +0 flail of protection. I hit *** on D4 but die shortly thereafter chain confused by sigmund.

Next up I am offered searing tongue at lvl 1 and get my second gift on my first kill on D2, throw flame. I die to natasha after being slowed. This was probably avoidable, and I could have just jibbed the early game with a +2 short sword of distortion. 

My final run i decide to give myself a potion of brilliance in addition to 35 piety. I want to be able to power though a fight against a unique. Sting is my starting gift and throw frost is offered on the last mob of D1

amalloy:
I'm closing this PR as it looks like this feature will not be merged without a substantial overhaul (and even then it seems unlikely).

PR #589, opened 2017-08-09
initfile.cc: Make sure unistd.h is included for Cygwin targets.

johnny-0:
Also, update other target checks in the #ifdef chain to check against
 an appropriate TARGET_OS macro.

Needed for the readlink() call in _find_executable_path().


rawlins:
Thanks!

PR #590, opened 2017-08-09
Make Dith hate all fire spells (bepbepimjep, #10969)

jmbjr:
See: https://crawl.develz.org/mantis/view.php?id=10969

Currently, not all fire spells are causing Dith to dish out
penance/wrath.

inner flame
ignite poison
bolt of magma
ignition
firestorm

Had to add some extra BEAMS to apply_beam_conducts() and then point
cast_ignition and cast_fire_storm to the same function.

Ignite_poison is handled similar to how SPELL_SWIFTNESS is handled for
the is_hasty conduct (check for the spell and send a DID_FIRE to
did_god_conduct).

### DIFF
--- a/crawl-ref/source/spl-cast.cc
+++ b/crawl-ref/source/spl-cast.cc
@@ -920,6 +920,11 @@ static void _spellcasting_god_conduct(spell_type spell)
     if (is_corpse_violating_spell(spell))
         did_god_conduct(DID_CORPSE_VIOLATION, conduct_level);
 
+    // not is_fiery_spell since the other ones handle the conduct themselves.
+    // need to handle ignite poison separately since it's not handled elsewhere.
+    if (spell == SPELL_IGNITE_POISON)
+        did_god_conduct(DID_FIRE, conduct_level);
+
     // not is_hasty_spell since the other ones handle the conduct themselves.
     if (spell == SPELL_SWIFTNESS)
         did_god_conduct(DID_HASTY, conduct_level);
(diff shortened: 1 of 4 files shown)


jmbjr:
I don't think this is quite ready yet, but wanted to get some comments. 
(Thanks to Neil for some late-night help)

DONE: verified that all fire spells now cause Dith to apply penance/wrath
Partially Done: verified that monsters casting the 5 spells in question don't trigger penance/wrath (so far just checked Cerebov)
Not Started: documentation cleanup

Question: bepbepimjep got credit in the mantis item for reporting this issue. How am I supposed to handle this in my PR? Should I have added (bepbepimjep) in my commit title?

crawl-ref/source/spl-cast.cc
```diff
@@ -920,6 +920,10 @@ static void _spellcasting_god_conduct(spell_type spell)
     if (is_corpse_violating_spell(spell))
         did_god_conduct(DID_CORPSE_VIOLATION, conduct_level);
 
+    // need to handle ignite_poison as a DID_FIRE
```
jmbjr: reviewing this, I think I should use a comment similar to SPELL_SWIFTNESS below.
// not is_fiery_spell since the other ones handle the conduct themselves

amalloy:
Yes, it would be polite to put `(bepbepimjep)` at the end of the first line of your commit message. That's easy enough for you to fix with `git commit --amend`, or for anyone to fix when merging this PR.

jmbjr:
I can't find any monsters who use Inner Flame, Ignite Poison, and Ignition.
I did confirm that monsters that cast Bolt of Magma do not create any odd penance issues.

Also, I tweaked the comment I mentioned above.

PR #591, opened 2017-08-10
Improve bear berserk

alexjurkiewicz:
Bears mostly die before they can berserk, or they berserk and then die a
turn or two later. Let them cast it instantly when in an emergency (like
orc battlecry) and double the cast frequency.

---
This change is [Reviewable](https://reviewable.io/reviews/crawl/crawl/591)

### DIFF
--- a/crawl-ref/source/mon-spell.h
+++ b/crawl-ref/source/mon-spell.h
@@ -284,7 +284,8 @@ static const mon_spellbook mspell_list[] =
 
     {  MST_BEAR,
       {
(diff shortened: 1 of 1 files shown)


rawlins:
Seems reasonable to change the frequency, but I'm not entirely sure about the instant part -- this will (I think) allow bears to with no warning be hitting at a level substantially higher than what xv shows.

Battlecry is not instant (and has never been that I can find - though I looked kind of quickly), but if it were, I'd have exactly the same concern about it (even more so, because its spikiness would be extremely dependent on move order in `world_reacts`).

ebering:
Github auto-close doesn't recognize "Closes 591", only "Closes #591". Oops.

PR #592, opened 2017-08-10
Give Frances & Harold better weapons

alexjurkiewicz:


isloat:
I'm not against standardizing URUG_WEAPONS into something slightly more generic, but I'm not sure that Frances and Harold were considered less difficult than appropriate?

alexjurkiewicz:
closing, no interest

PR #594, opened 2017-08-11
Improve trap interaction with fixed-position summons (Quatzecoatl, #11059)

jmbjr:
EDIT: I just saw this PR that addresses the same issue by simply allowing the prisms/spires to be placed on the traps in question: https://github.com/crawl/crawl/pull/536 , rather than the approach I was taking to warn about the hidden traps (similarly to the warnings about hidden monsters).  Is the other approach preferred? the code is simpler.

See https://crawl.develz.org/mantis/view.php?id=11059

Situation: when casting f. prism or lightning spire on a known or unknown teleport or shaft trap, the spell fizzles without really explaining why to the user.  "You see a puff of smoke. Nothing appears to happen".

Target: warn user when trying to cast f prism or spire on a known teleport or shaft trap.  give a flavor warning for unknown teleport or shaft traps (similar to the existing warning for invisible monsters).

Bonus: handle FIXME suggestion to turn off auto-pickup when you "find" an invisible monster.

This is still WIP but wanted to start getting some feedback.

DONE:
-updated cast_fulminating_prism to handle trap interaction.
1. Only special case teleport traps and shafts. Other traps (known and unknown) are not affected.
2. If trap is known, directly inform player it's not possible to place the prism and abort the spell.
--"You can't place the prism on this trap."
3. If trap is not known, provide a flavorful message, similar to trying to place on an invisible monster. (still charge the MP if the spell doesn't fail)
-- "You see a shimmering outline there, and the spell fizzles."
The idea is, you don't know if it's a tele trap or shaft. Might not be a huge deal but it seemed a bit too easy to just reveal the trap.

- Handle lightning_spire (refactor f. prism logic into new function that both spells can use)
- Pass caster info to function. No monsters use the spell, so it should just be to handle ghosts
- Reworded some messages and made them more generic to handle spire and prism
- Turn off auto-pickup when casting on invis monsters
- Add some docstrings/documentation

ToDo: more testing and get feedback.
Do I need more documentation? the two cast_ functions (f prism and l. spire) don't have docstrings, but neither do nearly all of the functions in spl-summoning.  Should I add them to the two functions I modified?

### DIFF
--- a/crawl-ref/source/spl-summoning.cc
+++ b/crawl-ref/source/spl-summoning.cc
@@ -63,6 +63,7 @@
 #include "teleport.h"
 #include "terrain.h"
 #include "timed-effects.h"
+#include "traps.h"
 #include "unwind.h"
 #include "viewchar.h"
 #include "xom.h"
@@ -2990,6 +2991,34 @@ spret_type cast_fulminating_prism(actor* caster, int pow,
                 canned_msg(MSG_GHOSTLY_OUTLINE);
         }
         return SPRET_SUCCESS;      // Don't give free detection!
+    }
+
+    // check for traps
+    // TODO: combine with above
+    const trap_def *ptrap = trap_at(where);
+    if (ptrap)
+    {
+        const trap_def& trap = *ptrap;
+        const bool player_knows_trap = (trap.is_known(&you));
+
+        // Only shaft and teleport traps should prevent prisms from being placed
+        if (trap.type == TRAP_TELEPORT || trap.type == TRAP_TELEPORT_PERMANENT || trap.type == TRAP_SHAFT)
+        {
+            // can you see the trap?
+            if (player_knows_trap)
+            {    // then tell them straight up they can't target it
+                if (caster->is_player())
+                    mpr("You can't place the prism on this trap.");
+                return SPRET_ABORT;
+            }
+            else // Flavor text
+            {
+                //canned_msg(MSG_GHOSTLY_TRAP);
+                mpr("GHOSTLY TRAP");
+            }
+            // we charge mana for the detection of the trap
+            return SPRET_SUCCESS;
+        }
     }
 
     fail_check();
--- a/crawl-ref/source/spl-summoning.cc
+++ b/crawl-ref/source/spl-summoning.cc
@@ -953,32 +953,9 @@ spret_type cast_summon_lightning_spire(int pow, const coord_def& where, god_type
         return SPRET_SUCCESS;
     }
 
-    // check for traps
-    const trap_def *ptrap = trap_at(where);
-    if (ptrap)
-    {
-        const trap_def& trap = *ptrap;
-        const bool player_knows_trap = (trap.is_known(&you));
-
-        // Only shaft and teleport traps should prevent spires from being placed
-        if (trap.type == TRAP_TELEPORT || trap.type == TRAP_TELEPORT_PERMANENT || trap.type == TRAP_SHAFT)
-        {
-            // if you can see the trap, tell the player straight up they can't target it
-            if (player_knows_trap)
-            {
-                mpr("You can't place the spire on this trap.");
-                return SPRET_ABORT;
-            }
-            //give a vague message. players won't know if it's a teleport or shaft trap
-            else
-            {
-                fail_check();
-                mpr("You see a shimmering outline there, and the spell fizzles.");
-            }
-            // we charge mana for the detection of the trap
-            return SPRET_SUCCESS;
-        }
-    }
+    spret_type handle_trap = handle_trap_at_target_location(where, fail);
+    if (handle_trap != SPRET_NONE)
+        return handle_trap;
 
     fail_check();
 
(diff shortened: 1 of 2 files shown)


jmbjr:
handle summon lightning spire similarly to f. prism (floated logic into new function).

I'm not certain I did it correctly, but my tests are working.

crawl-ref/source/spl-summoning.cc
```diff
@@ -948,10 +949,14 @@ spret_type cast_summon_lightning_spire(int pow, const coord_def& where, god_type
         fail_check();
 
         // invisible monster
-        mpr("Something you can't see is blocking your construction!");
+        canned_msg(MSG_GHOSTLY_OUTLINE);
```
jmbjr: Switched this to use the canned_msg that f. prism uses. Seemed like a reasonable thing to make them behave the same way.

crawl-ref/source/spl-summoning.cc
```diff
@@ -948,10 +949,14 @@ spret_type cast_summon_lightning_spire(int pow, const coord_def& where, god_type
...
         return SPRET_SUCCESS;
     }
 
+    spret_type handle_trap = handle_trap_at_target_location(where, fail);
+    if (handle_trap != SPRET_NONE)
+        return handle_trap;
+
```
jmbjr: is this the appropriate way to do this? It feels a bit wonky, but my intent was:

1. see if we need to handle traps at the target location.
2. if so, it will return something other than SPRET_NONE, so anything else we get we gotta return to the function above this.

I don't know if there's a better C++ pattern for something like this.

crawl-ref/source/spl-summoning.cc
```diff
@@ -3439,3 +3448,34 @@ int count_summons(const actor *summoner, spell_type spell)
 
     return count;
 }
+
```
jmbjr: this needs more work since I need to make the mpr messages more generic and need to add back the logic to handle whether a non-player is the caster (in the original f. prism code, first commit in this PR).

jmbjr:
The basic logic should be in place, I just need to add some comments and do more testing.

Mainly, I can't seem to create a ghost with fulminanting prism.

jmbjr:
I also need to squash the commits. most of them are just placeholders for now

jmbjr:
This is basically done. just need some fresh eyes on it to see what I'm missing.

neilmoore:
I'd suggest `can_place_on_trap` rather than the three-way or on the trap type... the only problem is that the function requires a mon_type, though that is only relevant to tentacle segments at the moment.

Purely formatting:
  * I see a lot of long (>80 character) lines
  * I'd also capitalise sentences and use complete punctuation, especially but not only in docstrings.  Especially in docstrings because those get reflowed when formatted by doxygen, so the missing punctuation isn't at the end of the line anymore.
  * I don't like putting a comment line between `}` and `else`.  I'd move it inside the `else { }` block. , This is mostly personal preference.
  * I don't like // without a space after it, but again that's personal preference.

jmbjr:
I handled all of the suggestions. Additionally, I renamed the function to _check_fixed_summon_loc (thanks Neil) and rewrote the docstring to indicate its purpose as more of a general handler for fixed position summons, rather than simply focused on traps (though for now, that's all it does).

Additionally, the caster argument was removed since monsters do not cast these two spells.

PleasingFungus:
I'd personally lean toward the solution in #536, for (a) simplicity and (b) because any kind of gameplay involving 'using spells to hunt for traps' sounds really bad to me.

Was there discussion in IRC about this? I'm rather out of the loop. If people do prefer, I'll take a look at this one; otherwise, I think I'd merge 536 in.

(Two weeks is more than enough to make a decision about this...)

jmbjr:
I did get some help in irc from neil and a few other devs over the course of the PR (see neil's comments above).  I pinged neil and a couple other devs in IRC to see if they could take another look at this.

a) agreed that simpler is usually good. I think in this case I added a few other things to the PR in addition to the root issue (see items c and d below). 

b) The approach I took was based on neil's original bug report.  You are still charged mana if the spell is successful, and it's similar to how detecting invisible monsters with those spells already works (give player a message indicating they found something that blocked the spell from being placed).  It will only work for teleport and shaft traps, not all traps, and you still don't know which one it is until you discover it or step on it, so I'm not sure how incentivized it really is, though I definitely agree with your concern since I had the same one when I was working on it.

c) if it ends up that #536 is merged, would it still be worth at least pulling in the FIXME's for toggling auto-pickup when detecting invisible monsters? (that feels like the crawl thing to do). There was also a minor pickup on using the same MSG_GHOSTLY_OUTLINE for the spire warning message. 

d) neil sounded like he liked the idea of the _check_fixed_summon_loc function as a good place for some  other fixed position checks (monster on the grid, etc) to be moved there eventually.  I'm not familiar with those checks yet though.

most of the IRC discussion was 8/10 to 8/16 or so, but like I said above, there wasn't a ton of discussion. Mostly with general help with the functions I was working with and a few good conversations with neil.

jmbjr:
I talked to Neil in IRC some more and decided to roll up the other fixed position summon checks into the new function (check for in-range and if the location is habitable to the summoned object).  This addresses and closes item (d) from my previous post.

I was hoping to get into the net positive millimarvin range, but didn't quite make it, which means I can't argue that this PR is net-simpler than the other PR :) 

My opinion is, the new function is probably worth considering (mainly for c and d, with a dash of a for the refactor) and if issue (b) with making hidden traps work the same as hidden monsters (give msg and charge mana) is still deemed to be a bad idea, it will be easy for me to make that change.

wheals:
Closing this since #536 was merged; feel free to open a new one with the refactoring.

PR #595, opened 2017-08-11
Several arrival vaults

giann:
First time writing vaults. I started with basic arrival vaults.

### DIFF
--- a/crawl-ref/source/dat/des/arrival/simple.des
+++ b/crawl-ref/source/dat/des/arrival/simple.des
@@ -2095,3 +2095,75 @@ E.....WWwwwWW.....E
 x....WwwwwwwwW....x
 xxxxxxxxxxxxxxxxxxx
 ENDMAP
+
+###############################################################################
+
+NAME:   giann_horns   
+TAGS:   arrival no_monster_gen
+ORIENT: float
+MAP
+x@xx...................xx@x
+x..xx.................xx..x
+x...xx..G..xx+xx..G..xx...x
+x....xx...xx...xx...xx....x
+x.....xxxxx.....xxxxx.....x
+x............{............x
+xx.......................xx
+ xxxxxxxxxxxx+xxxxxxxxxxxx
+ENDMAP
+
+NAME:   giann_bridges
+TAGS:   arrival no_monster_gen
+MONS:   plant
+ORIENT: float
+MAP
+.........................
+.1wwwwwwww1...1wwwwwwww1.
+.wwwWWWWwww...wwwWWWWwww.
+.wwwWWWWwww...wwwWWWWwww.
+.1wwwwwwww.....wwwwwwww1.
+............{............
+.1wwwwwwww.....wwwwwwww1.
+.wwwWWWWwww...wwwWWWWwww.
+.wwwWWWWwww...wwwWWWWwww.
+.1wwwwwwww1...1wwwwwwww1.
+.........................
+ENDMAP
+
+NAME:   giann_wishing_wells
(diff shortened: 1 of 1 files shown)


crawl-ref/source/dat/des/arrival/simple.des
```diff
@@ -2029,3 +2029,80 @@ x.G.G.G.x
...
+ xxxxxx....{....xxxxxx 
+      xx.......xx
+       xxxxxxxxx
+ENDMAP
+
+NAME:   giann_forest_graveyard
+TAGS:   arrival no_monster_gen
+MONS:   plant
+ORIENT: float
+: dgn.delayed_decay(_G, 'h', 'human skeleton')
```
alexjurkiewicz: if you wanted to be extra clever, this could be a skeleton of the same species as the player -- the memento_mori arrivals vault for implementation details

amalloy: I don't think we need another memento_mori: it's a cool gimmick once, but we don't need every skeleton to use the same trick.

Also, 16 skeletons is kinda a lot. Obviously it's no problem mechanically, but aesthetically it seems like overdoing it.

crawl-ref/source/dat/des/arrival/simple.des
```diff
@@ -2029,3 +2029,80 @@ x.G.G.G.x
...
+.1wwwwwwww.....wwwwwwww1.
+.wwwWWWWwww...wwwWWWWwww.
+.wwwWWWWwww...wwwWWWWwww.
+.1wwwwwwww1...1wwwwwwww1.
+.........................
+ENDMAP
+
+NAME:   giann_guarded_fountains
+TAGS:   arrival no_monster_gen no_item_gen no_rotate
+MONS:   hobgoblin
```
amalloy: Ouch, this is brutal. Make the player fight two hobgiblins as their first encounter, with no escape route? This will kill a lot of players. The decor looks nice but I don't like the mechanics.

Also I think you should replace most of the `.` floor tiles in the top half with ` ` empty tiles, telling the dungeon builder it's okay to put walls or whatever there. Having a huge open area like this is scary for the player, who relies on terrain to create favourable encounters.

Personally I'd remove the hobgoblins and the whole upper half of the map, and just have the lower half for some nice decor. If you're married to the "guarded" theme, I suggest some less formidable guards.

crawl-ref/source/dat/des/arrival/simple.des
```diff
@@ -2029,3 +2029,80 @@ x.G.G.G.x
...
+x..xx.................xx..x
+x...xx..G..xx+xx..G..xx...x
+x....xx...xx...xx...xx....x
+x.....xxxxx.....xxxxx.....x
+x............{............x
+xx.......................xx
+ xxxxxxxxxxxx+xxxxxxxxxxxx
+ENDMAP
+
+NAME:   giann_bridges
```
amalloy: Looks fine to me. The terrain is uncomfortably open, giving the player nowhere to hide, but at least if a large group of monsters try to engage you have pseudo-corridors to work with.

crawl-ref/source/dat/des/arrival/simple.des
```diff
@@ -2029,3 +2029,80 @@ x.G.G.G.x
 x.......x
 xxxx+xxxx
 ENDMAP
+
+###############################################################################
+
+NAME:   giann_horns
```
amalloy: Looks cool.

giann:
@amalloy updated the vaults with your comments in mind

PR #597, opened 2017-08-14
Toggle setting for modifier prefix keys / and *

shmup:
A new setting is introduced called `use_modifier_prefix_keys` that
enables and disables '/' behaving as SHIFT and '*' as CTRL.

It's set to true by default.

### DIFF
--- a/crawl-ref/source/cio.cc
+++ b/crawl-ref/source/cio.cc
@@ -75,28 +75,31 @@ private:
 };
 
 int unmangle_direction_keys(int keyin, KeymapContext keymap,
-                            bool fake_ctrl, bool fake_shift)
+                            bool allow_fake_modifiers)
 {
     // Kludging running and opening as two character sequences.
     // This is useful when you can't use control keys (macros, lua) or have
     // them bound to something on your system.
 
-    /* can we say yuck? -- haranp */
-    if (fake_ctrl && keyin == '*')
+    if (allow_fake_modifiers && Options.use_modifier_prefix_keys)
     {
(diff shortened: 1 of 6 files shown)


crawl-ref/source/cio.cc
```diff
@@ -77,26 +77,31 @@ class unwind_cursor
 int unmangle_direction_keys(int keyin, KeymapContext keymap,
                             bool fake_ctrl, bool fake_shift)
 {
+    const bool use_modifier_prefix_keys = Options.use_modifier_prefix_keys;
```
amalloy: What is this local for? You could just use `Options.use_modifier_prefix_keys` directly in the `if` condition.

Alternatively, I think it would be even better to combine the `fake_ctrl` and `fake_shift` arguments into a single `allow_fake_modifiers` argument (since they are never actually passed differently). In that case, you could simplify the decision tree:

```
if (allow_fake_modifiers && Options.use_modifier_prefix_keys)
{
    if (keyin == '*')
    {
        ...
    }
    ...
}
```

shmup: Both are good points, @amalloy

amalloy:
Nice, looks good to me. Not going to merge it immediately, to give |amethyst or anyone else time to review it.

neilmoore:
Looks good to me.  The options_guide.txt could explain a little more clearly (e.g. that you don't have to hold down * the way you hold Ctrl), but that can be improved later.

None of the built-in clua, config, or tests appear to use these prefixes, so they should be fine even with the option set to false.

PR #599, opened 2017-08-19
Display range for spells in monster description

Sandman25DCSS:
Please see
https://crawl.develz.org/tavern/viewtopic.php?f=8&t=23901
No colors yet, I am not sure how to do it properly.

image

### DIFF
--- a/crawl-ref/source/describe-spells.cc
+++ b/crawl-ref/source/describe-spells.cc
@@ -16,6 +16,7 @@
 #include "macro.h"
 #include "menu.h"
 #include "mon-book.h"
+#include "mon-cast.h"
 #include "monster.h" // SEEN_SPELLS_KEY
 #include "prompt.h"
 #include "religion.h"
@@ -516,9 +517,21 @@ static void _describe_book(const spellbook_contents &book,
         }
         else
         {
-            description.cprintf("%c - %s",
+           int pow = mons_power_for_hd(spell, hd, false);
+           int range = spell_range(spell, pow, false);
+           if (range > 0)
+           {
+                description.cprintf("%c - %s (%d)",
+                            spell_letter,
+                            chop_string(spell_title(spell), 25).c_str(),
+                            range);
+           }
+           else
+           {
+                description.cprintf("%c - %s",
                             spell_letter,
                             chop_string(spell_title(spell), 29).c_str());
+           }
         }
 
         // only display type & level for book spells


Sandman25DCSS:
Oh, it should be 24 instead of 25 to have all rows with same size. Sorry.

jmbjr:
Will there eventually be a title indicating that range is parentheses?

Sandman25DCSS:
I am not sure it is needed. We don't have it for Hexes chance.

rawlins:
I think it will need some kind of labeling -- this would be way too cryptic without it.  Percents are much easier to guess without explicit labeling, though that doesn't meant that they shouldn't have one too.  I would also suggest trying to get some kind of stylistic consistency between the percents and the ranges, either have both before the name or both after, with the same alignment.

For colors, I think you should be able to use `<white>...</white>` etc.  If those don't work there needs to be another wrapper function, I can check on the details later.

I like this change, though I'm more pro-numbers than some so I'm not sure what other devs will make of it.  But the fact that LCS range, vs. all the other spells, is so short is a very good example of information that perhaps should be revealed somehow.

Sandman25DCSS:
Yes, I need some detailed directions what to do here since I am not UI guy, especially for text games with limited space.

rawlins:
I can work on this directly at some point, but I won't get to it for a little while.  I think the simplest thing for labeling the ranges is to provide, before or after the full spell list, some version of the following text: `(the maximum range of the spell is in parenthesis)`. That's what I'd try first, at least.  It would also be helpful to provide a console screenshot with an 80x25 term, since that is the limiting case for size.

Sandman25DCSS:
I believe horizontal size is not changed, I shortened spell name by the same amount of characters as range takes.
Yes, it would be great if someone worked on it, I hope the patch is a good starting place and I really dislike working on UI with all those "let's move that label 3 characters to the left and capitalize it".

rawlins:
This is still on our list, and I think @aidanholm might already have a branch that does it.

aidanholm:
There is indeed: branch `pr-599`, which IIRC only needs a few spells to have their ranges fixed. It could probably be merged as is, since showing a range of (7) for spells like haste is a fairly minor bug.

ebering:
It seems @rawlins didn't get around to that. The branch needed rebasing to thread in some of the hex chance changes and for some new-ui things (I think; merge conflicts anyway). I've done that work, it is in `pr-599-rebase`.

Outstanding issues now are:

- [ ] Add the colour and range numbers to WebTiles
- [ ] Fix the issue @aidanholm mentioned with some monster self-targeting spells displaying a range of `(7)`.

ebering:
The second checkbox isn't completely fixed. The lone outstanding spell is Invisibility, which players can cast at LOS range (as can monsters), but monster AI will never attempt to cast invisible on anything other than the caster.

Options going forward include

- Make Invisibility self targeting only
- Put in a special case for this one spell
- Ignore the wart for now

I favor the first one.

ebering:
Merged, with tweaks and updates to apply cleanly to the current codebase in ab26df7839..9fa157057a

Sorry this took so long to get done, but as you can see it took a bit of doing in the end.

PR #600, opened 2017-08-21
Gnoll changes

Floodkiller:
Changes stat line for Gnolls to 10/10/10 and gives them the Shift Attributes ability.


crawl-ref/source/player-stats.cc
```diff
@@ -690,3 +691,132 @@ bool have_stat_zero()
 
     return false;
 }
+
+/*Gnoll ability to shift three points from one attribute to another,
+ *becoming drained if successful.
```
amalloy: Should have spaces after each `*`, and not have text on the same line as the opening `/*`.

crawl-ref/source/player-stats.cc
```diff
@@ -690,3 +691,132 @@ bool have_stat_zero()
...
+ *
+ *Cannot shift below stat zero, and cannot use if drained (checked in
+ *ability.cc)
+*/
```
amalloy: This `*` is supposed to line up with the others.

crawl-ref/source/player-stats.cc
```diff
@@ -690,3 +691,132 @@ bool have_stat_zero()
...
+spret_type gnoll_shift_attributes()
+{
+    stat_type source_attribute;
+    stat_type destination_attribute;
+    int keyin;
+#ifdef TOUCH_UI
```
amalloy: The TOUCH_UI stuff doesn't look right to me. I don't think this will even compile with `TOUCH_UI` defined, because you re-declare the same variable in both prompts. If it does compile it's a memory leak because you `new` up all these pointers and never `delete` them.

crawl-ref/source/player-stats.cc
```diff
@@ -690,3 +691,132 @@ bool have_stat_zero()
...
+    mprf(MSGCH_PROMPT, "Shift to (S)trength, (I)ntelligence, or (D)exterity? ");
+#endif
+    mouse_control mc2(MOUSE_MODE_PROMPT);
+#ifdef TOUCH_UI
+    keyin = pop->pop();
+#else
+    keyin = getchm();
+#endif
+
+    switch (keyin)
```
amalloy: Instead of repeating this switch, extract it out to a function.

crawl-ref/source/tags.cc
```diff
@@ -2365,6 +2365,15 @@ static void tag_read_you(reader &th)
             you.base_stats[i] = 7;
     }
 #endif
+#if TAG_MAJOR_VERSION == 34
```
amalloy: Save compat looks good to me.

crawl-ref/source/ability.cc
```diff
@@ -1826,6 +1829,17 @@ static spret_type _do_ability(const ability_def& abil, bool fail)
         }
         return frog_hop(fail);
 
+    case ABIL_SHIFT_ATTRIBUTES:
+    {
+        if (you.attribute[ATTR_XP_DRAIN] > 0)
```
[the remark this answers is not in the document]
Floodkiller: Removed the check as suggested.  It was needed in the original implementation of the ability to prevent abuse (shifting to Int to cast a spell, then shifting back immediately afterward), but the current drain cost should be sufficient enough to prevent that from happening.

crawl-ref/docs/crawl_manual.rst
```diff
@@ -1718,12 +1718,13 @@ Gnolls
...
-
-  On the bright side, their powerful noses have adapted to the Dungeon's scents,
-  allowing them to easily locate where treasures lay hidden.
+  effect on their bodies over the generations compared to other residents. Their
+  physical attributes have become severly impaired, unable to be affected by any
+  outside sources. However, stronger gnolls eventually learn to shift their
+  attributes around to adapt to any situation.
+  
+  Finally, their powerful noses have adapted to the Dungeon's scents to easily 
+  locate where rare treasures lay hidden.
```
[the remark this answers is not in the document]
Floodkiller: Fixed and rebased into the Shift Attributes commit.

crawl-ref/source/ability.cc
```diff
@@ -1826,6 +1829,17 @@ static spret_type _do_ability(const ability_def& abil, bool fail)
         }
         return frog_hop(fail);
 
+    case ABIL_SHIFT_ATTRIBUTES:
+    {
```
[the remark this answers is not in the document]
Floodkiller: Fixed and rebased into the Shift Attributes commit.

crawl-ref/source/player-stats.cc
Floodkiller: Fixed and rebased into the Shift Attributes commit.

crawl-ref/source/player-stats.cc
Floodkiller: Fixed and rebased into the Shift Attributes commit.

crawl-ref/source/player-stats.cc
Floodkiller: Fixed by moving TOUCH_UI and switch statements for choosing a stat into a single function instead.

crawl-ref/source/player-stats.cc
Floodkiller: Fixed by moving TOUCH_UI and switch statements for choosing a stat into a single function instead.

Floodkiller:
Rebased to merge minor commits fixing the Shift Attributes ability (based on requested changes) into the Shift Attributes commit.  Moving the stat choice prompt to a function was significant enough to leave as its own commit.

crawl-ref/source/player-stats.cc
```diff
@@ -97,28 +97,22 @@ int innate_stat(stat_type s)
...
-                                                  " dramatic" : "n");
-    crawl_state.stat_gain_prompt = true;
 #ifdef TOUCH_UI
-    learned_something_new(HINT_CHOOSE_STAT);
-    Popup *pop = new Popup("Increase Attributes");
+    if (increase_attribute)
+    {
+        learned_something_new(HINT_CHOOSE_STAT);
+    }
+    Popup *pop = new Popup(title);
```
amalloy: I still don't see a `delete` of this `new`ed Popup or any of the `MenuEntry`s. Is that handled automatically somehow by the touchui code? That would surprise me.

Floodkiller: If it isn't automatically deallocated by touchui or leaving the function, then the stated problem also exists in the master branch: attribute_increase() does not appear to delete the newed Popup or MenuEntrys after the player makes a choice.  I'll look into it further to see if I can find an answer, and figure out how to add in deletes if they aren't being deallocated already.

Floodkiller: Added a commit to manually delete the Popup and MenuEntrys when finishing the function.  Leave it out when merging if it turns out they are already being deleted elsewhere.

Floodkiller:
No longer applicable to current version, closing.

PR #601, opened 2017-08-21
Properly update shafted and banished monster timed-effects (PleasingFungus, #10971)

jmbjr:
see https://crawl.develz.org/mantis/view.php?id=10971

After discussing with gammafunk and |amethyst in IRC, came up with the following plan:

1. generalize timed-effects::update_level() code by creating a new function, update_monster, which handles updating a single monster.
2. add int transit_start_time to follower struct so we know when the monster was shafted or banished
3. call update_monster function after placing shafted/banished monster (via the_lost_ones list).
4. save compatibility
5. documentation


crawl-ref/source/timed-effects.cc
```diff
@@ -1438,6 +1416,41 @@ void update_level(int elapsedTime)
     delete_all_clouds();
 }
 
+bool update_monster(monster* mon, int turns)
```
amalloy: Should take a `monster&`.

crawl-ref/source/mon-transit.cc
```diff
@@ -164,6 +175,7 @@ static void _level_place_lost_monsters(m_transit_list &m)
                 // old loc isn't really meaningful
                 new_mon->apply_location_effects(new_mon->pos());
             m.erase(mon);
+
```
amalloy: Revert whitespace change.

crawl-ref/source/mon-transit.cc
```diff
@@ -142,7 +144,16 @@ void place_followers()
 static bool _place_lost_monster(follower &f)
 {
     dprf("Placing lost one: %s", f.mons.name(DESC_PLAIN, true).c_str());
-    return f.place(false);
+    if (f.place(false))
```
amalloy: Consider swapping the order of these two clauses, so that the error handling is simple and up-front, and then the main body of the function can be read all at once and less indented.

crawl-ref/source/mon-transit.cc
```diff
@@ -142,7 +144,16 @@ void place_followers()
...
+    {
+        //update monster
+        int turns = (you.elapsed_time - f.transit_start_time)/10;
+        mprf(MSGCH_WARN, "updating lost monster: %s, and simulating %i turns",
```
amalloy: %d, not %i.

jmbjr:
I updated update_monster to take monster&, then changed all the mon-> to mon.
Had to pass **mi from update_level().

However, things still aren't working correctly. 
I added some messages inside update_monster to show the HP pre and post healing, and the messages indicate that the monster *should* be healing back to max hp after waiting enough turns after it gets shafted before going to look for it.  Additionally, the message I added in timeout_enchantments also seems to show that poison should be timing out.  However, when I find my monster again, it is not at maxHP and is still poisoned.  If I shaft it again, wait 500 turns or so, and go find it, I either see the HP going back up to max in my debug messages like before, or I get a crash.

which is my second issue, I now get a crash sometimes in: mon-util.cc::monster_at.
ASSERT(mindex <= MAX_MONSTERS);

I did some debugging and most of the time mindex is under 30 or 40, but when it crashes, it's something like 11000.  I didn't run it enough times to see if the bad mindex was the same value since I ran into yet another issue, sometimes moving onto a space freezes crawl (no crash report).  

I had shafted  my test monster, used &d to find it (waited for my message when placing lost monsters triggered), &D to find it, then &B to teleport close to it.  then I walk away from it to get it in line to chase me into a new shaft trap. 
image

[Weeheir.cs.txt](https://github.com/crawl/crawl/files/1245882/Weeheir.cs.txt)

amalloy:
The fix I added here should fix at least one of your issues, probably both: you were writing to a monster* that no longer pointed anywhere useful, which is why your changes weren't reflected in game. This *might* fix the crash also: maybe you were writing to a dangling pointer that sometimes caused an access violation? Not sure.

At any rate, the new behaviour is still broken because the code decides not to heal or un-poison the monster. But that's better than deciding to do it but failing to.

jmbjr:
Still testing save-compat locally, but my tests so far seem to indicate the underlying issue has been addressed.  Banished and Shafted monsters now heal properly and timeout their enchantments upon being placed from the lost_ones monster transit list.

jmbjr:
Just need to add some documentation, but looking for feedback on whether I'm doing the saving and save compatibility correctly (last 2 commits).

amalloy:
The save compat stuff looks good to me. The only thing I'm not sure of is whether unsetting MF_JUST_SUMMONED in that function is the right answer; it works here, but I don't know the semantics of that flag well enough to say if it's right. Looks reasonable though, so if you think it's right and you can't find anyone else to argue with you I don't object.

jmbjr:
I left the unflagging of MF_JUST_SUMMONED as it was. I looked into it some more and the only other option I could see was to add another MF_SHAFTED flag and check for MF_SHAFTED and MF_BANISHED inside update_monster() instead of simply looking for MF_JUST_SUMMONED, but that seemed like a lot of work when the current solution seems to work.

crawl-ref/source/mon-transit.cc
```diff
@@ -146,12 +146,13 @@ static monster* _place_lost_monster(follower &f)
     dprf("Placing lost one: %s", f.mons.name(DESC_PLAIN, true).c_str());
     if (monster* mons = f.place(false))
     {
-        //update monster
+        // Figure out how many turns we need to update the monster
         int turns = (you.elapsed_time - f.transit_start_time)/10;
-        mprf(MSGCH_WARN, "updating lost monster: %s, and simulating %d turns",
+
```
amalloy: I don't think these two new blank lines make things more readable.

crawl-ref/source/tags.cc
```diff
@@ -1827,10 +1827,14 @@ static follower unmarshall_follower(reader &th)
     unmarshallMonster(th, f.mons);
 #if TAG_MAJOR_VERSION == 34
     if (th.getMinorVersion() >= TAG_MINOR_FOLLOWER_TRANSIT_TIME)
+#endif
```
amalloy: Yes, good catch.

crawl-ref/source/timed-effects.cc
```diff
@@ -1403,11 +1402,9 @@ void update_level(int elapsedTime)
 #ifdef DEBUG_DIAGNOSTICS
         mons_total++;
 #endif
-
+        // Try to update monster. If we return nullptr, we
```
amalloy: Forgot to finish this comment, it seems.

crawl-ref/source/timed-effects.cc
```diff
@@ -1417,28 +1414,31 @@ void update_level(int elapsedTime)
     delete_all_clouds();
 }
 
+/**
```
amalloy: :+1: Always love to see new docstrings added.

crawl-ref/source/timed-effects.cc
```diff
@@ -1417,28 +1414,31 @@ void update_level(int elapsedTime)
...
         return nullptr;
     }
 
-    // Following monsters don't get movement.
+    // Ignore monsters flagged to skip their next action
     if (mon.flags & MF_JUST_SUMMONED)
-        return &mon;
-
-
+        return nullptr;
```
amalloy: Are you sure about this? My plan was for `update_monster` to return the monster if it exists, regardless of whether it was updated by this function; returning nullptr only if the update causes the monster to not exist. If you make this change I think you'll need to also change at least one caller of `update_monster`.

crawl-ref/source/timed-effects.cc
```diff
@@ -1449,10 +1449,7 @@ monster* update_monster(monster& mon, int turns)
     mon.foe_memory = max(mon.foe_memory - turns, 0);
 
     if (turns >= 10 && mon.alive())
```
amalloy: I know you didn't add this code, but this could use a TODO about converting from the literal `10` to whatever constant we use for converting from turns to auts.

crawl-ref/source/timed-effects.cc
```diff
@@ -1417,28 +1414,31 @@ void update_level(int elapsedTime)
     delete_all_clouds();
 }
 
+/**
+ * Update the monster upon the player's return
+ *
+ * @param mon   The monster to update.
+ * @param turns How many turns the monster has been away from the player.
```
amalloy: Might be a good idea to explicitly state that it's turns and not auts. I have trouble sometimes when I just see "turns" because sometimes we say turns but mean auts. @neilmoore may disagree with me though.

jmbjr:
Ah, I think I get what you were doing for the return &mon vs return nullptr and I agree.  returning early has the same effect as the previous continue pre update_monster() function. I switched it back.

I also removed the blank lines and the debug message. it was causing some errors in the build and I don't think it's really needed anyway.

I added the FIXME for using the constant instead of the literal 10.

I also tweaked/removed a couple comments.

Thanks for the fast feedback

amalloy:
Great work. I'm happy to merge it in if you're done.

jmbjr:
Yea, I think it's done. Thanks again for the help.

PR #604, opened 2017-08-29
Clarify failure message for quaffing !invis while haloed (Yermak, #11104)

jmbjr:
See: https://crawl.develz.org/mantis/view.php?id=11104

Currently, self-haloing prevents !invis, even when umbraed.

The issue is, the msg to the player is vague and doesn't clarify that the reason the player can't go invisible at all is due to their weapon (EOS) or their Divine Halo (TSO or Wu Jian). Additionally, since the player is actually able to quaff !invis when non-self-haloed haloed (angels, sunshine, etc) + umbraed, it could be confusing and feel inconsistent. 

With this fix, the actual reason for preventing quaffing !invis is provided to the player.

I don't know if I'm using ASSERT appropriately, but my concern was a future halo source that wouldn't get properly handled by this message.  I'm assuming there may be a better way to do what I'm trying to do.

I also briefly considered just allowing quaffing !invis always if you were haloed (any source) + umbraed, but the code seems to intentionally make the distinction between self-halos and simply being backlit by an external halo source, so I didn't change the behavior.

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -4750,18 +4750,19 @@ bool invis_allowed(bool quiet, string *fail_reason)
 
     if (you.haloed() && you.halo_radius() != -1)
     {
-        if (player_equip_unrand(UNRAND_EOS))
-            reason = "Your weapon shines too brightly";
-        if (you.attribute[ATTR_HEAVENLY_STORM] > 0 ||
-            you.religion == GOD_SHINING_ONE)
-        {
-            if (reason.empty())
-                reason = "Your divine halo glows too radiantly";
-            else
-                reason = "Your weapon and divine halo glow too brightly";
-        }
+        bool divine = you.attribute[ATTR_HEAVENLY_STORM] > 0 ||
+                you.religion == GOD_SHINING_ONE;
+        bool weapon = player_equip_unrand(UNRAND_EOS);
 
-        ASSERT(!reason.empty());
+        if (divine && weapon)
+            reason = "Your weapon and divine halo glow too brightly";
+        else if (divine)
+            reason = "Your divine halo glows too radiantly";
+        else if (weapon)
+            reason = "Your weapon shines too brightly";
+        else
+            die("haloed by an unknown source");
+
         msg = reason + " to become invisible.";
         success = false;
     }
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -4751,15 +4751,18 @@ bool invis_allowed(bool quiet, string *fail_reason)
     if (you.haloed() && you.halo_radius() != -1)
     {
         if (player_equip_unrand(UNRAND_EOS))
-            reason = "weapon";
+            reason = "Your weapon shines too brightly";
         if (you.attribute[ATTR_HEAVENLY_STORM] > 0 ||
             you.religion == GOD_SHINING_ONE)
         {
-            reason = (reason == "" ? "Divine Halo" : reason + " and Divine Halo");
+            if (reason.empty())
+                reason = "Your divine halo glows too radiantly";
+            else
+                reason = "Your weapon and divine halo glow too brightly";
         }
 
-        ASSERT(reason != "");
-        msg = "Your " + reason + " prevents invisibility.";
+        ASSERT(!reason.empty());
+        msg = reason + " to become invisible.";
         success = false;
     }
     else if (you.backlit())
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -4745,11 +4745,21 @@ void dec_channel_player(int delay)
 bool invis_allowed(bool quiet, string *fail_reason)
 {
     string msg;
+    string reason = "";
     bool success = true;
 
     if (you.haloed() && you.halo_radius() != -1)
     {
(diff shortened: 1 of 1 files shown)


jmbjr:
note 1:  you could argue that the existing message, "Your halo.." already indicates you are self-haloed, but adding the actual reason, the weapon or Divine Halo, seems much more clear.

note 2: I didn't know if it was OK to capitalise Divine Halo as a way to reinforce it's due to a god effect.

crawl-ref/source/player.cc
```diff
@@ -4745,11 +4745,21 @@ void dec_channel_player(int delay)
...
-        msg = "Your halo prevents invisibility.";
+        if (player_equip_unrand(UNRAND_EOS))
+            reason = "weapon";
+        if (you.attribute[ATTR_HEAVENLY_STORM] > 0 ||
+            you.religion == GOD_SHINING_ONE)
+        {
+            reason = (reason == "" ? "Divine Halo" : reason + " and Divine Halo");
+        }
+
+        ASSERT(reason != "");
```
amalloy: `!reason.empty()`

crawl-ref/source/player.cc
```diff
@@ -4745,11 +4745,21 @@ void dec_channel_player(int delay)
...
 
     if (you.haloed() && you.halo_radius() != -1)
     {
-        msg = "Your halo prevents invisibility.";
+        if (player_equip_unrand(UNRAND_EOS))
+            reason = "weapon";
+        if (you.attribute[ATTR_HEAVENLY_STORM] > 0 ||
+            you.religion == GOD_SHINING_ONE)
+        {
+            reason = (reason == "" ? "Divine Halo" : reason + " and Divine Halo");
```
amalloy: `.empty()`.

amalloy: Also I think capitalising it is wrong. It's not an ability named "Divine Halo" (right?), it's a halo which is divine.

crawl-ref/source/player.cc
```diff
@@ -4745,11 +4745,21 @@ void dec_channel_player(int delay)
...
+        }
+
+        ASSERT(reason != "");
+        msg = "Your " + reason + " prevents invisibility.";
```
amalloy: Should use "prevent" if there is more than one reason.

crawl-ref/source/player.cc
```diff
@@ -4745,11 +4745,21 @@ void dec_channel_player(int delay)
...
 {
     string msg;
+    string reason = "";
     bool success = true;
 
     if (you.haloed() && you.halo_radius() != -1)
     {
-        msg = "Your halo prevents invisibility.";
+        if (player_equip_unrand(UNRAND_EOS))
+            reason = "weapon";
```
amalloy: I would say "weapon's halo" or something like that, to explain why the weapon makes invisibility impossible. Maybe even change the "your ____ prevents invisibility" style to "Your weapon shines too brightly to become invisible".

jmbjr: Great idea!

crawl-ref/source/player.cc
jmbjr: I take it this means this is a proper use of ASSERT?  (and doh on the .empty(), ty)

crawl-ref/source/player.cc
jmbjr: this actually is irrelevant now after the changes to msg upstream.

jmbjr:
I updated the messages. I really liked the suggestion to add a bit more flavor to it and vastly prefer the new messages now.

all other suggestions also addressed/fixed.

crawl-ref/source/player.cc
```diff
@@ -4745,11 +4745,24 @@ void dec_channel_player(int delay)
...
 {
     string msg;
+    string reason = "";
     bool success = true;
 
     if (you.haloed() && you.halo_radius() != -1)
     {
-        msg = "Your halo prevents invisibility.";
+        if (player_equip_unrand(UNRAND_EOS))
+            reason = "Your weapon shines too brightly";
```
amalloy: It feels unsatisfying to me to set this reason and then perhaps overwrite it later. In particular, this approach only works because you currently have exactly two halo sources. If there were more, you wouldn't be able to know which halo source had previously been set.

Better, I think, is to have non-overlapping `if` clauses, like this:

```
bool divine = (...);
bool weapon = (...);

if (divine && weapon)
  ...
else if (divine)
  ...
else if (weapon)
  ...
else
  die("haloed by an unknown source");
```

jmbjr:
Updated based on latest input.  I agree that it seems to flow better now.

I didn't know about die().  Is that a better way to handle the situation that I was concerned with on if a new halo source was added and not handled here? I removed the ASSERT(!message.empty()), but I wasn't sure if that was OK or if I should have both the die() and ASSERT.

amalloy:
IMO check with @neilmoore about whether die is good here. I think it provides the same information as assert but without the conditional, in which case it's perfect here.

amalloy:
I just looked at the source and indeed `die` is fine. It's the right thing to use when you want to crash because you've wandered into an impossible codepath as opposed to because a variable has taken on an impossible value. I approve these changes and will merge them in if you're done.

jmbjr:
I am done. Thanks!

PR #606, opened 2017-09-01
Remove monster spawns tweak level monster generation

gammafunk:
This PR is to review this branch for any technical issues (mostly a bunch of removed monster placement code) and discuss any needed changes to monster generation.

In this branch, monster spawns are gone from connected branches, taking place only on the Orb run in Abyss and Pan with the behavior in those branches essentially unchanged. Monsters no longer have a chance to try to spawn "from" staircases and using the monster lists of those stair's destination level. The code that did this was quite broken, placing fewer than intended "stair monsters", but it's generally just not a useful mechanic, since it either makes weaker monsters from earlier levels or out-of-branch monsters that could be done using vaults.  Also gone is a very weird and extremely rare mechanic that tried to push the player off stairs if the monster spawned on a staircase that the player occupied.

The recently addes xp_by_level table shows that 8-13% of XP in won games comes from spawns. This table is enabled by default in trunk and should be accurate for any game on or after 0.21-a0-218-g8aba26b (use keyword `xpinfo` with `!lg` to find such games). To compensate for XP loss from spawn removal, I've increased level monster generation from 3d10 to 3d12  in all branches except Crypt, with Crypt going from 3d8 to 3d10 instead and Vestibul's extra monsters going from 3d8 to 3d10 (the comment in the relevant commit incorrectly says 2d6 more for most branches). Objstat says this will give about 9% more XP for the Dungeon branch, but I will have to rerun this for an entire set of branches. 

You can find the morgue for a qw GrBe win from this branch [here](http://dpaste.com/3PTBSNA), and [here](http://dpaste.com/00BFNHR) is a qw GrBe win from trunk for comparison. It looks like nospawn is giving qw closer to 33% more XP. If this is not just variance betwen the runs, we might want to make the number of level monsters depend a bit on absdepth. A complete objstat run should give a better picture when I have time to run that.

Another issue is whether we want to do more to prevent players from lingering a long time on the level. The old OOD spawns were bad for just giving scumming potential, but it might be possible to use no-XP durable summons or perhaps even normal summons just spawned near the player. If find these solutions to be fairly annoying, especially durable summons, and we do have the food clock for the time being.

### DIFF
--- a/crawl-ref/source/dungeon.cc
+++ b/crawl-ref/source/dungeon.cc
@@ -1338,12 +1338,12 @@ static int _num_mons_wanted()
         return 0;
 
     if (player_in_branch(BRANCH_CRYPT))
-        return roll_dice(3, 13);
+        return roll_dice(3, 10);
 
-    int mon_wanted = roll_dice(3, 15);
+    int mon_wanted = roll_dice(3, 12);
 
     if (player_in_hell())
-        mon_wanted += roll_dice(3, 13);
+        mon_wanted += roll_dice(3, 10);
 
     if (mon_wanted > 60)
         mon_wanted = 60;
--- a/crawl-ref/source/dungeon.cc
+++ b/crawl-ref/source/dungeon.cc
@@ -1338,12 +1338,12 @@ static int _num_mons_wanted()
         return 0;
 
     if (player_in_branch(BRANCH_CRYPT))
-        return roll_dice(3, 10);
+        return roll_dice(3, 13);
 
-    int mon_wanted = roll_dice(3, 12);
+    int mon_wanted = roll_dice(3, 15);
 
     if (player_in_hell())
-        mon_wanted += roll_dice(3, 10);
+        mon_wanted += roll_dice(3, 13);
 
     if (mon_wanted > 60)
         mon_wanted = 60;
--- a/crawl-ref/source/dungeon.cc
+++ b/crawl-ref/source/dungeon.cc
@@ -3763,7 +3763,6 @@ static void _builder_monsters()
     int mon_wanted = _num_mons_wanted();
 
     const bool in_shoals = player_in_branch(BRANCH_SHOALS);
-    const bool in_pan    = player_in_branch(BRANCH_PANDEMONIUM);
     if (in_shoals)
         dgn_shoals_generate_flora();
 
@@ -3777,12 +3776,20 @@ static void _builder_monsters()
     for (int i = 0; i < mon_wanted; i++)
     {
         mgen_data mg;
-        if (!in_pan)
+        // Chance to generate the monster awake, but away from level stairs.
+        if (player_in_connected_branch()
+            // Exclude D:1 from generating these awake monsters.
+            && (!player_in_branch(BRANCH_DUNGEON) || you.depth > 1)
+            && one_chance_in(8))
+        {
+            mg.proximity = PROX_AWAY_FROM_STAIRS;
+        }
+        // Pan monsters always generate awake.
+        else if (!player_in_branch(BRANCH_PANDEMONIUM))
             mg.behaviour = BEH_SLEEP;
         mg.flags    |= MG_PERMIT_BANDS;
         mg.map_mask |= MMT_NO_MONS;
         mg.preferred_grid_feature = preferred_grid_feature;
-
         place_monster(mg);
     }
 
--- a/crawl-ref/source/dungeon.cc
+++ b/crawl-ref/source/dungeon.cc
@@ -1338,12 +1338,12 @@ static int _num_mons_wanted()
         return 0;
 
     if (player_in_branch(BRANCH_CRYPT))
-        return roll_dice(3, 8);
+        return roll_dice(3, 10);
 
-    int mon_wanted = roll_dice(3, 10);
+    int mon_wanted = roll_dice(3, 12);
 
     if (player_in_hell())
-        mon_wanted += roll_dice(3, 8);
+        mon_wanted += roll_dice(3, 10);
 
     if (mon_wanted > 60)
         mon_wanted = 60;
--- a/crawl-ref/source/mon-place.cc
+++ b/crawl-ref/source/mon-place.cc
@@ -511,53 +511,6 @@ bool drac_colour_incompatible(int drac, int colour)
     return drac == MONS_DRACONIAN_SCORCHER && colour == MONS_WHITE_DRACONIAN;
 }
 
-// Finds a random square as close to a staircase as possible
-static bool _find_mon_place_near_stairs(coord_def& pos,
-                                        dungeon_char_type *stair_type,
-                                        level_id &place)
-{
-    pos = get_random_stair();
-    const dungeon_feature_type feat = grd(pos);
-    *stair_type = get_feature_dchar(feat);
-
-    // First, assume a regular stair.
-    switch (feat_stair_direction(feat))
-    {
-    case CMD_GO_UPSTAIRS:
-        if (place.depth > 1)
-            place.depth--;
-        break;
-    case CMD_GO_DOWNSTAIRS:
-        if (place.depth < brdepth[place.branch])
-            place.depth++;
-        break;
-    default: ;
-    }
-
-    // Is it a branch stair?
-    for (branch_iterator it; it; ++it)
-    {
-        if (it->entry_stairs == feat)
-        {
-            place = it->id;
-            break;
-        }
-        else if (it->exit_stairs == feat)
-        {
-            place = brentry[it->id];
-            // This can happen on D:1 and in wizmode with random spawns on the
-            // first floor of a branch that didn't generate naturally.
-            if (!place.is_valid())
-                return false;
-            break;
-        }
-    }
-    const monster_type habitat_target = MONS_BAT;
-    int distance = 3;
-    pos = find_newmons_square_contiguous(habitat_target, pos, distance);
-    return in_bounds(pos);
-}
-
 bool needs_resolution(monster_type mon_type)
 {
     return mon_type == RANDOM_DRACONIAN || mon_type == RANDOM_BASE_DRACONIAN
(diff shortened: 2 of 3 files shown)


crawl-ref/source/dungeon.cc
```diff
@@ -1338,12 +1338,12 @@ static int _num_mons_wanted()
         return 0;
 
     if (player_in_branch(BRANCH_CRYPT))
```
amalloy: You could simplify this function while you're in here, right? Replace the random2avg for pan with an ordinary NdM dice roll (3d9 looks very close), and remove the doubled call to roll_dice for hell. That way every branch is just a single NdM dice roll, and you can make that explicit by only calling `roll_dice` in one place, with the various branches just having slightly different logic for calculating the number and size of dice to use.

crawl-ref/source/dungeon.cc
```diff
@@ -3777,12 +3776,20 @@ static void _builder_monsters()
     for (int i = 0; i < mon_wanted; i++)
     {
         mgen_data mg;
-        if (!in_pan)
+        // Chance to generate the monster awake, but away from level stairs.
+        if (player_in_connected_branch()
+            // Exclude D:1 from generating these awake monsters.
+            && (!player_in_branch(BRANCH_DUNGEON) || you.depth > 1)
```
amalloy: Can you use absdepth here to avoid needing to check both branch and level? The conditional is a little bit hard to read, and I think nothing has the same absdepth as D:1.

crawl-ref/source/mon-place.cc
```diff
@@ -74,11 +74,6 @@ static vector<bool> vault_mon_bands;
 #define VAULT_MON_BANDS_KEY   "vault_mon_bands"
 #endif
 
-// proximity is the same as for mons_place:
-// 0 is no restrictions
-// 1 attempts to place near player
-// 2 attempts to avoid player LOS
-
```
amalloy: I'm in favour of removing this pseudo-enum. But I'm curious where the code is that this used to refer to, and how it was updated to not need this pseudo-enum anymore.

crawl-ref/source/mon-place.cc
```diff
@@ -229,37 +224,6 @@ bool monster_can_submerge(const monster* mon, dungeon_feature_type feat)
...
-    return turns_on_level <= dropoff_start_turns ? base_value :
-           turns_on_level > dropoff_start_turns + dropoff_ramp_turns ?
-           final_value :
-
-           // Actual scaling, strictly linear at the moment:
-           (base_value +
-            (target_value - base_value)
-            * (turns_on_level - dropoff_start_turns)
-            / dropoff_ramp_turns);
-}
```
amalloy: Thank goodness we're getting rid of this mess.

crawl-ref/source/mon-place.cc
```diff
@@ -304,31 +264,14 @@ static void _apply_ood(level_id &place)
...
-    // genodds increases once you've spent more than 500 turns in Hell.
-    if (env.turns_on_level > taper_off_turn)
-    {
-        genodds += (env.turns_on_level - taper_off_turn);
-        genodds  = (genodds < 0 ? 20000 : min(genodds, 20000));
-    }
-
-    return genodds;
+     // On D:13 and deeper, something extreme:
+     if (place.absdepth() >= 12 && x_chance_in_y(2, 10000))
```
amalloy: Why 2/10000 instead of the simpler 1/5000?

crawl-ref/source/mon-place.cc
```diff
@@ -391,26 +324,22 @@ void spawn_random_monsters()
...
         return;
 
-    // Place normal dungeon monsters, but not in player LOS. Don't generate orb
-    // spawns in Abyss to show some mercy to players that get banished there on
-    // the orb run.
-    if (player_in_connected_branch()
-        || (player_on_orb_run() && !player_in_branch(BRANCH_ABYSS)))
+    // Orb spawns. Don't generate orb spawns in Abyss to show some mercy to
+    // players that get banished there on the orb run.
+    if (player_on_orb_run() && !player_in_branch(BRANCH_ABYSS))
```
amalloy: I don't think you should change it in this PR, but seeing this again made me think: is this really a necessary mercy? If I ever get banished during the orb run it's been a big relief, a nice chance to heal up. Maybe we should just let the orb run keep chasing you.

crawl-ref/source/mon-place.cc
```diff
@@ -391,26 +324,22 @@ void spawn_random_monsters()
...
     {
         dprf(DIAG_MONPLACE, "Placing monster, rate: %d, turns here: %d",
              rate, env.turns_on_level);
-        proximity_type prox = (one_chance_in(10) ? PROX_NEAR_STAIRS
-                                                 : PROX_AWAY_FROM_PLAYER);
+        proximity_type prox = PROX_AWAY_FROM_PLAYER;
```
amalloy: This `PROX_AWAY_FROM_PLAYER` is never used: on the next line you always overwrite it either with `PROX_CLOSE_TO_PLAYER` or `PROX_ANYWHERE`.

crawl-ref/source/mon-place.cc
```diff
@@ -391,26 +324,22 @@ void spawn_random_monsters()
...
 
         // The rules change once the player has picked up the Orb...
```
amalloy: Comment obsolete.

crawl-ref/source/mon-place.cc
```diff
@@ -794,15 +610,19 @@ static bool _valid_monster_generation_location(const mgen_data &mg,
...
+    if (mg.proximity == PROX_AWAY_FROM_PLAYER && close_to_player
+        || mg.proximity == PROX_CLOSE_TO_PLAYER && !close_to_player)
     {
         return false;
     }
+    // Check that the location is not proximal to level stairs.
+    else if (mg.proximity == PROX_AWAY_FROM_STAIRS)
+    {
+        for (distance_iterator di(mg_pos, false, LOS_RADIUS); di; ++di)
+            if (feat_is_stone_stair(grd(*di)))
```
amalloy: This applies both to up stairs and down stairs; is that intended? I would have guessed this was just about up stairs, for when a player first enters the level, and that it doesn't really matter if monsters spawn near a down stair.

Also, you probably want to consider branch entrances and/or exits.

crawl-ref/source/mon-place.cc
```diff
@@ -794,15 +610,19 @@ static bool _valid_monster_generation_location(const mgen_data &mg,
...
+    bool close_to_player = grid_distance(you.pos(), mg_pos) <= LOS_RADIUS;
+    if (mg.proximity == PROX_AWAY_FROM_PLAYER && close_to_player
+        || mg.proximity == PROX_CLOSE_TO_PLAYER && !close_to_player)
     {
         return false;
     }
+    // Check that the location is not proximal to level stairs.
+    else if (mg.proximity == PROX_AWAY_FROM_STAIRS)
+    {
+        for (distance_iterator di(mg_pos, false, LOS_RADIUS); di; ++di)
```
amalloy: Why distance_iterator instead of radius_iterator?

Also, this forbids areas that are on the other side of some solid walls, which are in a sense not actually near any stairs. Probably not a big deal, but if it's easy you might consider something like figuring out the pathfinding distance from a position to the nearest stair.

crawl-ref/source/mon-place.cc
```diff
@@ -964,82 +742,19 @@ monster* place_monster(mgen_data mg, bool force_pos, bool dont_place)
...
+    // Sanity check that the specified position is valid.
     else if (!_valid_monster_generation_location(mg) && !dont_place)
-    {
-        // Sanity check that the specified position is valid.
-        return 0;
-    }
+        return nullptr;
 
-    monster* mon = _place_monster_aux(mg, 0, place, force_pos, dont_place);
+    monster* mon
```
amalloy: This is a weird place to break a line. What about breaking after the nullptr instead?

crawl-ref/source/mon-place.cc
```diff
@@ -964,82 +742,19 @@ monster* place_monster(mgen_data mg, bool force_pos, bool dont_place)
...
+        = _place_monster_aux(mg, nullptr, place, force_pos, dont_place);
 
     // Bail out now if we failed.
```
amalloy: Comment is extraneous IMO. It's already quite clear what the next two lines do and why.

crawl-ref/source/mon-place.cc
rawlins: Have you been banished on the orbrun since banishment depth is variable?  At that point in the game you will almost certainly get banished to abyss 5, which isn't much of a break from the orbrun in my experience, especially on a non-stealthy character.  IIRC when this has happened to me the only viable way out was by fighting enough monsters to open a portal.

crawl-ref/source/mon-place.cc
gammafunk: These comments actually go back to the initial import in 673bdae75. I'm not sure specifically how these values were adapted into the proximity_type that mgen_data now uses, but it appears that happened back in Linley crawl.

crawl-ref/source/mon-place.cc
gammafunk: This could be simplified as you say, but elliptic and I agreed that it can simply be removed. It's a very low chance for monsters to be "super OOD" for absdepth equivalent to D:13 and later. It was more meaningful when it was scaled by turns, but vaults placing OOD through `8` and `9` monsters is already common in these later levels. Will remove this block when I rebase/merge.

crawl-ref/source/mon-place.cc
gammafunk: This prox type was added to handle level generation monsters that are generated awake so that the player wouldn't encounter these in LOS when first entering a level. So for that purpose, they should probably only affect stone stairs. It's not possible to first enter a level from a branch or a portal outside of wizard mode. Adding more restrictions increases the chance that a monster doesn't place properly (currently gets 45 tries).

crawl-ref/source/dungeon.cc
gammafunk: Yeah, I can check `env.absdepth0`, so I'll do that

crawl-ref/source/mon-place.cc
gammafunk: @rawlins I've actually just gotten the rune when this deep banishment happens on the orb run, since that's easy to do (esp without orb spawns), and it makes exits appear frequently, so I don't think killing monsters is the only way. Killing monsters may well be a faster way provided nothing too horrible appears.

Having orb spawns in Abyss would arguably be a nerf to Lucy's self-banishment, but I don't feel strongly either way. @amalloy If you find that there's agreement that this orb spawn protection is unecessary, feel free to remove it, but as you say I'll keep it in this PR.

crawl-ref/source/mon-place.cc
gammafunk: Yep, looks like a bit of code only relevant for the now-removed PROX_NEAR_STAIRS got left in, so I'll remove this line.

crawl-ref/source/mon-place.cc
gammafunk: Comment isn't exactly obsolete, since spawns having a special chance for being near the player are important to point out, but I can clean up this vague remark and put it with the comments above the block.

crawl-ref/source/mon-place.cc
gammafunk: Will remove.

crawl-ref/source/mon-place.cc
gammafunk: Yeah, I should probably only use that kind of line break when the right side is not breakable, will update.

crawl-ref/source/dungeon.cc
gammafunk: Good idea, I'll see if I can do that.

gammafunk:
These commits and the agreed upon changes along with some other small fixes were merged in trunk.

PR #608, opened 2017-09-04
Allow setting attitude of cloned monsters (damerell, #11171)

jmbjr:
See https://crawl.develz.org/mantis/view.php?id=11171

This implements option 3 from the bug report.
clone_mons() now accepts monster_attitude_type as a default parameter.  This allows us to directly set the monster attitude to handle the bug in question and maybe bypass some funky gymnastics related to the bug in question.

Additionally, an unused default parameter, pos, is removed from clone_mons() because it is not passed by any of the other locations that call clone_mons().

Finally, the call to clone_mons() from _phantom_mirror() has been updated to pass ATT_FRIENDLY.

### DIFF
--- a/crawl-ref/source/mon-clone.h
+++ b/crawl-ref/source/mon-clone.h
@@ -9,8 +9,7 @@
 bool mons_clonable(const monster* orig, bool needs_adjacent = true);
 monster *clone_mons(const monster* orig, bool quiet = false,
                     bool* obvious = nullptr,
-                    mon_attitude_type mon_att = ATT_SAME,
-                    coord_def pos = coord_def(0, 0));
+                    mon_attitude_type mon_att = ATT_SAME);
 
 void mons_summon_illusion_from(monster* mons, actor *foe,
                                spell_type spell_cast = SPELL_NO_SPELL,
--- a/crawl-ref/source/mon-clone.h
+++ b/crawl-ref/source/mon-clone.h
@@ -1,12 +1,16 @@
 #pragma once
 
+#include "mon-attitude-type.h"
+
 #define CLONE_MASTER_KEY "mcloneorig"
 #define CLONE_SLAVE_KEY "mclonedupe"
 
 // Formerly in mon-stuff:
 bool mons_clonable(const monster* orig, bool needs_adjacent = true);
 monster *clone_mons(const monster* orig, bool quiet = false,
-                    bool* obvious = nullptr, coord_def pos = coord_def(0, 0));
+                    bool* obvious = nullptr,
+                    mon_attitude_type mon_att = ATT_SAME,
+                    coord_def pos = coord_def(0, 0));
 
 void mons_summon_illusion_from(monster* mons, actor *foe,
                                spell_type spell_cast = SPELL_NO_SPELL,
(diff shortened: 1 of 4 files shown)


jmbjr:
What all should I be testing to ensure this is working properly?
Here's what I've tested so far:
1. The bug: enslave Mara, create OCS, and ensure that we only get ONE message : "Marking area around the orange crystal statue as unsafe for travelling." (confirmed)

2. used phantom mirror and ensure nothing "bad" happens. (confirmed. can still clone a monster and it's created as friendly).

3. created a few rakshasas and ensure nothing "weird" happens. (confirmed. rakshasa's behave normally and clone itself).

4. fired up arena and ensured that mara v ocs didn't cause any issues (mara still clones ocs properly).

I've never used arena mode much and I'm definitely not familiar with the arena code, so I'm not sure how to 100% ensure I didn't screw up the arena bookkeeping.

jmbjr:
I just took a closer look at the last recompile and I forgot to handle ATT_SAME in a bunch of switch blocks, so I still need to do that.

crawl-ref/source/mon-clone.h
```diff
@@ -1,12 +1,15 @@
 #pragma once
 
+#include "mon-attitude-type.h"
```
amalloy: Hm, it's unusual for us to include a header from another header. I see why you did it here: it seems necessary to have ATT_SAME in scope. Maybe @neilmoore has a suggestion of another way to do this.

jmbjr:
Yea we came up with a different implementation that avoids having to create ATT_SAME at all. I'm going to test and push hopefully later tonight. 

(Overload function and pass orig->temp_attitude() to the new function that doesn't have any default arguments)

jmbjr:
I guess I lied. I don't know how to remove the #include "mon_attitude_type.h" in mon-clone.h

I talked with geekosaur in IRC a bit, but the discussion went to preprocessors and quickly outpaced my current skill level.

I also did a force push to clean up some earlier commits.

neilmoore:
I think including the mon-attitude-type.h header from mon-clone.h is the right approach.  Otherwise you're just imposing a rule "you must #include mon-attitude-type.h before mon-clone.h", which just makes things for difficult for people trying to use mon-clone.h.

The problem would be with cycles of includes, but since mon-attitude-type.h doesn't include anything else, that's not a problem.

rawlins:
> I think including the mon-attitude-type.h header from mon-clone.h is the right approach.

I think this is actually pretty common, especially for enums that used to be in enum.h before it was broken up, such as mon-attitude-type.h.

    $ git grep "\#include" *.h | wc -l
    782
    $ git grep "\#include" *.h | grep \\\-type | wc -l
    205

jmbjr:
Ok. thanks or the input and clarification. Unless there's any other suggestions, I think I'm done with this one.

geekosaur:
The actual relevant IRC comment was about similar includes coming from `AppHdr.h` via `externs.h`. The preprocessor discussion was about how to track that kind of thing down.

wheals:
Looks good, only comment is that you could stand to squash the last commit onto the first one, so only the second implementation is in history (the other commits can stay on their own, though).

PR #611, opened 2017-09-11
Ghoul HP regeneration near submerged enemies

Velros:
While waiting in an area if there are submerged enemies nearby ghoul hp does not regenerate. Occurs very often in swamp with swamp worms.
Mutation text in game states: You do not regenerate when monsters are visible.

Changed the mutation to ignore submerged enemies.

(Issue's: 0011199, 0011197 on Mantis.)

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -1077,7 +1077,8 @@ bool regeneration_is_inhibited()
         {
             if (mons_is_threatening(**mi)
                 && !mi->wont_attack()
-                && !mi->neutral())
+                && !mi->neutral()
+                && !mi->submerged())
             {
                 return true;
             }


alexjurkiewicz:
Nice! What's the current behaviour with this mutation and invisible enemies (when the player has no SInv)?

Inhibited regeneration should probably ignore both sorts of monster hiding, or be affected by both.

Velros:
I think it needs to be clarified whether it should be active when there are nearby enemies or nearby visible enemies.

Currently, invisible enemies in LOS also prevent regeneration.  The issue with the submerged enemies is if you rest while there is one nearby you can spend a large number of turns (and can be confusing to players who don't know the reason why their health is not regenerating).  Invisible monsters will at least interrupt your resting and if you detect the disturbance of an invisible monster it prevents resting at all.

rawlins:
I think it makes sense based on the discussion to have submerged enemies behave differently than invisible enemies, so I'm going to merge this.  Thanks for the patch!

PR #615, opened 2017-09-15
Don't let players win by tripping out the dungeon (minmay, mong, #10708)

jmbjr:
See: https://crawl.develz.org/mantis/view.php?id=10708

Previously, confused players who tripped up the stairs could actually win,
despite getting a message saying they tripped back down.

This update modifies the returned travel destination after tripping if
the player is standing on the dungeon exit. Before, the destination
returned was the default branch with a depth of -1, which would be an
invalid destination. This was ok since this invalid destination would be
interpreted later to prevent going up-stairs after tripping.
However, there was special casing for standing on the dungeon
exit to skip being blocked from transitioning to the destination. This
logic was also updated.

I also refactored some of the logic that handles stopping players from transitioning to an invalid level.  There's also some new docstrings for the functions related to level transitioning, but I'm not happy with the current explanations and I would appreciate any improvements on wording or the formatting.

### DIFF
--- a/crawl-ref/source/stairs.cc
+++ b/crawl-ref/source/stairs.cc
@@ -427,7 +427,6 @@ static void _new_level_amuses_xom(dungeon_feature_type feat,
  *
  * @param how         How the player is trying to travel.
  *                    (e.g. stairs, traps, portals, etc)
- * @param whence      Not currently used
  * @param forced      True if the player is forcing the traveling attempt.
  *                    (e.g. forcibly exiting the abyss, etc)
  * @param going_up    True if the player is going upstairs.
@@ -439,7 +438,6 @@ static void _new_level_amuses_xom(dungeon_feature_type feat,
  *                    when attempting to leave the dungeon, depth = 1.
  */
 static level_id _travel_destination(const dungeon_feature_type how,
-                                    const dungeon_feature_type whence,
                                     bool forced, bool going_up,
                                     bool known_shaft)
 {
@@ -855,8 +853,8 @@ void take_stairs(dungeon_feature_type force_stair, bool going_up,
     // Latter case is falling down a shaft.
     const bool shaft = known_shaft || force_stair == DNGN_TRAP_SHAFT;
 
-    level_id whither = _travel_destination(how, old_feat,
-                           bool(force_stair), going_up, known_shaft);
+    level_id whither = _travel_destination(how, bool(force_stair), going_up,
+                                           known_shaft);
 
     if (_stop_transition(whither, old_feat, going_up))
         return;
--- a/crawl-ref/source/stairs.cc
+++ b/crawl-ref/source/stairs.cc
@@ -422,6 +422,22 @@ static void _new_level_amuses_xom(dungeon_feature_type feat,
     }
 }
 
+/**
+ * Determine destination level.
+ *
+ * @param how         How the player is trying to travel.
+ *                    (e.g. stairs, traps, portals, etc)
+ * @param whence      Not currently used
+ * @param forced      True if the player is forcing the traveling attempt.
+ *                    (e.g. forcibly exiting the abyss, etc)
+ * @param going_up    True if the player is going upstairs.
+ * @param known_shaft True if the player is intentionally shafting themself.
+ * @return            The destination level, if valid. Note the default value
+ *                    of dest is not valid (since depth = -1) and this is
+ *                    generally what is returned for invalid destinations.
+ *                    But note the special case when failing to climb stairs
+ *                    when attempting to leave the dungeon, depth = 1.
+ */
 static level_id _travel_destination(const dungeon_feature_type how,
                                     const dungeon_feature_type whence,
                                     bool forced, bool going_up,
@@ -469,6 +485,9 @@ static level_id _travel_destination(const dungeon_feature_type how,
         // going 'down' into the vestibule are twice as likely to fall, because
         // they have to pass a check here, and later in floor_transition
         // Right solution is probably to use the canonicalized direction everywhere
+
+        // If player falls down the stairs trying to leave the dungeon, we set
+        // the destination depth to 1 (i.e. D:1)
         if (how == DNGN_EXIT_DUNGEON)
             dest.depth = 1;
         return dest;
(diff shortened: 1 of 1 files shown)


crawl-ref/source/stairs.cc
```diff
@@ -422,6 +422,22 @@ static void _new_level_amuses_xom(dungeon_feature_type feat,
...
+ * @param forced      True if the player is forcing the traveling attempt.
+ *                    (e.g. forcibly exiting the abyss, etc)
+ * @param going_up    True if the player is going upstairs.
+ * @param known_shaft True if the player is intentionally shafting themself.
+ * @return            The destination level, if valid. Note the default value
+ *                    of dest is not valid (since depth = -1) and this is
+ *                    generally what is returned for invalid destinations.
+ *                    But note the special case when failing to climb stairs
+ *                    when attempting to leave the dungeon, depth = 1.
+ */
```
jmbjr: is it better to wrap the lines here like I'm doing? or should I break the 80 char max rule?

crawl-ref/source/stairs.cc
```diff
@@ -422,6 +422,22 @@ static void _new_level_amuses_xom(dungeon_feature_type feat,
     }
 }
 
+/**
+ * Determine destination level.
+ *
+ * @param how         How the player is trying to travel.
+ *                    (e.g. stairs, traps, portals, etc)
+ * @param whence      Not currently used
```
jmbjr: this param is not used. should I remove it? or is it better to just leave it here. Since it's not used, I didn't know the best way to explain what it really is. I kept getting confused on how to explain the difference between 'how' and 'whence' .

crawl-ref/source/stairs.cc
jmbjr: I'm not happy with the explanation on the return, so I'm very open to suggestions on clarifying it.

crawl-ref/source/stairs.cc
```diff
@@ -464,11 +480,18 @@ static level_id _travel_destination(const dungeon_feature_type how,
...
+    {
         // TODO: This probably causes an obscure bug where confused players
         // going 'down' into the vestibule are twice as likely to fall, because
         // they have to pass a check here, and later in floor_transition
         // Right solution is probably to use the canonicalized direction everywhere
+
+        // If player falls down the stairs trying to leave the dungeon, we set
+        // the destination depth to 1 (i.e. D:1)
+        if (how == DNGN_EXIT_DUNGEON)
+            dest.depth = 1;
```
jmbjr: I hope this is clear what I'm doing here and that this is OK. the issue is, when dest is returned as the default with branch = 0 and depth = -1, it's interpreted as invalid.  this is a problem because if we are inside this if block, it means the player slipped down the stairs and we want to prevent them from going_up.  This is OK for all up-stairs, EXCEPT leaving the dungeon. This is because the check to see if level transition should be blocked is special cased to only trigger when dest is invalid (i.e. the default) and the player is NOT trying to leave the dungeon.

Because of this, when slipping down the stairs when leaving the dungeon, I set the dest depth to 1 (which is actually unambiguously the actual destination in this situation). Then, this returned value is now valid and we can test for this condition to see if we need to block the transition.

crawl-ref/source/stairs.cc
```diff
@@ -529,6 +552,26 @@ static level_id _travel_destination(const dungeon_feature_type how,
...
+ * Check to see if the level transition should be stopped
+ *
+ * @param dest      The destination level (branch and depth).
+ * @param feat      The dungeon feature the player is standing on.
+ * @param going_up  True if the player is trying to go up stairs.
+ * @return          True if the level transition should be stopped.
+ */
+static bool _stop_transition(level_id dest, dungeon_feature_type feat,
+                             bool going_up)
+{
```
jmbjr: I made this a new function to try to clarify this logic since it seems a little clunky.  Is it better to leave in take_stairs()? I like it separate since it's easier to understand the overall flow in take_stairs()

crawl-ref/source/stairs.cc
MarvinPA: If it's unused then removing it sounds good - it might be worth doing a little code archaeology to figure out what (if anything) it was used for in the past, if possible.

crawl-ref/source/stairs.cc
MarvinPA: Wrapping like that looks good to me!

crawl-ref/source/stairs.cc
```diff
@@ -529,6 +552,26 @@ static level_id _travel_destination(const dungeon_feature_type how,
...
+    bool trying_to_win = feat == DNGN_EXIT_DUNGEON && going_up;
+
+    // When actually leaving the dungeon, dest is invalid, so if the player is
+    // trying to win, but the destination is valid, something caused the player
+    // to remain in the dungeon (i.e. slipped back downstairs while confused).
+    return (!dest.is_valid() && !trying_to_win) ||
+            (dest.is_valid() &&  trying_to_win);
```
rawlins: This setup, with the function + comment is _much_ better than the version I accidentally commented on a minute ago :).  I think the XNOR/iff logic now makes sense to me.

Maybe the bool should be named trying_to_exit, since you can escape without the orb.  Alternatively to the return condition here, you could also just return  `dest.is_valid() == trying_to_exit`.  One final thing to consider is if it would be clearer to factor the implicit negation in `stop` (to be explicit negation in take_stairs), and change the helper function to be something like `_level_transition_moves_player`, with the inverted semantics of what it has now.

jmbjr: great suggestions. thanks

crawl-ref/source/stairs.cc
jmbjr: looks like this variable may have been there since that function was created in a big refactor in 2015 09e009602f11412cbc. it wasn't used then either.

crawl-ref/source/stairs.cc
jmbjr: the logic is much simpler now.

crawl-ref/source/stairs.cc
```diff
@@ -422,6 +422,22 @@ static void _new_level_amuses_xom(dungeon_feature_type feat,
...
+ * @param whence      Not currently used
+ * @param forced      True if the player is forcing the traveling attempt.
+ *                    (e.g. forcibly exiting the abyss, etc)
+ * @param going_up    True if the player is going upstairs.
+ * @param known_shaft True if the player is intentionally shafting themself.
+ * @return            The destination level, if valid. Note the default value
+ *                    of dest is not valid (since depth = -1) and this is
+ *                    generally what is returned for invalid destinations.
+ *                    But note the special case when failing to climb stairs
+ *                    when attempting to leave the dungeon, depth = 1.
```
wheals: Should this say `depth = -1`? Or is it talking about the depth that you would be leaving the dungeon from? Since if the latter, that's not technically true, you can leave the dungeon from any level, though you won't find a `DNGN_EXIT_DUNGEON` feature outside of wizmode anywhere but D:1.

crawl-ref/source/stairs.cc
wheals: Seems fine to me!

rawlins:
Looks like this still needs some tweaking? I'll try to look it over more carefully at some point soon.

PR #616, opened 2017-09-18
Fix Webtiles Visible Monsters/Features/Items Menu

toastedzergling:
The Visible Monsters/Features/Items Menu has had a long standing bug where multi-line items wind up showing duplicated and tabbed inwards:

https://i.imgur.com/HUEZ2LA.png

So, I came up with a fairly brutal hack to fix it. so that my lovely Beogh Bros look much nicer:

https://i.imgur.com/WEYgWrf.png

I tried my best to make sure this didn't screw up any other menus.  I don't think it did, but would love an extra set of eyes.


rawlins:
Thanks for this fix.  I'm afraid this is too brutal for me, but it got me to track down the underlying bug.  I'm going to push something that fixes this in a slightly different way on the c++ side.  (Unfortunately the mouseover box won't work quite as nicely in my version, but it's very hard to get right in both webtiles and console simultaneously.)

PR #617, opened 2017-09-18
Fixed issue where draconian hat was not shown when equipped

efredericks:
Reordered `pack_doll_buf` enum and added special case to ensure draconians can always flaunt their fancy hats.


alexjurkiewicz:
Build failures seem like network issues rather than code problems

PR #619, opened 2017-10-02
Different messages for cloaks and scarves in _xom_pseudo_miscast

abstrakct:
First contribution, please be gentle.
This is just a simple fix for something I noticed while playing a Xom character.
The pseudo miscast code just looked for anything in the cloak slot, and would display "Your cloak billows in an unfelt wind" even if you had a scarf equipped.
It seemed appropriate to change it so that you get a different message if you have a scarf. I initially chose "Your scarf is twisted by an unfelt wind", but changed it to "Your scarf briefly wraps itself around your head!" which I think is funnier.
I have compiled and tested that it works as intended.

### DIFF
--- a/crawl-ref/source/xom.cc
+++ b/crawl-ref/source/xom.cc
@@ -2010,14 +2010,10 @@ static void _xom_pseudo_miscast(int /*sever*/)
     {
         item_def* item = you.slot_item(EQ_CLOAK);
 
-        if(item->sub_type == ARM_CLOAK)
-        {
+        if (item->sub_type == ARM_CLOAK)
             messages.emplace_back("Your cloak billows in an unfelt wind.");
-        }
-        else if(item->sub_type == ARM_SCARF)
-        {
+        else if (item->sub_type == ARM_SCARF)
             messages.emplace_back("Your scarf briefly wraps itself around your head!");
-        }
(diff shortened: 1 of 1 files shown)


crawl-ref/source/xom.cc
```diff
@@ -2007,7 +2007,18 @@ static void _xom_pseudo_miscast(int /*sever*/)
...
+    {
+        item_def* item = you.slot_item(EQ_CLOAK);
+
+        if(item->sub_type == ARM_CLOAK)
+        {
+            messages.emplace_back("Your cloak billows in an unfelt wind.");
+        }
+        else if(item->sub_type == ARM_SCARF)
+        {
+            messages.emplace_back("Your scarf briefly wraps itself around your head!");
```
rawlins: Thanks for the patch, good catch on this messaging issue.

Ridiculously picky comments: the house bracketing style would have no curly braces in these because the bodies of the if clauses are one line (you can use the util/unbrace command to enforce this) and we normally have a space after "if".

But this change looks good to me and if you don't have a chance to do these tweaks I (or someone else if they want) can easily add them and merge this.

abstrakct: Fixed!

PR #620, opened 2017-10-12
Tiles, WIP: improve layout spacing on main dungeon view

aidanholm:
General tiles UI streamlining:
 - The gridded tabs in the sidebar are now flush with the right side of the window.
 - The sidebar now has fixed width, and is always flush against the right side of the window.
 - The main dungeon view expands to fill all remaining space, showing parts of partially visible tiles.
 - The messages box sticks to the bottom of the screen.

There's some other fixes/enhancements as well. Some of them aren't really tiles related and should perhaps be in a separate PR...

### DIFF
--- a/crawl-ref/source/command.cc
+++ b/crawl-ref/source/command.cc
@@ -662,9 +662,6 @@ int show_keyhelp_menu(const vector<formatted_string> &lines,
         for (i = 0; i < blines.size(); ++i)
             cmd_help.add_item_formatted_string(blines[i]);
 
-        while (static_cast<int>(++i) < get_number_of_lines())
-            cmd_help.add_item_string("");
-
         // unscrollable
         cmd_help.add_entry(new MenuEntry(string(), MEL_TITLE));
     }
--- a/crawl-ref/source/command.cc
+++ b/crawl-ref/source/command.cc
@@ -677,20 +677,13 @@ int show_keyhelp_menu(const vector<formatted_string> &lines,
             string fname = canonicalise_file_separator(help_files[i].name);
             FILE* fp = fopen_u(datafile_path(fname, false).c_str(), "r");
 
-            if (!fp)
-                continue;
+            if (fp)
+            {
+                _add_file_to_scroller(fp, cmd_help, help_files[i].hotkey,
+                                    help_files[i].auto_hotkey);
+                fclose(fp);
+            }
 
-            // Put in a separator...
-            cmd_help.add_item_string("");
-            cmd_help.add_item_string(string(get_number_of_cols()-1,'='));
-            cmd_help.add_item_string("");
-
-            // ...and the file itself.
-            _add_file_to_scroller(fp, cmd_help, help_files[i].hotkey,
-                                  help_files[i].auto_hotkey);
-
-            // Done with this file.
-            fclose(fp);
         }
     }
 
--- a/crawl-ref/source/fontwrapper-ft.cc
+++ b/crawl-ref/source/fontwrapper-ft.cc
@@ -32,6 +32,26 @@
 # define dprintf(...) (void)0
 #endif
 
+class FontLibrary {
+public:
+    static FT_Library &get() {
+        static FontLibrary instance;
+        return instance.library;
+    }
+private:
+    FT_Library library;
+    FontLibrary ()
+    {
+        if (FT_Init_FreeType(&library))
+            die_noline("Failed to initialise freetype library.\n");
+    };
+    ~FontLibrary ()
+    {
+        if (FT_Done_FreeType(library))
+            die_noline("Failed to unload freetype library.\n");
+    };
+};
+
 FontWrapper* FontWrapper::create()
 {
     return new FTFontWrapper();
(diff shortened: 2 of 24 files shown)


crawl-ref/source/command.cc
```diff
@@ -796,14 +796,13 @@ static void _add_insert_commands(column_composer &cols, const int column,
 }
 
 static void _add_insert_commands(column_composer &cols, const int column,
-                                 const string desc, const int first, ...)
+                                 const string desc, ...)
```
rawlins: This was a very good find -- it's bad enough that |amethyst went ahead and implemented a much more drastic solution in 022a3cd6c2ec4970f0af811bda2f5684c8d63321 that eliminates the use of varargs altogether.  (As far as we can tell, nargs is a preemptive defense against the fact that it's very easy to use these versions of _add_insert_commands wrong and not terminate the varargs lists, in fact in a bunch of cases it was doing so.). There's lots more detailed discussion in the devlog from 10/16.

crawl-ref/source/fontwrapper-ft.cc
```diff
@@ -86,12 +104,11 @@ bool FTFontWrapper::load_font(const char *font_name, unsigned int font_size,
     if (!f)
         die_noline("Could not read font '%s'\n", font_name);
     unsigned long size = file_size(f);
-    FT_Byte *ttf = (FT_Byte*)malloc(size);
+    ttf = (FT_Byte*)malloc(size);
```
rawlins: I realize you're just editing code that was already there, but while you're in there I wonder if this malloc/free pair could be converted to new/delete?  There may be some arcane reason why not that I'm missing since I don't know much about freetype (though FT_Byte is probably just unsigned char), but I think we try to avoid malloc.  Especially given that ttf is now a member variable.

crawl-ref/source/fontwrapper-ft.cc
```diff
@@ -56,23 +76,21 @@ FTFontWrapper::~FTFontWrapper()
     delete[] m_glyphs;
     delete[] pixels;
     delete m_buf;
+    if (face) FT_Done_Face(face);
```
rawlins: tedious comment: should have a linebreak here

crawl-ref/source/fontwrapper-ft.cc
```diff
@@ -32,6 +32,24 @@
...
 
+class FontLibrary {
+public:
+    static FT_Library &get() {
+        static FontLibrary instance;
+        return instance.library;
+    }
+private:
+    FT_Library library;
+    FontLibrary () {
```
rawlins: tedious comment: `{` should be on a newline given our bracketing conventions (I won't comment all the cases of this but there are a bunch)

crawl-ref/source/command.cc
aidanholm: My preferred protection against not terminating varargs in this case would be GCC's `__attribute__ ((sentinel))`, but in C++ NULL != 0, so it caused a bunch of warnings when I tried it.

rawlins: Yeah, I think from the c++ perspective any use of varargs is considered not great, so |amethyst just converted it to a vector.

crawl-ref/source/fontwrapper-ft.cc
```diff
@@ -56,23 +76,21 @@ FTFontWrapper::~FTFontWrapper()
     delete[] m_glyphs;
     delete[] pixels;
     delete m_buf;
+    if (face) FT_Done_Face(face);
+    free(ttf);
```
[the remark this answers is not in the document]
rawlins: looking at it a bit more I guess this question applies to the already existing m_glyphs and pixels, too

crawl-ref/source/command.cc
aidanholm: Ah, found the IRC logs, which for some reason my IRC bouncer isn't working properly... I don't remember this causing a crash, I think it just popped up in valgrind while I was trying to track down one of the other resize issues. But in any case, glad it's fixed :)

crawl-ref/source/fontwrapper-ft.cc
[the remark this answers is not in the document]
aidanholm: `ttf` is set to null in the constructor, and free is well-defined when called with a null pointer, so it's fine

aidanholm:
I've fixed the braces/linebreaks, removed the varargs commit for the issue that's already been fixed in master, and rebased.

I've been playtesting and I've noticed an issue when using a scroll of identify: occasionally the "Identify which item?" screen is blank except for the header text, but I can still select an item with a keypress. Going up/down stairs fixes this and restores correct behavior. I'm not sure if this was caused by my changes, or by other changes in master since 0.21, but I've never noticed it before.

rawlins:
I'm getting glitching on OS X, I think because of the clipping stuff; only the lower-left quarter of the dungeon area is shown. This is a retina macbook in case it could be high-DPI-related, here's what it looks like: 

PR: https://www.dropbox.com/s/2dln8dqqoyh8gga/Screenshot%202017-10-17%2016.41.04.png?dl=0
master:

If you don't have access to os x, it might be easier for me to try to debug it myself, but let me know if you have any ideas.

rawlins:
Can confirm it's retina-related, things look fine on my mac mini with a normal monitor.

rawlins:
Merged, thanks very much for these fixes (and please do more if you want)!  I fixed the retina thing in dc75ea7c3988, it was just a matter of providing glScissor with device pixels rather than logical pixels.

PR #621, opened 2017-10-18
Fix tiles UI not being redrawn on expose

aidanholm:
Small fix this time :)

I use the awesomewm window manager, and when switching to a workspace
(tag) with a running instance of crawl-tiles, the tiles window takes
several seconds to redraw. Before it redraws, the window contains
whatever was in that section of the screen on the previous workspace.
This bug does not appear when minimizing and then restoring tiles.


rawlins:
Thanks, merged with a924934980fa! (Github's auto-detection failed because I rebased it I think).

PR #622, opened 2017-10-18
Fix message displayed when Gh tries to quaff potion of mutation

nlavsky:
After f20264c8 ghouls get the "You cannot mutate at present." message when they try to quaff an identified potion of mutation. The "at present" part is not correct, since it's a permanent restriction.

This commit fixes that and a typo in the function's description.


nlavsky:
Oops, didn't know about the "iff" thing.

PR #623, opened 2017-10-19
Fix wizmode mutation resistance bug

jrmartel:
Currently, the player can't give their character the mutation resistance mutation in wizard mode. This fixes that, by adding a way to invalidate unwind_var's if they are no longer needed.

Also, first PR, nice to meet you.

### DIFF
--- a/crawl-ref/source/unwind.h
+++ b/crawl-ref/source/unwind.h
@@ -32,7 +32,9 @@ public:
 
     ~unwind_var()
     {
-        val = oldval;
+        if (active) {
+            val = oldval;
+        }
     }
 
     /** Get the current (temporary) value of the wrapped lvalue. */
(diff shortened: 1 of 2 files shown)


jrmartel:
I originally had a more roundabout way of fixing this, but on further thought, mutate with force_mutation=true takes care of everything that setting mutation resistance was intended to do. 

The use of var_unwind was probably a mistake when it was added.

jrmartel:
Totally mystified on the build failure. My expectation would be that even if there is an error and some test scripts call this function (I don't see that any do), it would give the character the wrong mutations, not *segfault*.

Will close if unable to figure out.

rawlins:
One of the things that travis does is a bunch of random stress testing; this does include testing the mutation system, but it's not obvious that the test that crashed was related to your changes (`crawl-ref/source/test/los_csc.lua`), I'll try to take a more detailed look at some point. You can run these tests at the command line with `make test` if you're trying to debug something, or run a specific one (with a debug build) from the source directory by doing `./crawl -test los_csc.lua`.

rawlins:
Having looked at the crash a bit more, I don't see how the build failure could be related to this change, none of the tests run up to that point even call this code. It's probably either just travis flaking out (which does happen) or some rare memory corruption bug that you happened to hit (the memory situation described by the stack trace is a complete mess).

Thanks for the fix!

PR #624, opened 2017-10-19
Don't (auto)travel via stairs into exclusions

aidanholm:
This prevents auto-travel via staircases that lead to excluded areas,
and asks when traversing staircases into an excluded area. This works
only when the staircase's destination is known, and single-tile
exclusions on the destination staircase (indicating excluded travel in
the reverse direction) are ignored.

### DIFF
--- a/crawl-ref/source/travel.cc
+++ b/crawl-ref/source/travel.cc
@@ -2692,6 +2692,8 @@ static int _find_transtravel_stair(const level_id &cur,
 
     for (stair_info &si : stairs)
     {
+        if (stairs_destination_is_excluded(si))
+            continue;
 
         // Skip placeholders and excluded stairs.
         if (!si.can_travel() || is_excluded(si.position, li.get_excludes()))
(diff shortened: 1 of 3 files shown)


aidanholm:
This fixes https://crawl.develz.org/mantis/view.php?id=11137

crawl-ref/source/travel.cc
```diff
@@ -4777,3 +4779,22 @@ int travel_trail_index(const coord_def& gc)
...
+
+bool stairs_destination_is_excluded(const stair_info &si)
+{
+    level_pos dest = si.destination;
+    LevelInfo &dest_li = travel_cache.get_level_info(dest.id);
+
+    if (is_unknown_stair(si.position) || !is_excluded(dest.pos, dest_li.get_excludes()))
+        return false;
+
+    const exclude_set &excludes = dest_li.get_excludes();
```
amalloy: What is the purpose of this second check? You seem to want to consider only some exclusions as "counting" for purposes of whether the stair destination is excluded. I imagine you have a good reason, but I don't know what it is. I suggest adding a comment explaining it.

amalloy: I meant to hit "request changes" here since I'd be happy to merge it once that comment is addressed.

aidanholm: The purpose is to ignore single tile exclusions directly on the stair destination, which indicate that travel in one direction is excluded, but not the other (the current direction of travel). There might be such an exclusion in addition to other multi-tile exclusions on the destination floor, however, hence the loop. I'll add a comment.

neilmoore:
This caused a crash when taking stairs to a brand-new level, and possibly also when taking unvisited stairs to a known level (didn't test the latter).  I pushed a fixed version as 87dc6a9 (along with some reformatting for line length, and a range-based for loop instead of an explicit iterator loop, also for line length).

PR #628, opened 2017-10-20
Add auto_butcher_max_chunks option

thomcc:
If auto_butcher is set, and this option is greater than zero, it prevents auto-butchering from happening when you have more than this many chunks in your inventory. Defaults to zero, which is the current behavior.

I don't know if this code needs to handle vampires (does auto_butcher work for them?), but if it does, it probably does not handle them properly.

### DIFF
--- a/crawl-ref/source/initfile.cc
+++ b/crawl-ref/source/initfile.cc
@@ -268,6 +268,7 @@ const vector<GameOption*> game_options::build_options_list()
         new IntGameOption(SIMPLE_NAME(level_map_cursor_step), 7, 1, 50),
         new IntGameOption(SIMPLE_NAME(dump_item_origin_price), -1, -1),
         new IntGameOption(SIMPLE_NAME(dump_message_count), 20),
+        new IntGameOption(SIMPLE_NAME(auto_butcher_max_chunks), 0, 0),
         new ListGameOption<text_pattern>(SIMPLE_NAME(confirm_action)),
         new ListGameOption<text_pattern>(SIMPLE_NAME(drop_filter)),
         new ListGameOption<text_pattern>(SIMPLE_NAME(note_monsters)),
--- a/crawl-ref/source/items.cc
+++ b/crawl-ref/source/items.cc
@@ -2843,12 +2843,21 @@ static bool _is_option_autopickup(const item_def &item, bool ignore_force)
     return Options.autopickups[item.base_type];
 }
 
+static int _get_chunk_count() {
+    auto it = find_if(you.inv.begin(), you.inv.end(), [&](const item_def& item) {
+        return item.base_type == OBJ_FOOD && item.sub_type == FOOD_CHUNK && !is_bad_food(item);
+    });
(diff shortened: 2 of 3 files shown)


crawl-ref/source/initfile.cc
```diff
@@ -268,6 +268,7 @@ const vector<GameOption*> game_options::build_options_list()
         new IntGameOption(SIMPLE_NAME(level_map_cursor_step), 7, 1, 50),
         new IntGameOption(SIMPLE_NAME(dump_item_origin_price), -1, -1),
         new IntGameOption(SIMPLE_NAME(dump_message_count), 20),
+        new IntGameOption(SIMPLE_NAME(auto_butcher_max_chunks), 0, 0),
```
amalloy: Every new option should have an explanatory section in options_guide.txt.

crawl-ref/source/items.cc
```diff
@@ -2843,12 +2843,21 @@ static bool _is_option_autopickup(const item_def &item, bool ignore_force)
     return Options.autopickups[item.base_type];
 }
 
+static int _get_chunk_count() {
+    auto it = find_if(you.inv.begin(), you.inv.end(), [&](const item_def& item) {
```
amalloy: Should just be `[]` here, not `[&]`, since there are no locals we need to capture. Makes no difference to the compiler, of course, but looks a bit weird to a human.

crawl-ref/source/items.cc
```diff
@@ -2843,12 +2843,21 @@ static bool _is_option_autopickup(const item_def &item, bool ignore_force)
...
+}
+
 /// Should the player automatically butcher the given item?
 static bool _should_autobutcher(const item_def &item)
 {
     return Options.auto_butcher >= you.hunger_state
            && item.base_type == OBJ_CORPSES
-           && !is_inedible(item) && !is_bad_food(item);
+           && !is_inedible(item) && !is_bad_food(item) 
+           && (Options.auto_butcher_max_chunks == 0
```
amalloy: I'd rather create a local named `max_chunks` or something holding this value, so we don't have to refer to this really long option name twice.

ebering:
With vampires changed to not butcher, this should be easy to finish and bring in; I'm guessing the original submitter isn't going to make the requested changes but I'll try to do that in 0.25

thomcc:
Wow I completely forgot all about this, but appreciate you taking it up, sorry!

ebering:
Does this mean you'll handle the changes from the review?

thomcc:
Ah, ... Well, I hadn't planned on it, I was just apologizing for cluttering up your PR backlog for so long.

ebering:
It's not your fault we've been slow with merging; I'll take over the polishing.

alexjurkiewicz:
@thomcc FYI this option was enabled by default! See #1378.

thomcc:
That's great! Can't wait to try it out -- coincidentally just started playing crawl again a week or so ago (good quarantine game) and was even considering picking this up since it didn't seem to have any progress.

Glad to see that's not necessary, what timing!

PR #629, opened 2017-10-20
Make melee attack noise independent of damage done

alexjurkiewicz:


rawlins:
I'm not into this change -- not convinced it's a real problem, I'm not sure that just making the noise random is a good idea, and this particular random range in particular seems like it would be a (possibly big?) early game nerf since melee noise would be often audible beyond los right from D:1. If some other dev does get convinced it's a real problem I won't argue with that too much, but no one else has commented on this one and I think I'd still oppose this particular solution, so I'm going to go ahead and close this PR.

Also: out of context this PR is fairly uninterpretable because of the blank commit message (it happens that I've seen the relevant tavern thread, but that's just luck) -- please write commit messages :)

PR #630, opened 2017-10-20
Make dream sheep tile face left

aidanholm:
Also optimized the tile while I'm at it, why not...


rawlins:
N.b. this is for consistency with other lair herd animals (and other tiles in general).

PR #632, opened 2017-10-22
Refactor menu scrolling and pagination

aidanholm:
This PR has a giant refactoring of menu paging/scrolling, which is still fairly hacky but seems to be working on all three targets. There's a few preexisting menu bugs on webtiles that I haven't fixed, but hopefully I haven't added any new ones. I haven't yet tested spectating; might have broken something there.

I don't think this is quite ready for merging yet, but I'd appreciate any comments so far

Edit: moved misc. fixes to separate PRs

### DIFF
--- a/crawl-ref/source/menu.cc
+++ b/crawl-ref/source/menu.cc
@@ -1760,7 +1760,7 @@ int MenuHighlighter::entry_colour(const MenuEntry *entry) const
 // column_composer
 
 column_composer::column_composer(int cols, ...)
-    : pagesize(0), columns()
+    : columns()
 {
     ASSERT(cols > 0);
 
@@ -1781,11 +1781,6 @@ column_composer::column_composer(int cols, ...)
     va_end(args);
 }
 
-void column_composer::set_pagesize(int ps)
-{
-    pagesize = ps;
-}
-
 void column_composer::clear()
 {
     flines.clear();
@@ -1804,8 +1799,7 @@ void column_composer::add_formatted(int ncol,
     vector<formatted_string> newlines;
     // Add a blank line if necessary. Blank lines will not
     // be added at page boundaries.
-    if (add_separator && col.lines && !segs.empty()
-        && (!pagesize || col.lines % pagesize))
+    if (add_separator && col.lines && !segs.empty())
     {
         newlines.emplace_back();
     }
--- a/crawl-ref/source/menu.cc
+++ b/crawl-ref/source/menu.cc
@@ -377,7 +377,7 @@ void MenuDisplayTile::set_num_columns(int columns)
 }
 #endif
 
-Menu::Menu(int _flags, const string& tagname, bool text_only)
+Menu::Menu(int _flags, const string& tagname)
   : f_selitem(nullptr), f_keyfilter(nullptr),
     action_cycle(CYCLE_NONE), menu_action(ACT_EXAMINE), title(nullptr),
     title2(nullptr), flags(_flags), tag(tagname),
@@ -386,13 +386,7 @@ Menu::Menu(int _flags, const string& tagname, bool text_only)
     alive(false), last_selected(-1)
 {
 #ifdef USE_TILE_LOCAL
-    if (text_only)
-    {
-        ASSERT(!(flags & MF_USE_TWO_COLUMNS));
-        mdisplay = new MenuDisplayText(this);
-    }
-    else
-        mdisplay = new MenuDisplayTile(this);
+    mdisplay = new MenuDisplayTile(this);
 #else
     mdisplay = new MenuDisplayText(this);
 #endif
@@ -2002,6 +1996,10 @@ void column_composer::compose_formatted_column(
 
 formatted_scroller::formatted_scroller() : Menu(MF_NOSELECT)
 {
+#ifdef USE_TILE_LOCAL
+    delete mdisplay;
+    mdisplay = new MenuDisplayText(this);
+#endif
     set_highlighter(nullptr);
 }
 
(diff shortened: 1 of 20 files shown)


crawl-ref/source/menu.cc
```diff
@@ -249,6 +249,11 @@ void Menu::set_flags(int new_flags, bool use_options)
     if (use_options && Options.easy_exit_menu)
         flags |= MF_EASY_EXIT;
 
+#ifdef DEBUG
+    int sel_flag = flags & (MF_NOSELECT | MF_SINGLESELECT | MF_MULTISELECT);
+    ASSERT(sel_flag == MF_NOSELECT || sel_flag == MF_SINGLESELECT || sel_flag == MF_MULTISELECT);
+#endif
+
```
rawlins: Any particular reason this is debug-build specific?  We usually (though not always) have `ASSERT`s be production.

crawl-ref/source/menu.cc
```diff
@@ -97,16 +97,181 @@ int Popup::pop()
...
+    virtual void draw_more() = 0;
+    virtual void set_num_columns(int columns) = 0;
+    virtual bool scroll_page(int delta) = 0;
+    virtual bool scroll_to_item(int index) = 0;
+    virtual void get_visible_items_range(int &first, int &last) = 0;
+protected:
+    Menu *m_menu;
+};
+
+class MenuDisplayText : public MenuDisplay
```
rawlins: Why move/put these two class definitions in the cpp file?

crawl-ref/source/menu.cc
```diff
@@ -97,16 +97,181 @@ int Popup::pop()
...
+protected:
+    int m_pagesize;
+    int m_offset;
+    int m_first_entry;
+    int m_last_entry;
+
+    int recalculate_page_size();
+};
+
+class MenuDisplayTile : public MenuDisplay
```
rawlins: should this be in `#ifdef USE_TILE_LOCAL`?

crawl-ref/source/menu.cc
```diff
@@ -1415,63 +1606,29 @@ void Menu::draw_index_item(int index, const MenuEntry *me) const
...
 
 bool Menu::line_down()
 {
-    if (first_entry + pagesize < (int) items.size())
-    {
-        ++first_entry;
-        return true;
-    }
-    return false;
+    MenuDisplayText *text_display = dynamic_cast<MenuDisplayText*>(mdisplay);
```
rawlins: There's a bunch of static/dynamic casting of mdisplay to `MenuDisplayText *` that make me wonder if the encapsulation is quite right yet.

I'm also not sure I understand the motivation for removing line-by-line scrolling from MenuDisplayTiles, which would clean up this particular cast. For me at least, this is the default way I try to interact with crawl menus so I think it is a bit surprising from the user's perspective that it won't work.  I guess the problem is that hotkey assignment is tricky, but I would still rather keep some version of the current behavior.

crawl-ref/source/menu.cc
```diff
@@ -1945,7 +2102,10 @@ void formatted_scroller::draw_index_item(int index, const MenuEntry *me) const
     if (me->data == nullptr)
         Menu::draw_index_item(index, me);
     else
+    {
+        static_cast<MenuDisplayText*>(mdisplay)->cgotoxy_for_index(index);
```
rawlins: The way in which`formatted_scroller` implies that mdisplay must be a `MenuDisplayText` is a bit convoluted right now; static_cast behavior is actually undefined when you cast a pointer to a subtype that doesn't match, so the fact that this compiles isn't a verification that the cast is possible.  I'd prefer to make that more direct somehow, perhaps adding an ASSERT to the formatted_scroller constructor (and something in a docstring).   (Right now it seems like you could make one crash in tiles by calling the constructor with `MF_USE_TWO_COLUMNS` as a flag, but the crash happens in the Menu constructor in a way that I don't think will be very obvious as to the problem.)  It would be even better to do this somehow as a compile-time `static_assert`, though it's not obvious to me how.  Alternatively, change the encapsulation so that casting isn't needed here.

crawl-ref/source/menu.cc
```diff
@@ -2000,13 +2160,17 @@ string get_linebreak_string(const string& s, int maxcol)
     return r;
 }
 
-bool formatted_scroller::jump_to(int i)
+bool formatted_scroller::jump_to(int i, bool no_scroll)
 {
-    if (i == first_entry)
+    MenuDisplayText *display = static_cast<MenuDisplayText*>(mdisplay);
```
rawlins: see prev note on this cast

crawl-ref/source/menu.cc
aidanholm: I think my reasoning was that this would prevent a spurious unused variable warning if `ASSERT`s were disabled

crawl-ref/source/menu.cc
aidanholm: Mostly because I saw these as implementation details of the menu, rather than interfaces to the menu's functionality

aidanholm:
Re encapsulation/structure and casts, I agree that there's room for improvement. I'm not entirely happy with the fact that `formatted_scroller` is implemented as a menu at all, as there's a lot of extra baggage that comes with that.

My main reasoning for removing support for line-scrolling from MenuDisplayTiles was that I found it very unintuitive when there were two columns of items, and I kind of assumed it was an incidental effect; pretty silly of me in retrospect. It'd probably not be too much work to add it back in.

aidanholm:

PR #634, opened 2017-10-23
Tiles: add icons to lookup help for items, branches, and cards

aidanholm:
The only one missing icons now is statuses, but that will take a bit more work.


rawlins:
Thanks, looks great!

PR #636, opened 2017-10-26
Fix sidebar for large tile_font_stat_size

aidanholm:
Earlier changes hardcoded the sidebar pixel width. This change now bases
the sidebar width mostly on the width of the stat HUD (dependent on its
font size), but with a minimum width in tiles as well.

There's also a bunch of dead code removal, and the way the stat HUD is
fitted into the sidebar is improved slightly as well.

### DIFF
--- a/crawl-ref/source/tilesdl.cc
+++ b/crawl-ref/source/tilesdl.cc
@@ -897,6 +897,7 @@ void TilesFramework::do_layout()
     m_region_tile->dx = m_viewsc.x;
     m_region_tile->dy = m_viewsc.y;
     int message_y_divider = 0;
+    int sidebar_pw;
 
     // if the screen estate is very small, or if the option is set, choose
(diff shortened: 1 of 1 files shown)


crawl-ref/source/tilesdl.cc
```diff
@@ -897,6 +897,7 @@ void TilesFramework::do_layout()
     m_region_tile->dx = m_viewsc.x;
     m_region_tile->dy = m_viewsc.y;
     int message_y_divider = 0;
+    int sidebar_pw;
```
rawlins: I'm traveling for a few more days and so won't have time to actually look at this soon, but while restarting an errored travis job I noticed that sidebar_pw remains uninitialized if use_small_layout=true.  (From a compiler warning that might not show up for all compilers.)

aidanholm: Good catch

PR #640, opened 2017-10-28
Improve Maurice's message and fix #11241

Mindcrafter:
Now Maurice should give proper message in the form of "Maurice steals 1 of your 2 scrolls of immolation!"


Mindcrafter:
|amethyst has already approved this commit on irc when it was in the form of a patch.

Thanks to gammafunk, the patch became this pull request.

PR #641, opened 2017-10-28
Remove reference to cleared levels from "Out of the depths" paragraph.

Mindcrafter:
According to this change https://github.com/crawl/crawl/pull/606, OoD spawns were removed. It would be right to remove them from crawl manual, too.


amalloy:
As far as I know OOD monsters haven't been removed, only the timer that makes them spawn if you spend a long time on a level. They are still placed by vaults, and occasionally at level creation time.

Mindcrafter:
Oh, so this term applies to monsters in vaults, too! (I didn't know they also spawned at level creation time). I think we should still remove ", or cleared" from "also help to keep players on their toes by making shallow, or cleared, levels still not trivial"?

PR #642, opened 2017-10-29
Redesign/rework of UI, starting with high-scores screen

aidanholm:
This PR demonstrates a WIP system for replacing crawl's UI code with a hierarchical widget-based system. It only has hbox/vbox/text widgets at the moment, but they work on both console and tiles.

I've added a demo commit which uses the new APIs to show a screen before the highscores menu: the two text labels should be anchored to the top/bottom of the screen. Console crawl will still exit if you make the window too small.


alexjurkiewicz:
Could you add some screenshots? :)

aidanholm:
Sure, but there's not much to see just yet :) Ta-da:
out
The significance is that the UI looks and behaves the same in console tiles, and handles window/terminal resizes, with the same UI code; the UI layer abstracts over drawing font buffers or `cprintf()` calls.

This PR is more to gather feedback/input about the general approach taken here, so I've only implemented text and hbox/vbox widgets and a prototype event listener API. The layout push/pop and event pump APIs are designed to match the way UI code is currently written in crawl, and allow incremental replacement.

If you like the general approach, I think the next step is to add the image, grid layout, and scroll-view wigets that would be used for building most existing parts of the UI code. I'm still mulling over how higher-level UI concepts would be expressed: e.g. inventory menu as its own UI widget subclass, or just a hierarchy of scrollable > grid > image/text widgets with some event handlers.

aidanholm:

PR #644, opened 2017-11-01
Fix identification of floor items in edge cases.

Mindcrafter:
Items (wands and books) under player will be identified after using Shadow Step, transport, stairs or passage of Golubria.

Also Merfolk will now properly change its form after shadowstepping into water. moveto_location_effects() uses 'stepped = true' here because it prevents splash upon landing on water, which fits current behaviour. The only side effect here is that Tonado will run for its full duration while usual blink shortens it severely.

### DIFF
--- a/crawl-ref/source/god-abil.cc
+++ b/crawl-ref/source/god-abil.cc
@@ -3975,6 +3975,7 @@ bool dithmenos_shadow_step()
         }
     }
 
+    const coord_def old_pos = you.pos();
     // XXX: This only ever fails if something's on the landing site;
     // perhaps this should be handled more gracefully.
     if (!you.move_to_pos(tgt.landing_site))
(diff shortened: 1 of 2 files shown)


crawl-ref/source/god-abil.cc
```diff
@@ -3986,6 +3987,9 @@ bool dithmenos_shadow_step()
     const actor *victim = actor_at(sdirect.target);
     mprf("You step into %s shadow.",
          apostrophise(victim->name(DESC_THE)).c_str());
+    // Using 'stepped = true' here because it's Shadow *Step*.
+    // This helps to evade splash upon landing on water.
+    moveto_location_effects(grd(old_pos), true, old_pos);
```
rawlins: oh, I had the logic backwards when we discussed this in IRC, this looks right to me.

rawlins:
Thanks for the fix! I merged this commit in 01fcc59446125e075d23129dbf03c771da928f6d (via the command line, so github doesn't correctly detect that).  For future reference, could you put the kind of information you had in the PR comment here in the commit message itself? You can see how I did it in that commit (just copied the text + PR title from here and reformatted it). When you make a PR github will automatically pull the first commit message as a PR comment anyways.

PR #645, opened 2017-11-02
Improve messaging for passage of Golubria

Mindcrafter:
Now there are two reasonings for being unable to enter passage of Golubria:
1) "This passage doesn't lead anywhere!" for the case it's the only passage on the map;
2) "This passage seems to be blocked by something!" if there are other passages but they all are blocked by monsters standing on them.

The message "But it is blocked!" in the old code was never reached.

### DIFF
--- a/crawl-ref/source/traps.cc
+++ b/crawl-ref/source/traps.cc
@@ -457,13 +457,18 @@ vector<coord_def> find_golubria_on_level()
     return ret;
 }
 
-static bool _find_other_passage_side(coord_def& to)
+static bool _find_other_passage_side(coord_def& to, bool& is_blocked)
 {
     vector<coord_def> clear_passages;
     for (coord_def passage : find_golubria_on_level())
     {
-        if (passage != to && !actor_at(passage))
-            clear_passages.push_back(passage);
+        if (passage != to)
+        {
+            if (!actor_at(passage))
+                clear_passages.push_back(passage);
+            else
+                is_blocked = true;
+        }
     }
     const int choices = clear_passages.size();
     if (choices < 1)
@@ -504,7 +509,7 @@ void trap_def::trigger(actor& triggerer)
 
     // If set, the trap will be removed at the end of the
     // triggering process.
-    bool trap_destroyed = false, know_trap_destroyed = false;;
+    bool trap_destroyed = false, know_trap_destroyed = false;
 
     monster* m = triggerer.as_monster();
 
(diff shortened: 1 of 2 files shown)


Mindcrafter:
The message "But it is blocked!" in the old code was never reached.

crawl-ref/source/traps.cc
```diff
@@ -541,25 +546,22 @@ void trap_def::trigger(actor& triggerer)
...
-                know_trap_destroyed = you_trigger;
-            }
-            else
-                mpr("But it is blocked!");
+            triggerer.move_to_pos(to); // Should always be true.
+            place_cloud(CLOUD_TLOC_ENERGY, p, 1 + random2(3), &triggerer);
+            trap_destroyed = true;
+            know_trap_destroyed = you_trigger;
         }
+        else if (you_trigger)
```
amalloy: We require `{}`s wrapping any multi-line blocks like this `mpr`.

crawl-ref/source/traps.cc
```diff
@@ -541,25 +546,22 @@ void trap_def::trigger(actor& triggerer)
...
-                if (you_trigger)
-                    place_cloud(CLOUD_TLOC_ENERGY, p, 1 + random2(3), &you);
-                else
-                    place_cloud(CLOUD_TLOC_ENERGY, p, 1 + random2(3), m);
-                trap_destroyed = true;
-                know_trap_destroyed = you_trigger;
-            }
-            else
-                mpr("But it is blocked!");
+            triggerer.move_to_pos(to); // Should always be true.
```
amalloy: If it should always be true, it's a good candidate for an `ASSERT`.

crawl-ref/source/traps.cc
```diff
@@ -457,13 +457,18 @@ vector<coord_def> find_golubria_on_level()
     return ret;
 }
 
-static bool _find_other_passage_side(coord_def& to)
+static bool _find_other_passage_side(coord_def& to, bool& is_blocked)
```
amalloy: I don't love adding an out-parameter here. The two booleans are particularly confusing because you set `is_blocked` if *any* of the passages were blocked, which is something the caller doesn't care about at all: they only want to know if all passages were blocked, and therefore the trap couldn't trigger.

Returning a new enum type instead like PASSAGE_FOUND`/`PASSAGE_BLOCKED`/`PASSAGE_NONE` or something (feel free to choose better names) would be clearer than having two booleans.

crawl-ref/source/traps.cc
```diff
@@ -457,19 +457,30 @@ vector<coord_def> find_golubria_on_level()
     return ret;
 }
 
-static bool _find_other_passage_side(coord_def& to)
+enum _passage_type
```
amalloy: Neither the enum nor the enum values should have leading `_`s. We just use those for names with static linkage.

crawl-ref/source/traps.cc
```diff
@@ -541,24 +552,26 @@ void trap_def::trigger(actor& triggerer)
...
-                    place_cloud(CLOUD_TLOC_ENERGY, p, 1 + random2(3), &you);
-                else
-                    place_cloud(CLOUD_TLOC_ENERGY, p, 1 + random2(3), m);
-                trap_destroyed = true;
-                know_trap_destroyed = you_trigger;
-            }
-            else
-                mpr("But it is blocked!");
+            // Should always be true.
+            bool assert_check = triggerer.move_to_pos(to);
```
amalloy: Name should describe what the value means. I'd call it `moved`, since then `ASSERT(moved)` is quite clear.

PR #647, opened 2017-11-02
Fixed no ability description on R-click (#11231)

Filip-Michalowski:
The workaround is a bit ugly - I had to use getchm() in order to let the
screen appear, and I'm not sure what show_as_menu() does.

This change is a response to [this Mantis bug report](https://crawl.develz.org/mantis/view.php?id=11231),

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -2935,15 +2935,13 @@ void describe_spell(spell_type spelled, const monster_info *mon_owner,
 void describe_ability(ability_type ability)
 {
 #ifdef USE_TILE_WEB
-    // XXX I have no idea what show_as_menu() does, can't find definition
     tiles_crt_control show_as_menu(CRT_MENU, "describe_ability");
 #endif
 
-    string desc = get_ability_desc(ability);
-    print_description(desc);
+    print_description(get_ability_desc(ability));
 
     mouse_control mc(MOUSE_MODE_MORE);
-    getchm();// FIXME description screen wouldn't show up without getchm()
+    getchm();// description screen wouldn't show up without getchm()
 }
 
 
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -16,6 +16,7 @@
 #include <sstream>
 #include <string>
 
+#include "ability.h"
 #include "adjust.h"
 #include "areas.h"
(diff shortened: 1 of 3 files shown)


crawl-ref/source/describe.cc
```diff
@@ -2926,6 +2927,26 @@ void describe_spell(spell_type spelled, const monster_info *mon_owner,
...
+/**
+ * Examine a given ability. List its description and details.
+ *
+ * @param ability   The ability in question.
+ */
+void describe_ability(ability_type ability)
+{
+#ifdef USE_TILE_WEB
+    // XXX I have no idea what show_as_menu() does, can't find definition
+    tiles_crt_control show_as_menu(CRT_MENU, "describe_ability");
```
amalloy: `show_as_menu` is not a function at all, it's the name you're giving for your new object of type `tiles_crt_control`. Since you never refer to the object at all, you could just as well call it `asdfkjuke`. Or, better, choose a name that describes what it's for.

crawl-ref/source/describe.cc
```diff
@@ -2926,6 +2927,26 @@ void describe_spell(spell_type spelled, const monster_info *mon_owner,
...
+#endif
+
+    string desc = get_ability_desc(ability);
+    print_description(desc);
```
amalloy: Inlining the `desc` variable looks clearer to me.

Filip-Michalowski: I agree. However, this is my first time doing a pull request - do I have to inline this in my fork and submit another pull request, or are you going to inline it yourself?

crawl-ref/source/describe.cc
rawlins: `show_as_menu` actually does give some clues as to what it does -- it's a RAII thing that tells webtiles to use the menu mode for drawing (as opposed to the plain CRT mode, which is almost but not quite never used any more).

crawl-ref/source/describe.cc
rawlins: You can make the change yourself in the same branch and push, which will cause the PR to update; and then (optionally) rebase so it's all one commit. (If you don't do that someone will probably eventually do it for you.)

Filip-Michalowski:
So... Is there anything else I should do? Or are you simply busy with Pull Requests with higher priority?

PR #648, opened 2017-11-02
Improve WJC messaging (sokol815)

Mindcrafter:
Fixes #11275 on Mantis and also improves messages for losing **** and ***** abilities.

### DIFF
--- a/crawl-ref/source/religion.cc
+++ b/crawl-ref/source/religion.cc
@@ -340,13 +340,15 @@ const vector<god_power> god_powers[NUM_GODS] =
 
     // Wu Jian
     { { 1, "attack monsters by moving around them",
-           "no longer perform spinning attacks" },
+           "perform spinning attacks" },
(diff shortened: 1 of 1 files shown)


crawl-ref/source/religion.cc
```diff
@@ -340,13 +340,15 @@ const vector<god_power> god_powers[NUM_GODS] =
 
     // Wu Jian
     { { 1, "attack monsters by moving around them",
-           "no longer perform spinning attacks" },
+           "perform spinning attacks" },
```
amalloy: Why not just remove these extra entries entirely, as, for example, Hepliaklqana does?

Mindcrafter: "You can no longer perform airborne attacks." sounds better than "You can no longer perform distracting airborne attacks by moving against a solid obstacle."
This holds true for all other abilities except maybe the first one.

gregbillings:
Looks good! Nice and simple change.

PR #649, opened 2017-11-09
Remove reference to Haste

Mindcrafter:
This spell no longer exists and potions of haste don't contaminate you anymore.


Mindcrafter:
'Irradiate' isn't enchantment, thus the change "enchantments" -> "spells". But with this change we no longer mention that invisibility from other sources is harmful too (though the initial contamination part is the biggest on the spell). Could it be worded better?

Edit: I forgot to add [skip ci].

PR #650, opened 2017-11-10
Add size to '%' screen

Mindcrafter:


Mindcrafter:
When I added 'Size' section to manual I erroneously thought monster description used 'little', not 'very small'. The commit also fixes this.

PR #652, opened 2017-11-13
WJC rework

PabloMansanet:
# WJC Rework 
This WJC revision is the product of discussions involving gammafunk, |amethyst and myself, including comments from other devs and raised by elliptic's concern over the state of the god.

After taking some time out and reflecting on the god's design, I've come to agree with some of the dev and experienced player criticisms that I couldn't wrap my head around before, especially about the mandatory nature of movement combat. After reading the direction in which |amethyst wanted to take the god, and finding myself agreeing with it, I decided I wanted to propose a full rework that will hopefully achieve the objectives I list below, which I think align with what the dev team is looking for. After |amethyst and gammafunk gave it their OK, I got to work on the patch itself.

# Goals:

1. Ensure martial attacks are not strictly better than tab, to guarantee a player is never forced to engage in tedious, repetitive maneuvering to obtain an advantage.
2. Give martial attacks particular niche uses that propose real gameplay choices. Reward the player for correctly identifying situations in which using a martial attack provides an advantage, but make sure these situations are infrequent enough not to be aggravating.
3. Push towards determinism; reduce reliance in random effects such as distract and instead reward tactical thinking, with mechanics that benefit from planning.
4. Reduce UI friction by making martial attacks simple and straight forward to use, with minimal restriction.
5. Reinvigorate Serpent's Lash by making it cheaper and more impactful, moving it to the center of the god's strategy.
6. Pivot the theme of the god towards "alpha strike". Rather than glass cannon per se, this revision of WJC rewards making risky plays that eliminate the threat before it can strike back.

With these goals in mind, I've proposed and implemented the following changes for each of WJC's abilities:

EDIT: Changed the unlock order to guarantee casting Serpent's Lash does not drop you below the piety level needed to use any martial attack, as it leads to upsetting scenarios.
## Serpent's Lash

- Serpent's Lash now unlocks at *** piety, from **** (all martial attacks have been moved one pip lower).
- Serpent's Lash cost has been halved (2 piety) and its exhaustion duration reduced by 33%.
- Serpent's Lash now deals slightly lower bonus damage (50% -> 40%).
- Martial attacks initiated during Serpent's Lash effect cannot miss.
- Cosmetic: Killing monsters with Serpent's Lash active blows them up (this effect was reserved to hitting distracted enemies, but the distracted mechanic is no more).

**Justification**
I believe Serpent's Lash has always been a well designed ability, but it was sadly let down by the design issues with the martial attacks themselves. At its cost, it was often reserved as an escape ability to reach the next set of stairs, and often underutilized. This is a shame, since Serpent's Lash offers an interesting twist to the god's mechanics by allowing you to plan ahead and ensuring that enemies will maintain their position for the player to execute their plan.

The cost reduction and early unlock are meant to incentivize using Serpent's Lash more frequently. Also, while the damage modifier on older Serpent's Lash was significant, the uncertainty about actually hitting an enemy with it pushed players to reserve it for escapes, since it was hard to guarantee that it would eliminate the threat. Having Serpent's Lash never miss increases determinism and pushes players to use it aggressively, furthering the "alpha strike" theme of this revision. As explained below, Whirlwind now allows the player to gain one tile of distance relatively easily, so the temptation to save Serpent's Lash is much lower, and even in the case it's used in such a way, it's probably better to do so by attacking and retreating once the effect is over.

## Lunge

- Lunge is now the first martial attack to unlock, at -.
- Lunge's damage modifier has been reduced slightly (130% -> 120%)

**Justification**
Lunge is largely untouched as it was considered the best designed of all three moves. The damage bonus has been reduced slightly to tone down its effectiveness for stabber builds, and to lower the explosiveness when combined with the new Serpent's Lash. The very early unlock allows the player to start getting used to martial attacks, and guarantees they know how to use them when Serpent's Lash becomes online.

## Whirlwind

- Whirlwind is now the second martial attack to unlock, at *.
- Whirlwind deals 80% weapon damage.
- Whirlwind adds the "pinned down" status to monsters around your attack path. This status prevents monster movement for one turn after their reaction. This only affects actual movement; the monster is still free to attack or perform any other action.

**Justification**
Reducing whirlwind damage ensures that constant whirlwind is never the optimal means of fighting, returning tab to the its place as the default (and well automated) melee action. Regardless of the number of enemies, a player that simply chooses to whirlwind against all of them is losing on the damage/incoming damage ratio. However, Whirlwind retains the important role of allowing the player to reposition without losing *all* damage. 

The pin down effect serves two main purposes; the first is to offer an implicit risk/reward tradeoff by allowing the player a safe retreat (one tile of distance at will) in exchange for fighting in open terrain, which is generally less safe. The second is to stop monsters in place for the player to execute a plan. For this second purpose, it works much better than the old distract mechanic as it is 100% deterministic. 

It is worth noting that the 80% figure has been chosen precisely to be the inverse of the 120%  for lunge, and this is important to make sure that the "triangle" movement strategy is not optimal (whirlwind to the side, step back diagonally, lunge forward). Whirlwind and Lunge are never better than tab against a single enemy. However, against multiple enemies it is possible to combine whirlwind and lunge in interesting ways to maximize damage, but it's neither mandatory nor immediately obvious, or reducible to a simple pattern.

## Wall Jump

- Wall Jump is now the third martial attack to unlock, at **.
- Wall Jump loses all target restrictions. It can be used anywhere.
- Wall Jump takes the same amount of time to arrive to the location as walking (basically, performing a Wall Jump takes two turns worth of time).
- Wall Jump performs twice as many martial strikes on landing, to compensate for the slower execution.
- Wall Jump loses Distraction. It deals 100% of weapon damage per strike and has no additional effects.
- Serpent's Lash allows for an instant wall jump, or discounts 50% of the wall jump time if the player had already moved once.

**Justification**
Wall Jump has always been the most polarizing of all of WJC's abilities. Opinions on it are generally extreme, it's either the most fun ability of the kit or a degenerate, flawed idea. Its detractors have often pointed out that allowing the player to traverse space faster at no cost has negative implications for turn count speed running, kiting, and general exploitation of movement mechanics in a way that is repetitive and aggravating. 

To fix this, MarvinPA added a positional restriction to the ability, restricting it to only be used as an attack. While this was effective at solving the problems above, it upset the Wall Jump supporters as it hindered the feeling of freedom associated to the ability. Despite being exploitable and problematic, its non-combat uses like traversing lava or deep water pits are an integral part of the ability's identity, and I was sure that if I were to undertake another WJC rework, having a freely available Wall Jump should have to be part of it.

This iteration of Wall Jump fixes the problem by simply slowing down to the pace of walking. The mechanics are identical as before, the difference being that two turns worth of time will have elapsed after using the ability, which allows monsters to catch up, and prevents the ability from being abused for map traversal.

On the surface, this may sound like it nerfs Wall Jump to the point of being useless in combat. Interestingly it just offers a different tradeoff. The slower execution means (as it often does with WJC martial attacks) that the player will perform a higher number of attacks per landing. Averaged over time, wall jump would deal equivalent damage to tabbing, but sequenced differently: two turns worth of outgoing damage in exchange for two turns of incoming damage. This is both good and bad; it may allow the player to eliminate a monster before it gets a chance to retaliate (especially squishy ranged monsters thanks to the distance gained), but it also relinquishes fine control over the fight, precluding the player from escaping quickly if they are hurt or another enemy enters LOS. With or without Serpent's Lash, the new Wall Jump aligns with the "alpha strike" focus of this proposal by offering the option of a reckless attack that eliminates a group of enemies before they can react.

## Heavenly Storm

Heavenly Storm is untouched, as it seems to be generally well received and not too problematic, aside from back-end code concerns (it might need a cleanup) and graphics (it wouldn't hurt if someone more art inclined would draw some clouds that don't look like mustard).

# Conclusion
I hope this proposal and the accompanying code made sense. They are the result of taking some time away from the game, reflecting on the god's design and collecting feedback, particularly from developers. I credit |amethyst for putting me on the right path by expressing some of the god's design issues very clearly. I'm really looking forward to testing this, and hopefully finding a sweet spot that satisfies everyone.


rawlins:
Thinks for the commits and the detailed note in this PR!  I merged this mostly as-is in  d87a9b2, just added some UI indicators for pin.  There's been a bit of discussion in the dev channel since the merge that you may want to check out.

PR #653, opened 2017-11-19
Wand charge merging and other wand-related changes

gammafunk:
Making a PR since this is a bunch of commits that contain a good deal of potential breakage. I have done some basic testing but still need to make sure saves transfer alright; there's no actual need for save compat code, so this should be OK.

In general feedback welcome with respect to the implementation and of course for any gameplay concerns. It may be possible to salvage wand charge identification, but based on past discussions, I doubt this is the case.


Mindcrafter:
Scrolls of recharging are still on 'Items not yet recognised' list.

DenialAdams:
Reddit thread discussing this:

https://www.reddit.com/r/dcss/comments/7e6ldx/wands_reworked/

Also, a bug reported there (haven't confirmed it):

> Major bug in new change, identify still brings up wands to be identified. If you started with a wand, they ALL start with 15 charges. new wands you pick up come up as identify target even though they list the charges. I suspect when you use them once they drop from the list of "unidentified".

gammafunk:
All three of these issues have been fixed in trunk, thanks.

Closing this PR, since it was intended for technical review of the code and any design issues, but I ended up merging it not long after the PR was opened so we could find problems like these. If there are any further wand-related bugs, please file a bug in [mantis](https://crawl.develz.org/mantis/bug_report_page.php).

PR #659, opened 2017-12-07
Add uncollected rune/orb tiles for } menu

aidanholm:
Preview; runes and the orb 'light up' once they have been collected:

preview

The uncollected rune/orb tiles were made with the following imagemagick
command:

  convert in.png -normalize +level 12%,18% -fx intensity out.png

Brightness normalization isn't perfect, and perhaps the parts of each
tile included in the greyed-out version should be restricted; e.g.
mnoleg's rune is a little spotty.


crawl-ref/source/tilepick.cc
```diff
@@ -2573,12 +2606,16 @@ tileidx_t tileidx_item(const item_def &item)
             return _tileidx_corpse(item);
 
     case OBJ_ORBS:
+        if (item.quantity <= 0)
```
rawlins: I guess the crawl idiom here and below would be to use a single return statement with `? ... : ...`. (I'm just going to merge this PR though, this comment is only for reference.)

aidanholm: Yup; I think I didn't do that because of the resulting line length (then the one below was for consistency).

PR #661, opened 2017-12-08
Require a single keypress to abort ?amnesia

Mindcrafter:
This commit puts ?amnesia in line with other scrolls that bring up a menu.


Mindcrafter:
There are still inconsistencies left though: any letter keypress that isn't legal exits ?amnesia's menu but not other scrolls' menus; any digit keypress does the opposite: doesn't exit ?amnesia's menu but exits other scrolls' menus; the first string of ?amnesia's menu is coloured in blue (Forget which spell?), the first string of other menus is in white (Identify which item?), the formatting is different too.

PR #663, opened 2017-12-12
Normalize number of attacks under Serpent's Lash

PabloMansanet:
As it stands right now, fast races are unfairly punished and slow races rewarded when using Serpent's Lash.

There is an aspect of SL that is unavoidable and makes it valuable for slow characters; a free move is worth more for them. However, since number of attacks scales proportionally to the character's move delay, they also enjoyed the additional advantage of hitting more often when lashing.

This fix makes it so all characters calculate their number of attacks based on a normalized 10 aut move delay when using Serpent's Lash, regardless of their movement speed. Keep in mind that the formula still factors in attack delay as usual, so it won't allow a character to perform multiple hits with a long attack delay weapon.

### DIFF
--- a/crawl-ref/source/god-passive.cc
+++ b/crawl-ref/source/god-passive.cc
@@ -1596,7 +1596,10 @@ static bool _can_attack_martial(const monster* mons)
 // made the same amount of attacks as tabbing.
 static int _wu_jian_number_of_attacks(bool wall_jump)
 {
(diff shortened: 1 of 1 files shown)


rawlins:
Thanks for the update. In the future, can you put text like this in the commit message?  Github will automatically copy the first commit message in a PR to the PR comments.

PR #665, opened 2017-12-13
General descriptions refactoring

aidanholm:
Fixes mantis issue #11287. Removed books aren't listed any more, and `print_description()` is replaced with `show_description()` which does basically the same thing but with a formatted_scroller. `show_description()` also does toggling to the description quote.

This depends on #677's line wrap fixes: there's a copy of that PR on the front of this one.


aidanholm:

PR #671, opened 2017-12-18
Expanded Item Descriptions

JoelMarler:
I've expanded some of the item descriptions (the ones with X's before their titles).

If you have anything that needs to be written, tell me.

### DIFF
--- a/crawl-ref/source/dat/descript/items.txt
+++ b/crawl-ref/source/dat/descript/items.txt
@@ -36,13 +36,13 @@ An amulet that substantially increases the wearer's normal rate of magic
 regeneration. In order to function, it must first attune itself to the wearer's
 body with full reserves of magic.
 %%%%
-amulet of nothing
+XXXXXXXXXXXXXXXXXamulet of nothing
 
-An amulet with no special properties.
+An amulet with no special properties. Often worn by humans.
 %%%%
 amulet of rage
 
-An amulet that enables its wearer to attempt to enter a state of berserk rage.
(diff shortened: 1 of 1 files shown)


aidanholm:
Hi! These look interesting. I only speak for myself, and not for the dev team, but I do have some reservations about many of the additions:
 - the descriptions of amulets as "carved from ruby", "made of .. silver", "covered in lichen" will conflict with the amulet tile variations randomly chosen for each amulet. I also don't feel it's compatible with ID'ing amulets: e.g. if an amulet is covered in lichen, why isn't it ID'ed automatically?
 - implied effect on the player: "Reading the tome leaves you baffled", "uncomfortably hot to the touch", "holding it stings a little" etc. could be misconstrued as having actual gameplay effects, rather than just being flavour text.
- incompatible with crawl "lore": for cantrips, a beginner's book doesn't really fit "backbone of magic academies"; there's never any reason to knowingly drink a potion of degeneration (except xom maybe?) so the new text feels off, and scrolls of noise are much louder than "raspberry" suggests.

JoelMarler:
That's very fair criticism: I'll change all of those additions over the
next few days.

I'd also like to create vaults by the way, but I can't create an account on
Mantis and submit them there (which I understand is the usual method): how
would you like me to do that? I did add a few by editing the 'simple' and
'large' arrival vault folders on GitHub, but perhaps that's a little
intrusive.

MarvinPA:
Going to go ahead and close this, as they don't really fit Crawl's description style - we used to have a lot of descriptions along these lines (fire spellbooks all being warm, ice ones being cold, etc) that have since been removed in favour of having descriptions be more concise. (Some day I'll finish working on a style guide and upload that, too.)

PR #676, opened 2017-12-19
Translation of UI elements is now possible.

Mnoleg:
Translation of UI elements is now possible.
The translation engine is ready only for the "Place" and "Noise" elements.
Data loading from /dat/descript/ui.txt.


amalloy:
I like the idea of this PR, but I'm worried about what will happen with translated strings that are longer than the existing English defaults. The UI is laid out quite rigidly, and I would not expect it to work well with items that are larger than expected. Even for items smaller than expected it may end up looking poor.

rawlins:
It might work (& be relatively simple) to make output.cc explicit about some of the size limits, and just do string chopping or padding as needed -- simplest version would be to add min/max lengths to the `get_ui` function in this PR.  (I'm not immediately sure if min lengths are needed at all.) Then it's up to the translator to get their text right, but they wouldn't have the option of breaking the HUD.

rawlins:
One more things -- because output.cc isn't written very well it is very helpful to anyone working on that file to be able to see the English text right there, so rather than having an English file, for now I might prefer if `get_ui` takes as an argument the English text to use if no translated version is found.

aidanholm:
@amalloy Actually, rewriting the UI to be more flexible about layout is exactly what I'm working on! There's a WIP in #642; although it has a few hacks I have yet to iron out, almost all of the functionality required to implement the sidebar stats view is there, and it'd be flexible wrt content size automatically. My current plan is to implement nicer description popups first, then begin attacking the stats view. In the meantime, chopping+padding strings to the right size would be workable.

WRT the PR itself, it's going to break if there are formatting specifiers in the translated strings, since it's passed directly to `cprintf()` (which should probably have a printf attribute). I also feel that the colon shouldn't be part of the translated strings (unless that varies by language?).

Mnoleg:
Thank you all for the feedback!
I will close this PR because I ruined my crawl repo by changing autor info. 
I will create a new fork and a new PR if I can improve my idea.
Perhaps the new improved UI is really the right solution in this situation.

PR #678, opened 2017-12-20
Borgnjor's Vile Clutch and removal of Control Undead and Cigotuvi's Embrace

gammafunk:
A quick PR to discuss any issues with this new spell and removals of Control Undead and Cigotuvi's Embrace, before I merge these changes.


crawl-ref/source/actor.cc
```diff
@@ -516,6 +542,47 @@ void actor::stop_constricting_all(bool intentional, bool quiet)
...
+}
+
+/**
+ * Stop directly constricting all defenders.
+ *
+ * @param intentional True if this was intentional, which affects the language
+ *                    in any message.
+ * @param quiet       If True, don't display a message.
+ */
+void actor::stop_directly_constricting_all(bool intentional, bool quiet)
```
amalloy: Good function. Your fix for my complaint about the obtuseness of `stop_constricting_all(true, false)` is better than what I had in mind.

crawl-ref/source/beam.cc
```diff
@@ -5824,8 +5838,11 @@ bool bolt::explode(bool show_more, bool hole_in_the_middle)
         loudness = explosion_noise(r);
 
         // Not an "explosion", but still a bit noisy at the target location.
-        if (origin_spell == SPELL_INFESTATION)
+        if (origin_spell == SPELL_INFESTATION
+            || origin_spell == SPELL_BORGNJORS_VILE_CLUTCH)
+        {
```
amalloy: `origin_spell`, not `SPELL_INFESTATION`

crawl-ref/source/dat/descript/spells.txt
```diff
@@ -201,6 +201,11 @@ related to) power.
 It is powerful enough to cancel the effects of Death's Door, although doing so
 will briefly paralyse the caster. It has no effect on the undead.
 %%%%
+Borgnjor's Vile Clutch spell
+
+Calls forth hands from corpses interred beneath the dungeon floor. Enemies
```
amalloy: It occurs to me now that this description may mislead players into thinking this only works on areas that have corpses nearby. I don't immediately have a good suggested fix, although maybe the word "ancient" could be used somehow to reinforce this is not about recent corpses?

crawl-ref/source/mon-ench.cc
```diff
@@ -326,6 +326,12 @@ void monster::add_enchantment_effect(const mon_enchant &ench, bool quiet)
              name(DESC_THE).c_str());
         break;
 
+    case ENCH_BORGNJORS_VILE_CLUTCH:
```
amalloy: This is currently being printed separately for each monster in the area of effect, right? How hard would it be to combine these messages? "Zombie hands grab the quokka, the juggernaut, and Cerebov" seems better than spending three whole lines on it, but if it's not easy to do I don't think it's a huge deal.

gammafunk: It would be nice, but probably some significant work to accumulate those messages. I'm not sure if we have any good facility for combining messages over a beam explosions, unless the targets all have the same name, in which case the message system already does the message combination since they are identical.

crawl-ref/source/dat/descript/spells.txt
gammafunk: I suppose it's possible for someone to get confused about that, but it does say "interred beneath the dungeon floor". I'll run the addition of "ancient" by mikee, the vile clutch description master.

crawl-ref/source/beam.cc
gammafunk: Doh!

PR #681, opened 2017-12-21
Remake god description UI with formatted_scroller

aidanholm:
This commit moves the describe-god UI to a formatted_scroller, and
reimplements the religion join yesno prompts as a small state machine.
This helps get rid of cgotoxy/the CRT abstraction: the god description
functions now build formatted_strings.

The god joining prompts are stuffed into the more string of the
formatted_scroller, which is... not ideal, but should work until I get a
replacement for formatted_scroller up and running.

Fixes:
 - on webtiles, ^ will now show the religion UI in a popup box,
   instead of clearing the full screen: this matches ?/G behaviour.
 - religion join prompts can no longer fill up the screen.
 - the page indicator is now hidden when asking the player to join.

Changes:
 - Message about yes/no only no longer changes based on the key that's
   pressed. To compensate, adjusted the string slightly to indicate that
   uppercase Y is required, while N and n are acceptable.
 - Pressing escape now always leaves the join-religion prompt loop and
   closes the menu.

Regressions:
 - on webtiles, switching between religion pages now closes the popup
   before reopening it to the new page.
 - religion prompts now appear at the bottom of the screen, with the
   page indicator: not a problem on console, but is sub-optimal on local
   tiles (until I make local tiles menus boxed, like webtiles menus).

### DIFF
--- a/crawl-ref/source/describe-god.cc
+++ b/crawl-ref/source/describe-god.cc
@@ -25,6 +25,7 @@
 #include "libutil.h"
 #include "macro.h"
 #include "menu.h"
+#include "message.h"
 #include "religion.h"
 #include "skills.h"
 #include "spl-util.h"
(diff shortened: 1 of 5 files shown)


aidanholm:

PR #689, opened 2017-12-31
Do not hear sounds from nowhere when changing levels

mboeh:
Fixes https://crawl.develz.org/mantis/view.php?id=11279

You can reproduce the bug consistently by entering a floor with a portal, leaving the floor, waiting for a while, and returning to that floor. While catching the floor up to the current time, the portal message is triggered while the player is not yet on the floor.

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -995,7 +995,8 @@ bool player_equip_unrand(int unrand_index)
 
 bool player_can_hear(const coord_def& p, int hear_distance)
 {
-    return !silenced(p)
+    return in_bounds(you.pos())
+           && !silenced(p)
            && !silenced(you.pos())
            && you.pos().distance_from(p) <= hear_distance;
 }


aidanholm:
Thanks for the fix; this causes the messages to not be displayed at all, however. I have a separate fix that postpones timed portal messages until after the level has finished loading, and will merge this after that's in.

mboeh:
Thanks! I noticed e27d229730f9 and came by to check if this PR was still necessary. Feel free to close it if you think it's redundant.

aidanholm:
The fix looks good to me, but it might have unintended side-effects that other devs would be aware of; I'm not terribly familiar with that part of the code. Personally I'd prefer to log a warning when the caller is passing incorrect parameters, but crawl code doesn't really do that AFAIK.

ebering:
This is causing initial portal vault messages to not be displayed at all. I don't have time to investigate further so I've reverted for now.

PR #690, opened 2018-01-01
Add random food flavours to rations.

Floodkiller:
Adds pizza, jerky, royal jelly, and fruit flavour text as potential
responses when eating ratoins.

### DIFF
--- a/crawl-ref/source/dat/database/miscname.txt
+++ b/crawl-ref/source/dat/database/miscname.txt
@@ -324,3 +324,188 @@ You hear words spoken in a strange and terrible language...
 
 You hear diabolical laughter!
 %%%%
+eating_ration
+
+That ration really hit the spot!
+
+That jerky was delicious!
+
+w:1
+That royal jelly was slimy, yet satisfying!
+
+w:1
+That jerky was jerk-a-riffic!
+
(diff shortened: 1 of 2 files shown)


aidanholm:
I like the additional flavour, but I feel that this conflicts with the description for rations, which suggests that each ration is comprised of a mix of foods, and any two rations have basically the same ingredients.

I think it'd be best if the ration description were also updated to remove this implication, but other devs may have other opinions.

rawlins:
I like this idea but I think it might be even more confusing than random ration sprites already are, because the random message is unrelated to the sprites the player has seen. I'd prefer to do something like, track the `item.rnd` values that the player has seen on rations and use that to choose the messages, but this would be somewhat more involved to implement. I might still merge this as is or build on it, but I'll have to think about it.

(Also, just so you know, this PR came after the feature freeze so probably it won't get acted on for a few weeks.)

aidanholm:
WRT random ration sprites, which I personally found pretty odd, one alternative I thought of was replacing the random food tiles with a single tile that depicts a variety of food. The rations description would probably need to be tweaked to imply that individual rations can differ / _not_ imply that rations are all the same.

PR #691, opened 2018-01-01
Mark partially identified holy weapons as useless for demonspawn/undead

mboeh:
Fixes https://crawl.develz.org/mantis/view.php?id=9545

Let me know if there's a more conventional way to break out complex conditionals like this.


aidanholm:
I think this check would be better done in `is_holy_item()` itself

PR #696, opened 2018-01-20
A simple Bitcoin meme arrival vault

MalcolmRose:
I haven't used git in awhile, hopefully I didn't do anything wrong here. Let me know.


MarvinPA:
I don't think this is a good vault idea.

MalcolmRose:
It depends on what you mean. Ascii art? Sure. But there are lots of symbols in Crawl. This is just a symbol. There's a dollar sign Gozag altar vault, ampersands and such (which admittedly have to do with roguelike history, but still). I'm unclear on how this is any different.

alexjurkiewicz:
This is a real world reference, which crawl generally tries to avoid. It
also references a meme and a specific technology, which will date over time.

MalcolmRose:
There is a new PR now based on the feedback given here. https://github.com/crawl/crawl/pull/699

PR #697, opened 2018-01-21
Goldify books

wheals:
Dusting off #485, with some cleaning up of Doesnty's work for code convention nitpicks/clean history. Credit to aidanh for rebasing the branch in the first place.


rawlins:
I didn't do a thorough code pass (beyond checking some of the code quality stuff that bugged me in the original PR) but I've loaded this up in a real save that is in depths to get a sense for what it would be like in the lategame. Here's some thoughts:

* It's probably mergeable without addressing any of the stuff below right away, it seems quite usable and the first version has been in hellcrawl for a while with no complaints beyond dealing with large spell lists.
* I'm not sure if the hide/unhide thing really works for me as a complete system for making a very large spell list manageable, but maybe loading a lot of spells in at once is not the best test case, since you'd curate your hidden list over time. It does seem useful as a piece of it though.  (I think I'd most naturally keep most spells hidden, with a narrow selection of current possibilities visible.)
* I'd prefer to have at least 3 sort options -- name, fail rate (the current), and school. It might also be fun to have a sort by book option, though I'm not sure this would add much beyond flavor. (Book information is still available with ?/, which might be enough.  Side note -- ?/ for a spell tells you what books it can come from, not sure why the regular description doesn't. Second side note, ?/ for books seems to double the spell list.)
* Probably a search by name (and maybe book) would be good too, since right now ctrl-f has this role. This is also the word from hellmonk about how this is working in hellcrawl, search is probably needed.
* I can't find any way to look at unmemorizable spells, possibly they should be shown in red or something on the hidden pane? In master, they are viewable from the book itself, though not in M. Example: portal projectile on a felid.
* Spells unmemorizable because of Ru sacrifices are not viewable at all, they don't even get the red "n spells unmemorisable" note at the bottom; if all the spells you've seen are in this category the M menu won't even open.  This is probably an old inconsistency that is more noticeable when M is the main way to view spells you've seen.
* It might be nice to integrate book flavor text somehow, but I'm not sure what's best. Right now in this PR the text is still there but hard to find, you can access it from `?/` and from `r,`. Some ideas: could show it when the player reads a book. Could show it in a hypothetical sort-by-book view.

yrmvgh:
> Some ideas: could show it when the player reads a book.

I would hope that the default case is to not dump flavor text at the user, but an rcfile option would be fun for flavor folk - some of them already enjoy the rcfile option to use fake langs.

wheals:
Just rebased it, with a few fixes in the menu with the new commit. Main thing I want to do is add some more features to it, especially searching and sorting.

wheals:
Unmemorisable spells (including ones restricted by Ru) now show up in the menu, and you can see why they can't be learned in the describe section.

I added the search, and fixed some of the outstanding issues from the original PR. 

?/ for books duplicating the list seems to have been fixed. As far as `I` or `m!` not showing the book list, I assume it's because it no longer matters once you have actually obtained the spell, and even less so now that books cease to exist once you pick them up.


I think I'd like to go ahead and merge this as is, and see whether there is demand for a sort option -- I think it might clutter the UI unnecessarily. For example, I've never seen anyone using the sort option in shops, though I'm sure some people do.

crawl-ref/source/dat/descript/backgrounds.txt
```diff
@@ -70,8 +70,8 @@ armour and a short sword.
 %%%%
 Ice Elementalist
 
-Ice Elementalists start with the book of Frost and the Freeze spell. The book's
-other spells are quite versatile.
+Ice Elementalists start the Freeze spell. The spells in their library when they
```
aidanholm: s/start/start with/

PR #699, opened 2018-01-21
A generic gold themed vault

MalcolmRose:
Evolution of a previous idea. This basic arrival vault is gold-themed.


MarvinPA:
This isn't really a workable vault concept unfortunately - it's going to be immediately recognisable to anyone spoiled that there's multiple annoying slow monsters to avoid. There's also already an arrival vault about running away from a worm, and that's not really a good vault either. For the most part, D:1 arrival vaults should be simple and purely decorative, without placing monsters or loot.

MalcolmRose:
Yeah, I think I'm done submitting vaults. Thanks for your time, Chris.

PR #700, opened 2018-01-22
Spooky vaults

rawlins:
This branch implements one of the ideas for ghost reform developed first in the 0.21 plan, ghost vaults. This proposal has three main aims: (i) keep ghosts, (ii) make ghost griefing harder, as well as make certain kinds of scumming pointless (e.g. shaftrobin), and (iii) provide more (or at least a different kind of) player agency in how to deal with ghosts. Under this proposal, player ghosts place only in vaults, and in particular in sealed vaults that the player can see into. This proposal will also allow buffing of ghosts, and so far this branch includes one such buff -- ghosts can use stairs.

Technical changes included in this branch:
1. Putting "player ghost" in vault will attempt to place a ghost from a bones file for that level.
2. Ghosts are placed only with vaults, rather than with a hard-coded post-levelgen placement system. I've attempted to keep the probabilities mostly the same.  See `dat/des/variable/ghost.des` for a few basic examples.  All of these are sealed vaults, either with runed doors or with transporters. (N.b. this isn't enforced in the code or anything.)
3. For each level, in addition to the regular bones database, the game keeps around a "permastore" of player ghosts, in order to have a semi-permanent stock of ghosts to place. In extreme circumstances, the game will fall back on mirroring the player to make a ghost. The point of this is to make it so that the probability of ghost vault placement is independent (as far as possible) from what has happened with previous games. Names in the permastore are randomized for online play at the moment. The permastore does slowly change. One idea that came up in chat is to hand-pick some ghosts to seed this with.

Side note: oldghosts could be easily implemented in the context of these technical changes by having a "dummy" vault that just places a ghost. This would still be better than the status quo because I think being able to explicitly place ghosts could be useful in vault design in any case. (Prior to this change, putting "player ghost" in a vault would parse, but only produce a mirrored ghost based on the player.)

A few issues:
* vault design is fairly basic at the moment. I haven't added any ghosts to existing vaults, one could do this with some high level runed door things for example. I also haven't included any multi-ghost vaults which could also be done. The vaults I wrote do place loot, which I suspect is too much.
* technical change 3 is not well tested and I don't know how it'll work on servers in practice.
* this change naturally would go with other changes to the ghost creation algorithm. Should the 0.21 speed nerf be reverted?
* I suspect that at the moment, a ghost vault placing will tend to crowd out a minivault, which is not ideal.

### DIFF
--- a/crawl-ref/source/dat/des/branches/vaults_rooms_ghost.des
+++ b/crawl-ref/source/dat/des/branches/vaults_rooms_ghost.des
@@ -1,0 +1,51 @@
+################################################################################
+# vaults_rooms_ghost.des: Ghost subvaults for the Vaults.
+#
+# Content: These are the subvaults that place a player ghost. All vaults here
+# should call the function ghost_vault_setup(). Until the Vaults layout can be
+# adapted to specially place these vaults, we tag them all with vaults_room so
+# they they can have a common rate of placement. Set a different WEIGHT to make
+# a specific vault more or less common
+################################################################################
+
+: require("dlua/ghost.lua")
+
+NAME:   advil_vaults_ghost_guards
+TAGS:   vaults_room
+KITEM:  12 = superb_item / star_item
+KMONS:  1 = vault guard / ironbrand convoker / ironheart preserver / \
+            vault warden
+KMONS:  2 = vault warden
+KMONS:  3 = player ghost
+SHUFFLE: 13
+: ghost_vault_setup(_G)
+MAP
+........
+........
+..nnnn..
+..n31n..
+..n12n..
+..n==n..
+........
+........
+ENDMAP
+
+# based on st_lines
+NAME:   advil_vaults_ghost_lines
+TAGS:   vaults_hard
+KITEM:  123 = superb_item / star_item
+KMONS:  1 = vault guard / ironbrand convoker / ironheart preserver / \
+            vault warden
+KMONS:  2 = vault warden
+KMONS:  3 = player ghost
+SHUFFLE: 1239
+SUBST: 9 = 98
+: ghost_vault_setup(_G)
+MAP
+........
+nnn==nnn
+..=92=..
+..=31=..
+nnn==nnn
+........
+ENDMAP
--- a/crawl-ref/source/dat/des/branches/vaults_rooms_ghost.des
+++ b/crawl-ref/source/dat/des/branches/vaults_rooms_ghost.des
@@ -165,3 +165,29 @@ MAP
 ..nDxDxDx
 ..xxxxxxx
 ENDMAP
+
+NAME:   gammafunk_vaults_ghost_fury_of_okawaru
+TAGS:   vaults_room vaults_orient_s
+KFEAT:  O = altar_okawaru
+KMONS:  O = player ghost
+MONS:   centaur warrior / tengu warrior / orc knight / naga warrior / \
+        merfolk impaler / merfolk javelineer / deep elf knight / \
+        deep elf archer
+MONS:   minotaur / stone giant / orc warlord
+MONS:   fire giant / frost giant / deep elf blademaster / \
+        deep elf master archer / titan
+SUBST:  defgh = |*, i = *%, j = %$..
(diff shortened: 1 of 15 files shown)


rawlins:
Also, I think this needs some feedback before merging, and might be a good candidate for trying in an experimental for a while.

jmbjr:
This sounds exciting!

I can set up an experimental on cbro asap if you'd like

crawl-ref/source/dat/des/variable/ghost.des
```diff
@@ -0,0 +1,77 @@
...
+#
+##############################################################################
+
+###############################################################################
+# Two basic ghost vaults
+NAME:   basic_ghost_runed
+CHANCE: 25%
+DEPTH:  D:3-, Lair, Elf, Orc, Snake, Shoals, Swamp, Spider, Depths, Vaults, Crypt, Zot, Dis, Geh, Tar, Coc
+TAGS:   player_ghost luniq_player_ghost chance_player_ghost no_tele_into no_trap_gen allow_dup
+KFEAT:  + = runed_door
```
amalloy: The glyph `=` is already a runed door. I'd rather use that than this KFEAT on `+`.

crawl-ref/source/dat/des/variable/ghost.des
```diff
@@ -0,0 +1,77 @@
...
+ .P@S.
+ENDMAP
+
+NAME:   vaults_ghost_runed
+WEIGHT: 50
+TAGS:   vaults_room player_ghost luniq_player_ghost no_trap_gen no_tele_into allow_dup
+KFEAT:  + = runed_door
+KITEM:  1 = any useful good_item / nothing
+KMONS:  ! = generate_awake player ghost
+SHUFFLE: 123!
```
amalloy: I don't see a 2 or 3 here. Does this mean that this "ghost" vault can generate with or without an item, and with or without a ghost? I'm not sure what the value in that is.

crawl-ref/source/describe.cc
```diff
@@ -3876,6 +3876,8 @@ void get_monster_db_desc(const monster_info& mi, describe_info &inf,
 
     case MONS_PLAYER_GHOST:
         inf.body << "The apparition of " << get_ghost_description(mi) << ".\n";
+        if (mi.props.exists(MIRRORED_GHOST_KEY))
+            inf.body << "It looks just like you...spooky!\n";
```
amalloy: :+1:

crawl-ref/source/files.cc
```diff
@@ -1704,14 +1703,27 @@ void save_game_state()
...
+{
+    return _get_bonefile_directory() + _make_ghost_filename(true);
+}
+
+// Bones files
+//
+// There are two kinds of bones files: temporary bones files and the
+// permastore. Temporary bones files are ephemeral: ghosts will be reused only
+// if they are on the floor where the player dies. The permastore is a more
+// permanent stock of ghosts (per levle) to use as a backup in case the
```
amalloy: s/levle/level/

crawl-ref/source/files.cc
```diff
@@ -2382,6 +2491,69 @@ static FILE* _make_bones_file(string * return_gfilename)
...
+#ifdef DGAMELAUNCH
+        permastore[rewrite_i].name = make_name();
+#endif
+        rewrite = true;
+    }
+
+    if (rewrite)
+    {
+        string permastore_file = _bones_permastore_file();
+        FILE *gfil = lk_open("wb", permastore_file);
```
amalloy: Surely we can afford at least `gfile`, although personally I'd prefer `ghost_file`.

crawl-ref/source/dat/des/branches/vaults_rooms_ghost.des
```diff
@@ -32,7 +32,7 @@ ENDMAP
 
 # based on st_lines
 NAME:   advil_vaults_ghost_lines
-TAGS:   vaults_hard
+TAGS:   vaults_room
```
alexjurkiewicz: is this change intentional? It's part of an unrelated commit

PR #701, opened 2018-01-24
Health potions give 100%

jeremygurr:


rawlins:
uh, I think you might have clicked the PR button in the wrong repo?

jeremygurr:
Sorry. Is there a way to tell github not to default pull requests to parent of the fork?

rawlins:
ah, I see what happened, this seems to be a newish feature they've added to the PR interface.  I can't find any workarounds except basically rebuilding the repository so it's not a fork on github, which is probably non-ideal.  [One answer here](https://stackoverflow.com/questions/45499374/how-can-i-change-base-fork-on-github) suggests that github support might be able to change this, though.

PR #704, opened 2018-01-25
Rebalancing Abyss

alexjurkiewicz:
I figure now is a good time in the development cycle to open a large PR like this :)

The big reason I think Abyss deserves some love is because its danger level scales weirdly through the game. At low levels it's deadly, in midgame there's a moderate challenge, and in late game you don't care at all.

There's also a few secondary issues:
* It takes too long to get out of the Abyss when you get banished
* You can scum for items

In terms of danger level, I think it's fine that *getting the rune* becomes easier as your level increases. But I do think monster banishment should remain approximately equally dangerous at all points in the game.

To that end, I've made a few changes:
* Abyss depth danger scaling is increased. With the goal of keeping Abyss:3 about the same as current, Abyss 1 is much easier, while Abyss 5 should be exceptionally dangerous. This is done by changing the number of monsters and making the pick list much more discrete
* Fewer items generate, and more generate the deeper you go
* When banished, the maximum depth is capped based on XL for <10 XL characters, and an exit is generated nearby
* The Abyssal rune is moved to depth 4+

---
This change is [Reviewable](https://reviewable.io/reviews/crawl/crawl/704)


alexjurkiewicz:
I'm closing this. It doesn't seem like anyone on the dev team is particularly interested in this change. Some of the feedback was:

* being banished is meant to be differently scary at different points in the game. Specifically, it's meant to be very scary in the early game
* the difficulty of an early game banishment is about right
* deeper abyss danger level is about right

PR #712, opened 2018-02-19
Fix traps not being activated by wind-blasted actors

aidanholm:
This works, but I'm pretty unfamiliar with this part of the code so I'd appreciate a double-check.

### DIFF
--- a/crawl-ref/source/evoke.cc
+++ b/crawl-ref/source/evoke.cc
@@ -1124,6 +1124,7 @@ void wind_blast(actor* agent, int pow, coord_def target, bool card)
     wind_beam.is_tracer       = true;
 
     map<actor *, coord_def> collisions;
+    vector<pair<actor *, coord_def>> moved_actors;
 
     bool player_affected = false;
     counted_monster_list affected_monsters;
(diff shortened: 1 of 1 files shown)


crawl-ref/source/evoke.cc
```diff
@@ -1147,6 +1147,8 @@ void wind_blast(actor* agent, int pow, coord_def target, bool card)
                     && act->is_habitable(newpos))
                 {
                     act->move_to_pos(newpos);
+                    if (auto ptrap = trap_at(newpos))
```
rawlins: I don't really know this code either, but maybe it should instead use `actor::apply_location_effects`, which will handle traps plus. I suspect from reading through the existing code that a bunch of other stuff isn't handled as well right now, for example that a Mf getting windblasted into water won't transform (though I haven't tested).

Also, for the sake of message ordering this position could be a bit early?

aidanholm: It seems like the best fix would be to collect a list of (agent, new position) pairs in the loop that is currently moving agents, and loop through them at the end of the function. That's easy enough, but I'm not sure about the wind_beam bolt; if it's just used for raycasting, it's fine where it is now, but if the wind beam could potentially kill monsters directly it'd also have to be moved to the end for proper message ordering.

PR #714, opened 2018-03-04
Add auto_hide_spells option

Doesnty:
A friend was bugging me to make the library work nicely in reverse, so here's an option to do that. Off by default, it hides spells from spellbooks you find on the ground so you can go into the list and unhide the ones you want to eventually learn, as opposed to hiding all of the many spells you don't care about.

Because you usually want most of the spells in your background, it won't hide spells from that.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -572,7 +572,6 @@ auto_hide_spells = false
         When set to true, spells added to the library will automatically be
         hidden.
 
-
 3-b     Passive Sightings (detected or remembered entities).
 ------------------------------------------------------------
 
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -24,7 +24,8 @@ The contents of this text are:
 3-  Interface.
 3-a     Dropping and Picking up.
                 autopickup, autopickup_exceptions, default_autopickup,
-                pickup_thrown, assign_item_slot, pickup_menu_limit, drop_filter
+                pickup_thrown, assign_item_slot, pickup_menu_limit,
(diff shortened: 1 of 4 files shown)


crawl-ref/docs/options_guide.txt
```diff
@@ -567,6 +568,11 @@ drop_filter += <regex>, <regex>, ...
         set/clear selection of items that match the filter
         expression(s).
 
+auto_hide_spells = false
+        When set to true, spells added to the library will automatically be
+        hidden.
+
+
```
amalloy: IMO one newline too many, since the old file just had one blank line here.

crawl-ref/source/items.cc
```diff
@@ -1841,6 +1841,8 @@ static void _get_book(const item_def& it, bool quiet)
                 spellnames.push_back(spell_title(st));
             else
                 you.hidden_spells.set(st, true);
+            if (Options.auto_hide_spells && allow_auto_hide)
+                you.hidden_spells.set(st, true);
```
amalloy: I don't love the duplication here, but I don't see a clean way to remove it. Maybe:

    if (memorise(...))
      ...;
    if (hide || !memorise(...))
      hide();

Duplicates the can_memorise check instead, which seems a bit better to me but I wouldn't mind if you disagreed.

Doesnty:
```
if (memorize)
{
  spell_names.push_back();
  if (hide)
    hide();
}
else
  hide();
```

Removes the duplication in execution but is a bit less easy to read. Thoughts?

wheals:
Merged in 5fdf0d7d63fc57cdf596bb2c4f24f273a332687c

PR #715, opened 2018-03-05
Randomise the placement of the rune in dis_grunt

NormalPerson7:
Stealing runes is one of the most effective ways to do 15 runes currently, and one of the easiest runes to steal is the iron rune of dis_grunt. This PR makes that somewhat less easy to do by randomising the placement of Dispater and the iron rune, having it occasionally place in the central structure and an equal chance of placing in any one of the four corners, instead of just the two corners nearest to the entrance, as was previously the case. It also adds an extra Hell Sentinel to the vault, just to make everything symmetrical with the same number and difficulty of enemies in all the different arrangements.

(Also apparently I messed up the formatting of the commit comment, in that it should be 72-characters per line. Sorry about that, I don't know how to easily fix it.)

### DIFF
--- a/crawl-ref/source/dat/des/branches/dis.des
+++ b/crawl-ref/source/dat/des/branches/dis.des
@@ -823,21 +823,24 @@ MAP
 ............................7.........7.............................
 ENDMAP
 
-NAME:   dis_grunt
-MONS:   patrolling Dispater, Hell Sentinel, iron dragon / iron giant w:5
-MONS:   rust devil / iron imp w:20
-MONS:   iron troll, quicksilver dragon, cacodemon
-NSUBST: G = 4:8 / *:G
-TILE:   G = dngn_statue_iron_golem
-COLOUR: G = cyan
-KMONS:  8 = iron golem
-KMONS:  9 = war gargoyle
-ITEM:   plate armour good_item / crystal plate armour good_item w:3
-ITEM:   acquire weapon
-KMONS:  F = Hell Sentinel
-KITEM:  F = |
-KMONS:  E = Hell Sentinel
-KITEM:  E = $
+NAME:    dis_grunt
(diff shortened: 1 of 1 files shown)


NormalPerson7:
Hmm yes I tried `SHUFFLE` but it wouldn't guarantee the placement of the rune / sometimes the rune would be placed multiple times. Feel free to tell me how you can guarantee that SHUFFLE will place Dispater and the rune exactly once.

NormalPerson7:
I will open this PR again using `SHUFFLE`, thanks a lot for the advice.

NormalPerson7:
This second commit changes the randomisation method to use SHUFFLE, and fixes the formatting on the commit message to be max-72-character-lines, as you prefer.

crawl-ref/source/dat/des/branches/dis.des
```diff
@@ -823,31 +823,34 @@ MAP
...
-KITEM:  F = |
-KMONS:  E = Hell Sentinel
-KITEM:  E = $
+NAME:    dis_grunt
+MONS:    patrolling Dispater, Hell Sentinel, iron dragon / iron giant w:5
+MONS:    rust devil / iron imp w:20
+MONS:    iron troll, quicksilver dragon, cacodemon
+SHUFFLE: AP / BQ / CR / DS / ET
+SUBST:   A = 1, B = 2, C = 2, D = 7, E = 7
+SUBST:   P = O, Q = ., R = ., S = ., T = .
```
amalloy: Would this be clearer as

    SUBST: P = O, QRST = .

? I could see arguments either way: it makes it clear that only P matters, but on the other hand it hides the correspondence between A/P, B/Q, and so on. 

Similarly the previous line could be shortened if you think that's better.

NormalPerson7: Yes I think that's simpler and more concise, I don't think the correspondence really matters.

PR #716, opened 2018-03-05
Simplify code for changing pluses on mace of Variability.

amalloy:
All ye who think ifs are better for control flow than functions, speak now or forever hold your peace.

### DIFF
--- a/crawl-ref/source/art-func.h
+++ b/crawl-ref/source/art-func.h
@@ -495,13 +495,8 @@ static void _VAMPIRES_TOOTH_equip(item_def *item, bool *show_msgs, bool unmeld)
 
 static void _VARIABILITY_world_reacts(item_def *item)
 {
-    if (x_chance_in_y(2, 5))
-        item->plus += random_choose(+1, -1);
-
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
"sword"? 😛

amalloy:
@alexjurkiewicz A sword is just a mace with one very pronounced spike. 

Fixed.

alexjurkiewicz:
Did you see the rework that @Hellmonk made for Hellcrawl? IMO it would be a better approach for vanilla too
https://github.com/Hellmonk/hellcrawl/commit/fd85efbddbe62bffad6dc0b8656ef4e665fadfce

amalloy:
I agree the existing mechanics for Variability seem pretty dumb to me, for the reason hellmonk gives. The fix seems like a reasonable direction, although I don't know crawl's combat mechanics well enough to say that the particular numbers chosen result in the mace having an appropriate power level and variance. I'm not really inclined to port that change unilaterally, but if you mention it in `##crawl-dev` or open a new PR for it then someone who knows combat better than I do can express an opinion.

PR #719, opened 2018-03-08
Make the text on the god granted powers ^ screen make more sense.

NormalPerson7:
This PR aims to make gods' powers summaries make sense on both the ^ screen and when they are seen in the message log from reaching/losing that piety level, and also make them consistent overall. This is done by avoiding the continuous present at all costs, and preferring "you can..." or "this god will..." constructions over anything else. There are still a few places where these constructions are not used, namely Jiyva's 5* expel jellies, both of Qazlal's 4* powers, and Ru's 2* power, because these would probably require complete rephrasing to work into a "can..." or "will..." structure.


amalloy:
Why is future tense better than present? Some of these I agree with, like "is protecting" is bad, but "protects" seems okay to me.

NormalPerson7:
To me it felt like the future was more consistent than the present, and slightly clearer in both of the two contexts the messages are seen. I agree that it is really insignificant though. If you would prefer, I can certainly tame back some of these changes to the present.

(Overall the best solution is to separate out the messages for the two different contexts, but we don't (yet) have the functionality for that - eg. it would be better to say "protects" on the ^ screen probably, to be consistent with "you can" in the present, but "is now protecting" would be the best way to do it upon gaining a continuous power, where this construction is awful on the ^ screen.)

I believe `"Ashenzari now supports your skills"`/`"Ashenzari is now supporting your skills"` is better than `"Ashenzari will support your skills"` is better than `"Ashenzari supports your skills"` is better than `"Ashenzari is supporting your skills"` solely for when you gain that power, because "supports" has no sense that this has just started to happen, whereas "will support" implies that it is going to start to happen. You can disagree, and this is probably far more detail than necessary...

ebering:
I think that if you're going to go to this trouble, you should go ahead and add functionality for the split you propose, providing the correct context. There are some grammar functions in the codebase but I'm not super familiar with them, it would be splendid if you could put them to use here.

At the very least, this patch doesn't compile as-is. You've got some missing commas or something. It has also aged to the point of gaining merge conflicts with newnem (sorry).

If you don't want to do a lot of work, a much more conservative patch just adding the few missing gift descriptions in present tense would be welcome.

NormalPerson7:
I can certainly look at this in the coming days. I didn't actually realise this didn't compile it's been so long since I looked at it, so I can easily fix that.

NormalPerson7:
Apologies for all the force-pushes/changes to the title, but I have now implemented an extra variable for the text on the `^` screen to decouple it from the text on gaining an ability, as you requested.

crawl-ref/source/religion.cc
```diff
@@ -226,12 +251,18 @@ const vector<god_power> god_powers[NUM_GODS] =
...
+           "Jiyva will no longer mutate your body and modify your attributes.",
+           "Jiyva will mutate your body and modify your attributes as you gain piety." },
+      { 3, "Jiyva is now protecting you from corrosive effects.",
+           "Jiyva will no longer protect you from corrosive effects.",
+           "Jiyva protects you from corrosive effects." },
       { 4, ABIL_JIYVA_SLIMIFY, "turn your foes to slime" },
       { 5, "You may now expel jellies when seriously injured.",
-           "You will no longer expel jellies when injured." },
-      { 5, ABIL_JIYVA_CURE_BAD_MUTATION, "call upon Jiyva to remove your harmful mutations" },
+           "You will no longer expel jellies when injured.",
```
ebering: It looks like these are backwards for gain/loss.

NormalPerson7: If you mean the "You may now expel jellies" lines, they're not:
`"You may now expel jellies when seriously injured"` is `gain`
`"You will no longer expel jellies when injured"` is `loss`
`"You may expel jellies when seriously injured"` is `general`, the text on the ^ screen.

ebering: Oh, I read the diff wrong. My bad.

PR #720, opened 2018-03-09
Protect player from last tile of bounced spell

Patashu:
Inspired by https://crawl.develz.org/tavern/viewtopic.php?f=8&t=25173

When spells like Lightning Bolt have odd ranges (which can happen suddenly when spellpower increases), it is harder to use them to hit enemies with a double bounce and yourself 0 times. So protect the player against such cases, so bouncing spells get monotonically better, rather than strangely worse sometimes.

Tracer and targeter work correctly too.


Patashu:
Ah shit I meant to pull request this onto gooncrawl and pull requested it onto mainline crawl instead. Disregard plz

PR #721, opened 2018-03-15
Datify species definitions

alexjurkiewicz:
After seeing a mini-explosion of new/revived species in many active forks and lots of proposals (with patches!) on the forum, I thought it would be nice to make building a new species faster and easier. One way to do this would be convert species definitions into data files, similar to `art-data.txt`, which is what these commits do.

I would like to data-ise `species-data.h` and `aptitudes.h` to start with (longer term this approach could also apply to backgrounds and monsters).

I wanted to open this early to get some feedback on the approach before I spent more time on it. What do you think?

---
This change is [Reviewable](https://reviewable.io/reviews/crawl/crawl/721)


alexjurkiewicz:
cc/ @Hellmonk @jeremygurr @Floodkiller

crawl-ref/source/dat/species/centaur.yaml
```diff
@@ -7,7 +7,7 @@ aptitudes:
   hp: 1
   mp_mod: 0
   mr: 3
-size: SIZE_LARGE
```
rawlins: when reading this PR I'm sort of vacillating on whether it is better to prefer the c names or some kind of more readable thing, for datifying. One advantage of using C names is that `git grep SIZE_LARGE` works uniformly. But on the other hand, hopefully all enums will become enum classes so the names might change anyways. Not sure what to do...

rawlins:
Broadly speaking I think this is a great idea, and there's a lot of things that could be productively datified if we nail down a solid system for doing it. This does make it potentially more complicated/brittle to change anything about species other than what the template allows, but when browsing the commit history for species-data.h I don't really see all that many changes like this. This needs some kind of automated validation, but I think that might just be part of the build system in a straightforward way (like it is for tiles stuff).

The main custom code I spot in species-data.h is the major-version-specific expired species, so that's something to handle. I would also prefer to keep the "Entry format" comment.

alexjurkiewicz:
The broad positive reaction is great! 👍 I'll try to do more work this
weekend.

> when reading this PR I'm sort of vacillating on whether it is better to
prefer the c names or some kind of more readable thing, for datifying

The art-data.txt format is weird and doesn't map well to the underlying
enums a lot of the time, and I've spent time coding unrands so I get where
you're coming from.

As one of my key goals for this change is to open up species creation to a
wider audience, I'd prefer to keep all enum stuff invisible. For backwards
comparisons we'll certainly need escape hatches (like a way to mark
obsolete species, weird things like demigod levelup field and so on) but
I'd like to keep them the exception rather than the rule.

rawlins:
Note for later: if this eventually gets merged, we'll need to update or remove https://github.com/crawl/crawl/blob/master/crawl-ref/docs/develop/species_creation.txt

alexjurkiewicz:
This PR changes the habitat type of Grey Draconian from `HT_AMPHIBIOUS` to `HT_WATER`. This doesn't seem to have any effect, but it would be nice if someone more experienced can confirm this.

alexjurkiewicz:
I think this is ready for review. Sorry that it's so large! Rewriting every species definition will do that :)

Some high level aspects I'm especially interested in feedback on:

* The structure / layout of species YAML files. Could it be improved? [Here's the documentation.](https://github.com/alexjurkiewicz/crawl-ref/blob/datafiles/crawl-ref/docs/develop/species_creation.md)
* The changes made to the Makefile (species-gen.py creates four files, Make doesn't seem to handle 1 command -> 4 outputs very well)
* The SP_FIRST_NONBASE_DRACONIAN removal (e2e696a) was to make my life easier but I think it was a good idea anyway. Any issues with it?

crawl-ref/source/util/species-gen.py
```diff
@@ -0,0 +1,682 @@
...
+        species['short_name'] = s.get('short_name', s['name'][:2])
+        species['adjective'] = quote(s['adjective']) if 'adjective' in s else 'nullptr'
+        species['genus'] = quote(s['genus']) if 'genus' in s else 'nullptr'
+        species['species_flags'] = species_flags(s.get('species_flags', []))
+        species['xp'] = s['aptitudes']['xp']
+        species['hp'] = s['aptitudes']['hp']
+        species['mp'] = s['aptitudes']['mp_mod']
+        species['mr'] = s['aptitudes']['mr']
+        species['aptitudes'] = aptitudes(s['aptitudes'])
+        species['monster_name'] = s.get('monster', "MONS_%s" % s['name'].replace(' ', '_').upper())
```
aidanholm: Perhaps `s.replace(' ', '_').upper()` should be extracted to an `enumify()` function, which takes a prefix (`MONS_` in this example).

crawl-ref/source/util/species-gen.py
```diff
@@ -0,0 +1,682 @@
...
+        species['habitat'] = 'HT_LAND' if not s.get('can_swim') else 'HT_WATER'
+        species['undead'] = s.get('undead_type', 'US_ALIVE')
```
aidanholm: There doesn't appear to be validation here, and the `US_` prefix has to be included; I'd suggest implementing and using `enumify()`, probably with an optional list of permissible enums for validation.

crawl-ref/source/util/species-gen.py
```diff
@@ -0,0 +1,682 @@
...
+
+class Species(dict):
+    """Parser for YAML definition files.
+
+    If any YAML content is invalid, the relevant parser function below should
+    raise ValueError.
+    """
+    @staticmethod
+    def from_yaml(s):
+        species = Species()
```
aidanholm: This bit seems to be missing some validation, especially with numerical fields.

alexjurkiewicz:
Thanks for the second set of eyes! I will take a look at everything ASAP

crawl-ref/source/util/species-gen.py
rawlins: I'm still not totally sold on the motivation for not just using the enum names here.  From an earlier discussion, the idea is to make things as easy as possible, which is nice and all. But from a practical perspective, no matter how easy, there won't be *that* many species proposals, and the people who have to maintain this and keep track of the mapping from enum names to data keywords is still us. So all this is to say that I'm still on the fence about introducing a whole new set of keywords for this format which are not the same as the enum names.

But if this is to be merged with this, the mapping needs to be absolutely as transparent as possible, and so abstracting this into a function is a good suggestion.

crawl-ref/docs/develop/species_creation.md
```diff
@@ -0,0 +1,47 @@
+# DCSS Species Definitions
+
+DCSS uses YAML files to define species, which can be found in [source/dat/species](../../source/dat/species).
```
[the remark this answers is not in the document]
alexjurkiewicz: fixed

crawl-ref/docs/develop/species_creation.md
```diff
@@ -0,0 +1,47 @@
...
+
+To create a new species, you just need to put an appropriate YAML file in there and compile DCSS.
```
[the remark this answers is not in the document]
alexjurkiewicz: fixed

crawl-ref/source/player.cc
```diff
@@ -2864,15 +2864,20 @@ void level_change(bool skip_attribute_increase)
                     do
                     {
                         you.species = static_cast<species_type>(
-                                       random_range(SP_FIRST_NONBASE_DRACONIAN,
-                                                    SP_LAST_NONBASE_DRACONIAN));
+                                       random_range(0,
```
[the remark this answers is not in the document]
alexjurkiewicz: fixed

crawl-ref/source/player.cc
```diff
@@ -2864,15 +2864,20 @@ void level_change(bool skip_attribute_increase)
...
+                           || you.species == SP_MOTTLED_DRACONIAN);
 #endif
 #if TAG_MAJOR_VERSION > 34
-                    you.species = static_cast<species_type>(
-                                       random_range(SP_FIRST_NONBASE_DRACONIAN,
-                                                    SP_LAST_NONBASE_DRACONIAN));
+                    do
+                    {
+                        you.species = static_cast<species_type>(
+                                        random_range(0,
```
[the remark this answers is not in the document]
alexjurkiewicz: good point, fixed!

crawl-ref/source/species.cc
```diff
@@ -211,20 +200,10 @@ const vector<string>& fake_mutations(species_type species, bool terse)
  * @return          An action to be printed when the player prays at an altar.
  *                  E.g., "coil in front of", "kneel at", etc.
  */
-string species_prayer_action(species_type species)
+string species_prayer_action(species_type sp)
```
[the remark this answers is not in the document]
alexjurkiewicz: good point, fixed

crawl-ref/source/util/species-gen.py
alexjurkiewicz: I added an `enumify` function as suggested

crawl-ref/source/util/species-gen.py
alexjurkiewicz: Added `enumify` function as suggested, although validation of acceptable values is kept to field-specific functions.

crawl-ref/source/util/species-gen.py
aidanholm: Yeah there's an argument to be made against massaging input into an enum; if so, the function should at least verify that the format is correct.

crawl-ref/source/util/species-gen.py
alexjurkiewicz: Added a lot of validation in ecfd8d3fde -- there may be a little more I could do

crawl-ref/source/player.cc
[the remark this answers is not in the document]
rawlins: no need for comments just saying "fixed", we (that is, every dev, unless they have mute the entire PR) get an email each and every time :)

alexjurkiewicz:
I pulled all template fragments into separate files, which simplifies `species-gen.py`. Addressed most of the issues but the two hidden reviews above for `species-gen.py` are still "active" -- since fixing them is somewhat a judgement call.

I will do a little more work on this before it's ready for re-review.

crawl-ref/source/util/species-gen.py
alexjurkiewicz: After sleeping on it, I agree. Enum values are used elsewhere in the code, people adding or modifying a species will want to be able to find the enum name of a species somewhere other than an auto-generated file. I will change the enum value to be a required field and validate it matches `SP_[A-Z]+`

alexjurkiewicz:
I'm interested in review of 27845c7, because I feel the code I've updated is ugly and there's probably a more idiomatic way to write it.

Specifically:

* `random_starting_species` function's algorithm is a copy-paste of the old code and it seems ugly / unnecessarily in-place
* the use of `species_type temp` variables in two places seems bad?
* The code for `SP_VIABLE`/`SP_RANDOM` doesn't handle the case where no suitable combination can be found, is the best way to implement this a loop over all species in-order to generate a list of candidate selections, then bailing out if the list is empty, otherwise randomly picking an element?

alexjurkiewicz:
Thanks, I've cleaned up `newgame.cc` more properly as part of fixing random/viable character building that was broken after 27845c7.

One thing I turned up is an undocumented(?) Crawl convention. Species without any recommended jobs, and jobs without any recommended species are considered "disabled" and do not show on the character select screen. I decided to codify this and made it part of the species data format (`recommended_jobs` is mandatory, for advanced use you can make it optional by setting `starting_species` to `False` (I thought it better to have an explicit override for advanced users rather than allowing blank `recommended_jobs` because that seemed like something that would trip up new contributors.))

alexjurkiewicz:
There is no further pending work I'm aware of for this branch! Hopefully there's no more rework needed, because this PR is getting a little ridiculous.

alexjurkiewicz:
Cool!

I'll do some squashing today.

Thanks for all your help 😀

What do you prefer to do-random-while? I saw filter_enum which might be
useful.

alexjurkiewicz:
I made one small update:

> Species without any recommended jobs, and jobs without any recommended species are considered "disabled" and do not show on the character select screen.

I had added two ways to control this: `difficulty: False` or `starting_species: False`. I fixed this by removing `starting_species` and updating the docs for `difficulty` and `recommended_jobs`.

aidanholm:
Sorry, forgot to reply before! My preference is for a single random() call selecting a random element of a vector, where the vector is a static constexpr of the desired items. This increases memory usage a little, but makes selection simple, and having no valid choices will abort the program instead of spinning endlessly.

I hadn't noticed `filter_enum`; thanks! It could be used to replace all of the do-while loops in the diff; if you don't mind making this change, I think it'd improve the code quality a lot. No need to worry about them being created at runtime; that's less important.

The `difficulty`/`starting_species` change sounds good; I'd prefer `difficulty` be mandatory, however, as there's no obviously appropriate default value.

alexjurkiewicz:
Will re-squash this PR once we clean up these last two things.

>  I'd prefer difficulty be mandatory, 

Done.

> I hadn't noticed filter_enum; thanks! It could be used to replace all of the do-while loops in the diff; if you don't mind making this change, I think it'd improve the code quality a lot. No need to worry about them being created at runtime; that's less important.

So there are three approaches:

1. Keep the same code as now
2. Use `filter_enum` to build a smaller list and then select an item out of that. Problem: the filtering function takes both species and job, would need to create a `filter_enum_1arg` or similar to deal with this. 
3. Build a compile-time table of recommended species / non-banned species and use that. A structure like `map<species_type, vector<job_type>>` or `array<pair<species_type, job_type>, NUM_SPECIES * NUM_JOBS>`. But we had some errors using these approaches.

I'm happy to do 2 or 3, but with 2 I'd like feedback if it's appropriate to create `filter_enum_1arg` or if there's a better way, and for 3 I'm going to need a big hint as to what the structure looks like (I can dynamically fill it in with codegen easily enough).

aidanholm:
WRT 2), `filter_enum` can be used with a lambda function just fine; no modification necessary:
```c++
auto jobs = filter_enum(job_type, [&](job_type job) {
    return job_allowed(ng.species, job) != CC_BANNED;
});
auto job = job[random2(jobs.size())]; // <- note: this doesn't handle an empty array properly
```

WRT 3), if std::array isn't working, you could try C-style arrays:
```c++
pair<species_type, job_type> foo[] = { make_pair(SP_HUMAN, JOB_FIGHTER) };
```
With both 2) and 3), you'll still need to filter them, so there's not too much difference.

alexjurkiewicz:
I've implemented `filter_enum` in the first new commit. In the second commit I de-duplicated the logic further between SP_VIABLE/SP_RANDOM (ditto for JOB_) -- I think it's better, but a `switch (ng_choice.job)` inside a `switch (ng_choice.job)` feels weird. Is there a better approach?

crawl-ref/source/newgame.cc
```diff
@@ -182,32 +182,37 @@ static void _resolve_species(newgame_def& ng, const newgame_def& ng_choice)
...
         {
-            do
-            {
-                ng.species = random_species();
-            } while (!job_recommends_species(ng.job, ng.species));
+            auto species_recommended_by_job =
+                filter_enum(species_type, [&](species_type species) {
+                    return job_recommends_species(ng.job, species);
+                });
+            // Note: we know the array has at least one element because all
```
aidanholm: Might want to `ASSERT(species_recommended_by_job.size() > 0)` here; `random2` has the unfortunate property that a range of size 0 will always return 0, instead of faulting from div-0. Even if it's guaranteed by other checks, defense in depth never hurts.

crawl-ref/source/newgame.cc
```diff
@@ -228,27 +233,34 @@ static void _resolve_job(newgame_def& ng, const newgame_def& ng_choice)
...
-            {
-                ng.job = random_job();
-            } while (!species_recommends_job(ng.species, ng.job));
+            auto job_recommended_by_species =
+                filter_enum(job_type, [&](job_type job) {
+                    return species_recommends_job(job, ng.species);
+                });
+            // Note: we know the array has at least one element because all
+            // starting species must recommend at least one opposite job.
+            ng.job = job_recommended_by_species[
```
aidanholm: ditto

crawl-ref/source/newgame.cc
```diff
@@ -182,38 +182,34 @@ static void _resolve_species(newgame_def& ng, const newgame_def& ng_choice)
...
-                random2(species_allowed_by_job.size())];
+            switch (ng_choice.species)
+            {
+            case SP_VIABLE:
+                // Note: we know the array has at least one element because all
+                // starting jobs must recommend at least one species.
+                candidate_species =
+                    filter_enum(species_type, [&](species_type species) {
+                        return job_recommends_species(ng.job, species);
+                    });
```
aidanholm: No break here; if this is intentional, add a fallthrough comment

crawl-ref/source/newgame.cc
```diff
@@ -233,38 +229,34 @@ static void _resolve_job(newgame_def& ng, const newgame_def& ng_choice)
...
-                random2(jobs_allowed_by_species.size())];
+            switch (ng_choice.job)
+            {
+            case JOB_VIABLE:
+                // Note: we know the array has at least one element because all
+                // starting species must recommend at least one job.
+                candidate_jobs =
+                    filter_enum(job_type, [&](job_type job) {
+                        return species_recommends_job(job, ng.species);
+                    });
```
aidanholm: missing break

crawl-ref/source/newgame.cc
```diff
@@ -233,38 +229,34 @@ static void _resolve_job(newgame_def& ng, const newgame_def& ng_choice)
...
+            case JOB_RANDOM:
+                candidate_jobs =
+                    filter_enum(job_type, [&](job_type job) {
+                        return character_is_allowed(ng.species, job);
+                    });
+                if (candidate_species.size() == 0)
```
aidanholm: this test should be outside the inner switch, so it tests for both VIABLE and RANDOM cases

crawl-ref/source/newgame.cc
```diff
@@ -182,38 +182,34 @@ static void _resolve_species(newgame_def& ng, const newgame_def& ng_choice)
...
+                candidate_species =
+                    filter_enum(species_type, [&](species_type species) {
+                        return job_recommends_species(ng.job, species);
+                    });
+            case SP_RANDOM:
+                candidate_species =
+                    filter_enum(species_type, [&](species_type species) {
+                        return character_is_allowed(species, ng.job);
+                    });
+                if (candidate_species.size() == 0)
```
aidanholm: this test should be outside the inner switch, so it tests for both VIABLE and RANDOM cases

crawl-ref/source/newgame.cc
```diff
@@ -182,38 +182,34 @@ static void _resolve_species(newgame_def& ng, const newgame_def& ng_choice)
...
+                    die("Selected job allows no species?!");
```
aidanholm: missing break

crawl-ref/source/newgame.cc
```diff
@@ -233,38 +229,34 @@ static void _resolve_job(newgame_def& ng, const newgame_def& ng_choice)
...
+                    filter_enum(job_type, [&](job_type job) {
+                        return species_recommends_job(job, ng.species);
+                    });
+            case JOB_RANDOM:
+                candidate_jobs =
+                    filter_enum(job_type, [&](job_type job) {
+                        return character_is_allowed(ng.species, job);
+                    });
+                if (candidate_species.size() == 0)
+                    die("Selected species allows no jobs?!");
```
aidanholm: missing break

crawl-ref/source/newgame.cc
alexjurkiewicz: The messaging is different ("no recommended jobs" vs "no allowed jobs"), so I'll keep it inside the inner switch.

alexjurkiewicz:
Fixed everything addressed. Fixed another bug I hadn't noticed -- random species/jobs picked by random included disabled ones.

Also added a default die() case to the inner loop because clang wasn't able to figure out the only two values were VIABLE/RANDOM and warned about non-exhaustively checking the variable.

aidanholm:
Instead of using `die()` for the default, it's preferable to use `__builtin_unreachable()`; it won't generate code for a case that should never be handled. That works on clang and gcc, and msvc has `__assume(false)`, so a wrapper macro is needed. I'm surprised crawl doesn't have one of these already, actually.

alexjurkiewicz:
Can we avoid creating this macro as part of this PR? I feel it is unfocused
enough already 😂

aidanholm:
Yeah, no worries; I don't have any remaining comments, except that these fixes should be squashed, but advil might have some feedback too

rawlins:
I've been working on this in a branch in the main repo, because as I started going through the details there were a bunch of changes I wanted to make that were easier for me to just do as I went. The branch is: https://github.com/crawl/crawl/tree/datafiles

This is a bit logistically non-ideal because @alexjurkiewicz can't commit there, if you want you could set this PR branch to be editable by maintainers and I will push the changes here instead.

A lot of it is various refactoring/cleanup issues, but I have found a few missing things so far: demigod statup didn't do quite what the current species-data.h does, and undead type wasn't set correctly.

rawlins:
Unfortunately since servers need to be able to build everything, I'm not sure that a python 3 requirement is realistic. We aren't even sure what will happen if a python 2 dependency shows up for all servers, though presumably anything running webtiles has it *somewhere*.  (gammafunk checked CDO and reported that it had 2.7, but not 3; I think that's the only public server that doesn't run webtiles).

Some specific testing that needs to happen before this could be merged:
 * server builds (possibly try installing the branch as an experimental on CBRO just to see if it will build, and/or on CPO)
 * windows msys2 builds
 * windows cross-compiling

crawl-ref/docs/develop/species_creation.md
```diff
@@ -29,7 +29,7 @@ You can view existing files to get a sense of the format, it's quite straightfor
 | int | `int` | **Yes** | Starting intelligence. |
 | dex | `int` | **Yes** | Starting dexterity. |
 | levelup_stat_frequency | `int` | **Yes** | How frequently (in XLs) the species' periodic stat-up occurs. You can specify `28` to ensure the species never gains stats on levelup. Note that all species gain a user-selectable stat every three levels. |
-| levelup_stats | `array of strings` | No | Which stats the species' periodic stat-up can randomly select from. Defaults to `[str, dex, int]`. |
+| levelup_stats | `array of strings` or `str` | No | Which stats the species' periodic stat-up can randomly select from. Defaults to `[str, dex, int]`. Set to `none` for cases where levelup stats happen outside of the usual system. |
```
alexjurkiewicz: We should ensure that if `levelup_stats` is `none`, `levelup_stat_frequency` is `28`. ...Right? I imagine the game will crash otherwise.

alexjurkiewicz:
I don't have a windows build environment set up -- @gammafunk are you able to try compiling? You'll need python + pyyaml

gammafunk:
I was able to cross-compile the installer using mingw to test windows tiles+console, and everything appeared to work fine when briefly trying a test character. This is something easy to set up in your linux environment yourself for testing purposes @alexjurkiewicz. On debian-based systems the development environment package you need is just `mingw64`. Once that's installed and then you can try the command, which is the exact one we use to compile the final widows installer:
```
   make CROSSHOST=i686-w64-mingw32 package-windows
```

You can install that with wine and also test through wine, or you can of course test it on windows.

I can try msys2 through a VM at some point later, but your branch doesn't have the necessary submodule update in 1871b151c, which I think means msys2 builds will fail. You might want to simply merge master into this branch.

Having pyyaml as dependency further requires setup by admins after this commit is merged. So because of this and there are some major issues that have to be addressed:

1. If this script is actually going to require python3, we need to hear from each and every admin as to whether that's going to cause a problem. CDO does not have python3, so that's at least one server that will need new software installed. It might be a lot more convenient to simply have the script be compatible with both python2 and python3. If we learn that every server but CDO either has python3 or can easily update, it might not be so bad to have a hard requirement for python3. Probably CDO could be updated eventually with python3, but someone would have to check with Napkin to be sure.

2. Admins have to at least install pyyaml, otherwise trunk will fail to build. We might need to plan a specific merge date for this or just wait to hear from each admin that they've installed pyyaml into their chroot's python. We absolutely want to avoid simply merging trunk-breaking commits without our admins knowing about this ahead of time. 

The CWZ admin does his updates manually every so often after merging local language-related patches, so that server won't spontaneously update and break trunk if we have trouble reaching him for some reason. LLD is similar, but dplusplus is easy to reach on discord.

rawlins:
I more or less vetoed the python 3 requirement already, I've tested the script in python 2.7 and it runs fine.

alexjurkiewicz:
Right. The YAML module is still an external dependency though. Server admins will need to `$package_manager install python-yaml` or `pip install pyyaml`.

alexjurkiewicz:
Tested compiling in MSYS2, and it works! You can even change level, for example if you encounter a kobold with dagger of distortion on d:1:
screen shot 2018-05-18 at 4 07 46 pm
screen shot 2018-05-18 at 4 17 13 pm
(The graphical artifacts are due to running in a VM, I'm pretty sure. They occurred with a vanilla build too.)

Added some pyyaml install docs for MSYS2.

aidanholm:
Aren't servers mostly running Ubuntu/Debian? pacman won't work (isn't installed) on those setups.

gammafunk:
We're talking about msys2 @aidanholm, which is for compiling from windows to build windows-native binaries. This is not relevant to servers.

gammafunk:
Oops, accidentally hit "Close and comment" instead of "Comment", ignore that :)

aidanholm:
Oops, I should have read the context first; thanks for clearing up my confusion

alexjurkiewicz:
rebased

rawlins:
I'll aim to merge this this weekend.

ebering:
This is in the wiki, but I figure I'd notify people via commenting on the PR. @alexjurkiewicz and I have set a Flag Day for merging this of 19 January (UTC). We will both be available in the window 2300 19 Jan. - 0100 20 Jan. UTC should things go wrong.

My planned timeline is:
- Merge at 2200. CPO and CKO will be the crash canaries with 15 minute build schedules.
- CUE rebuilds at 2300. Once it has, check that it is working.
- At this point we should be flying smoothly, or if there's a major problem, be aware of it and already backed out.
- Once CUE is ok, one-by-one rebuild and test the remaining servers we have rebuild access to.
- Hang out and hope nothing explodes.

alexjurkiewicz:
I might not be available the full two hours

PR #724, opened 2018-03-21
Faerie Dragon species add

RealzHS:


crawl-ref/source/player.cc
```diff
@@ -5863,7 +5901,7 @@ int player::racial_ac(bool temp) const
     // drac scales suppressed in all serious forms, except dragon
     if (species_is_draconian(species)
         && (!player_is_shapechanged() || form == transformation::dragon
-            || !temp))
+        || !temp))
```
NormalPerson7: Erm that formatting was correct...

crawl-ref/source/player.cc
```diff
@@ -3167,6 +3179,13 @@ int player_stealth()
             stealth += STEALTH_PIP;
     }
 
+    //Faerie Dragons' bright wings reduce stealth.
+    if (you.species == SP_FAERIE_DRAGON && (you.form == transformation::none
+        || you.form == transformation::appendage))
```
NormalPerson7: Did you forget blade hands?

crawl-ref/source/player.cc
RealzHS: Fixed.

crawl-ref/source/player.cc
RealzHS: Great catch, I did. And dragon form. Fixed for both.

crawl-ref/source/dat/clua/automagic.lua
```diff
@@ -240,7 +240,11 @@ end
 
 local function mp_is_low()
   local mp, mmp = you.mp()
-  return (100*mp <= AUTOMAGIC_STOP*mmp)
```
aidanholm: Why not use a ternary? I think it makes it clearer:
```lua
local mp, mmp = you.mp()
mp = mp + (you.race() == "Faerie Dragon" and 1 or 0)
return 100*mp <= AUTOMAGIC_STOP*mmp
```
or else just
```lua
if you.race() == "Faerie Dragon" then mp = mp + 1 end
```

crawl-ref/source/item-use.cc
```diff
@@ -701,6 +701,15 @@ bool can_wear_armour(const item_def &item, bool verbose, bool ignore_temporary)
         return false;
     }
 
+    if (you.species == SP_FAERIE_DRAGON && slot == EQ_BODY_ARMOUR)
+    {
+        if (verbose)
+        {
```
aidanholm: No need for the {} around the mprf call here

crawl-ref/source/status.cc
```diff
@@ -929,7 +929,16 @@ static void _describe_airborne(status_info* inf)
     const bool perm      = you.permanent_flight();
     const bool expiring  = (!perm && dur_expiring(DUR_FLIGHT));
     const bool emergency = you.props[EMERGENCY_FLIGHT_KEY].get_bool();
-    const string desc   = you.tengu_flight() ? " quickly and evasively" : "";
+    string desc          = "";
+
+    if (you.tengu_flight())
```
aidanholm: No need for {} here

crawl-ref/source/status.cc
```diff
@@ -929,7 +929,16 @@ static void _describe_airborne(status_info* inf)
...
+    {
+        desc = " quickly and evasively";
+    }
+    else if (you.faerie_dragon_flight())
```
aidanholm: ditto

crawl-ref/source/tiledoll.cc
```diff
@@ -286,11 +286,12 @@ void fill_doll_equipment(dolls_data &result)
         tileidx_t ch;
         switch (you.species)
         {
-        case SP_CENTAUR: ch = TILEP_TRAN_STATUE_CENTAUR;  break;
```
aidanholm: It's not hugely important, but in general, it's better not to align items like this; it hinders using tools like git-blame. Probably best if you just add the faerie dragon line and let it stick out the side.

crawl-ref/source/tiledoll.cc
```diff
@@ -286,11 +286,12 @@ void fill_doll_equipment(dolls_data &result)
...
-        case SP_NAGA:    ch = TILEP_TRAN_STATUE_NAGA;     break;
-        case SP_FELID:   ch = TILEP_TRAN_STATUE_FELID;    break;
-        case SP_OCTOPODE:ch = TILEP_TRAN_STATUE_OCTOPODE; break;
-        default:         ch = TILEP_TRAN_STATUE_HUMANOID; break;
+        case SP_CENTAUR:       ch = TILEP_TRAN_STATUE_CENTAUR;       break;
```
aidanholm: ditto

crawl-ref/source/tiledoll.cc
```diff
@@ -300,11 +301,12 @@ void fill_doll_equipment(dolls_data &result)
     case transformation::lich:
         switch (you.species)
         {
-        case SP_CENTAUR: ch = TILEP_TRAN_LICH_CENTAUR;  break;
-        case SP_NAGA:    ch = TILEP_TRAN_LICH_NAGA;     break;
-        case SP_FELID:   ch = TILEP_TRAN_LICH_FELID;    break;
-        case SP_OCTOPODE:ch = TILEP_TRAN_LICH_OCTOPODE; break;
-        default:         ch = TILEP_TRAN_LICH_HUMANOID; break;
+        case SP_CENTAUR:       ch = TILEP_TRAN_LICH_CENTAUR;       break;
```
aidanholm: ditto

crawl-ref/source/main.cc
```diff
@@ -447,6 +447,12 @@ NORETURN static void _launch_game()
 
     _prep_input();
 
+    if (you.species == SP_FAERIE_DRAGON && you.num_turns == 0
```
aidanholm: These conditionals should be moved into `float_once()`, as the faerie dragon check is currently duplicated. It should also be given a more descriptive name; consider making it static to main.cc and calling it `_faerie_dragon_begin_game_flying()`. You might want to consider using `no_messages` to silence `float_player()`, if the player is supposed to be already flying when they begin the game.

crawl-ref/source/player.cc
```diff
@@ -4869,13 +4887,30 @@ void float_player()
     }
     else if (you.tengu_flight())
         mpr("You swoop lightly up into the air.");
+
```
aidanholm: No need for the empty lines before/after this case.

MarvinPA:
Closing this, per gammafunk's comment.

PR #726, opened 2018-03-24
Tileschat mute

rawlins:
Implement some basic tileschat admin commands for the player, mainly aimed at dealing with spamming; this allows muting and unmuting, and will save mute lists on the server. This branch is ready for some review.

These changes may require a server admin to update code that's not usually updated on a new version (for the webtiles server).

### DIFF
--- a/crawl-ref/.gitignore
+++ b/crawl-ref/.gitignore
@@ -183,6 +183,7 @@ Thumbs.db
 
 # Webtiles stuff
 passwd.db3
+user_settings.db3
 *.ttyrec
 /source/rcs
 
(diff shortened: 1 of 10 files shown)


floraline:
Can you explain your thought process for line 219 on server.py? What exactly does `dgl_mode` mean, and why did you choose not to put the database function inside this condition?

rawlins:
@floraline `dgl_mode` is an option that lets you run the webtiles server without the lobby / logins / without dgl integration (I think it might have been intended for local play on webtiles?). I put this db check outside the if because there are still player names without dgl, and there might in principle be some user settings worth saving there even without logins (though the mute ones specifically don't seem very useful if this mode is intended for single player, it's true). I didn't mess with the existing check because dgl needs the password database specifically.

I have to admit I didn't really test with `dgl_mode = False` though, since it's so rarely used, and definitely not used on any public servers.

rawlins:
(Followup, chat seems to be entirely broken with `dgl_mode = False`, and spectating isn't possible anyways.)

PR #729, opened 2018-03-27
Add player LUA callback skill_training_needed

dokempf:
The new skill target mechanism and its lua bindings
are a very nice way of implementing automated skill
advancement logic. However, when the last skill target is met
while gaining experience the skill menu is opened
and needs user interaction. The patch introduces a
callback `skill_training_needed` in this case. If it
returns true, the skill menu is not opened.

### DIFF
--- a/crawl-ref/source/skills.cc
+++ b/crawl-ref/source/skills.cc
@@ -15,6 +15,7 @@
 #include <sstream>
 
 #include "ability.h"
+#include "clua.h"
 #include "describe-god.h"
 #include "evoke.h"
 #include "exercise.h"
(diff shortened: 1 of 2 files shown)


aidanholm:
So the idea is to set skills to train from the `skill_training_needed` callback? This sounds good, but I'd prefer to check whether any skills have been selected; it's not clear what happens if the callback returns true without doing anything.

dokempf:
Yes, that was the idea. But I agree, a check is safer. Will look into implementing that tomorrow.

dokempf:
The code now doublechecks skill training after the callback. Glad to hear more feedback.

aidanholm:
I'd prefer to avoid the `do_lua_callback` recursion; instead, it'd be better to recompute `trainable_skill` and/or `could_train` after the lua call; computing those should then be moved to a helper function.

rawlins:
Thanks for getting this started! I went ahead and cleaned up the supporting code necessary to do the check non-recursively, which was a bit of a mess (and found a bug in the process). I kept your first commit as b03524a33725.

PR #731, opened 2018-04-09
Make Zonguldrok's Shrine non-trivial.

NormalPerson7:
For a long time now, Zonguldrok's Shrine has been one of the easiest wizlabs. It has killed a meagre 21 players since the release of 0.20, compared to Lehudib's which has over 70 kills. This pull request aims to fix this by making the spawns over time from the gravestones on the outside much more dangerous, and create a more challenging boss battle. There were three stages to this:

1. Buff the boss battle: add two extra liches in addition to the antique lich when moving towards the loot.
2. Buff and vary the derived undead. Instead of three possible derived undead, there are now eighteen base types which can each spawn as zombies, simulacra or spectral things - which base type spawns is dependent on the tier of spawn, which increases over number of spawns already made.
3. Buff and vary the non-derived undead.  Instead of four possible derived undead, there are now twelve, which have a variety of different powers: invisibility, hexes, MR/2 attacks, slowing attacks, fearsome and vampiric melee, summoning and offensive spells, as opposed to the skeletal warriors and 1 ancient champion that were the most dangerous things that spawned previously.

Please could you provide general feedback where possible, in addition to specific feedback regarding:
(a) Code style: feel free to tell me how the code could be clearer or more efficient;
(b) Is the boss battle too dangerous now, and would there be something more fitting other than more liches?
(c) How do I set up set_random_mon_list - it seems not to actually affect anything when it comes to gameplay?
(d) Is it even necessary at all to have a cap on the number of non-derived undead that can spawn from each tier? Given that there is now a much greater variance in the monsters that spawn, I am wondering if this can be entirely removed.
(e) The sword of Zonguldrok doesn't actually ever seem to spawn; instead I have found a randart double sword every time that I have tested. Is this a feature or a bug, because it seems like the sword should always generate?


NormalPerson7:
Erm it looks like I need to reinvestigate how to update my local NormalPerson7/crawl branch, and once I am not comparing stuff to early March, re-do this PR.

Feel free to give feedback on the changes **only to wizlab.des** in the meantime.

NormalPerson7:
The new PR can be found at #732.

PR #732, opened 2018-04-10
Buff Zonguldrok's Shrine

NormalPerson7:
For a long time now, Zonguldrok's Shrine has been one of the easiest wizlabs. It has killed a meagre 21 players since the release of 0.20, compared to Lehudib's which has over 70 kills. This pull request aims to fix this by making the spawns over time from the gravestones on the outside much more dangerous, and create a more challenging boss battle. There were three stages to this:

**Buff the boss battle**: add two extra liches in addition to the antique lich when moving towards the loot.
**Buff and vary the derived undead**. Instead of three possible derived undead, there are now nine base types which can each spawn as zombies, simulacra or spectral things - which base type spawns is dependent on the tier of spawn, which increases over number of spawns already made.
**Buff and vary the non-derived undead.** Instead of four possible derived undead, there are now thirteen, which have a variety of different powers: invisibility, hexes, MR/2 attacks, slowing attacks, fearsome and vampiric melee, summoning and offensive spells, as opposed to the skeletal warriors and 1 ancient champion that were the most dangerous things that spawned previously.

Please could you provide general feedback where possible, in addition to specific feedback regarding:
(a) Code style: feel free to tell me how the code could be clearer or more efficient;
(b) Is the boss battle too dangerous now, and would there be something more fitting other than more liches?
(c) _How do I set up set_random_mon_list - it seems not to actually affect anything when it comes to gameplay?_ **Done, thanks for explaining this gammafunk.**
(d) _Is it even necessary at all to have a cap on the number of non-derived undead that can spawn from each tier? Given that there is now a much greater variation in the monsters that spawn, I am wondering if this can be entirely removed._  **I have now removed this, in commit 3.**
(e) The sword of Zonguldrok doesn't actually ever seem to spawn; instead I have found a randart double sword every time that I have tested. Is this a feature or a bug, because it seems like the sword should always generate?

This is my second attempt at creating this PR. After accidentally apparently changing 262 files with an attempted but failed merge commit, I decided to start again from scratch. It appears to be fixed now, but that is why there are a few erroneous commits from before. The relevant commits are all numbered.


crawl-ref/source/dat/des/portals/wizlab.des
```diff
@@ -1138,8 +1204,25 @@ SUBST:      LZ = '
...
+    .. [[ cyclops zombie         / cyclops simulacrum         / spectral cyclops         ]]  
+    .. [[ swamp dragon zombie    / swamp dragon simulacrum    / spectral swamp dragon    ]]
+    .. [[ deep troll zombie      / deep troll simulacrum      / spectral deep troll      ]]
+    .. [[ stone giant zombie     / stone giant simulacrum     / spectral stone giant     ]]
+    .. [[ titan zombie           / titan simulacrum           / spectral titan           ]]
+    .. [[ ettin zombie           / ettin simulacrum           / spectral ettin           ]]
+    .. [[ juggernaut zombie      / juggernaut simulacrum      / spectral juggernaut      ]]
+    .. [[ golden dragon zombie   / golden dragon simulacrum   / spectral golden dragon   ]]
+    -- non-derived undead have total weight 270
+    .. [[ vampire w:20 / necrophage w:20 / shadow w:20 / wraith w:20 ]]
```
NormalPerson7: I... can't add up. This should say 280.

crawl-ref/source/dat/des/portals/wizlab.des
```diff
@@ -161,41 +161,97 @@ function wizlab_zonguldrok_spawn_zombies(data, triggerable, triggerer,
...
+        "cyclops", "two-headed ogre", "swamp dragon",          -- tier 1
+        "stone giant", "deep troll", "storm dragon",           -- tier 2
+        "juggernaut", "ettin", "golden dragon"                 -- tier 3
+    }
+    
+    -- These are the non-derived undead to be spawned (tier 4 is always an ancient champion)
+    local list_regular_undead = {
+        "shadow", "vampire", "wraith", "necrophage"                                  -- tier 1
+        "skeletal warrior", "vampire mage", "freezing wraith", "phantasmal warrior", -- tier 2
+        "ghoul", "vampire knight", "jiangshi", "shadow wraith"                       -- tier 3
```
NormalPerson7: note to self: add comma on necrophage

regret-index:
Self notes of public record more than anything else, I guess, since the contributor working here hasn't been seen in quite a few years.

bc928ea 's Borg wizlab stole most of what this rework could focus on, by being mini-Crypt except with even fancier monsters between lost souls and halazid warlocks. I'm not entirely fond of wizlab_zonguldrok this much later multiple ways over- wizlabs ended up learning into magic school skill titles and spell names more than the one unrand, and out of all the mechanical focuses to give to a portal for distinction purposes, a second mini-Crypt near the time one might do Crypt isn't selling it much either. Maybe a subvaulted "unrand expo" portal for Lair branches would be neat for giving this more justified purpose and giving those more than just ice caves for threat portals (if probably redundant with the many guarded_unrand series of vaults). Maybe making ghost vaults into subvaulted portals and then always having zombie waves before one can even see a choice of ghosts would be a better use of trying to make the presented undead work the whole game long.

In terms of the work done here, outside of doing such efforts: I do like the derived zombie upgrading, but the wraiths in the non-derived set are unimpressive and alongside the spectrals they wander away from the unrand being a Reaping sword, not a Death Channel sword. Adding plain liches after bc7614a and 6612197 is awkward, since there's not much room for their orbs. With the precedent of the new head instructor alongside the cloud mage and hellbinder, I'd rather make a melee lich (two melee liches?) so the wizlab about a sword can feature a boss trying to kill the player with a sword instead.

There's a pretty large amounts of conflicts from how this one's five years old. I'll see if it's possible to squash most of this work down into one commit and then do my work in a follow-up bit, but if Git's non-cooperative with this and I probably will be then I'll just cite this pull request in my own work, alas.

PR #733, opened 2018-04-10
The unrand trident "Rift" with distortion and extended reach.

aolowin:


rawlins:
No strong opinions about the unrand itself, but I'm pretty nervous about extra-reach from a maintainability/code as well as ui perspective, reaching is already a bit messy in implementation and kind of painful UI-wise for cases where tab doesn't work.

Anyone have thoughts about the unrand?

ebering:
I'm interested in the concept. It's definitely a cool idea. I think there will be space in the unrand pole arms for such a weapon; we're currently in the process of reviewing and revising existing unrands so I'll add this to the list of things to consider. @gammafunk do you have thoughts on this?

Adding an extra square of range is a big bonus, so I'd probably not make it with such a large plus. On the other hand it's a trident. I'd probably make it something like a +3 demon trident.

In terms of the maintainability, it seems like it refactors the reach code in a sensible way, but I'm less familiar with that part of the codebase so that's something to keep in mind.

gammafunk:
I haven't really looked over the details of the implementation, but actually extending reach another tile seems like it's going to introduce a lot of logistical and technical issues for no real benefit in game-play. It's not really interesting for the player to consider reach attacks through two monsters with abysmal accuracy. If people like the extra range reach idea, probably a smarter way to do it is just a range 3 smite target melee attack. Then it's buffed vs normal reach and has an additional range on top of this with what's probably going to end up a much better UI.

As far as strength goes, it's easiest to compare it to something like a longbow with limited range. If we're talking about making it have a demon trident base type, it's going to do a bit less AveEffDam than that with much less range. Its benefit over the longbow is costing a lot less XP and having smite targeting (assuming we go with the smite targeting idea). The reduced XP cost seems like the biggest advantage, so I guess I could see giving it strongly reduced enchant, but something like +5 through +7 would also be reasonable. It's going to outclass e.g. a vorpal longbow, but it's not better than Zephyr. If you keep it as a trident, it should definitely just have high enchantment if not a bit of over-enchantment (I'm not not sure about damage numbers with distortion).

ebering:
Looking back at this, I like the smite targeted range 3 pole-arm plan for this, with a demon trident base type.

If you're still interested @aolowin I'd be happy to merge a modified patch. If not, and someone else is interested in taking up that direction, please send us a PR.

aolowin:
I'll try and get a modified patch together when I get some free time. Probably getting the smite targeting to work with autofight is the biggest unknown. You really want to be able to use TAB with distortion.

aolowin:
PR949 has a new version. This one can be closed.

PR #734, opened 2018-04-12
Fix statues' description not having a reference to their inability to regenerate. (#11448)

NormalPerson7:
This fixes https://crawl.develz.org/mantis/view.php?id=11448.

The first three commits can be ignored --  I still don't know how to clear irrelevant commits from a pull request. The only commit which does anything is the one labelled so.


crawl-ref/source/dat/des/portals/wizlab.des
```diff
@@ -1895,4 +1895,4 @@ ccccccccccc.....cccccc....c*...*c....ccccc''...........''ccc
 ccccccccccccc.rcccccccccccccbPbcccccccccccc'''''''''''''cccc
 ccccccccccccccccccccccccccccbbbccccccccccccccccccccccccccccc
 cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
-ENDMAP
+ENDMAP
```
NormalPerson7: no idea what this is

crawl-ref/source/mon-data.h
```diff
@@ -5909,7 +5909,7 @@ DUMMY(MONS_GOLEM, '8', LIGHTGREY, "golem", TILEP_MONS_IRON_GOLEM)
 
 { // always redefined
     MONS_STATUE, '8', LIGHTGREY, "statue",
-    M_STATIONARY | M_SPEAKS | M_ARCHER | M_DONT_MELEE | M_NO_POLY_TO,
+    M_STATIONARY | M_SPEAKS | M_ARCHER | M_DONT_MELEE | M_NO_POLY_TO | M_NO_REGEN,
```
aidanholm: `mons_can_regenerate()` checks the monster class's flags as a fallback, so you should be able to remove all other `M_NO_REGEN` flags and just keep the one on `MONS_STATUE`. Haven't tested this though.

NormalPerson7: I tested this and it does not appear to work - Only `MONS_STATUE` and its derivatives keep the "It cannot regenerate" text, all other statues do not get it.

aidanholm:
Thanks for the fix, manually merged at https://github.com/crawl/crawl/commit/c608dd7fdd4d

PR #735, opened 2018-04-13
Fix the player being prompted twice to put on *Drain/*Contam rings 

NormalPerson7:
Previously the player would be asked "Really put blarggh on?" both before and after they selected which ring to swap it for. This PR changes it to only prompt the player before. 

Feel free to tell me of a cleaner way to pass the argument `check_for_inscriptions` from `JewelleryOnDelay::finish()` to `_puton_item` without having `puton_ring` take it only for the function call to `_puton_item`, because this seemed messy to me.

The first three commits can be ignored.


crawl-ref/source/item-use.cc
```diff
@@ -1708,13 +1709,16 @@ static bool _puton_item(int item_slot, bool prompt_slot)
 
     // It looks to be possible to equip this item. Before going any further,
     // we should prompt the user with any warnings that come with trying to
-    // put it on.
-    if (!check_warning_inscriptions(item, OPER_PUTON))
+    // put it on, except when they have already been prompted with them
+    // from switching rings.
+    if (check_for_inscriptions)
```
aidanholm: These could probably be combined into a single `if (... && ...)`

crawl-ref/source/item-use.cc
```diff
@@ -1845,9 +1849,8 @@ bool puton_ring(int slot, bool allow_prompt)
     if (prompt_failed(item_slot))
         return false;
 
-    bool prompt = allow_prompt ? Options.jewellery_prompt : false;
```
aidanholm: No need to delete the extra line here (and the one above, before "const bool is_amulet"), as it bloats the diff.

crawl-ref/source/item-use.cc
NormalPerson7: I am not 100% sure how C++ works so feel free to correct me. I assumed that if it was all in one `if(...)`, the function check_warning_inscriptions would be run even when check_for_inscriptions was false, and it's this function which causes the prompt -- we do not want to call this function if check_for_inscriptions is false.

crawl-ref/source/item-use.cc
```diff
@@ -1845,9 +1847,9 @@ bool puton_ring(int slot, bool allow_prompt)
...
+    bool prompt = allow_prompt ? Options.jewellery_prompt : false;
```
aidanholm: Extra space at the end of the line here; you can use `git diff --check` to check for problems like this.

aidanholm:
Merged at 30dfbe6fde

PR #736, opened 2018-04-13
Fix the message order for unwielding Fragile/*Contam/etc artefacts (#11427)

NormalPerson7:
This PR aims to make the message order when handling randarts make more sense. 

I still haven't figured out how to avoid having these first three commits in my PRs - `git rebase master` isn't changing things. Hence the first three commits can, once again, be ignored.


crawl-ref/source/player-equip.cc
```diff
@@ -632,10 +632,9 @@ static void _unequip_weapon_effect(item_def& real_item, bool showMsgs,
 
     // Fragile artefacts may be destroyed, so make a copy
     item_def item = real_item;
-
-    // Call this first, so that the unrandart func can set showMsgs to
-    // false if it does its own message handling.
-    if (is_artefact(item))
+    // If item is an unrand, call this first, so that the unrandart func
```
NormalPerson7: This variable is probably not even needed now, but I'll leave it here in case some Fragile unrand one day gets added.

aidanholm:
I'm not entirely happy with this approach. I think a better (although still not great) solution would be to:
1. move the `ARTP_FRAGILE` check and its contents to a separate `_unequip_destroy_fragile_artefact()` helper
2. only call it from `_unequip_artefact_effect()` if the caller is `_unequip_weapon_effect()`
3. always call it from `_unequip_weapon_effect()` for artefacts.

NormalPerson7:
Thanks, that does sound better, you're right -- I will do that later.

NormalPerson7:
In fact I will do this in a different branch, so I can start again from scratch and remove these erroneous commits, so I will close this PR then.

PR #738, opened 2018-04-14
Add a reference to * direction being an alternative to ctrl+direction.

NormalPerson7:
A really minor thing that annoyed me when I didn't know it was an option -- since this is only mentioned for this specific case of ctrl in the command list and as such it is easy to miss.

The first five commits can be ignored.


NormalPerson7:
Ok you devs will be pleased to know, I'm sure, that this kind of mass commit idiocy will never happen again. I finally figured out how to use `git rebase`, thanks for the help aidanholm and gammafunk.

crawl-ref/source/main.cc
```diff
@@ -2921,8 +2921,13 @@ static void _move_player(coord_def move)
             // Don't choose a random location to try to attack into - allows
             // abuse, since trying to move (not attack) takes no time, and
             // shouldn't. Just force confused trees to use ctrl.
+#ifdef USE_TILE_WEB
             mpr("You cannot move. (Use ctrl+direction or * direction to "
```
NormalPerson7: Can someone confirm for me that #ifdef USE_TILE_WEB returns true only for WebTiles games and not online console games? aidanh was worried that it might accept all online games.

NormalPerson7: It does accept online console games, thanks |amethyst, but I'm not really sure whether this is an issue at all.

PR #739, opened 2018-04-17
Remove books from shopping list on pickup (Kipster)

nlavsky:
Usually items are removed from the shopping list when they are put into
inventory. But after the introduction of spell library books no longer
take up inventory space, so the usual code path,
    _put_item_in_inv() -> _got_item() -> cull_identical_items(),
no longer works for them.


aidanholm:
Thanks!

PR #740, opened 2018-04-19
Add an option to dump the player's options file.

NormalPerson7:
Since last December, the speedrunning community has realised the potential for the use of multi-action macros made in-game, and lua macros constructed in the RC file, and come to the consensus that this should not be allowed in the rules of realtime speedrunning. However, there has been no way to actually check that players saying they are not using such multi-action macros are not.

This PR aims to go one step towards achieving this goal, by adding an option to dump the options file.
At some point another PR will need to be made, in order to dump a player's in-game macros into the RC file.

~~I will probably have to do a little more work on this to deal with extra options from the command line, but this isn't strictly necessary to get this option to work.~~ (Done)

### DIFF
--- a/crawl-ref/source/initfile.cc
+++ b/crawl-ref/source/initfile.cc
@@ -216,6 +216,7 @@ const vector<GameOption*> game_options::build_options_list()
         new BoolGameOption(SIMPLE_NAME(default_manual_training), false),
         new BoolGameOption(SIMPLE_NAME(one_SDL_sound_channel), false),
         new BoolGameOption(SIMPLE_NAME(sounds_on), true),
+        new BoolGameOption(SIMPLE_NAME(dump_options), false),
         new ColourGameOption(SIMPLE_NAME(tc_reachable), BLUE),
         new ColourGameOption(SIMPLE_NAME(tc_excluded), LIGHTMAGENTA),
         new ColourGameOption(SIMPLE_NAME(tc_exclude_circle), RED),
--- a/crawl-ref/source/initfile.cc
+++ b/crawl-ref/source/initfile.cc
@@ -1584,13 +1584,17 @@ void read_init_file(bool runscript)
     UNUSED(lua_builtins);
     UNUSED(config_defaults);
 #endif
+    // Clear Options.file_contents here to avoid duplication in dump
+    Options.file_contents.clear();
 
     // Load early binding extra options from the command line BEFORE init.txt.
     Options.filename     = "extra opts first";
     Options.basefilename = "extra opts first";
     Options.line_num     = 0;
+
     for (const string &extra : SysEnv.extra_opts_first)
     {
+        Options.file_contents += extra + "\n";  // Add extra options to dump
         Options.line_num++;
         Options.read_option_line(extra, true);
     }
@@ -1635,7 +1639,7 @@ void read_init_file(bool runscript)
     if (f.error())
         return;
     string option_file_contents = Options.read_options(f, runscript);
-    Options.file_contents = option_file_contents;
+    Options.file_contents += option_file_contents;
 
     if (Options.read_persist_options)
     {
@@ -1653,6 +1657,7 @@ void read_init_file(bool runscript)
     Options.line_num     = 0;
     for (const string &extra : SysEnv.extra_opts_last)
     {
+        Options.file_contents += extra + "\n";  // Add extra options to dump
         Options.line_num++;
         Options.read_option_line(extra, true);
     }
--- a/crawl-ref/source/initfile.cc
+++ b/crawl-ref/source/initfile.cc
@@ -1625,6 +1625,8 @@ void read_init_file(bool runscript)
     }
 #endif
 
+    FileLineInput f(init_file_name.c_str());
+
     Options.filename = init_file_name;
     Options.line_num = 0;
 #ifdef UNIX
@@ -1633,18 +1635,10 @@ void read_init_file(bool runscript)
     Options.basefilename = "init.txt";
 #endif
 
-    // Read in the options file
-    FILE *f = fopen_u(init_file_name.c_str(), "r");
-
-    unsigned long size = file_size(f);
-
-    char *options_contents = new char[size];
-
-    fread(options_contents, 1, size, f);
-    fclose(f);
-
-    read_options(options_contents, runscript);
-    Options.file_contents += options_contents;
+    if (f.error())
+        return;
+    string option_file_contents = Options.read_options(f, runscript);
+    Options.file_contents += "\n" + option_file_contents + "\n\n";
 
     if (Options.read_persist_options)
     {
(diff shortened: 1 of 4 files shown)


crawl-ref/source/initfile.cc
```diff
@@ -1748,7 +1751,7 @@ void save_player_name()
 void read_options(const string &s, bool runscript, bool clear_aliases)
 {
     StringLineInput st(s);
-    Options.read_options(st, runscript, clear_aliases);
+    string test = Options.read_options(st, runscript, clear_aliases);
```
NormalPerson7: Feel free to tell me of a better way to do this, because it seems incredibly messy to create a variable which is never used again.

aidanholm: I'm not sure why you create a variable here; can't you just ignore the return value?

In any case, I'm not a huge fan of teetering abstraction, but you could create a class deriving from `FileLineReader` and overriding `get_line` that additionally fills a string with the file contents. The API would look like this:
```c
CopyingFileLineReader cflr(f, Options.file_contents); // second param is the string to fill
Options.read_options(cflr, runscript);
```
If you do this, please get the file size and pre-expand the string before filling it.

crawl-ref/source/initfile.cc
```diff
@@ -1788,6 +1791,9 @@ void game_options::read_options(LineInput &il, bool runscript,
 
         trim_string(str);
 
+        content += str;
```
aidanholm: `content += str + "\n";`

crawl-ref/source/initfile.cc
NormalPerson7: Is this necessary, because this would probably take a while for me to properly understand?

crawl-ref/source/initfile.cc
NormalPerson7: You think it's better on one line than two? I mean I can do that... I just thought it _really_ didn't matter.

crawl-ref/source/initfile.cc
NormalPerson7: Ok so if I understand this correctly, that would actually just be adding those two lines instead of `Options.read_options(blargh)`? I will do that then.

NormalPerson7: Never mind, it isn't that easy; I'll leave it how it is currently, and just add the command line extra options into `Options.file_contents`.

crawl-ref/source/initfile.cc
```diff
@@ -1584,13 +1584,17 @@ void read_init_file(bool runscript)
...
     UNUSED(config_defaults);
 #endif
+    // Clear Options.file_contents here to avoid duplication in dump
+    Options.file_contents.clear();
 
     // Load early binding extra options from the command line BEFORE init.txt.
     Options.filename     = "extra opts first";
     Options.basefilename = "extra opts first";
     Options.line_num     = 0;
+
```
NormalPerson7: Sorry about the extra blank line...

crawl-ref/source/initfile.cc
aidanholm: If you're less familiar with C++, creating such a class might be more difficult.

A third option, probably better than the current, is to read the file contents into a string, construct a StringLineInput from that, and pass that to Options.read_options instead of the FileLineReader currently used; then you have the file contents as a string as a side effect.

NormalPerson7:
I saw there was some discussion on ##crawl-dev about the implications of using StringLineInput -- is the first method acceptable or do you want me to try the second method, seeing as it appears the StringLineInput method isn't good enough?

aidanholm:
Yeah; I think out of the options, reverting the most recent commit is probably best.

crawl-ref/source/initfile.cc
```diff
@@ -1632,7 +1637,8 @@ void read_init_file(bool runscript)
 
     if (f.error())
         return;
-    Options.read_options(f, runscript);
+    string option_file_contents = Options.read_options(f, runscript);
+    Options.file_contents += "\n" + option_file_contents + "\n\n";
```
brianlheim: this will result in 2 copies of `option_file_contents` - one for the first `+`, and one for the `+=`. It would be better to append each element separately using `+=`. The other uses of `+` in this diff will similarly do some extra allocation you could avoid.

sorry for lurking, I just really like this project and sometimes take a peek at what y'all are up to :)

NormalPerson7: Are you saying that there will be two copies of the options file in the dump? Because some simple testing reveals that is not the case... Otherwise I don't really understand your point.

Glad to hear you like the project.

NormalPerson7: Just to clarify my understanding (which I believe to be correct, I'm no expert):
If options_file_contents contains one thing, eg `"test"`, then line 1641 creates the string `"\ntest\n\n"`, and appends this to `Options.file_contents`. I don't know where you think it is being duplicated.

aidanholm: I believe @brianlheim is referring to temporary copies of the string containing the file contents, caused by the order of operations used: the string "\n" is created, then the file contents is copied onto the end of that, then "\n\n" is copied onto the end of that, and the result is then copied onto `Options.file_contents`; so the full contents of the file are copied twice. They are absolutely correct that this is less than maximally efficient, but this is somewhat less of a concern in this particular case.

brianlheim: @aidanholm - that is correct. You're right this is not a huge concern in this particular instance, but being neglectful of temporary copies can lead to bad performance overall if the mistake is made repeatedly. And it is made here several times already: `content += str + "\n"` in the worst case may cause a reallocation when `"\n"` is appended to `str`, which is pessimistic when both are going to be copied to `content` in the end anyway.

NormalPerson7: Right, that makes sense, thanks guys.

brianlheim: Np! Sorry, I know it's kind of rude of me to interject out of the blue like this. I should say that in general I think you guys do a better job keeping this codebase clean than I probably would. :) It gives me hope that if/when I start to contribute it'll be fun and easy!

aidanholm: If you think that's bad, just look at line 1798 of this diff :)

NormalPerson7:
So is this PR ready now, or is there anything else that you think should be done/would prefer me to do beforehand?

NormalPerson7:
I am closing this PR now, because the realtime speedrunning community will now use video evidence to vet the very fastest of runs as had previously been suggested by many of the development team, so such option file dumping is less of a priority (even though the basic functionality may be useful for other purposes). It seemed like the best way forward for dumping the options file for other purposes was to use hashes to check when (and how) the options file was changed, and I don't feel I am the right person to fiddle around with that.

The basic functionality will remain in this PR, and I will not delete the NormalPerson7/crawl:speedrun-rc branch in case anyone else wants to pick up the pieces for this in the future. It is largely no longer required for the realtime community though.

PR #745, opened 2018-04-25
Update full_describe_view() to fix an edge case with apportation

Jmeigs1:
Bugfix for issue #11358


Jmeigs1:
After discussion in IRC and more testing I've decided to dig further and find a more fundamental fix for stashtracker getting out of sync.  The issue is not limited to ^X but all things that depend on stashtracker.

PR #746, opened 2018-04-27
Fix disconnected areas in erik_orc_3 (Leszczynek, #11460)

NormalPerson7:
This vault had several locations where one-or-more tiles could become disconnected from the rest of the vault. This PR fixes that.


NormalPerson7:
A similar fix was committed by amalloy in commit 2a24f12, so I will close this.

PR #747, opened 2018-04-28
Update ozocubu's armour to break for non-standard forms of movement

Jmeigs1:
Includes stairs, portals, shafts, wu-jian abils, and frog jump

### DIFF
--- a/crawl-ref/source/Makefile.obj
+++ b/crawl-ref/source/Makefile.obj
@@ -163,6 +163,7 @@ mon-tentacle.o \
 mon-transit.o \
 mon-util.o \
 monster.o \
+movement.o \
 mutation.o \
 nearby-danger.o \
 newgame.o \
--- a/crawl-ref/source/god-abil.cc
+++ b/crawl-ref/source/god-abil.cc
@@ -59,6 +59,7 @@
 #include "mon-poly.h"
 #include "mon-tentacle.h"
 #include "mon-util.h"
+#include "movement.h"
 #include "mutation.h"
 #include "notes.h"
 #include "ouch.h"
@@ -76,9 +77,9 @@
 #include "spl-goditem.h"
 #include "spl-monench.h"
 #include "spl-summoning.h"
-#include "spl-wpnench.h"
 #include "spl-transloc.h"
 #include "spl-util.h"
+#include "spl-wpnench.h"
 #include "sprint.h"
 #include "state.h"
 #include "stringutil.h"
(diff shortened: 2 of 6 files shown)


MarvinPA:
I'm not really a fan of having movement abilities be a factor here, for the same reasons as given in this PR for barbs: https://github.com/crawl/crawl/pull/459

I do think breaking it on level transitions (stairs/shafts/portals) is good though (in which case barbs should probably also be consistent with that).

Jmeigs1:
My apologies, I did not realize this had been previously discussed in some form.

A few points back:

I would argue that Wu Jian leap should be kept regardless on the basis that it should function consistently regardless of whether it is cast via move keys or direct ability cast.

I had not considered barbs and I agree keeping them consistent would be ideal.  Barbs is much worse because traditionally we warn the player when they attempt to self inflict damage.  There is not the same expectation for ice armor.

I'm not sure I agree that the player experience is deeply impacted having these effects trigger for physical movement abilities, provided the standard warnings for self damage are maintained.  The underlying issue seems to me to be that the movement code is in need of a refactor as it doesn't offer a clean way to implement this.

rawlins:
>it should function consistently regardless of whether it is cast via move keys or direct ability cast.

If there's a case where it's not consistent that's a bug (independent of this PR), can you clarify what's happening now?

Jmeigs1:
There are two ways to trigger leap atm.  Pressing a movement key when next to a wall and pressing a(letter) next to a wall.  

When the movement key is pressed the ice armor break, and barbs would both proc as part of standard movement code.  When the ability is cast they do not.

rawlins:
ok, I see; those should be resolved to be consistent one way or the other.

@MarvinPA since wall jump doesn't have any restrictions on re-use, it seems like a different case than hop/power leap, no? My default preference on seeing this bug would be that both forms of wall jump break ozo's, because (i) I think this is what you'd expect just from knowledge about how WJC works, and(ii) it seems like it's going to lead to optimalplayers always moving around with walljump to avoid the restriction. Barbs my feelings are less strong about.

Jmeigs1:
This was discussed briefly via IRC but I feel we quickly got sidetracked into a discussion about the playability/state of WJC.

Purely as a plan forward for this specific PR I would like to:
1. Remove the armor break from hop.
2. Implement a similar barbs check for level traversal
3. Leave the WJC armor break without a barbs check for the time being.

I intend to take a deeper dive in another PR addressing movement code and WJC.

Jmeigs1:
Updated the PR to have the above logic.  Added new movement file that I intend to reorganise player movement into.

PR #748, opened 2018-04-29
Clarify Engulf's movement penalty (Ge0ff, #11458)

NormalPerson7:
This commit adds a ?/T description for the Engulf status and changes the message when moving while engulfed to explicitly reference the fact that it took longer.

The ?/T description right now is convoluted and I would appreciate feedback as to whether (a) it is even necessary to have separate descriptions (I thought it seemed misleading to state that unbreathing chars couldn't read scrolls etc) and (b) if there's a way not to have to list all the species. It may just be better to append a note about how swimming chars get no penalties and unbreathing chars just get the movement penalty, rather than cutting it out of the description altogether.


crawl-ref/source/dat/descript/status.txt
```diff
@@ -333,6 +333,36 @@ You are silenced. This makes reading scrolls, casting spells, invoking divine
 abilities, or shouting impossible. The oppressive, unnatural effect greatly
 hampers your stealth.
 %%%%
+Engulf status
+
+{{
+    desc = "You are engulfed in water"
+
+    if you.race() == "Octopode"
```
aidanholm: Consider using `if you.race():match("Octopode|Merfolk|Barachi") ...`; it can be clearer, and the current indentation isn't very nice.

crawl-ref/source/dat/descript/status.txt
```diff
@@ -333,6 +333,36 @@ You are silenced. This makes reading scrolls, casting spells, invoking divine
...
+        desc = desc .. "."
+    elseif you.race() == "Ghoul"
+           or you.race() == "Mummy"
+           or you.race() == "Vampire"
+           or you.race() == "Gargoyle"
+           or you.race() == "Grey Draconian" then
+        desc = desc .. ". It will take longer to move and escape the water" ..
+                       " elemental engulfing you."
+    else
+        desc = desc .. " and are unable to breathe. You will take increasing" ..
```
aidanholm: Using `[[` and `]]` for multiline string literals might be nicer

crawl-ref/source/dat/descript/status.txt
```diff
@@ -333,6 +333,36 @@ You are silenced. This makes reading scrolls, casting spells, invoking divine
 abilities, or shouting impossible. The oppressive, unnatural effect greatly
 hampers your stealth.
 %%%%
+Engulf status
+
+{{
+    desc = "You are engulfed in water"
```
aidanholm: If `desc` can be local, it should be

crawl-ref/source/main.cc
```diff
@@ -3160,7 +3160,8 @@ static void _move_player(coord_def move)
                 mpr("You deftly slip free of the water engulfing you.");
             else //Unless you're a natural swimmer, this takes longer than normal
             {
-                mpr("With effort, you pull free of the water engulfing you.");
+                mpr("It takes you longer to move as you pull free of the water "
```
aidanholm: Perhaps "You slowly pull free ..." instead of "With effort, you pull free ..."?

NormalPerson7: Yes that would also work.

crawl-ref/source/dat/descript/status.txt
NormalPerson7: Sure thanks, that's the method I was looking for :)

crawl-ref/source/dat/descript/status.txt
NormalPerson7: That still requires concatenation with `.. [[ "text on line" ]]`, no?

MarvinPA:
Documenting this is great but this also seems like a good prompt to simplify the effect itself, it's very overcomplicated as-is. Removing the movement penalty would be a good start, and there's definitely no good reason for it to specifically target Zin recite - it should just be full silence probably. That would also mean less granularity in its effects - either you're fully immune (swimming/unbreathing) or fully affected.

NormalPerson7:
I mean I completely agree with that, I just didn't necessarily feel it was right for me to be the one to initially request such a simplification -- however I will now make such simplification as well as change the original post to reflect that.

NormalPerson7:
There is probably an issue in how the status label for Engulf is now handled -- it should be dark grey for unbreathing chars now, not yellow as it used to be. I haven't yet figured out how to change that.

PR #749, opened 2018-04-29
Update poisonous vapours to not leak armour brand via warning

Jmeigs1:
Now only considers the monster base resists for the warning.


NormalPerson7:
I think this information leak occurs to other poison spells like mephitic cloud and OTR, so you should also check them.

Jmeigs1:
OTR doesn't have a problem.

Mephitic cloud does leak but the issue is different and related to the beam/targeting class. I believe there are beam targeting spells from other schools that have this issue as well.  It's complex enough that it warrants its own PR.

Tested all other poison spells from here: http://crawl.chaosforge.org/Poison_Magic

MarvinPA:
This is no longer necessary as monster equipment is identified on sight instead in 6bcb69f0587a50dd59af.

PR #755, opened 2018-05-07
Fix Stomp description

Doesnty:
Stomp's description is totally wrong; the damage it deals is based on a percentage of current HP, not max. The percentage also does not increase with Invocations skill; it is always 1/6, with the added component independent of max HP rising with skill (from 2d2 to 2d15.5). I'm not sure if this is a problem with the description or the implementation, since looking back the two have conflicted since Uskayaw was first added, but the unambiguous comment in _get_stomped() indicates the description is in the wrong.

### DIFF
--- a/crawl-ref/source/dat/descript/ability.txt
+++ b/crawl-ref/source/dat/descript/ability.txt
@@ -755,8 +755,8 @@ Accepting a sacrifice will reset the timer to the standard length.
 Stomp ability
 
 Stomps down in time with the rhythm of the dance, sending a shockwave through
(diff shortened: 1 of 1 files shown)


Doesnty:
>send simple text fix pr
>it sits untouched for 23 days with no comment
>someone else sends basically the same pr
>they realize it's a dupe, delete the pr
>gammafunk tells that person to undelete it so they can accept it
>my now redundant pr is still sitting untouched

??????

I'll just shelve this for you as one last gesture of goodwill, then.

PR #761, opened 2018-05-15
Add stab messages for blobs

Muffindrake:
I spend a lot of time in crawl stabbing things and felt like an opportunity had been wasted to implement yet another food-related stab message.

I find that stale pudding is a good approximation/compromise between the various blobs, which range from viscuous ooze to jiggly jello. Perhaps not stale in the sense that it has completely lost its fresh pudding qualities, but something between those two extremes.

Chance is one in three in accordance with the turkey attack message in another damage type.

### DIFF
--- a/crawl-ref/source/melee-attack.cc
+++ b/crawl-ref/source/melee-attack.cc
@@ -1574,6 +1574,19 @@ void melee_attack::set_attack_verb(int damage)
                 attack_verb = "attack";
                 verb_degree = "'s weak point";
             }
+            else if (get_mon_shape(defender_genus) == MON_SHAPE_BLOB
+                     && one_chance_in(3))
+            {
+                static const char * const pierce_desc[][2] =
+                {
+                    {"prod", "like stale pudding"},
+                    {"plough into", "like fruit mash"},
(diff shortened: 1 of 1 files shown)


aidanholm:
"poke" feels a little underwhelming for a description of an especially strong attack. I don't really have an opinion about "stale pudding" either way.

Muffindrake:
> You poke the death ooze like stale pudding!!!!

"Poke" is a powerful enough description of an attack against something that has been reduced to food.
I'm not too deep into the food mutilation business however, so I may have missed something better to use here.

aidanholm:
Perhaps "plough into" or "plough through"?

Muffindrake:
That one doesn't really work with an utensil that is not far removed from a kitchen knife and fits better with a larger and more bulky implement, if we ever get a hoe weapon in crawl.

aidanholm:
"plough" doesn't necessarily have to be used with farming implements; it also means "strike with great force", and when used with food it can mean "to begin eating quickly", which matches the food imagery.

MarvinPA:
I'd tend towards closing this unless anyone else has any input, I don't really think the proposed messages sound good unfortunately.

PR #762, opened 2018-05-21
A knot of arrival vaults.

ebering:


crawl-ref/source/dat/des/arrival/simple.des
```diff
@@ -2150,3 +2150,59 @@ ttt...ss.............ss...ttt
  tt.......................tt
 xxxxxxxxxxxxxx.xxxxxxxxxxxxxx
 ENDMAP
+
+NAME:   ebering_hopf_link
```
alexjurkiewicz: without looking in game and probably being too sensitive.... is this perhaps swastika-like? 😅

rawlins: The vault overall looks ok to me, but for this one it probably depends on what it looks like when the player is at { with the rest of the vault not revealed -- I suspect it's ok but it's worth checking.

ebering: Here's the view. If it doesn't pass muster the `SUBST` line can be changed to not use opaque terrain types. The vaults in this pr fit in a single field of view with the player on `{`.

    ##
    #.
    #.
    #.##
    #.@#
    #..#
    ##.#
     ..#
     ..#
    ####

PR #764, opened 2018-05-22
Refactor elemental colours and recolour awoken forests in console.

ebering:
Make awoken forests an animated RED/LIGHTRED (angry trees) so that the
effect stands out to the player.

Old behavior is awoken forests are an animated GREEN/LIGHTGREEN
alternation. This sticks out a bit in swamp, where trees are
GREEN/BROWN, but is still not as dramatic as the berserk faces now in
tiles. It is also much less noticable outside of the swamp (some lair
ends in particular) where trees are GREEN/LIGHTGREEN to begin with.


gammafunk:
When I tested this in console, it seemed that 1/3 of the time in the UI states a tree would get colored as simply green.

PR #769, opened 2018-05-28
Correct Usk stomp description

ebering:


ebering:
Sorry, this duplicates #755 . Closing.

ebering:
Re-opening at gammafunk's request because #755 mentions the exact fraction, which is not current crawl philosophy.

ebering:
Re-opening at gammafunk's request because #755 mentions the exact fraction, which is not current crawl philosophy.

PR #770, opened 2018-05-30
Sif Muna adjustments

RealzHS:
Sif Muna adjustments:
 - Divine Energy removed.
 - Channel Magic now at 1*.
 - Spell miscast protection still at 2*, and now also refunds half the MP cost of the miscast spell, rounded up.
 - Book gifts now start at 3*.
 - Amnesia still at 4*. 
 - New ability at 5* - Divine Focus - permits 3 instant spell casts, but at a high piety cost and with a minusCast timeout after (timeout duration reduced w/ Invo skill, minimum 3-4 turns).
 - Overall piety gain slightly increased (total gain is now about 2/3rds or a little less that of a 'normal' god like Makhleb).

  \- Realz

[aidanholm edit: formatted as list]


RealzHS:
Overall, these are relatively light changes and the god is mostly the same, but it's hopefully enough to be exciting.

The general ideas are:

1) - Sif is a little soft from 1-3*. Earlier access to Channel Magic and the slight piety bump provide a nice boost in this regard
2) - Book gifts are great, but ideally they come online early enough to more heavily influence build decisions for a character
3) - Sif is a generalist caster god who is mostly passive, and a second combat ability will help further distinguish Sif from the all-passive and laser-focused Vehumet

The goals for the active are:

1) - Give Sif casters the ability to do something that no other spellcaster+god combo can
2) - Put spells in a new context within which to evaluate them
3) - Provide opportunities for players to feel creative and clever, both with their spell builds and tactically

As an aside, I've heard mention that piety gain is possibly due for a revision/reduction, and I welcome that. I'm interested to see what a 10-20% reduction in piety growth across the board would do, but I imagine it's desirable to make a universal change rather than tweak all non-Sif piety. Given that, I think it makes sense to first (or simultaneously) allow Sif to get closer to (read: still significantly less than) other gods' piety growth.

I hope this gets a chance to be tested in trunk. If it doesn't go well and it gets pulled, maybe it can at least inform future Sif-related work =)

Cheers,

Realz

Hellmonk:
sif is real heckin' bad right now so I support this rework

ebering:
I'm interested in taking a look at Sif in 0.23 based on some of these ideas. I've updated your brainstorm page with some thoughts and an idea hellmonk and I kicked about half jokingly for a Sif active.

Specific remarks:
  - I agree that something about piety gain needs to be addressed, but I think the issue is with communication and feel, not the actual gain number. 80% of Sif max piety milestones in 0.21 occur at XL <= 20. 
  - Miscast effects are currently hard to reason about, and can be very severe indeed even at low XL. Probably not worth rolling the dice.
  - @gammafunk did the last revision. I couldn't find any comments from him on the divine energy/channeling changes you suggest.
  - Previous IRC feedback on Divine Focus pointed out that without a high piety cost it's possibly overpowered.

If you're still interested in further developing these ideas come back by ##crawl-dev or let's set to work on the wiki brainstorm page.

gammafunk:
@ebering Some comments and thoughts here since you pinged me:

The current weakness of Sif mostly stems from how slow her piety gain is and how two of her powers are related to MP restoration. Slow piety gain is a problem for getting to 1* compared to other gods, and it's a problem for keeping 100% miscast protection. Probably the best fix for the piety issue is to just award piety on kills instead of having this hybrid approach. With no other adjustments this would improve Sif's early game a lot and make miscast protection more useful.

It's not actually critical that Sif has an additional active if it's relatively easy to get to full miscast protection; not every god needs multiple actives. It's true that miscast protection is not an especially popular passive among players, but it is strong when used well. If Sif does have another high cost active, to some extent it can't mesh well with miscast protection. There are sort of two routes to directly addressing that that I could imagine. The first would be to find a passive to replace miscast protection that didn't depend so strongly on high piety levels so that the higher cost active doesn't effect availability of two passives. The second is to just keep miscast protection and try to make the active something that people will want to use fairly rarely. The triple cast ability in this PR seems like it's taking this approach, since it looks like it's going to be better to just cast a spell three over three turns in a very large majority of situations, given the cool-down. (_Technical side note: Divine Focus uses a formula for the cool-down that introduces breakpoints for influence of Invocations skill, so it needs to use another approach there)_

Alternately you could just avoid directly addressing it simply have a high-cost active that people will want to use a bit more commonly and just let them decide if they want to sacrifice their passives at least temporarily. A multiple casting ability like the one in the PR but without any -Cast cool-down would be one example of doing this. I don't love the multiple casting abilities since they feel like somewhat obfuscated "spell finesse", but they're workable idea, I guess. Another thing to keep in mind with these proposed actives is how strong are we trying to make Sif? If we're just trying to catch her up a bit to Vehumet, a weak god in many ways, then changing piety gain to be on kills possibly along with just moving channel to 1* and removing divine energy would do this. Any new active can easily make the god functionally a lot stronger, so if the motivation for said active is mostly "give players another cool use for piety", you have to walk a fine line to avoid buffing the god too much.

I don't think the proposed MP-restore miscast protection change is good. Sif just shouldn't have two abilities related to restoring MP (one problem with current Sif), and this PR replaces the more powerful miscast protection with simple MP restoration. Sif doesn't need more MP restoration if she grants access to relatively cheap channeling.

I also don't agree that book gifts need to change. This idea about influencing build decisions seems to be targeting users who don't start with spells, but this is just always going to be a very small set of Sif worshipers. A vast majority of characters worshiping Sif have starting books that already do an excellent job of carrying the player to this point. Currently Sif starts gifting spells basically at the perfect time, at some point during Lair or perhaps even before depending on skilling and branch generation. I think the main results of gifting spells earlier will be to grant even more spells and to make gifting seem less exciting. Additionally Vehumet and Kiku already cover the niche of gifting spells early.

ebering:
Sif got overhauled in 0.24, and this was a great starting point for the conversation. Thank you!

However, with the changes in it's time to close this PR.

PR #771, opened 2018-05-30
Refactor attack punctuation

Muffindrake:
That function stood out like a splinter of hot garbage in a sore thumb while I was casually browsing the code base. Here's an equivalent replacement (as well as axes for duplicate code).


aidanholm:
Probably should say "remove" not "axe" in commit messages, in case anyone greps for axe-related stuff.

Muffindrake:
@aidanholm Noted.

PR #772, opened 2018-06-01
magic system for ogres or trolls or something else

pubby:
I've been messing around with a special magic system. It's a bit janky, but I've been playing games with it and it seems pretty fun/unique. Thought I'd post it here for funsies.

The gist of it is:

- You don't learn spells from spell books, instead you 'discover' spells by gaining experience. The formula is 1 random spell every 2 levels, up to level 20.
- Because of this, you only discover a few (10) spells per game.
- The only spell school is spellcasting. All spells are "spellcasting" type.

I've put this system on the ogre species for now, but I think it would work OK on trolls or a new species. I hope someone (gammafunk) will try it out.


alexjurkiewicz:
2018-06-01 06:27:31,604 INFO: #59683 P25667 ERR: ASSERT(you.species == SP_FELID) in 'ng-wanderer.cc' at line 169 failed.

(When starting as a Wanderer)

pubby:
@gammafunk

pubby:
use the oni branch

PR #773, opened 2018-06-03
Grant Qazlal cloud immunity to allies

RealzHS:
This fixes the non-combo circumstance where your own clouds kill your Qazlal Elementals. Furthermore, Qazlal's noise is already a sufficient penalty and denying most summons is an unnecessary limitation on tactical and build choices.

### DIFF
--- a/crawl-ref/source/cloud.cc
+++ b/crawl-ref/source/cloud.cc
@@ -918,6 +918,12 @@ bool actor_cloud_immune(const actor &act, const cloud_struct &cloud)
         return true;
     }
 
+    if (!player && have_passive(passive_t::cloud_immunity)
+        && (act.as_monster()->friendly() || act.as_monster()->neutral()))
+	{
+        return true;
+    }
+
     return false;
 }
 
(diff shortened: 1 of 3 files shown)


crawl-ref/source/cloud.cc
```diff
@@ -918,6 +918,12 @@ bool actor_cloud_immune(const actor &act, const cloud_struct &cloud)
         return true;
     }
 
+    if (!player && have_passive(passive_t::cloud_immunity)
+        && (act.as_monster()->friendly() || act.as_monster()->neutral()))
+    {
+        return true;
+    }
+
```
NormalPerson7: These braces are unnecessary when you have a single line inside the `if`.

RealzHS: I'll forward this by pointing out I'm still relatively new to programming and my C++ experience starts and ends with the few months I've learned via modding Crawl. In this case I borrowed the formatting from the block just below this one, and although I recognized it as unnecessary, I thought it helped with readability following a multi-line if statement. Is this a reasonable approach, or is a hard rule of no bracing a standard practice I should be following regardless of circumstance?

alexjurkiewicz:
In crawl code, any single line block should not have braces. You can run
util/unbrace to automatically apply this change. But if a PR is accepted
it's easy enough for devs to run it themselves as a follow up commit.

ebering:
Closed in the mentioned commit.

PR #780, opened 2018-06-08
Don't re-check forgotten map squares (10461)

ebering:
Prevents annotations for runed doors and transporters from stacking up.

### DIFF
--- a/crawl-ref/source/travel.cc
+++ b/crawl-ref/source/travel.cc
@@ -4750,8 +4750,15 @@ bool check_for_interesting_features()
         if (env.map_knowledge(p).flags & MAP_SEEN_FLAG
             && !env.map_seen(p))
         {
+            // Update the shadow map
             env.map_seen.set(p);
-            _check_interesting_square(p, discoveries);
+
+            // But don't stop if we knew about it previously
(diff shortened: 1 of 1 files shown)


crawl-ref/source/travel.cc
```diff
@@ -4750,8 +4750,15 @@ bool check_for_interesting_features()
         if (env.map_knowledge(p).flags & MAP_SEEN_FLAG
             && !env.map_seen(p))
         {
+            // Update the shadow map
             env.map_seen.set(p);
-            _check_interesting_square(p, discoveries);
+
+            // But don't stop if we knew about it previously
+            if (env.map_forgotten.get()
+                && !((*env.map_forgotten.get())(p).flags & MAP_SEEN_FLAG))
```
aidanholm: unique_ptr implements `operator bool` and `operator*`, so the `.get()` calls here should be superfluous

ebering: Thanks for the tip. There are superfluous get()s elsewhere in the crawl code.

ebering:
Force-pushed with the fixup commit squashed.

PR #782, opened 2018-06-08
Clean up uses of unique_ptrs

ebering:
operator bool and operator* allow a lot of superflous .get() calls to be
removed, as well as a few ASSERTs that will never trigger and some
braces.


aidanholm:
Thanks for the fixups! Merged this one manually, as there were a few minor adjustments/bug fixes needed.

PR #783, opened 2018-06-09
Fix grammar in Zin's recite description (0011491)

shieldal:
Fixes a grammar mistake that affected Mennas' description. Previously read "Reciting Zin's laws will not affect he". Reworded to more closely resemble the other Zin's laws description.


rawlins:
Thanks for the fix, I'll go ahead merge this as is. For future reference of anyone who's interested, this could also have been fixed by replacing `pronoun` with `mi.pronoun(PRONOUN_OBJECTIVE);`

PR #784, opened 2018-06-09
Add password reset function to webtiles

floraline:
Hi, this is my first PR for the crawl project. The other day I tried to log on to one of the webtiles servers and realized I forgot my password. The admin was happy to help me and it was taken care of quickly, but I was surprised that there isn't an automated process for this. (I've done my best to make sure this feature doesn't already exist; if I missed it, I'll be very sad)

This PR adds a password request function to the webtiles server. A user can open the forgot password form to request a reset link in their email. When they click on the link, they are sent back to the server with a unique parameter that allows them to enter a new password. Of course, an SMTP server is required to send the emails.

I've only tested this as much as I had to to keep working on it. I'm very new to this project, and I am also new to Python and SQLite. I understand that this PR may never be merged, but please give feedback where you can. I would like to learn everything I can from this. I do have a couple of comments that I would most like addressed:

- Are usernames unique in the webtiles password DB? I did not notice any unique constraint
- Are emails unique?
- Is there a better way to know the host (URL and port) from within the server? I added it as a config option "lobby_url"
- ~~I was not able to test the Python SMTP function over SSL~~ I've done this
- ~~I also was not able to test the Python SMTP Login function, and as a result I'm not sure if the config values I added are enough~~ I've done this
- Is it important that I check to guarantee that the secret key is unique before sending it via email?

These things are bothering me:

- What is the proper way to arrive at the password reset form, after the user clicks the link in their email? I added a dialog INSIDE the go_lobby function because this was the only way I could get it to work, but something about this feels wrong. The socket is loaded in its entirety while you're entering your new password, but I have to completely reload the page in order to get rid of the secret key parameter, so the socket is loaded twice for no reason

- This PR adds two new columns in the database. I added the columns to the 'ensure_user_db_exists' function, but what about servers that already have a database? I included a .sql file that will upgrade an existing database, but I don't know how this is normally done.

Thank you so much for reviewing this PR. All suggestions, corrections, concerns, and other comments are welcome.


aidanholm:
Hi, and thanks for the contribution! This looks like it'll be a great addition to the webtiles functionality.

There is no need to check that a randomly generated token is not already present in the database. The chances of a collision are infinitesimal.

Unfortunately, your token generation mechanism is vulnerable; the output of `random.getrandbits()` (which uses the Mersenne Twister RNG) is directly included in the password reset email. A malicious user could reset their password many times to recover the state of the RNG, enabling them to predict all future tokens. They could then send a password reset email for any user, predict its token, and change that user's password. I haven't had time to look too closely, but there might be other security issues as well.

To fix this, you should use a secure RNG. You also currently generate 64 random bits, then stretch them to 256 bits with SHA-256; I would prefer that you generate 256 random bits directly and remove the use of SHA-256 entirely.

rawlins:
Thanks for the contribution! I agree that this would be great to have. I don't know much about crypto so I won't comment on that. Re sqlite injection, my understanding is that execute calls that use string substitution for `?` parameters are safe, so the way it is done in this PR looks fine to me at a quick glance. (See [here](https://docs.python.org/2/library/sqlite3.html) for documentation.) What wouldn't be safe would be to construct a query using python string substitution.

One potential issue is that servers might not even validate email format, let alone check that it's a real email. So it would be nice to do at least some validity checking before sending the password reset email. I'm not sure if there are any standard tricks for this or best practices for password reset emails along these lines, but if there are, we should find out about them and use them (to avoid the server's smtp server getting blacklisted if nothing else). It's possible that some server operators would be interested in validating email up front, too, so that's a change that might go naturally with this PR.

> Are usernames unique in the webtiles password DB? I did not notice any unique constraint

They should be unique per server in virtue of the way they are created/accessed.

> This PR adds two new columns in the database. I added the columns to the 'ensure_user_db_exists' function, but what about servers that already have a database?

I think what you'd want to do is add code to the `ensure_user_db_exists` function that checks for the columns and adds them as necessary. I do wonder if it might be better to put these columns in a different table, though? That would (a) make it easier to completely clear out or reset of something goes wrong, and (b) help with the fact that this database/table is also used by dgl in a separate process from webtiles (I'm not immediately sure if anything would go wrong but dgl is a bit complicated so I worry about it). It should be relatively easy to test upgrading code with a webtiles build, just checkout master, do a webtiles build, and create a user or two via the webtiles interface, and it will create the default db format in the webtiles directory under source as `passwd.db3`. Then make a copy of that so you can swap it back in as needed, go to your branch, do a webtiles build, and try out the upgrade code. Incidentally the way I dealt with the concern about dgl in https://github.com/crawl/crawl/pull/726 was to add an entirely separate db, but it probably doesn't make sense to do that for these columns.

Some of your other questions I don't have immediate answers to and may ultimately need to be ironed out in practice with the help of some server admin(s).

floraline:
>One potential issue is that servers might not even validate email format, let alone check that it's a real email. So it would be nice to do at least some validity checking before sending the password reset email.

My understanding here is that the only proper way of validating email addresses is to send a validation email. If they are capable of receiving that email then we know it's a valid address. Regex checking of email addresses usually falls short of proper validation.

I used the same error-checking in the forgot password form that is used in the registration form, which basically just ensures that there are no spaces. It might be appropriate to improve validation of email addresses in both places in this PR, and it may also be necessary to add upfront email validation. Does anyone have thoughts on this?

On putting the new columns in a separate table, I'm not completely opposed to this if there is a consensus. The reason I chose to put the columns in the user table is that both new values are directly tied to a user. This guarantees that each user will only ever have one token. It also guarantees that the database size will not grow indefinitely; repeated reset requests will only write to these two fields, rather than potentially inserting new rows in another table. This means the database will only grow to at most X new bytes per user. One issue I have with creating a separate table is that you are adding a relationship that must be maintained. I think this hurts maintenance rather than helping it, but I am open to discussion on this. One factor I didn't consider is how this could affect dgl.

floraline:
I have altered the code to use os.urandom which according to [this documentation](https://docs.python.org/2/library/os.html#os.urandom) is "suitable for cryptographic use."

floraline:
I started using a real SMTP server in order to verify that the multipart email is assembled correctly. When I did this, I noticed that smtplib's `sendemail` is a blocking call. I started getting this in the console whenever an email is sent:

    IOLoop blocked for 0.500000 seconds

Can anyone confirm for me that this Tornado warning message is, in fact, only referring to the thread for my specific request? This blocking call is not going to affect anything else on the server, right? This could probably be improved by somehow sending the email asynchronously, but I am perfectly happy with a blocking inline call as long as it only affects the user who is making the request.

floraline:
Is there agreement that it's appropriate to check for these columns and automatically alter the database schema during runtime? It would ensure that the database does not get out of sync with the code, but something could go wrong and data could be lost. If we instead provide a script to upgrade the database, administrators would have full control over the process. They can make sure that a proper backup exists beforehand. They would also have the option of not performing the upgrade, and leaving this feature disabled.

Does anyone have any thoughts on this? I can see good reasons to go either way.

alexjurkiewicz:
Additing a column will be a blocking operation but with the small size of
dcss servers I can't see that being a problem.

aidanholm:
Using `os.urandom` is acceptable. You're absolutely right about SHA-256 protecting against an attacker with read-only access; I was distracted by the key-stretching aspect. I think it's probably preferable to use something like base62 instead of a numerical representation, to reduce the token length.

It'd be a good idea to at least verify that the email address contains an '@'.

One problem with keeping only one token per user is that a user who triest to reset their token multiple times will first receive an email with an invalid token. This is a plausible scenario if the user has a slow email provider with long arrival delays. Resending the first token is not an option as only the hash is stored. The usual way to fix this would be to store tokens in a separate table and permit a reasonable number of tokens per user.

aidanholm:
> some user has stale tokens out there that could be used to maliciously reset their password

Could you describe how such an attack might be carried out? Tokens expire after one hour, and the attacker does not have access to the target user's email.

floraline:
You're right, the attack window would be very small. The only other argument I have is that it's more complex and therefore more difficult to maintain. I guess it would also be easy to spam the server with requests and fill this table up, but you could just do the same with registrations so that's not really a new problem.

I'll go ahead and start working on splitting it out into a separate table, and if there are no other comments by then, I'll make the commit.

aidanholm:
Apart from the one hour limit, I just can't think of a reasonable threat model in which having multiple tokens in a database is less secure than having a single token.

There is a complexity increase, but keeping tokens in their own table is better w.r.t. separation of concerns. While it might be good to do some basic limiting of token requests, rate limiting is really the job of a dynamic firewall.

floraline:
I've moved the tokens into a separate table, for your review.

The only thing left that concerns me is whether we should implement email validation in this PR. I think it's a valuable addition, but I don't know whether it belongs in this PR.

alexjurkiewicz:
This is great, thanks! For email validation, please just ensure the email address matches `.+@.+`. If users have an email address without an `@`, most Linux mail servers will deliver to `@hostname` which might lead to abuse (filling up the disk with an unmonitored local email spool).

floraline:
I'm using this utility function to validate emails:

    def validate_email_address(address):
      if not address: return "Email address can't be empty"
      if " " in address: return "Email address can't contain a space"
      if "@" not in address: return "Expected email address to contain the @ symbol"
      if not re.match(r"[^@]+@[^@]+\.[^@]+", address): return "Invalid email address"
      return None

floraline:
I'm not aware of any forthcoming changes at this point, awaiting further review. Please let me know if this PR can  be improved in any way.

aidanholm:
Just from reading the code, it seems tokens are only validated once the user fills out the token request form. It'd be better to validate the token before showing the password reset form.

I haven't had time to actually test this out; I should be able to do so in a few days.

rawlins:
Thanks again for the PR, and sorry to take so long to merge it.

I've tested this as carefully as I can with a local dummy smtp server (python smtpd), but not in a production setting (which for obvious reasons is hard to do unless you already have a server). So hopefully it won't be too much of an adventure for server operators who enable it. @floraline I assume you tested this and I know you're running an unofficial server, just to check -- have you tested it with a real smtp configuration?

floraline:
I have not tested this on my server, but I did test it in my development copy using a real smtp server and real inboxes. I also used local dummy smtp and got the same good results with both. I'm not super familiar with email so I don't know about potential spam prevention problems that this code might not be avoiding properly.

I will test this as soon as possible on my server proper now that this is in master. Thanks for taking the time to deal with this PR.

alexjurkiewicz:
Hey Floraline, are you running a public server? What is its url? 🎉

floraline:
I'm running a server at kelbi.org, but it's not public really there are just a couple of my friends playing on it. I'm just doing it for the experience.

PR #786, opened 2018-06-10
Make the estimate of XLs to mindelay more accurate (0011490)

shieldal:
Previously the estimated XLs to mindelay in the item description of shields and weapons didn't take the progress toward the next level into account. No matter how much progress you had made toward the next level it counted that as a whole level. With this change it only adds how much exp need to get to the next level to the estimate and matches the desired behavior outlined in the bug report.


rawlins:
Thanks for the fix! I merged this in https://github.com/crawl/crawl/commit/5786b86ef4da and then reworked it further in the resolving commit.

PR #790, opened 2018-06-17
Add a lightning-themed vault for Depths

Muffindrake:


gammafunk:
I took a more careful look at this vault, and I have some quick edits and suggestions that might be useful for a future version, but I also recommend you substantially rework the layout/idea of this vault. More on that after the edits/suggestions.  Here's a quick edit to clean up DES and randomize numbers of monsters, mostly preserving the average:

```
NAME:   muffindrake_ghost_lightning
DEPTH:  Depths
ORIENT: float
KMONS:  O = player_ghost
MONS:   air elemental, black draconian, storm dragon, spark wasp
MONS:   spriggan air mage, titan
: if crawl.one_chance_in(4) then
MONS:   sojobo band, titan
: else
MONS:   titan
: end
NSUBST: - = 1 / 2=1- / 2=2 / 4=2- / 2=3 / 2=3- / 2=4 / 2=4-/ 5 / 2=5- / -
NSUBST: - = 6 / 6- / 7 / O / -
NSUBST: _ = 1 / 1_ / 2 / 2=2_ / 3 / 2=3_ / 6 / 6_ / O / _
FTILE:  12345678O-_$*| = floor_pebble_blue
: ghost_setup(_G)
MAP
      .............
     ..nnnnnnnnnnn..
    ..nn$$$$$n*|*nW..
    .nn$$$$$n*||*nWW.
    .n$$$$$n--nnnnwW..
   ..n$$$nn----nnwwWW.
   .nn$$n------nwwwwW.
   .n$$n------nnwwwWW@.
  ..nnn-------nwwwwwWW.
  .nn--------wnwwwwWWW..
  .n-----m---nnwwwWWwWW.
 ..n----m---wnwwwwwwwwW.
 .nn---m----wnwwwnnnnnn.
 .n----m----n=nnnn____n.
 .n---m----wnnnwm_m___n.
 .n--m----nnww--m_____n.
..n--m-----w---wnnn___n.
.nn----------wwnnn___nn.
.n--------wnnn=nn__m_n..
.n-----wwnnnwwwn____nn.
.n--wnnnnnwwwwnn____n..
.nnnnnwwwwwwwnn____nn.
.WWwwwwwwwwnnn_____n..
..Wwwwwwwwnn_____nnn.
 .WWwwwwwwn__m_nnn...
 ..WWwwwwnn___nn...
  .@Wwwwnn___nn..
   .Wwwnn____n..
   .WWwn___nnn.
   ..Wwn__nn...
    .WWn=nn..
    ..nWWWWn.
     ..nnn@..
      .....
ENDMAP
```

I took out drac stormcallers after remembering what their abilities are, since they're really not very lightning themed monsters in terms of what they actually do, it's more that their name is evocative of lightning. Spriggan air mages are a more reasonable choice, since they're native to Depths, have a lightning theme, just with a different defensive profile.

Some other things you could consider is making a variant of this for Zot but using elec golems instead of Titans and possible blizzard demons instead of spriggan air mages. Zot tends to have a few cross-branch things so you could sort of decide if other things like spark wasps need replacing based on what you'd like to see in the vault.

Finally, this would be a cool vault to have some chance of placing one or two unrands like the arc blade, storm bow, amulet of the air, lightning scales, or maybe other less directly related things like: glaive of the guard, crown of Dyrovepreva, cloak of starlight, maxwell's etheric cage, or robe of clouds. Picking just two or three that you might like to see in the vault and making the chance for each not too high would work pretty well.

### Vault rework
In terms of the vault itself, I think the layout has a mix of ideas that don't work well. It has a side chamber with a runed door and easy access without flight and a hoard of monsters and no reword. The apparent idea is that you'll fight all those monsters so you get some slightly favorable place to dig into the larger chamber. Then the other entrances are supposed to be hard to get to since they require deep water traversal. But this is a pretty false choice since it's generally not hard for players at this depth to have flight (the ring is common and many flight potions tend to be available).

It would be good to pick one major layout idea for the vault and stick with that one. If relatively difficult access requiring flight is the theme you like most, try to make a layout that requires progressive/repeated access with flight using some kind of spiral or switchback layout where you fight progressively harder pockets of monsters, culminating in a hard fight in the final chamber. Maybe the ghosts would only be seen at the end chamber (with a second runed door), or you could put ghosts along the way (even one in each pocket of land) and give the pockets  each their own runed door and glass. Whether you allow digging inside this kind of vault is up to you, but of course you need to not allow the ghosts to be dug out by e.g. a random monster with a dig wand. Adding `no_item_gen` and the already added `no_tele_into` can help you there if you use stone on the outside of the vault.

Alternately if you really want two entrances and flight access isn't what you most care about, it's better to simply have the vault be fully connected so monsters from the two areas can co-mingle. The two paths can just provide somewhat different layouts and monster distributions to choose from, but ultimately the player will have to deal with the whole vault.

Muffindrake:
I'll think on the vault for a while, thanks for your feedback.

PR #791, opened 2018-06-19
Webtiles: Use X-Forwarded-Proto header for websocket protocol

floraline:
If the webtiles server is running behind a proxy server which terminates SSL connections, and http is used between the proxy and webtiles, the X-Forwarded-Proto header can be set by the proxy to indicate to the webtiles server that the client is using a secure connection. This allows webtiles to generate the correct protocol wss:// even it sees http.

1. Client connects to proxy server with https
2. Proxy connects to webtiles with http
3. Webtiles thinks the client wants to use insecure websocket
4. Client will refuse the websocket

This PR fixes this issue and allows the proxy to inform webtiles of the protocol that is really being used by the client: https.

### DIFF
--- a/crawl-ref/source/webserver/server.py
+++ b/crawl-ref/source/webserver/server.py
@@ -18,10 +18,7 @@ import userdb
 class MainHandler(tornado.web.RequestHandler):
     def get(self):
         host = self.request.host
-        proto = self.request.headers.get_list('x-forwarded-proto')
-        if len(proto) > 0 : proto = proto[0]
-        else: proto = None
-        if self.request.protocol == "https" or proto == "https":
+        if self.request.protocol == "https" or self.request.headers.get("x-forwarded-proto") == "https":
             protocol = "wss://"
         else:
             protocol = "ws://"
--- a/crawl-ref/source/webserver/server.py
+++ b/crawl-ref/source/webserver/server.py
@@ -18,7 +18,10 @@ import userdb
 class MainHandler(tornado.web.RequestHandler):
     def get(self):
         host = self.request.host
-        if self.request.protocol == "https":
+        proto = self.request.headers.get_list('x-forwarded-proto')
+        if len(proto) > 0 : proto = proto[0]
+        else: proto = None
+        if self.request.protocol == "https" or proto == "https":
             protocol = "wss://"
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
This is solved a little differently in webtiles-changes (the python3 branch I use): https://github.com/crawl/crawl/blob/webtiles-changes/crawl-ref/source/webserver/static/scripts/comm.js#L226

crawl-ref/source/webserver/server.py
```diff
@@ -18,7 +18,10 @@
 class MainHandler(tornado.web.RequestHandler):
     def get(self):
         host = self.request.host
-        if self.request.protocol == "https":
+        proto = self.request.headers.get_list('x-forwarded-proto')
+        if len(proto) > 0 : proto = proto[0]
```
alexjurkiewicz: nit: the space after `:` is not like existing code style

crawl-ref/source/webserver/server.py
```diff
@@ -18,7 +18,10 @@
 class MainHandler(tornado.web.RequestHandler):
     def get(self):
         host = self.request.host
-        if self.request.protocol == "https":
+        proto = self.request.headers.get_list('x-forwarded-proto')
```
alexjurkiewicz: Since you only care about the latest value of this header if it is present, you can simplify the code to:

```py
if self.request.protocol == "https" or self.request.headers.get("x-forwarded-proto", "http") == "https":
```

floraline: I couldn't figure out how to do that so simple like headers.get()! The documentation described it as a "dictionary-like object" and get_list() was the only getter they told me about.

crawl-ref/source/webserver/server.py
floraline: Whoops, this is my day job sneaking through

floraline:
I don't need this anymore, but this is still a useful change for anyone who wants to put webtiles behind an ssl-terminating proxy. Without this PR, or something like it, webtiles will not work properly on a server configured in that way.

It looks like `webtiles-changes` side-steps this issue, I don't know how easy it would be to replicate that here.

PR #792, opened 2018-06-22
New species: Oni

pubby:
This is an alternative to #772 (my ogre magic thing) that puts the gimmick on a new species called Oni rather than on Og/HO. Trying to put it on an existing species worked fine, but the aptitudes and flavor was not ideal and so this was worth trying.


pubby:
Postin the final version.

cc @gammafunk and friends

PR #793, opened 2018-06-22
Smith god: Igni Ipthes

pubby:
This is an overhaul of an idea dpeg and I worked on years ago.

The gist is:

- God wants you to enchant your gear
- If you enchant your gear, you can craft artefacts out of them (subject to limitations)

Besides that, there's a few active abilities which keep the game interesting in-between:
-  Use fire fortress to disarm enemies, but also to damage and control movement
 - Use brand weapon for some old school crusader action
 - Use scrap to round-out bad RNG.

The code seems to work, so I'm posting this for playtesting.


pubby:
Closing this. The X-crawl folks and hellmonk are much better to work with.

Thanks again Alex for hosting it.

PR #797, opened 2018-06-23
Show Enchanter as recommended for Formicid

Muffindrake:
+2 Hexes, +0 Short Blades and +3 Stealth, as well as ability to easily branch out to other schools of magic given rather neutral aptitudes easily qualifies this background for this race.


ebering:
Aptitudes alone aren’t enough to recommend a background, we want the recommendations to guide a new player into play styles that match racial innate abilities. An En won’t take advantage of the four strong arms. It’s ok for some non-recommended combos to be good, take SpBe for an example.

PR #799, opened 2018-06-24
Bugfix:  Qazlal cloud immunity is taken into account in messages.

jnooner:


jnooner:
Notes to self:

[21:35] <+amalloy> also i don't like the idea of "feeling that X is unaffected". it's like feeling yourself continuing to weigh as much as you did a moment ago
[21:35] <+amalloy> i'd rather just not print any message when nothing happens, but i guess there should be some signal to the player about what kind of scarf (or whatever) it is
[21:35] <+MarvinPA> i guess it could maybe just be that there's no message at all if you do have the qazlal passive?
[21:35] <+MarvinPA> or yeah, hmm
[21:36] <+amalloy> if you already have the immunity, it could check whether it was already identified, and if not print "It is a scarf of cloud immunity." or something
[21:37] <+amalloy> and otherwise print nothing
[21:38] <+MarvinPA> Alcon: also for any changes like this, it's no problem to rewrite git history when you're making a pull request (so that it's just one commit rather than multiple commits with minor fixes)
[21:38] <+MarvinPA> if you're familiar with how to do that in git
[21:39] <+MarvinPA> i should absolutely not still be awake though, so i can take another look tomorrow probably!

PR #800, opened 2018-06-24
Add option to disable gzip in webtiles

floraline:
This change allows for configuring tornado to disable gzip compression.

### DIFF
--- a/crawl-ref/source/webserver/server.py
+++ b/crawl-ref/source/webserver/server.py
@@ -121,7 +121,7 @@ def bind_server():
             (r"/", MainHandler),
             (r"/socket", CrawlWebSocket),
             (r"/gamedata/(.*)/(.*)", GameDataHandler)
-            ], gzip=use_gzip, **settings)
+            ], gzip=getattr(config,"use_gzip",True), **settings)
 
     kwargs = {}
     if http_connection_timeout is not None:
--- a/crawl-ref/source/webserver/server.py
+++ b/crawl-ref/source/webserver/server.py
@@ -121,7 +121,7 @@ def bind_server():
             (r"/", MainHandler),
             (r"/socket", CrawlWebSocket),
             (r"/gamedata/(.*)/(.*)", GameDataHandler)
-            ], gzip=True, **settings)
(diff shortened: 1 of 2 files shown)


floraline:
Just so server operators have the option. I personally would prefer to disable gzip to avoid SSL vulnerabilities, and to use less CPU in exchange for more bandwidth.

alexjurkiewicz:
I think http compression should be either always on or always off. I'm in
favour of leaving compression enabled, but if people think a breach attack
is likely, we should be disabling it globally.

floraline:
I agree that it should be consistent among official servers, but this should still be an option for private servers which may not necessarily use the same setting.

crawl-ref/source/webserver/server.py
```diff
@@ -121,7 +121,7 @@ def bind_server():
             (r"/", MainHandler),
             (r"/socket", CrawlWebSocket),
             (r"/gamedata/(.*)/(.*)", GameDataHandler)
-            ], gzip=True, **settings)
+            ], gzip=use_gzip, **settings)
```
rawlins: This seems reasonable to me to have this as an option (though you may be underestimating how big some of the map json is), but as-is it'll run into the usual problems that are involved in adding a new value to `config.py`, because servers won't get the `config.py` part of this change. Should use `getattr(config, "use_gzip", True)` or something similar, cf. 84a5ed292d.

Ozaq:
> Just so server operators have the option. I personally would prefer to disable gzip to avoid SSL vulnerabilities, and to use less CPU in exchange for more bandwidth.
> 
> I think I was also having benign Tornado errors reported constantly in my logs when using gzip over SSL.

@floraline your comment made me very curious about what vulnerability you are referring to. 5 min of google only turned up BREACH (http://www.breachattack.com) I would be grateful for any input that helps me understand your comment. Cheers!

floraline:
I think what you found there is all I know on the subject, that's one of the things I found when researching this option so I figured I'd bring it up. I was really more concerned with the CPU/bandwidth trade-off.

PR #803, opened 2018-06-27
Adding a variety of new arrival vaults

psywombats:
First time vault submission, edits/feedback welcome if anything looks nonstandard. I've been working on a Tiled (.tmx) utility to make generation a little more visual and am hoping to submit more vaults for other areas in the future.

### simple:
 - psy_arrival_basics - two exits and kite route
 - psy_arrival_conway - features a plant glider pattern
 - psy_arrival_dodecagon - floating star, can generate with 1 or 12 sides open to leave
 - psy_arrival_heroes - statue hall
 - psy_arrival_radial - two variable maps with radial symmetry
 - psy_arrival_whirlpool - shallow water paths

### small:
 - psy_arrival_batcave - a cave, with bats in it... batcave
 - psy_arrival_fountain - grove with a spring, half blocked off by plants
 - psy_arrival_solomon - variable map based on the seal of Solomon design, can generate as transparent or three separate rooms
 - psy_arrival_spokes - open 3-fork vault, some stuff in middle as it's vulnerable
 - psy_arrival_statuary - statues/fountains on a checkered floor design
 - psy_arrival_terrarium - some closed areas with water/plants

### large:
 - psy_arrival_wood_between_worlds - low-weight greenery vault with some id'd tele traps

### DIFF
--- a/crawl-ref/source/dat/des/arrival/large.des
+++ b/crawl-ref/source/dat/des/arrival/large.des
@@ -1013,3 +1013,40 @@ xttttwwwwwwwwwwwwwwwwwwW
 xtwwwwwwwwwwwwwwwwwwwwwW
 xWWWWWWWWWWWWWWWWWWWWWWW
 ENDMAP
+
+################################################################################
+NAME:       psy_arrival_wood_between_worlds
+TAGS:       arrival no_monster_gen no_trap_gen no_pool_fixup
+WEIGHT:     3
+ORIENT:     float
+FTILE:      t.qG^P{ = floor_grass
+KMONS:      P = plant / bush
+KFEAT:      ^ = known teleport trap
+KFEAT:      p = W
+KMONS:      p = plant
+NSUBST:     p = 10:p / *:W, q = 10:P / *:.
+MAP
+ xxxxxxxxx+++xxxxxxxxx 
+xxtttttttx...xtttttttxx
+xttpW.qttx...xttq.Wpttx
+xtGWW^...........^WWGtx
+xttpW.qtqqq.qqqtq.Wpttx
+xxttttttttq.qttttttttxx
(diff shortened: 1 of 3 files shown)


ebering:
These were put in a PR shortly before the 0.22 release, I'm sorry we didn't get to them then.

I'll try to review them this week.

At a cursory reading they look good. I'm a bit concerned that bat_cave places monsters. Some of our older arrival vaults do this, but it should be avoided for new vaults. If you think you can suggest the theme with bat corpses that could work, or you could add a small amount of loot and make it a regular early D vault.

ebering:
I've merged some of these, with some tweaks. The tweaks were setting COLOUR (or removing COLOUR), and fixing up whitespace. When using custom tiles, it's nice to set colour to match.

Trailing whitespace is against crawl code style, this can be checked for and fixed automatically with util/checkwhite, something to use if you submit future vaults.

I did not merge the following vaults:
* psy_arrival_batcave As I said in my previous comment, we want to move away from arrival vault monsters.
* psy_arrival_solomon As a symbol this runs very close to the "no ascii art" rule. We have some more abstract symbols in the arrival vaults, but don't use ones with broader meaning.

Thanks for the submission!

ebering:
P.S. How would you like to be listed in CREDITS.txt?

psywombats:
@ebering Thanks for the feedback + merge, you patched things up before I could get home to remove the batty one haha

If arrival vaults merit credit mention then I wouldn't mind. Will hopefully be adding more soon. If there's a better channel to submit vaults just let me know. Thanks!

ebering:
@psywombats Yes, it does merit credit mention. My question was what name would you like to be mentioned as?

PRs are a good avenue for vault submission, though it's helpful for them to each be in a separate commit, and we'll squash when merging into master.

PR #804, opened 2018-06-28
Fix item ego cloud immunity message with Qazlal (0011500)

shieldal:
Made it so "You feel vulnerable to the effects of clouds." and "You feel immune to the effects of clouds" messages from unequiping/equiping cloud immunity items don't appear when you have Qazlal cloud immunity.

Mantis Issue: https://crawl.develz.org/mantis/view.php?id=11500


rawlins:
Did you mean to close #816? This fix is still relevant I think, if you reopen I'll merge it.

PR #806, opened 2018-06-29
Fix zot traps casting friendly malign gateways (0011042)

shieldal:
Previously zot traps would cast a friendly malign gateway when a non-player triggered the trap. This has been changed to always cast a hostile malign gateway.

Mantis issue: https://crawl.develz.org/mantis/view.php?id=11042


shieldal:
You can credit me as Alexandra Shields.

PR #808, opened 2018-07-01
Attempt God conversions on sight.

ebering:
Currently three gods change monster attitudes on sight: Beogh, Gozag
(under bribe), and Jiyva.

When a player sees a monster, Gozag and Jiyva have already decided
whether they will convert the monster, but prior to this patch the
conversion did not happen until the player took an action with the
monster in LOS. This led to an unfortunate bug in the following
situation: if a bribed monster walked into LOS of a player bribing a
branch with Gozag no messaging would be displayed and the player could
attack the monster. This would have the effect of betraying the bribe
and turn any bribed monsters in LOS hostile for what appeared to be no
reason to the player. (The problem is less severe under Jiyva since not
attacking slimes is part of the conduct, but it does cause the
occasional 'hostile' slime to appear.)

Beogh's conversion behavior is different, and requires the monster be
awake, clear-headed, and targeting the player. As a result this check
needs to happen every turn the orc is reacting to the player (Beogh sets
a flag to make only one conversion attempt per-orc).

This patch makes conversion attempts happen in both relevant code-paths
using a do_conversions() function. While this causes some minor
execution duplication, each conversion function handles being called
multiple times on the same monster correctly, and abstracting the
conversion attempts to attitude-change make it easier to keep track of
them for the future (however unlikely more ally religions are).

I'm fixing the bug this way and not by changing how gozag_break_bribe
operates to avoid the mini-game of "I won't attack this monster that
just walked into my LOS in case I'm about to bribe it". The Beogh
conversion rules should probably be re-thought as their current state
has a bit of this mini-game.


ebering:
I've been thinking about this a bit more. In particular, the XXX this commit leaves suggesting that Beogh conversion should be done on sight and without the complicated extra checks. After further reflection I think that all god conversions should be a passive that occurs as soon as the player spots the monster. An update force-push will land soon unless someone favors the conservative approach.

PR #821, opened 2018-07-17
Replace outdated commands in FAQ and hints

Pekkekke:
Multiple instances of 'p'raying still existed in FAQ and hints. These have been replaced with references to using > to pray rather than 'p'.


ebering:
Fixes bug https://crawl.develz.org/mantis/view.php?id=11512

PR #827, opened 2018-07-24
Change deep water to shallow in aerie_lemuel (11554)

ebering:
This could create teleport or flight closets otherwise.


ebering:
@rawlins I replied to gammafunk that I'd like to think more about rescuing this vault, but I don't have the time to do so in the foreseeable future. So this should probably be closed and the vault removed.

PR #828, opened 2018-07-28
Fixup artp tweaks and revise some unrand blades.

ebering:
SHIELDING (giving SH) is now a possible Artefact property, but it was not possible to assign this in art-data.txt, nor was it possible to add it using &t in wizard mode. Fix that first.

Then, three unrandart blade revisions based on irc conversations with mikee and gammafunk:

- Give Morg 1.5 * Necromancy skill for pain damage, as well as Harm and *Drain. The Harm part was met with some controversy but I think it poses an interesting choice, as elaborated on in the commit message. Another possibility would be to make it a necro enhancer, though that is less interesting. Remove the Int+5 and MR+.

- Make the Spriggan's knife always stab at Tier 1, remove Dex+4 and MR+. This makes it very very good at hitting things that can be stabbed, but still bad otherwise in melee.

- Make the Singing Sword not depend on tension, instead give the damaging noise effect on hits (with a chance). Simplify and buff the damage formula.

### DIFF
--- a/crawl-ref/source/actor.cc
+++ b/crawl-ref/source/actor.cc
@@ -286,7 +286,10 @@ bool actor::reflection(bool calc_unid, bool items) const
 
 bool actor::extra_harm(bool calc_unid, bool items) const
 {
-    return items && wearing(EQ_AMULET, AMU_HARM, calc_unid);
+    const item_def *wp = primary_weapon();
+    return items &&
+           (wearing(EQ_AMULET, AMU_HARM, calc_unid)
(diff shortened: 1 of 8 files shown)


ebering:
gammafunk brought in a change to the singing sword in 2a92f5cd5cf4f48e6d11ec406b8c9fa68a8b77ec after refactoring los spell code.

I'm conflicted on Morg (theres a different plan for it that's been floated). The remaining commits have been picked and merged, so I'm closing this pr.

PR #829, opened 2018-07-30
Fix the spell targeter for Gell's Gravitas

NormalPerson7:
This is inspired by the Tavern post made by Doesnt outlining the targeter's problems, which can be found here: https://crawl.develz.org/tavern/viewtopic.php?f=17&t=25497&p=329839&hilit=gell%27s#p329839

Previously, the formula used to calculate the targeted area was
completely different to that which the spell itself used, resulting in
the targeter blatantly lying to the player about the tiles which could
and could not be affected.

The current behaviour of Gell's Gravitas is entirely predictable based
on spell power, with no randomness involved, therefore the new targeter
is entirely in the darker yellow colour to indicate the certainty. The
power breakpoints are:
\>0 power: at radius 1 from the centre, the monster is pulled 1 tile;
\>0 power: at radius 2, the monster is pulled 1 tile;
\>60 power: at radius 2, the monster is pulled 2 tiles; and
\>80 power: at radius 3, the monster is pulled 1 tile.

The previous spell targeter suggested that, at low power levels,
monsters at radius 2 could never be affected, and that at exceptionally
high power levels, targets at radius 4 could be affected, both of which
are incorrect. Hence, this commit changes the targeter to reflect the
fact that monsters at radius 2 are always affected and monsters at
radius 3 are always affected when, and only when, above 80 power.

An argument could be made for having the targeted tile (the centre of
the "gravitational reorientation") not appear coloured, given that this
tile is not affected, however this would require significantly more
work.

This commit removes the gravitas_range function, which was previously
only used for making the targeter, because it is no longer required.
Additionally, this commit removes the player's tile being highlighted,
given the player is never affected.

Note: the new targeter does not change at 60 power, because the radius that is affected does not change. An argument could be made for having tiles at radius 2 be pale yellow to indicate that enemies are not dragged right to the centre when below 60 piety (and similarly for tiles at radius 3 above 80 piety). Please let me know if you would prefer this behaviour.

### DIFF
--- a/crawl-ref/source/spl-cast.cc
+++ b/crawl-ref/source/spl-cast.cc
@@ -1172,8 +1172,8 @@ static unique_ptr<targeter> _spell_targeter(spell_type spell, int pow,
                                              range);
     case SPELL_GRAVITAS:
         return make_unique<targeter_smite>(&you, range,
-                                           pow >= 80 ? 3 : 2,
-                                           pow >= 80 ? 3 : 2,
+                                           gravitas_range(pow),
+                                           gravitas_range(pow),
                                            false,
                                            [](const coord_def& p) -> bool {
                                               return you.pos() != p; });
--- a/crawl-ref/source/spl-cast.cc
+++ b/crawl-ref/source/spl-cast.cc
@@ -1172,8 +1172,11 @@ static unique_ptr<targeter> _spell_targeter(spell_type spell, int pow,
                                              range);
     case SPELL_GRAVITAS:
         return make_unique<targeter_smite>(&you, range,
-                                           gravitas_range(pow, 2),
-                                           gravitas_range(pow));
+                                           pow >= 80 ? 3 : 2,
+                                           pow >= 80 ? 3 : 2,
+                                           false,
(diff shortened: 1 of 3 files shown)


rawlins:
Thanks for the PR, this is a very good fix (and helpful towards figuring out how to deal with gell's in the long run, something on my agenda).

My preferred way to do this would be to, instead of removing the range function, unify range calculations between `fatal_attraction` and the targeter call in such a function so that there's no code duplication (which is probably what led to this issue in the first place). It's been a while since this PR so feel free to do this if you want / have time, and I'll merge it, but otherwise I can do it.

NormalPerson7:
I've pushed another commit, but I don't think it really properly addresses the code duplication issue. I'll see what you think.

rawlins:
I guess you mean because `strength` is still calculated separately, but I think this is still better and I don't see a very clean way around that, and the two could be independent anyways. (Though right now strength could in principle be calculated based just on `gravitas_range` and `range`).

I think later I'd like to buff this spell a bit and possibly tweak the breakpoint locations so I can look into it more when I do that, I can merge this as-is after travis checks finish.

PR #838, opened 2018-08-13
New unrand: Forest's Gift (+MP on block)

alexjurkiewicz:


alexjurkiewicz:
Sorry. The work is based on this thread:
https://crawl.develz.org/tavern/viewtopic.php?f=8&t=25639

Some discussion in ##crawl-dev was that this initial version was quite weak. The item would be reasonably niche, used by warrior-mages and gspirit-wearing fighters. The former don't use a lot of magic during a fight so it's not actually so useful, the latter do but would probably prefer a shield or large shield.

I think the next step is to make the item stronger, by increasing the MP given and upsizing to a normal shield.

I was also considering the on hit effect to be several turns of boosted MP regen, and giving the item gspirit (because it's hard to see the item being used unless you find it and the amulet right now).

alexjurkiewicz:
This has been reworked. Commit message:

> A large shield with guardian spirit and MP-on-block.
> Per-block, the amount of MP regenerated scales from ~1.1 to ~3.8 based
> on XL. As gspirit is a relatively desirable effect the MP regeneration
> has been kept low and the shield has no enchantment.

Needs tiles. Two other things:

1. I'm on the fence about whether to use a shield or large shield. I picked large shield just because there's one large shield unrand and two shields.
2. The balance of the item is pretty unknown. I tried to err on the side of underpowered (as my past PRs have more often been accused of being buffs than nerfs), but that might be a mistake for an unrand.

alexjurkiewicz:
Closing. I'm not happy with the lore here, but I think the mechanical idea is worth pursuing.

PR #842, opened 2018-08-18
Order background screen by initial difficulty

alexjurkiewicz:
The current ordering groups backgrounds thematically, which is good for
helping players find related roles, but doesn't give a lot of guidance
for new players.

As was done for the species selection screen, divide the backgrounds
into three groups, based on difficulty for new players to learn.

The order of backgrounds in each column can probably be improved. At the
moment it's sort of random.

### DIFF
--- a/crawl-ref/source/newgame.cc
+++ b/crawl-ref/source/newgame.cc
@@ -1158,33 +1158,50 @@ void job_group::attach(const newgame_def& ng, const newgame_def& defaults,
 static job_group jobs_order[] =
 {
     {
-        "Warrior",
-        coord_def(0, 0), 15,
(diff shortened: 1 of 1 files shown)


ebering:
I'm a strong negative on this. In thinking about the various crawl rating proposals, it's clear that the function `difficulty : species x backgrounds -> N` does not have a product decomposition in terms of functions `f: species -> N` and `g: backgrounds -> N`. It would be a bad idea to suggest that it does.

I think that the recommended backgrounds system works better for indicating the partial functions `difficulty(sp, - ) : backgrounds -> N`. It's low on granularity, which is a feature to the extent that the new game menu doesn't recommend things that are strong but offbeat (like SpBe), since recommendations are aimed at new or inexperienced players.

ebering:
Closing this. Other devs haven't expressed interest and I listed my strong negative reasons in a previous comment.

PR #845, opened 2018-08-29
Allow Trog to berserk all species

alexjurkiewicz:
Currently formicids and the undead cannot go berserk (except Full+
vampires) from any means. This commit allows Trog's divine rage to
stir anger into the heart of even these species.

Most god abilities override species restrictions, for example Trog's
Hand, Bend Space, and Potion Petition working where the player might not
expect them to. There are cases where god abilities do not override
species restrictions, for example Jiyva not mutation the undead, and
Nemelex cards respecting various species restrictions.

In future, it might be worth looking at these two approaches to the
species restriction vs divine ability priority and deciding if there
should be more consistency in how they are resolved (either changing all
cases to one or the other approach, or defining an inutitive test to
tell whether an ability should override restrictions or not).

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -941,7 +941,6 @@ ability_type fixup_ability(ability_type ability)
         return ability;
 
     case ABIL_EVOKE_BERSERK:
-    case ABIL_TROG_BERSERK:
         if (you.is_lifeless_undead(false)
             || you.species == SP_FORMICID)
         {
@@ -1294,12 +1293,11 @@ static bool _check_ability_possible(const ability_def& abil, bool quiet = false)
         }
     }
 
-    if ((abil.ability == ABIL_EVOKE_BERSERK
-         || abil.ability == ABIL_TROG_BERSERK)
-        && !you.can_go_berserk(true, false, quiet))
-    {
-        return false;
-    }
+    if (abil.ability == ABIL_EVOKE_BERSERK)
+        return you.can_go_berserk(true, false, quiet);
+
+    if (abil.ability == ABIL_TROG_BERSERK)
+        return you.can_go_berserk(true, false, quiet, nullptr, true, true);
 
     if ((abil.ability == ABIL_EVOKE_FLIGHT
          || abil.ability == ABIL_TRAN_BAT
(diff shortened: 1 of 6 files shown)


alexjurkiewicz:
Some more comments (copied from Tavern):

Mechanically speaking, I think the game would be better with a cleaner definition of how divine abilities and species restrictions work.

In general, I think god abilities should override species restrictions. At the moment, some god abilities do and some do not.

My proposed rule would be simply: "divine abilities ignore species restrictions, divine gifts do not".

With this rule, Potion Petition should work for all species, as should (perhaps more controversially) Trog's Berserk, and (definitely controversially) Jiyva would mutate the undead. Nemelex, Oka, and Trog gifts would allow follow existing restrictions as they are "gifts".

(nb, Jiyva mutations are coded as "gifts", but I don't think they should be thought of as the same type of thing as item gifts)

alexjurkiewicz:
And from IRC:

04:25 < alexjurkiewicz> I think, in general, it's more fun/interesting to have
                        divine abilities bypass species limitations. It's just
                        a matter of if the bypass breaks the game somehow
04:26 < alexjurkiewicz> The other big one raised was Jiyva's mutation gifts. I
                        originally was thinking it would be nonsensical to
                        mutation the undead. But after thinking about it, I'm
                        coming around to the idea being OK. If Jiyva is the
                        only way undead can be mutated, that's sort of cool.
                        And I don't think it makes undead species "more like"
                        living ones
04:26 < alexjurkiewicz> Gn/Transfer Knowledge would be super OP though, so I
                        think from a balance perspective it has to be restricted

rawlins:
For the specific case of berserk, I'm not really sure I see how it improves things (in terms of fun, balance, or really even consistency -- given how complex the overall situation is) to let fo/mu berserk, but I don't feel strongly enough to close this PR.

Echoing some of the ##crawl-dev discussion, I think this is better handled on a case-by-case basis. I don't really think a blanket rule or principle about divine abilities vs. species is necessary, as long as the information about what abilities will work is not spoilery.

ebering:
I find arguments from consistency across multiple species and deities unconvincing. Adopting a principle like you suggest would subsequently bind our hands (e.g. its implications for Gn/Ash), so I think we'll stick to the case by case approach.

When it comes to Trog berserk on Fo/undead, I don't think Trog needs a buff for these species, so I'm going to close this.

PR #847, opened 2018-09-10
Upgrade orb run spawns

alexjurkiewicz:
The orb run is relatively safe for players, but the difficulty is wildly
variable across multiple games.

Part of the reason for this is that the spawn list is so varied. An orb
run spawn might be a Pandemonium lord or a crimson imp. This change aims
to make the spawn list more consistent, and more dangerous.

Mid-tier demons used to make up over a third of orb run spawns. Cut that
down to 10% and give the excess spawn chance to orb guardians and
greater demons. Other numbers are slightly shuffled to rounder figures.

```
Monster          Old   New
---------------+-----+----
Pan lord       |  9% | 10%
Tier 1/2 Demon | 12% | 20%
Tier 3/4 Demon | 34% | 10%
Tier 5 Demon   |  4% |  0%
Orb Guardian   | 13% | 30%
Pan spawn      | 27% | 30%
```

### DIFF
--- a/crawl-ref/source/mon-place.cc
+++ b/crawl-ref/source/mon-place.cc
@@ -303,7 +303,7 @@ void spawn_random_monsters()
     }
 
     if (player_on_orb_run())
-        rate = have_passive(passive_t::slow_orb_run) ? 16 : 8;
+        rate = have_passive(passive_t::slow_orb_run) ? 32 : 16;
 
     if (player_in_branch(BRANCH_ABYSS))
     {
@@ -324,7 +324,7 @@ void spawn_random_monsters()
              rate, env.turns_on_level);
 
         mgen_data mg(WANDERING_MONSTER);
-        mg.proximity = one_chance_in(3) ? PROX_CLOSE_TO_PLAYER : PROX_ANYWHERE;
+        mg.proximity = PROX_CLOSE_TO_PLAYER;
         mg.foe = MHITYOU;
         // Don't count orb run spawns in the xp_by_level dump
         mg.xp_tracking = XP_UNTRACKED;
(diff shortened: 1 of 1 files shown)


MarvinPA:
Making the spawn variation less wild definitely sounds good to me, but I'm not sure about having orb guardians be such a big chunk of the spawn table. The player will already have dealt with a pretty large number of them, and they're kinda boring.

I think probably another part of the high variation is that 2/3 of orb run spawns are just being placed randomly, so maybe it'd be better to make most (or all) orb run spawns be close to the player but reduce the spawn rate accordingly?

alexjurkiewicz:
I specifically like orb guardians because they aren't particularly dangerous solo but being fast and durable they force the player to spend precious time dealing with them, leading to more time on floors for other monsters to come me along.

Agree about the placement issue. I'll add a commit to remove all non-nearby spawns and reduce the rate commensurately.

alexjurkiewicz:
Squashed and updated. New spawn table:

```
    Monster          Old   New
    ---------------+-----+----
    Pan lord       |  9% | 10%
    Tier 1/2 Demon | 12% | 25%
    Tier 3/4 Demon | 34% | 10%
    Tier 5 Demon   |  4% |  0%
    Orb Guardian   | 13% | 15%
    Pan spawn      | 27% | 40%
```

What do you think?

MarvinPA:
At a glance that looks good to me in terms of the spawn table, I'm not so sure about increasing the spawnrate on top of that though. I'll try and take a proper look locally when I get a chance but I think it'd be better to keep the spawnrate the same-ish rather than trying to make the orb run too much harder.

alexjurkiewicz:
I'll leave any rate changes to you to make as desired, let me know if you have any other feedback

MarvinPA:
Merged with a slight reduction to the spawnrate, thanks for the PR!

PR #848, opened 2018-09-11
Reduce player paralysis maximum duration

alexjurkiewicz:
From 2-7 to 2-5 turns. Even with this change, paralysis is one of the
worst things to happen to the player. There are similar effects (sleep,
petrification) which have longer durations but aren't as tough to deal
with.


MarvinPA:
I think the current duration is pretty fine - the thread that prompted this was about as extreme an edge case as you can get. It might be more reasonable to address the weird combination of sleep and paralysis that came up there (possibly so that sleep combined with paralysis always just results in paralysis, no matter which order they were applied?)

alexjurkiewicz:
I've updated this PR to change behaviour of sleeping, as you suggested. Being paralysed or undergoing any petrification stage will wake up the player.

PR #849, opened 2018-09-14
Halve the max MP contribution from evocations

alexjurkiewicz:
About 1 MP every 4 levels of evocations.

Over time, fewer evocable abilities have required MP. The ones that
remain now are: blink, flight (1 MP), invisibility (2 MP), and the
unrand abilities fog and ratskin (3 MP).

A lot of characters train evocations for MP to use for other purposes.
Most notably, MP for guardian spirit or for DD Heal Wounds ability.
Arguably, this makes evocations a little too good, so reduce the amount
of MP one can theoretically get from the skill.

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -3988,7 +3988,7 @@ int get_real_mp(bool include_items)
 
     int spell_extra = spellcasting; // 100%
     int invoc_extra = you.skill(SK_INVOCATIONS, 1 * scale, true) / 2; // 50%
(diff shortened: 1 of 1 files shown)


ebering:
I'm going to close this, it's been open for some time without any devteam member moving on it. I'm a weak no, for the following considerations:

- Many gods permit training invocations, which gives MP, even when not all god abilities for every god use lots of MP. A much bigger change would be to only have spell casting give Max MP; but that has serious balance implications and thinking them through is out of scope for this PR thread.
- Guardian spirit amulets are a magical device so training a skill to make them better isn't unheard of.

PR #852, opened 2018-09-28
Fix a couple of wall jump inconsistencies (11687, 11690)

NormalPerson7:
Firstly, allow wall jumping via the ability whilst berserk. Previously, the ability menu simply would not open whilst berserk. (11687)

Secondly, disallow wall jumping while stuck in nets or webs - previously this was impossible via movement but the ability method was successful. (the second issue reported in 11690)

I am working on fixing the first issue in https://crawl.develz.org/mantis/view.php?id=11690, and may have a fix for that soon. For this issue, it seems logical that the player gets their attacks when landing in a web, so the only actual issue that needs fixing is the message order. This is more complicated than it looks though.

For now I thought I would post these two fixes, since these can be merged anyway. If there are any code style issues, feel free to post them and I can fix them. I have tested both of these commits and they seem to work fine for me.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -1623,6 +1623,42 @@ static bool _check_ability_possible(const ability_def& abil, bool quiet = false)
         }
         return true;
 
+    case ABIL_WU_JIAN_WALLJUMP:
+    {
+        // TODO: Add check for whether there is any valid landing spot
+        if (you.is_nervous())
+        {
+            if (!quiet)
+                mpr("You are too terrified to wall jump!");
+            return false;
+        }
+        if (you.attribute[ATTR_HELD])
+        {
+            if (!quiet)
+            {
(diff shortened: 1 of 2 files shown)


ebering:
The fix for 11687 looks ready to go. 

For the first part of 11690. I'm not a U.I. person so I don't know what's better: allow the option of wall jump and give an explanatory failure message or to simply disable it in the ability menu when the player is webbed. @aidanholm do you have an opinion here?

NormalPerson7:
The current behaviour gives a failure message as soon as the ability is selected; it does not ask for a target. However, I didn't add anything in particular to the _check_ability_possible function so the ability probably isn't greyed out in the a menu (I didn't check).

ebering:
It is not greyed out. I think it should be.

NormalPerson7:
I have just pushed an extra case for the wall jump ability to _check_ability_possible.

PR #856, opened 2018-10-04
Moved some actions from main.cc

ImmemorConsultrixContrarie:
Doing this "TODO: Move player movement code to its own file".
Transfered everything for move action plus close-door action.
And since I heard only "lump it altogether", there's the lump in the "movement.h/.cc"


ebering:
Thanks for the PR. In the future, if there's justification we would like it included in the commit message(s). I'll try to review this tonight, and can update the commit message then.

PR #859, opened 2018-10-08
Adjust minmay_gold_behind_fungi

biasface:
Replace fungus with toadstools so less time is spent punching fungi to get the gold.


ebering:
Thanks for the PR. Could you git commit --amend and put the justification from the PR into the commit message and then force push?

gammafunk:
Toadstools decay quickly (in 250 turns or so), so you'd want to place mostly fungus and then sprinkle in some toadstools randomly. This will generally mean that players find the vault with gaps in the fungi, but as far as the flavor of the loot vault goes, that will still be there.

PR #860, opened 2018-10-14
Remove references to loud noises breaking mesmerisation and fear

nlavsky:
Since this mechanic was removed in 74fa8348.


ebering:
Thanks for the patch!

PR #861, opened 2018-10-14
Draconian Color Rework/Addition

Spaninq:
The main idea is to synchronize draconians with their dragon brethren.  Comparing the two, we see
Draconian <-> Dragon
Red              |     Fire
White           |     Ice
Green           |     Swamp
Yellow          |    Acid
Grey             |    Iron
Black            |     Storm
Purple          |    Quicksilver
Pale              |     Steam
???                |     Shadow
???                |     Pearl
???                |     Bone
???                |     Golden

First, I'm going to suggest changing Black Draconian's color to Blue, to match how storm dragons are depicted.

Next, the new Black Draconian would be the Shadow Draconian equivalent.
Skills: [+2 Spellcasting (to+1)], +2 Necromancy
Resists: rN+, [maybe Stealth++, too?]
Breath: [Bolt of draining (that's what shadow dragons have), alternatively breath out a cloud of either negative energy or miasma that has damage nerfed, Spectral fog is probably too OP to consider]
In addition, Draconians that worship a good god when they reach XL 7 cannot become Black Draconians.

[Pearl (Lustrous? Shiny?)] Draconian 
Skills: [+2 Invocations (to +3)?]
Resists: rHoly (apparently it exists, although it's really rare), rN+
Breath: Essentially Holy Breath, but nerfed to a single Cloud of Blessed Fire, and probably damage nerfed
In addition, Draconians that worship an evil god when they reach XL 7 cannot become [Pearl] Draconians

Bone Draconian
Skills: [+2 Fighting (to+3), +2 Shields, -2 Spellcasting (to-3)]
Resists: Immediately Gains Rank 1 of Large Bone Plates Mutation (+4 SH).  [In addition, gains Rank 3 of Large Bone Plates Mutation (+8 SH) at XL 14.]
Breath: None
(The idea is not that they're skeletons, but Draconians whose scales have become exoskeletons)

I'm not going to suggest a Gold Draconian because that just sounds too OP.  
[Instead, I'm going to tentatively suggest a color that should have a much lower probability of spawning assuming that is possible [unless you worship Xom before XL7 :P].
Teal Draconian (a la Faerie Dragon Scales)
Skills: [None]
Resists: [rCorr] rMut (to make Xom more bearable), Clarity
Breath: a single Cloud of Seething Chaos (what could possibly go wrong?)
This particular color is more of an idea that might make CK more viable/fun and can easily be removed if the people wish.]

Anything in [] is up for debate, and number values may (probably) need tweaking.

Of course, these would also need monster variants as well (except Teal [unless you want to throw one or two in Mnoleg's Realm]).  The new version of Black Draconians would show up in dark/evil themed vaults, and [Pearl] Draconians would show up in holy themed vaults (like The Shining One's Fortress). It appears most draconians have roughly the same stats, so that shouldn't be too hard to copy/paste.


ebering:
This looks like its based for x-crawl, not mainline.

PR #864, opened 2018-10-16
Surround nagdon_arrival_double_spiral with walls

alexjurkiewicz:
If a ranged monster (say a dart slug) wanders into your LoS while you
are navigating this arrival vault (for example if the walls are replaced
with lava or deep water) it may get several free shots at you before you
can escape the vault. Prevent this by surrounding the entire vault with
rock walls and a door.


ebering:
Thanks for the patch. I ended up revising the vault a different way, removing the lava and using shallow water instead.

PR #867, opened 2018-10-21
Add a lighthouse-themed bazaar vault

kenranunderscore:
I just found this old content that I created and forgot about and now want to try adding it (again, and slighty modified). Below is the original description.
**Note:** I remember talking to one of the devs who wasn't fond of the ophan in the middle I think, so any suggestions as to what to replace it with are very welcome as well (maybe a TSO altar?)! Let me know what you think :)

This is an(other) attempt at straying away a bit from the classical bazaar vault looks by using some more irregular shapes. The vault `bazaar_kenran_lighthouse` features a circular hut built on an island containing 4 random shops. Its aim is to make the player think of a lighthouse in the sea.

The shoreline, as well as the transitions from shallow to deep water, and from deep water to open sea, respectively, vary slightly in each generated map.

At the moment, the centre of the lighthouse contains a trapped ophan (as to provide a source of light). Of course this is rather gimmicky, but I found it fun. It can easily be replaced by another shop or decorative interior though, should monster generation in this case be frowned upon.

Here are screenshots to show how it looks in tiles/console mode:
console
enter
full
interior

### DIFF
--- a/crawl-ref/source/dat/des/portals/bazaar.des
+++ b/crawl-ref/source/dat/des/portals/bazaar.des
@@ -1845,6 +1845,63 @@ x.......x     x.......x
 xxxxxxxxx     xxxxxxxxx
 ENDMAP
 
+NAME:   bazaar_kenran_lighthouse
+TAGS:   no_pool_fixup
+ORIENT: encompass
+WEIGHT: 1
+KMONS:  1 = ophan
+SUBST:  m =.WWW
+SUBST:  d = wwW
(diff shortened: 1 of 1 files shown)


ebering:
This isn't in the vault guideline documents (our fault), but the use of permarock is strongly discouraged outside of map boundaries.

I also don't like placing a monster for flavor for the light, but I do like the idea of having a light. I suggest using a statue and giving it a halo, though I'm not sure that can be done with a MONS specification.

PR #868, opened 2018-10-21
Display (Immune) instead of chance to defeat MR in monster descriptions when the player is intrinsically immune to a hex

NormalPerson7:
When a player is intrinsically immune to a particular hex, through
innate abilities, equipment, god abilities or transformations, make the
chance to defeat MR indicator next to the spell on the monster
description screen read "(Immune)" instead. It is misleading to imply
that these abilities can succeed on the player through printing the
chance to defeat MR, as was the case previously.

This applies to:
 - Paralysis and slowing as a formicid,
 - Confusion with clarity or under Zin's vitalisation,
 - Petrification as a gargoyle, in statue form or under Zin's
   vitalisation,
 - Sleep and fear as undead or nonliving, with clarity or while berserk,
 - Virulence when immune to poison,
 - Porkalator when unable to transform,
 - Teleport and blink as a formicid or with -Tele equipment, and
 - Mesmerisation with clarity or while berserk.

This commit creates a new member function in the player class,
immune_to_hex. This method also returns true for hex-like spells that do
not have an MR check, such as paralysis gaze and blink other, however
these spells do not have the MR indicator and as such also do not have
the immunity indicator.

(If the only reason the player cannot be affected by a hex is because
they have sufficient MR to resist it always, the MR indicator will still
read "(0%)", to maintain the distinction between MR and intrinsic
immunity.)

Please make me aware of any hex immunity cases that I have missed, or any code style issues you have, and I will aim to fix them.

### DIFF
--- a/crawl-ref/source/describe-spells.cc
+++ b/crawl-ref/source/describe-spells.cc
@@ -508,11 +508,20 @@ static void _describe_book(const spellbook_contents &book,
 #endif
             && (get_spell_flags(spell) & SPFLAG_MR_CHECK))
         {
-            int chance = hex_chance(spell, hd);
-            int ch_len = to_string(chance).length();
-            description.cprintf("%c - (%d%%) %s",
-                            spell_letter, chance,
-                            chop_string(spell_title(spell), 25-ch_len).c_str());
+            if (you.immune_to_hex(spell))
+            {
+                description.cprintf("%c - (Immune) %s",
+                                spell_letter,
+                                chop_string(spell_title(spell), 20).c_str());
+            }
+            else
+            {
+                int chance = hex_chance(spell, hd);
+                int ch_len = to_string(chance).length();
+                description.cprintf("%c - (%d%%) %s",
+                    spell_letter, chance,
+                    chop_string(spell_title(spell), 25-ch_len).c_str());
+            }
         }
         else
         {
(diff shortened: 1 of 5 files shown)


ebering:
Quick comment: for display consistency with other immunities it should probably be (immune) with a lowercase I.

ebering:
I think this is a good feature to add. However, when you press a letter corresponding to a spell, you get the spell's description followed by range and the chance to beat MR (for example this is from a Lich's petrify):

```Range : L------>
Chance to beat your MR: 11%```

This display should also indicate that the player is immune to the spell.

NormalPerson7:
Right, agreed, I'll do that. Not sure what's going on with the same message 4 times, mind.

NormalPerson7:
Right, agreed, I'll do that. Not sure what's going on with the same message 3 times, but never mind.

NormalPerson7:
Right, agreed, I'll do that. Not sure what's going on with the same message 3 times, but never mind.

NormalPerson7:
I've pushed a commit dealing with the spell descriptions, so the basic change should now be ready to merge.

I looked through l-you.cc regarding adding this immune_to_hex function to clua, and it doesn't look very practical, because there is currently no way to lookup a spell by its name. If there's a way around this / an easy way to add such a lookup function, then the actual function in l-you.cc shouldn't be too difficult.

PR #869, opened 2018-10-23
Announce the starting items and spells for wanderers (11725), and start wanderers with a spell if they have only one level 1 spell

NormalPerson7:
First commit:

Previously, players that did not look at their M screen would not be
aware of if they had spells in their spell library.

Therefore, print the player's inventory and spell library to the message
log at the start of the game. Also, take a note of the spell library to
include in the dump. This is done by expanding the note_wanderer_items
function and renaming it to note_wanderer_equipment (to be more
accurate).

This commit also adds support to print any spells the player begins with
memorised at the start, although this currently never happens.

Second commit:

If wanderers have more than one level 1 spell, they may not have enough
spell levels to memorise all of them, and so the player should get to
choose which spell to memorise.

If, however, wanderers have only 1 level 1 spell in their library, and
they have a spell level available, it makes sense for the player to have
that spell automatically memorised. Hence, in this case, memorise the
player's only level 1 spell.

Please feel free to give comments on either of these commits, or any comments on how the code could be improved.

### DIFF
--- a/crawl-ref/source/main.cc
+++ b/crawl-ref/source/main.cc
@@ -572,11 +572,10 @@ static void _wanderer_note_equipment()
                                  return item.name(DESC_A, false, true);
                              }, ", ", ", ", mem_fn(&item_def::defined));
 
-    // XXX: Currently wanderers don't start with any of their spells memorised,
-    // so this never triggers.
+    // Wanderers start with at most 1 spell memorised.
     const string spell_str =
         !you.spell_no ? "" :
-        "; and the following spells memorised: "
+        "; and the following spell memorised: "
         + comma_separated_fn(begin(you.spells), end(you.spells),
                              [] (const spell_type spell) -> string
                              {
--- a/crawl-ref/source/main.cc
+++ b/crawl-ref/source/main.cc
@@ -560,16 +560,53 @@ static void _show_commandline_options_help()
 #endif
 }
 
-static void _wanderer_note_items()
+// Announce to the message log and make a note of the player's starting items,
+// spells and spell library
+static void _wanderer_note_equipment()
 {
     const string equip_str =
-        you.your_name + " set off with: "
+        "the following items: "
         + comma_separated_fn(begin(you.inv), end(you.inv),
                              [] (const item_def &item) -> string
                              {
                                  return item.name(DESC_A, false, true);
                              }, ", ", ", ", mem_fn(&item_def::defined));
-    take_note(Note(NOTE_MESSAGE, 0, 0, equip_str));
+
+    // XXX: Currently wanderers don't start with any of their spells memorised,
+    // so this never triggers.
+    const string spell_str =
+        !you.spell_no ? "" :
+        "; and the following spells memorised: "
+        + comma_separated_fn(begin(you.spells), end(you.spells),
+                             [] (const spell_type spell) -> string
+                             {
+                                 return spell_title(spell);
+                             },
+                             ", ", ", ",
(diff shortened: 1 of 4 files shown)


crawl-ref/source/main.cc
```diff
@@ -560,16 +560,52 @@ static void _show_commandline_options_help()
...
                              }, ", ", ", ", mem_fn(&item_def::defined));
-    take_note(Note(NOTE_MESSAGE, 0, 0, equip_str));
+
+    // Wanderers start with at most 1 spell memorised.
+    const string spell_str =
+        !you.spell_no ? "" :
+        "; and the following spell memorised: "
+        + comma_separated_fn(begin(you.spells), end(you.spells),
+                             [] (const spell_type spell) -> string
+                             {
```
NormalPerson7: Perhaps this `comma_separated_fn` can be simplified to just `you.spells[0]` if wanderers start with at most 1 level 1 spell memorised? Should there also be an assert somewhere that you.spell_no = 0 or 1?

MarvinPA:
I haven't had time to look into it properly but it doesn't look like memorise_wanderer_spell is working as intended, I rolled up a few characters and they could still start with a single L1 spell without it being memorised.

NormalPerson7:
I tested it and it seemed to work for me, but in any case I'll do a bit more testing to check I didn't accidentally break something. It won't work if the player doesn't have any spell levels, perhaps that's what you were experiencing?

crawl-ref/source/ng-setup.cc
```diff
@@ -509,6 +509,9 @@ static void _setup_generic(const newgame_def& ng)
     // Must be after _give_basic_knowledge
     add_held_books_to_library();
 
+    if (you.char_class = JOB_WANDERER)
```
MarvinPA: This should be an "==" rather than an "=".

crawl-ref/source/ng-wanderer.cc
```diff
@@ -747,3 +747,24 @@ void create_wanderer()
...
 }
+
+void memorise_wanderer_spell()
+{
+    // If the player got only one level 1 spell, memorise it. Otherwise, let the
+    // player choose which spell(s) to memorise and don't memorise any.
+    auto const available_spells = get_sorted_spell_list(true, true);
+    if (available_spells.size())
+    {
+        int num_level_one_spells;
```
MarvinPA: It looks like the problem I was having with spells not memorising is this being uninitialised, initialising it at 0 makes it work as expected for me.

crawl-ref/source/ng-setup.cc
NormalPerson7: Oh whoops, always need a second pair of eyes :)

crawl-ref/source/ng-wanderer.cc
NormalPerson7: Hmm that's weird, does this vary between compilers?

rawlins: yes, it definitely can; and specificaly clang often behaves differently than gcc w.r.t. uninitialized variables.

NormalPerson7: That probably explains why mine worked fine then, thanks.

MarvinPA:
Merged, thanks! (with the minor fixes squashed in). I think there's probably room to improve the messaging - it's a bit of a wall of text when you also have a bunch of spells, but it looks like it's all working as intended and is definitely a good improvement.

PR #870, opened 2018-10-23
Fix HP+ faerie dragon scales causing HP-scaling transmutations to overscale the player's current HP (11728)

NormalPerson7:
When wearing faerie dragon scales with the HP artefact property and
entering or leaving a transformation that modifies HP, HP would be
scaled incorrectly. This was because:
 - On entering a transformation, equipment was melded after the
   transformation took effect, which meant the new max HP was used as
   the finishing value for the HP change from melding the scales
   (resulting in a much higher than expected final value).
 - On exiting a transformation, equipment was unmelded before the max HP
   was downscaled, which meant the old max HP was used as the starting
   value for the HP change from unmelding the scales (resulting in a
   much lower than expected final HP).

Hence, in transform(), meld equipment before changing the value of
you.form, and in untransform(), downscale HP from the transformation
ending before unmelding equipment.

### DIFF
--- a/crawl-ref/source/transform.cc
+++ b/crawl-ref/source/transform.cc
@@ -1729,12 +1729,12 @@ bool transform(int pow, transformation which_trans, bool involuntary,
     // Give the transformation message.
     mpr(get_form(which_trans)->transform_message(previous_trans));
 
+    _remove_equipment(rem_stuff);
+
     // Update your status.
     you.form = which_trans;
     you.set_duration(DUR_TRANSFORMATION, _transform_duration(which_trans, pow));
     update_player_symbol();
-
-    _remove_equipment(rem_stuff);
 
     you.props[TRANSFORM_POW_KEY] = pow;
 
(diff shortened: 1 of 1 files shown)


stenellad:
Can you share the reason for the PR? I can't re-create the bug.

NormalPerson7:
Ah, not properly tested... I guess this will require some more work then.

NormalPerson7:
I've just read some of the discussion about this from last night stenellad, and I'd just like to clarify that `calc_hp` and `get_real_hp` only deal with adjusting the max HP. The player's current HP is scaled through the call to `_extra_hp` for transforming, and in `untransform` itself when untransforming. However, `_remove_equipment` and `_unmeld_equipment` calls a series of item handling functions, the last of which, when the HP artefact property is being (un)melded, is `recalc_and_scale_hp`. This function, `recalc_and_scale_hp` then proceeds to use the new max HP for the (un)transformation for its scaling multiplier, which results in the player's HP being scaled incorrectly at this function.

The best way I've come up with so far is to add a property to one of the classes, that would indicate whether the game is in the process of transforming the player, and then `recalc_and_scale_hp` would be able to use the correct scaling multiplier. (I don't know in what class or property vector is best to put this property.)

stenellad:
It's more complicated than that. The entire thing needs to be rewritten, I think.

When you transform, the original call to meld the artefact during the transformation also calls recalc_and_scale_hp(), which then calls calc_hp(), which calls get_real_hp(true,false) (setting your max HP to your transformed maximum). Then it's called again during the transform's own calc_hp() call in _extra_hp. 

Essentially, you're scaled up twice. 

But the max HP for the transformation is also wrong, because it's calculated inside get_real_hp(true,false),  which considers equiped artefacts only when the first argument is true (resulting in HP growth based on the additional HP from the artefact).

It's a little difficult to write out and explain, I hope this wasn't too difficult to understand.

PR #871, opened 2018-10-24
MSVC2017 Changes

stenellad:
See #876 

<Close this>


crawl-ref/source/shopping.cc
```diff
@@ -2350,10 +2350,10 @@ void ShoppingList::display(bool view_only)
             }
 
             del_thing_at_index(index);
-            mtitle->quantity = list->size();
+            mtitle->quantity = ShoppingList::list->size();
```
rawlins: I think these should be `this->list` rather than `ShoppingList::list`. The problem here seems to be that this is inside a lambda that doesn't capture `list`, so it's dependent on how the compiler handles the fact that `this` is captured.

stenellad: Changes made

PR #872, opened 2018-10-24
Known traps

ebering:
This branch is an experimental pass at some trap changes. The purpose is to separate the "forward progress/tension creation" aspect of traps from the "tactical terrain" aspect. To this end, all traps are revealed, explore-based trap effects are introduced, alarm traps now always target the player, and a new "dispersal trap" which random blinks everything in LOS of the trap when triggered.


DenialAdams:
First of all, I'm very excited for these changes. I love that traps are being made more strategic, and dispersal traps seem really interesting.

I'm confused as to the state of teleport traps after these changes. Why can teleport traps not be triggered by exploration? You mentioned dispersal traps as a replacement for teleport traps, are teleport traps being removed altogether? It didn't look like it based on the commits, but perhaps I am missing something.

ebering:
Teleport as an explore effect isn't that exciting. It starts out with a chance to be as exciting as a shaft, but as the player explores the chance of ending in a safe area after a teleport increases.

The reference to dispersal traps replacing teleport traps is the result of this being edited by me a couple of times before posting it publicly. I had thought about replacing teleport traps with dispersal traps entirely, but decided against it.

Issue #874, opened 2018-10-25
HP Rewrite


stenellad:
#870 is more complicated than initially expected. 

[recalc_and_scale_hp()](https://github.com/crawl/crawl/blob/master/crawl-ref/source/player.cc#L2716)
[calc_hp()](https://github.com/crawl/crawl/blob/master/crawl-ref/source/player.cc#L3635)
[get_real_hp()](https://github.com/crawl/crawl/blob/master/crawl-ref/source/player.cc#L3928)

are problematic. In particular, calc_hp() is called twice during a transformation, if the user is wearing HP artefacts. This would be fine, but get_real_hp() inside of calc_hp() returns the transformed HP value both times. 

For example, 57/148 wearing +15 faerie dragon scales, casting dragon form. After the transform's [remove_equip](https://github.com/crawl/crawl/blob/master/crawl-ref/source/transform.cc#L1737) code, including a call to [recalc_and_scale_hp()](https://github.com/crawl/crawl/blob/1f986f14360838e08eb608ce4f4c2868ef75de54/crawl-ref/source/player-equip.cc#L45) due to the scales, the character is at 77/148 (since recalc_and_scale_hp() thinks the max HP is already 199 due to the call to get_real_hp(true,false)). Then the second call within [_extra_hp](https://github.com/crawl/crawl/blob/master/crawl-ref/source/transform.cc#L2039) sets us to 119/199. 

The issue stems from unintelligent calls to get_real_hp(), as well as unintelligent code within get_real_hp(). 

The code originates from [this commit](https://github.com/crawl/crawl/commit/3a533adeb52d)

~~#875 is a working patch for the original problem.~~
#877 is a better patch, and also solves the original problem. 

Closed because I wrote it

PR #877, opened 2018-10-25
HP Simplification & Transform Bugfix

stenellad:
In the same vein as #875, meant to fix #870 and work on #874. Credit to NP7 for assistance with planning.

Commit 1
-
Since the default parameter for rotting is only used once (of now 7 references), remove it from the ghost function (i.e. define it explicitly).

Since deflate_hp()'s second parameter is always false, make it default.

Make rotting ask if we're including it, rather than not including it.
Add a parameter to check whether or not we're counting equipment (unnecessary but probably good)

Add a check to get_real_hp() for transform power to confirm transformation (and fix the bug)
Use recalc_and_scale_hp() instead of extraneous functions

Commit 2
-
Replace all instances of recalc_and_scale_hp() with validate_hp(true), except in the case of removing armour, in which case current health is not rescaled. (e.g. 84/100 HP -> 86/115 HP due to regen)
Replace all instances of calc_hp() with validate_hp().
Combine the two functions.
Add *another* check for transform power, to make sure that not only does it exist, but it's not 0.

Commit 3
-
May as well use the function for Berserk. 

Commit 4
-
I guess we're going all the way. All HP changes are now managed by validate_hp(). Death's Door now actually sets your maximum HP to the maximum allowed HP, instead of just pretending to. Since that's the last use of deflate_hp(), we can remove it in favor of the check inside validate_hp().

All HP management is done through calling validate_hp(); there are some extraneous calls to get_real_hp(), but these are never for modifying a player's HP.

Discussion
-
Concerns = quite a few functions now call validate_hp(). It may be wise to move some of the more common calls (e.g. the scoreboard) to a different function, to make it easier to debug this in the future. This concern is not a reason to not pull these changes (provided they aren't buggy, though they shouldn't be), as the two functions do roughly the same thing and were chosen between seemingly at random.

### DIFF
--- a/crawl-ref/source/art-func.h
+++ b/crawl-ref/source/art-func.h
@@ -1284,7 +1284,7 @@ static void _ETHERIC_CAGE_world_reacts(item_def *item)
 
 static void _ETERNAL_TORMENT_equip(item_def *item, bool *show_msgs, bool unmeld)
 {
-    validate_hp();
+    calc_hp();
 }
 
 static void _ETERNAL_TORMENT_world_reacts(item_def *item)
@@ -1296,7 +1296,7 @@ static void _ETERNAL_TORMENT_world_reacts(item_def *item)
 
 static void _ETERNAL_TORMENT_unequip(item_def *item, bool *show_msgs)
 {
-    validate_hp();
+    calc_hp();
 }
 
 ///////////////////////////////////////////////////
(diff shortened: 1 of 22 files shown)


NormalPerson7:
You've got some indentation issues going on I think.

ebering:
Merged in trunk (1406302)

PR #878, opened 2018-10-26
MSVC 2017 Compatability

stenellad:
Changes for MSVC2017 compatability
Not included: Changes to submodule project files.

Use of all changes (and non-included submodule files) successfully builds 32-bit release tiles on MSVC2017 with no errors. Tested using Windows 8.1. Work required on freetype project files for 64-bit and debug to work.

No discussion needed:
-
Any of the project files
zlib.props (a project file)
debug.h
dungeon.cc

Discussion:
-
- AppHdr.cc

SDL2 binaries use the name SDL2. This is an MSVC file.

- MSVC/include/stdint.h

Definitions required for ratio and chrono to work properly. This might be properly defined elsewhere.

 - acquire.cc

MSVC doesn't like implicit use of && and "filter". Can probably be solved with good code. Change is a hack and shouldn't be used.

- glwrapper-ogl.cc
- windowmanager-sdl.cc

This change was required locally, I don't know if this breaks anything. If it doesn't then it should be fine.

- geom2d.h
- hash.h

If this tag isn't necessary it shouldn't be used here. Can post relevant errors if these are missed.

- domino.cc
- loading-screen.cc
- wizard.cc

Precompiled header file changed to be first line

- viewchar.cc
- initfile.cc

These are more changes here than intended. The actual changes are removal of all unicode characters from the file, in favor of their equivalent code. 

- libutil.h

Self explanatory

- player-reacts.cc
- spl-transloc.cc
- spl-transloc.h

MSVC had trouble telling the difference between this and std::disjunction. My change is a hack. Consider a different name, or a convention?

- tile.cc

Use of VLAs is strictly not supported. My change is a hack and should not be kept.

- shopping.cc

Complaints about std::list()

- tile_page.cc
- tile.cc
- tile_colour.cc
- tile_list_processor.cc
- tile_page.cc

<algorithm> included for complaints about a lack of max() and min() definitions. May not be necessarily.

- fixedvector.h

Not sure if this change is necessary

Sorry about the duplicates.

### DIFF
--- a/crawl-ref/source/geom2d.h
+++ b/crawl-ref/source/geom2d.h
@@ -12,13 +12,13 @@ struct vector
         : x(_x), y(_y) {}
 
     const vector& operator+=(const vector &v);
-    vector operator+(const vector &v) const PURE;
-    vector operator-() const PURE;
+    vector operator+(const vector &v) const;
+    vector operator-() const;
     const vector& operator-=(const vector &v);
-    vector operator-(const vector &v) const PURE;
+    vector operator-(const vector &v) const;
 };
 
-vector operator*(double t, const vector &v) PURE;
+vector operator*(double t, const vector &v);
 
 struct form
 {
@@ -75,7 +75,7 @@ struct lineseq
     lineseq(double a, double b, double o, double d)
         : f(a,b), offset(o), dist(d) {}
 
-    double index(const vector &v) const PURE;
+    double index(const vector &v) const;
 };
 
 struct grid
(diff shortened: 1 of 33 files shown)


crawl-ref/source/geom2d.h
```diff
@@ -12,13 +12,13 @@ struct vector
...
-    vector operator-() const PURE;
+    vector operator+(const vector &v) const;
+    vector operator-() const;
     const vector& operator-=(const vector &v);
-    vector operator-(const vector &v) const PURE;
+    vector operator-(const vector &v) const;
 };
 
-vector operator*(double t, const vector &v) PURE;
+vector operator*(double t, const vector &v);
```
amalloy: I'm sad to be losing these PURE annotations just to appease a compiler that doesn't support them. Can we leave them in, and instead redefine the PURE macro to expand to nothing while under MSVC?

PR #880, opened 2018-10-28
Add submitted vaults from mantis

ebering:
As we've drifted to GitHub a bunch of vault submissions have wallowed on mantis. I went digging to try to get what's good to be gotten and close the rest.

There are a lot of vaults here. I've tested them and made adjustments, but I'm still a somewhat novice vaultsmith so I'd like another set of eyes on these before we land them.


MarvinPA:
From a quick glance just at the entry vaults:
wander_arrival_estate - probably shouldn't place butterflies, they're mostly just annoying and we've removed them from various altar vaults in the past.
wander_arrival_abandoned_jail - defines a monster but doesn't place any (and also probably shouldn't)

NormalPerson7:
We could add https://crawl.develz.org/mantis/view.php?id=11757 to this PR.

ebering:
I definitely intend to handle those separately.

PR #882, opened 2018-10-30
Leave Dual-Wielders Alone

### LINKED ISSUE #858: Tukima's Dance used on dual-wielding monsters bug re: sprite appearance.
Platform: Webtiles
Version: Trunk, 0.22

Using Tukima's Dance on dual-wielding monsters such as two-headed ogres or ettins occasionally causes both weapons to disappear from the monster sprite, but upon inspection the game will note that they are still wielding a single weapon and by casting Tukima's Dance a second time this "invisible" weapon can be animated as normal. All seems to be fine with the code apart from the sprite appearance.

Edit: Added screenshot example.
tukimasdancespritebug

stenellad:
Fixes #858

If a dual-wielder's primary weapon was removed by Tukima's
(determined by a coin-flip in monster::weapon()) , their mcache
entry would be invalided by this check, which didn't consider the
valid option of an alt-hand-wielded weapon.

### DIFF
--- a/crawl-ref/source/tilemcache.cc
+++ b/crawl-ref/source/tilemcache.cc
@@ -1242,8 +1242,14 @@ bool mcache_monster::valid(const monster_info& mon)
     bool have_shield_offs = (mon.type == MONS_PLAYER
                              && Options.tile_shield_offsets.first != INT_MAX)
         || get_shield_offset(mon_tile, &ox, &oy);
+    //Return true if the tile has a weapon offset and has a weapon,
+    //a shield offset and a shield, or is a dual-wielder and has a 
+    //shield offset and an off-hand weapon (a valid edge case)
     return (mon.inv[MSLOT_WEAPON] && have_weapon_offs)
-        || (mon.inv[MSLOT_SHIELD] && have_shield_offs);
+            || ((mon.inv[MSLOT_SHIELD]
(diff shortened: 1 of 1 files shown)


crawl-ref/source/tilemcache.cc
```diff
@@ -1243,7 +1243,10 @@ bool mcache_monster::valid(const monster_info& mon)
                              && Options.tile_shield_offsets.first != INT_MAX)
         || get_shield_offset(mon_tile, &ox, &oy);
     return (mon.inv[MSLOT_WEAPON] && have_weapon_offs)
-        || (mon.inv[MSLOT_SHIELD] && have_shield_offs);
+            || ((mon.inv[MSLOT_SHIELD]
+                    || (mon.inv[MSLOT_ALT_WEAPON]
+                            && mons_class_wields_two_weapons(mon.type)))
+                 && have_shield_offs);
```
rawlins: If the position of `have_shield_offs` is right in this condition (not sure just from reading) then it probably merits a code comment, because that implies that `get_shield_offset` is used for second weapons as well as shields. (Is it? I couldn't tell by just reading that function either, so if so, it could use an edit to the docstring as well.)

stenellad: It is, yes.

PR #883, opened 2018-10-31
Properly explode Wooden doors

stenellad:
Fixes https://crawl.develz.org/mantis/view.php?id=11695

There are currently no stone doors in the game.

Damage unchanged.

### DIFF
--- a/crawl-ref/source/spl-damage.cc
+++ b/crawl-ref/source/spl-damage.cc
@@ -2104,7 +2104,7 @@ bool setup_fragmentation_beam(bolt &beam, int pow, const actor *caster,
     case DNGN_METAL_WALL:
         if (what)
             *what = "metal wall";
-        // fall through
+        // fall-through
     case DNGN_GRATE:
         if (what && *what == nullptr)
             *what = "iron grate";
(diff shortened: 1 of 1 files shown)


rawlins:
I noted this in IRC, but it was easily missed -- if doors are made of wood then thematically LRD shouldn't work (relative to both the idea as I understand it, and all the current flavor text / manual description). So either doors shouldn't be targetable at all, or the flavor for LRD needs to be changed too. I'm not sure making LRD work on wood in general is the right way to go though, not that we have that much of it. One option would be reflavoring just the message here so that it works on the door frame, which though small on the tiles, appears to be made of stone. I think this latter one is my preferred approach, because having it not work on doors would be kind of a fiddly exception; not sure if devs who have more of a stake in LRD (e.g. @gammafunk) have preferences here.

stenellad:
Commit modified according to your comments.

I was under the impression that LRD still broke doors, and that your concerns about the wooden nature of doors didn't matter thematically, since the wooden doors themselves are what shatter into pieces. Since this is not the case, I completely agree, and have modified the text to make more sense.

PR #884, opened 2018-10-31
 Properly notify non-users of Gourmand

stenellad:
https://crawl.develz.org/mantis/view.php?id=11692

No gameplay effects.

### DIFF
--- a/crawl-ref/source/player-equip.cc
+++ b/crawl-ref/source/player-equip.cc
@@ -1281,21 +1281,20 @@ static void _equip_jewellery_effect(item_def &item, bool unmeld,
         break;
 
     case AMU_THE_GOURMAND:
-        if (you.species == SP_SPRIGGAN 
-         || you.species == SP_VAMPIRE
-         || you.species == SP_MUMMY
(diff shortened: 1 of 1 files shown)


crawl-ref/source/player-equip.cc
```diff
@@ -1281,9 +1281,22 @@ static void _equip_jewellery_effect(item_def &item, bool unmeld,
         break;
 
     case AMU_THE_GOURMAND:
-        // What's this supposed to achieve? (jpeg)
-        you.duration[DUR_GOURMAND] = 0;
-        mpr("You feel a craving for the dungeon's cuisine.");
+        if (you.species == SP_SPRIGGAN
```
rawlins: this conditional works, but can it be replaced with more abstracted functions of some kind, rather than direct species checks? Spriggans already have MUT_HERBIVOROUS inherently, and the other two are undead. For the next conditional block all the species should inherently have MUT_CARNIVOROUS. That way this function wouldn't need to be edited when new species with these properties are added.

PR #886, opened 2018-10-31
Handle Skill Targets with Potion of Experience

stenellad:
https://crawl.develz.org/mantis/view.php?id=11352

Restore/disable met targets after quaffing a potion of experience

### DIFF
--- a/crawl-ref/source/skill-menu.cc
+++ b/crawl-ref/source/skill-menu.cc
@@ -814,6 +814,7 @@ void SkillMenu::finish_experience(bool experience_change)
         }
         m_skill_backup.restore_training();
         m_skill_backup = skill_state();
+        check_training_targets();
     }
 }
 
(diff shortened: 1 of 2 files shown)


rawlins:
Correct bug link: https://crawl.develz.org/mantis/view.php?id=11352

It would be helpful to be a bit more verbose in commit messages. For example, in this commit, it's not obvious just from reading the code what the `skills.cc` change is for, relative to this bug -- is the xp potion menu affecting training targets in a way that makes them need restoring? (Alternatively, a comment in the code might make sense depending on what is going on.) Also, in general it's good practice when fixing a mantis bug to put the bug number in the commit title or at least the commit message.

i.e., you're submitting enough PRs that we, or at least I, am going to start bugging you about style stuff before merging that we don't necessarily enforce right away :-)

stenellad:
Commit modified according to your comments

PR #887, opened 2018-10-31
Describe Aux Tailswipe in Mutations

stenellad:
Should probably do all of the aux attacks at once if you're going to do it, so this shouldn't be merged until that conversation is had. I didn't realize that it was consistently absent. (it also isn't retaliatory). It's also better handled using the checks that already exist in  _extra_aux_attack in player.cc
-

https://crawl.develz.org/mantis/view.php?id=11154

### DIFF
--- a/crawl-ref/source/mutation.cc
+++ b/crawl-ref/source/mutation.cc
@@ -693,6 +693,9 @@ string describe_mutations(bool drop_title)
     if (player_res_poison(false, false, false) == 3)
         result += "You are immune to poison.\n";
 
+    if (you.fishtail == true)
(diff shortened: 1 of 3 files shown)


MarvinPA:
The tailslap is just an aux attack like horns, fangs etc, it's not retaliatory (as noted by alexjurkiewicz in IRC).

ebering:
Closing this and making a note in the bug report. A patch addressing the general situation to communicate to the player that they have aux attacks in all situations that they do would be welcome, though I'm not sure the 'A' screen is the best place to list the offhand punch.

PR #888, opened 2018-11-01
Don't give stat zero'd slow players false hope

stenellad:
https://crawl.develz.org/mantis/view.php?id=11476

If a player is slowed in some way, then stat zero'd, when the slow runs out, they'll be told they've been sped up, despite still being slow.

Check to make sure players aren't being slowed by stat zero before telling them they're fast again.

Whichever condition ends second will send the message as expected.


rawlins:
please use max 72 char width on commit messages, if you're doing them in vim, you can do this with `:set tw=72`.

Code-wise this one looks ok to merge; I might consider abstracting these bits of code into something like a function that checks if player speed has recovered (by any means), but that's not necessary for this PR.

PR #889, opened 2018-11-01
Remove Twisted Resurrection

Doesnty:
So you can use Yred's Enslave Soul to convert a deep elf high priest (or death mage), which will go around casting Twisted Resurrection. It largely works as the removed player spell, and is degenerate in the worst ways. The specter will only cast it in combat, and will only cast it if two or more corpses are in view, but it creates permanent and powerful allies that follow you through staircases and ignore Yred's usual limits.

Twisted Resurrection has another, less important problem; with the new butchering mechanics, it is physically impossible to butcher corpses that you can't eat. While this is a good UI change, it makes zero sense in-universe that spriggans and mummies are missing this defense against Twisted Resurrection.

Given both of these I think the spell really ought to go. I would also eliminate crawling corpses and macabre masses, but there are a couple vaults that make use of those, so I'm not quite sure what to do with them.

I also think deep elf death mages are kinda boring with this change and could do with Haunt as a replacement spell, but I decided to omit that from the PR since it's a bit of a creative liberty. Apologies for the redundant commits.

Credit to MalcolmRose for alerting me to the existence of the first issue.


alexjurkiewicz:
I like twisted resurrection! But the Tried part does sound degenerate.
Maybe it could just be disabled on enslaved souls?

stenellad:
Why not just fix what you don't like, instead of removing it from the game?

rawlins:
I don't have super strong opinions about the right way to fix the main issue, and it doesn't address the butchering problem, but something like @alexjurkiewicz 's suggestion has precedent in 2959f2344f0.

ebering:
Someday (maybe soon? who knows) we'll remove food and in most of those scenarios monster TR will get a minor buff because butchering will go with food removal. For that reason I'd like to keep TR and go with @alexjurkiewicz suggestion that it be disabled for enslaved souls. If you want to force push a patch to this PR that goes in that direction I'd merge it, otherwise I'll leave this open for issue tracking until I find time to write such a patch or someone who disagrees with me merges this one.

PR #893, opened 2018-11-04
Clean up some code

NormalPerson7:
Also fix an issue with 394fb52 not properly accounting for trolls.


rawlins:
thanks, yeah I was lazy about enforcing spacing when merging the original one.

PR #894, opened 2018-11-04
Don't corrode or appear to damage acid immune enemies with Punk or acidic bite (11746)

NormalPerson7:
Acidic attacks, from the acidic bite mutation and Punk, previously
displayed the "[Enemy] is splashed with acid" line against all targets,
which suggested to the player that acid immune targets still took extra
damage from these attacks even though any damage was in fact resisted.
Additionally, this was able to corrode acid immune enemies, which does
not seem thematically correct.

Therefore, in `monster::splash_with_acid`, add a check at the start of the
method for acid immunity, which if true, return early from without any
opportunity to damage or corrode the monster.

Acid resistant enemies correctly take half-damage from acid splash and
resist corrosion half of the time both before and after this change.

### DIFF
--- a/crawl-ref/source/monster.cc
+++ b/crawl-ref/source/monster.cc
@@ -4370,6 +4370,10 @@ bool monster::corrode_equipment(const char* corrosion_source, int degree)
 void monster::splash_with_acid(const actor* evildoer, int /*acid_strength*/,
                                bool /*allow_corrosion*/, const char* /*hurt_msg*/)
 {
+    // Splashing with acid shouldn't do anything to immune targets
+    if (res_acid() == 3)
+        return;
+
     const int dam = roll_dice(2, 4);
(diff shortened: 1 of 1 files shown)


stenellad:
While this is obviously correct from a lore perspective, it's important to note that there aren't that many enemies with corrosion resistance. It may come as quite a shock to randomly have punk do zero damage, if it's the person's only weapon. Is it still capable of killing things like shrikes?

NormalPerson7:
The only acid immune enemies I know of are jellies, and the only acid resistant enemies I know of are yellow draconians; caustic shrikes are not acid resistant. It's not a big deal from a balance perspective I think

Also, no acid splash damage actually previously happened, because all monsters with three ranks of a resistance get all damage negated; it's just that the message was no different (and corrosion could occur) which made it appear like damage was being done.

PR #899, opened 2018-11-10
Mention Ijyb's and Maurice's wand-carrying in their descriptions

NormalPerson7:
Just a flavour thing; this should have no impact on gameplay now since wands are always announced.

Suggestions on how to improve the incorporation of wands into their descriptions are very much welcome (especially Maurice's, whose I've just added with no real additional flavour text).


NormalPerson7:
Yea I agree that wording is way better :D

PR #901, opened 2018-11-11
Abyss Recall

stenellad:
Upon further inspection, this appears to be a design choice.
https://github.com/crawl/crawl/blob/master/crawl-ref/source/god-companions.cc#L116

Discussion required?

Regardless, this should not be merged, as the solution would be to remove the above-linked lines, or add an exception for Hep's ancestor.

--
Bug description:

Leaving the Abyss without the Ancestor, then re-entering makes
it impossible to recall the Ancestor for a short period of time,
failing silently. This fixes the check that causes it to fail.

### DIFF
--- a/crawl-ref/source/god-companions.cc
+++ b/crawl-ref/source/god-companions.cc
@@ -184,9 +184,12 @@ bool companion_is_elsewhere(mid_t mid, bool must_exist)
     if (companion_list.count(mid))
     {
         return companion_list[mid].level != level_id::current()
-               || (player_in_branch(BRANCH_PANDEMONIUM)
-                   && companion_list[mid].level.branch == BRANCH_PANDEMONIUM
-                   && !monster_by_mid(mid));
+            || (player_in_branch(BRANCH_PANDEMONIUM)
+                && companion_list[mid].level.branch == BRANCH_PANDEMONIUM
+                && !monster_by_mid(mid))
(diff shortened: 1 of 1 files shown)


NormalPerson7:
Well failing silently is a bug, even if it failing at all is a design choice. My personal opinion though is that the  ancestor should be successfully recalled in all cases where the ancestor is not dead, ie. that Hep's ancestor should be an exception to other cases.

stenellad:
This code was written before Hep was added, so it seems like the silent failing would be appropriate in cases where you're trying to recall e.g. Beogh allies from the abyss. 

That said, it's a pretty simple oversight, and easy to fix. The only question is how, specifically.

ebering:
Closing this because the request description states "this should not be merged".

Just removing the mentioned lines in god_companion will not fix the interaction between the Hep ancestor and the abyss. The following behaviors (in addition to the depart/reenter weirdness) need to be addressed:

1. Entering via a portal, the ancestor does not follow.
2. Leaving the abyss with the ancestor in it may render the ancestor unrecallable temporarily.
3. Losing LOS of the ancestor (via abyss shift or teleport) in the abyss silently renders the ancestor unrecallable. 

There are gameplay issues with the interaction between the abyss and allies that should be thought over carefully. "Quick fixes" are not the approach to take here.

PR #905, opened 2018-11-23
Let the sturdy frame mutation reduce stealth/dodging penalties (11768)

NormalPerson7:
See mantis bug report 11768: https://crawl.develz.org/mantis/view.php?id=11768

Previously, the sturdy frame mutation was taken into consideration only
for the adjusted_armour_evasion_penalty method, not the
unadjusted_armour_evasion_penalty. The latter of the two methods is used
for stealth penalties, Jiyva stat shuffling, autotraining and some parts
of the dodging penalty calculations, which means these things were not
affected by the sturdy frame mutation. This meant the claimed ER-2 of
the sturdy frame mutation did not have the full effect of reducing the
ER by 2.

Hence, move the sturdy frame mutation's effect to
unadjusted_armour_evasion_penalty.

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -5701,7 +5701,8 @@ void player::ablate_deflection()
  * Used as the base for adjusted armour penalty calculations, as well as for
  * stealth penalty calculations.
  *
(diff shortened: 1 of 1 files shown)


ebering:
Thanks for the patch!

PR #906, opened 2018-11-23
Don't halve the MR of magic immune creatures from AF_VULN (11754)

NormalPerson7:
Previously phantasmal warriors were able to halve the MR of magic immune
creatures, which was announced and appeared to have an effect even
though it really had no effect whatsoever.

Hence, add a check to exit the case for AF_VULN early if the defender is
immune to hostile enchantments.


ebering:
Thanks for the patch?

PR #908, opened 2018-11-24
MSVC gitignore

stenellad:
Exclude MSVC outputs
Modify vcproj exclude to include vcxproj files


crawl-ref/.gitignore
```diff
@@ -148,12 +148,16 @@ makefile.dep
 # Visual C++ cruft
 *.suo
 *.ncb
-*.vcproj.*.user
+*.vc*proj.user
```
aidanholm:
```suggestion
*.vc{x,}proj.user
```

PR #909, opened 2018-11-25
Simplify Pacification and Display Pacification Chance

ebering:
This series of patches changes pacification to a simpler formula that depends only on monster class, not the monster, and then displays the success chance in the targeter.

Because of the mc change, ghost_demon type monsters are made immune.


gammafunk:
I think the last commit excludes ugly things, very ugly things, and dancing weapons from pacification, which is probably not what you intended. No comment on the variability of these monsters' HP, but it might be best to just exclude the specific two monsters you care about instead of checking for ghost demon, since that flag is purely a technical one related to how a monster is created.

PR #911, opened 2018-11-27
Fix messaging when Maurice steals a weapon (Salmon)

nlavsky:
When Maurice steals a weapon, the games says
`  Maurice steals <weapon>...`
`  Maurice wields <weapon>...`
`  It bursts into flame!/Its appearance distorts for a moment./etc.`
even though the weapon always goes into his alternative weapon
slot (unless Maurice's disarmed).

The same happens if you give an "alt"-inscribed weapon to a
monster via the xo wizard command.

Fix this, and also simplify monster::equip_weapon().

### DIFF
--- a/crawl-ref/source/monster.cc
+++ b/crawl-ref/source/monster.cc
@@ -806,79 +806,76 @@ bool monster::likes_wand(const item_def &item) const
 
 void monster::equip_weapon(item_def &item, bool msg)
 {
-    if (msg)
-    {
-        const string str = " wields " +
-                           item.name(DESC_A, false, false, true, false,
-                                     ISFLAG_CURSED) + ".";
-        msg = simple_monster_message(*this, str.c_str());
-    }
+    if (!msg) return;
+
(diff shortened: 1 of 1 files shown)


stenellad:
To anyone else that reads through this, the bugfix is at the bottom at line 6440. The rest of the changes aren't actually changes at all, just removal of the silly if(msg) wrapping the entire function.

stenellad:
Looking at the new changes:
`equip_weapon`
`equip_armour`
`equip_jewellery`
are only ever called within `equip(). `

`equip()` is called at lines: (and never outside of monster.cc)

1170
1195
1248
2155
6440

All but 6440 seem to have been handled as expected (switching from passing the boolean msg to the function to an if statement before the function call). 

6440 has been changed to include the bugfix from the original commit, making it from always true to 
`if(mslot != MSLOT_ALT_WEAPON || mons_wields_two_weapons(*this))`

tl;dr looks good to me.

PR #912, opened 2018-11-27
MSVC support cont.

stenellad:
Update gen-all.cmd to be up to date with the current makefile,
namely adding aptitudes and crawl_manual.txt, which stops
MSVC builds from crashing when they can't find the files.

Add some more MSVC build outputs to gitignore
Copy CREDITS.txt from the root directory to docs/CREDITS.txt
to match the makefile/release versions; add it to gitignore.

Change MSVC/stdint.h to match MSVC/include/stdint.h , which
we changed earlier in the first MSVC compatability commits.

### DIFF
--- a/crawl-ref/.gitignore
+++ b/crawl-ref/.gitignore
@@ -150,6 +150,7 @@ makefile.dep
 *.ncb
 *.vcproj.user
 *.vcxproj.user
+/source/MSVC/*.user
 
 # Visual C++ outputs
 /source/*.exe
@@ -157,8 +158,12 @@ makefile.dep
 /source/*.dll
 /source/*.exp
 /source/*.lib
+/source/*.iobj
+/source/*.ipdb
 /source/MSVC/.vs/*
 /source/contrib/.vs/*
+*.coverage
+*.vsp
 
 # Packaging
 /source/util/release_ver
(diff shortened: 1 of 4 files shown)


crawl-ref/source/MSVC/stdint.h
```diff
@@ -125,7 +125,7 @@ typedef uint64_t  uintmax_t;
 
 
 // 7.18.2 Limits of specified-width integer types
-
+#define __STDC_LIMIT_MACROS
```
amalloy: Did you intend to leave this in? If so, the next line should be changed, and probably a comment should be added explaining the purpose of this line.

stenellad: Yep! I changed it to match https://github.com/crawl/crawl/pull/878/files#diff-13a5ee50343a26462bbfe00974d4c914 ; this file wasn't changed in the past because it wasn't used by anything. I'm using it now for projects I haven't pushed yet.

There's no harm in changing it, since it's only used by MSVC. I don't know why it works, but without the line, chrono and ratio throw hundreds of errors.

I'd rather not mess with the remainder of the file, since this bugfix is the limit of my experience with it.

See 
https://github.com/ftylitak/qzxing/issues/54
https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/3a3dc941-af59-4788-b75e-74e585a25f3e/compile-issue-vs2015-undeclared-its-in-stdinth-intmaxmax?forum=vcgeneral

amalloy: My point is that it's silly to have code that looks like:

    #define x
    #if defined(x) 
    // body...
    #endif

The entire body is now included unconditionally and all the preprocessor stuff is just noise. 

But it looks like this file wasn't really written as part of crawl at all, but is basically part of the standard library - I would be quite suspicious of making any change at all to stdint.h. Rather, we should work with its existing interface, which is to define `__STDC_LIMIT_MACROS` before including this file, to customise its behaviour. My understanding of the article you linked is that it suggests that the "right" fix is to include `<cstdint>` instead of including `<stdint.h>` directly - cstdint will take care of defining this flag for us, if necessary, before including stdint.h itself. Have you tried making that change to all the DCSS source files, and found that it does not fix the problem for MSVC?

stenellad: Completely agree.

Changing it to <cstdint> didn't work, but explicitly defining __STDC_LIMIT_MACROS in the preprocessor list did; while effectively the same thing, it's definitely the better location for the explicit definition.

PR #914, opened 2018-11-30
Create Mac Tile Download

servayboy:
Making a mac tile download would be very helpful for me personally


ebering:
You meant to make this pull request to bloatrawl.

PR #915, opened 2018-11-30
Don't grant exploration piety for Deep water, Lava, and the Temple

ebering:
Granting exploration piety for deep water, lava, and "endless" features
encouraged players with flight or deep water swimmers to wander around
the non-threatening boundaries of shoals and certain lava vaults.

Granting exploration piety in the Temple encouraged a player to
autoexplore the threat-free temple.

I'm not totally happy with the way this is communicated to the player in this branch, and wanted to get some feedback before merging it. However I think the mechanical change is a good one to make.


floraline:
Personally I don't think the text in `god-conduct.cc` needs to change at all. To "explore the world" already implies exploring *terrain* to me, I didn't even know these tiles counted.

PR #916, opened 2018-12-02
Revised^n Report on the god Nemelex

ebering:
This is currently a WIP revision of Nemelex along the lines of what I've written on the [wiki](../wiki/Goldify-Nemelex-Decks). It's still rough around the edges, but runs and is playable.

### Remaining Todo pre-experimental

- [x] Make drawing each deck a separate `a`bility
- [x] Have the deck chooser use the ability hot-keys for the corresponding decks within triple draw, deal four, and stack five.
- [x] Add tiles for drawing decks
- [x] Add tiles to the deck select menu (similar iconography to the above).
- [x] Deck description for abilities

### Remaining Todo

- [ ] Add a deck display to `^` (on hold until a general auxiliary pane added to `^`)
- [x] Show cards remaining in the ability menu or deck description or somewhere quick to get to
- [x] Streamline the Stack Five UI, or restrict it to a single deck.
- [x] Adjust piety values and ability costs
- [x] (Maybe unnecessary) Save compat for games saved in the uncancellable state for stack five and triple draw.
- [x] set default `force_more`s for some gifts (alexjurkiewicz) every gift is probably overboard, but 0->1 and 12->13 are important.
- [x] show max cards in deck descriptions
- [x] clarify the order of stacked decks in the describe screen.
- [x] do something about Velocity, maybe look at other card effects (**danger zone** don't spiral into a full card review)
- [x] consider adjusting the card gift type distribution
- [x] Fix the bug pointed out below with cancelling cards.
- [x] Show next card in the stack in the `a` menu.


ebering:
From @alexjurkiewicz bug report: if you get a confirm prompt while
                        targeting a destruction spell (eg "your line of fire to
                        the x is blocked by a bush") and answer N, the card is
                        cancelled rather than returning you to the targeter.

ebering:
@MarvinPA Thanks for the feedback. I agree, and will be pushing the second alternative (if < 4 just deal what's left) momentarily.

I've also reduced the piety cost to zero: since it clears the deck the ability is already indirectly piety gated. The cost was (debatably) too high for old Nem, but definitely too high for this iteration.

b-crawl:
Deal 4 doesn't seem to actually clear decks. I'll be keeping it that way, but you might want to change it.

PR #920, opened 2018-12-09
Add dispersal traps to Zot:5

NormalPerson7:
This commit creates a new function is_regular_trap in traps.cc, which
returns whether a particular trap can spawn from a vault which places
"any trap" (that is, a trap that can be spawned from TRAP_RANDOM).

The primary purpose of this is to add dispersal traps to the
hall_of_Zot, in the hope that it will create interesting and dangerous
situations in Zot:5 with the player having to deal with themself and
their enemies being thrown around everywhere. Regarding the forced trap
locations in Zot:5, dispersal traps should be less punishing than
teleport or Zot traps, since the player can get over the other side
simply by repeatedly entering the trap.

I would appreciate feedback on how to make this code less awkward.

### DIFF
--- a/crawl-ref/source/dungeon.cc
+++ b/crawl-ref/source/dungeon.cc
@@ -5875,6 +5875,7 @@ static void _place_specific_trap(const coord_def& where, trap_spec* spec,
                 env.placing_vault.c_str());
     }
 
+    // find an appropriate trap for TRAP_RANDOM
     while (spec_type >= NUM_TRAPS
 #if TAG_MAJOR_VERSION == 34
            || spec_type == TRAP_DART || spec_type == TRAP_GAS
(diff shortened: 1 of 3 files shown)


crawl-ref/source/dungeon.cc
```diff
@@ -5875,14 +5875,19 @@ static void _place_specific_trap(const coord_def& where, trap_spec* spec,
...
 #endif
            || !is_valid_shaft_level() && spec_type == TRAP_SHAFT)
     {
-        spec_type = static_cast<trap_type>(random2(TRAP_MAX_REGULAR + 1));
+        do
+        {
+            spec_type = static_cast<trap_type>(random2(NUM_TRAPS));
+        }
+        while (!is_regular_trap(spec_type))
     }
```
NormalPerson7: Having a `do...while` loop inside a `while` loop like this seems really ugly and inefficient. Is there a better way to do this so that only the `is_regular_trap` condition is checked at the end of every loop whereas all of the other conditions are checked at the start?

NormalPerson7:
Heads up: I haven't actually checked it works yet... I'll test it now.

crawl-ref/source/dungeon.cc
ebering: This loop only runs for TRAP_RANDOM. I'd replace the `spec_type >= NUM_TRAPS` clause by wrapping the whole loop in an explicit `if (spec_type == TRAP_RANDOM)` block and moving the remainder of the conditions to the single `do...while`

NormalPerson7: Ah yea that's neater. I missed the fact that shafts are handled in an earlier `if` and so this loop doesn't run for shafts on the final floors of branches.

ebering:
Thanks for the patch!

PR #922, opened 2018-12-14
Refinement of Amulet of the Acrobat

bjobae:
EV bonus from the amulet of the acrobat is shown more consistently (see https://crawl.develz.org/mantis/view.php?id=11436).
Also, it's EV bonus is not applied while the player is constricted or held.

### DIFF
--- a/crawl-ref/source/main.cc
+++ b/crawl-ref/source/main.cc
@@ -1120,14 +1120,6 @@ static void _input()
 #endif
         const command_type cmd = you.turn_is_over ? CMD_NO_CMD : _get_next_cmd();
 
-        // Clear "last action was a move or rest" flag.
-        // This needs to be after _get_next_cmd, which triggers a tiles redraw.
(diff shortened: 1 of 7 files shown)


ebering:
I know @rawlins expressed some skepticism on IRC that this really made things clearer. Having a status that only affects the player during the monster turn subject to certain conditions is always going to be difficult to communicate well. @cbuchananhowland might have insight into the original implementation choices and why acrobat didn't work this way initially.

Independent of the duration change, the not granting acrobat while held or constricted seems sensible. It would be nice if you could split your changes into two separate commits (`git add --patch` is a helpful command for doing this)

bjobae:
Thanks for your feedback. I wanted to address (11436) and thought the new behaviour would be appropriate. Maybe I am missing something or it's just a matter of taste. I do not expect the second commit to be merged but I'll leave it here just in case.

ebering:
I had been worried that this would be abusable, but after thinking about it this is definitely an improvement over the status quo (appearing but disappearing after a zero time action). Merged with some tweaks, thanks for the patch!

PR #924, opened 2018-12-17
Fix display of rotted HP on WebTiles (11759)

nlavsky:
Commit 1406302deb reversed the meaning of the get_real_hp()'s second
parameter and updated all instances of this function, except for one
in tileweb.cc.


ebering:
The hero we need right now. Thanks for the patch!

PR #927, opened 2018-12-24
Convert spret_type to an enum class

NormalPerson7:
Change spret_type to an enum class, and change its name to spret.

If this is appreciated, I would be happy to continue to class-ify enums, because I think it is the general opinion that enum classes are superior to enums for type security reasons.

This PR should now be ready to be merged.

### DIFF
--- a/crawl-ref/source/spl-summoning.cc
+++ b/crawl-ref/source/spl-summoning.cc
@@ -90,7 +90,7 @@ static mgen_data _pal_data(monster_type pal, int dur, god_type god,
     return _summon_data(you, pal, dur, god, spell);
 }
 
-spret_type cast_summon_butterflies(int pow, god_type god, bool fail)
+spret cast_summon_butterflies(int pow, god_type god, bool fail)
 {
     fail_check();
     bool success = false;
@@ -109,10 +109,10 @@ spret_type cast_summon_butterflies(int pow, god_type god, bool fail)
     if (!success)
         canned_msg(MSG_NOTHING_HAPPENS);
 
-    return spret_type::success;
+    return spret::success;
 }
 
-spret_type cast_summon_small_mammal(int pow, god_type god, bool fail)
+spret cast_summon_small_mammal(int pow, god_type god, bool fail)
 {
     fail_check();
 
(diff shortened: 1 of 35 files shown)


ebering:
Classifying-enums is definitely something that will improve type safety and code health, so thanks for the work. My personal suggestion for the enum class name is simply `spret` so that code reads similar to the old enum (e.g. `SPRET_ABORT` becomes `spret::abort`). This should make proofreading easier.

Things I'd like to see before merging:

- [ ] Squash this to a pair of commits (use `git rebase --interactive` to put your fixups into the appropriate commits)
- [ ] Run qw on the result. While qw doesn't cast, it does evoke and use god abilities, so it will hopefully catch any weirdness.

NormalPerson7:
Yea I thought you'd probably prefer `spret` :)

Also thanks for the advice about qw; I was wondering if there was an easy way to test running the game.

NormalPerson7:
I have tried to run qw on this, and it gives me `XX failed despite returning true.` errors when qw tries to do some things, such as berserking with -Berserk and also trying to chop some corpses, both before and after this change, so it would seem that there are currently issues with qw that prevent easy playtesting.

ebering:
The qw checked into the main repo is a bit out of date, but even elliptic's up-to-date qw has a couple issues. I've sent him some PRs to fix those and I've got it running smoothly now.

crawl-ref/source/spl-summoning.cc
```diff
@@ -1890,30 +1890,30 @@ spret_type cast_animate_skeleton(god_type god, bool fail)
...
  * Have the player cast simulacrum.
  *
  * @param pow The spell power.
  * @param god The god casting the spell.
- * @param fail If true, return SPRET_FAIL unless the spell is aborted.
- * @returns SPRET_ABORT if no viable corpse was at the player's location,
- *          otherwise SPRET_TRUE or SPRET_FAIL based on fail.
+ * @param fail If true, return spret::fail unless the spell is aborted.
+ * @returns spret::abort if no viable corpse was at the player's location,
+ *          otherwise SPRET_TRUE or spret::fail based on fail.
```
ebering: Comment here has one unconverted enum value.

NormalPerson7: Right, because SPRET_TRUE was never an enum value in the first place, but yea you're right I should've run `git grep SPRET` to check.

NormalPerson7:
Any idea where I went wrong with qw then?

ebering:
Thanks for the patch.

PR #928, opened 2018-12-25
Add three tile colour options to the options guide (11802)

NormalPerson7:
Add the options:
    tile_transporter_col,
    tile_transporter_landing_col, and
    tile_explore_horizon_col
to options_guide.txt.
Adjust whitespace to retain alignment of the other tile colour options.


ebering:
Thanks for the patch!

PR #929, opened 2018-12-27
Make cloud duration more apparent to console players

ufshaikh:
Tile players can see when certain clouds (e.g., fire from conjure flame)
are about to expire due to the tile changing. This commit reveals
similar information to console players in two ways. First, pressing | to
raise the layers menu will print a number indicating remaining duration
rather than the usual cloud glyph. Second, an option is added to modify
the cloud glyph or color conditional on its remaining duration.


crawl-ref/source/initfile.cc
```diff
@@ -1480,6 +1481,83 @@ void game_options::add_feature_override(const string &text, bool prepend)
...
+        string key = lowercase_string(mapping[0]);
+        cloud_decay_state cd_key;
+
+        if (key=="terminal")
+            cd_key = CD_TERMINAL;
+        else if (key=="weak")
+            cd_key = CD_WEAK;
+        else if (key=="fading")
+            cd_key = CD_FADING;
+        else
```
NormalPerson7: It's probably slightly neater to use `switch ... case "terminal: ... default:` here, but it's a matter of preference.

Ozaq: Except that you cannot switch over strings :)

NormalPerson7: Oh, right :)

Ozaq: Although I did look over the commit a bit more and ther is a new enum introduced. Dedicated to string / from string functions instead of ad hoc conersion would greatly improve the PR. Does that make sense or should I elaborate on that?

ebering:
At an actual computer now instead of on mobile, I suggest that you add three new dchar types in `dungeon-char-type.h` for the decaying cloud types, and in `viewchar.cc` have them all display as 0 in CSET_ASCII and use the following for CSET_DEFAULT

| cloud state | code point | char |
| --- | --- | --- |
| weak |  U'\x263C'  | ☼ |
| fading | U'\x25CB' | ○ |
| terminal | U'\xB0' | ° |

ufshaikh:
Thank you for the feedback ebering. When you have a chance, please take a look at the new commit.

> I think that making the layered view behave differently is not a good approach: clouds should be their cloud glyphs by default.

I use CSET_DEFAULT, and naturally I am happy to have the cloud glyphs change by default. Perhaps using layers is not the right approach, but I hope something can also be done for those who use CSET_ASCII.

ebering:
The final state of this looks good. As you can tell I'm trying to develop while away from my development environment. If you could rebase and squash it all into a single commit and force push I think it should be mergeable; if you don't have time I'll do that when I next have time.

If another team member wants to squash and rebase this please do.

ufshaikh:
Done.

ebering:
Thanks for the patch!

PR #930, opened 2018-12-28
Remove useless function declaration

ufshaikh:
This function is never defined or called.

(Apologies---this was supposed to be a part of the previous cloud patch, but in the second version I forgot to bundle it in.)


ebering:
Thanks! Cleanup like this is best as its own commit.

PR #932, opened 2018-12-31
Improve spell description of memorised uncastable spells (11808)

NormalPerson7:
"You cannot memorise this spell because ..." was unsuitable when the
player had already memorised an unusable spell before it was made
unusable.

Therefore, change this message to "You cannot cast this spell because ..."
when the player had previously memorised the spell.

It is apparently necessary to cast the first string in the list of strings to concatenate
explicitly to type `string` to avoid compilation errors. This is a little inelegant, so I'd
be happy to change it to an 
`if (you.has_spell(spell)) { description += ... ; } else { description += ... ; }`
block.

Fix was inspired by damerell's comment in the bug report.


crawl-ref/source/describe.cc
```diff
@@ -2967,7 +2967,9 @@ static string _player_spell_desc(spell_type spell)
 
     if (!you_can_memorise(spell))
     {
-        description += "\nYou cannot memorise this spell because "
+        description += (string) "\nYou cannot "
```
ebering: I'm confused as to why you need the cast here. What compiler version and on what system?

`desc_cannot_memorise_reason` has type string, so the compiler should be able to figure out that you're building a  `string` to append to description. 

That said, it would probably be better to re-factor this whole method to use an `ostringstream` instead of a bunch of implicit string appends with `+` and `+=`.

NormalPerson7: Before appending things to `description`, the code first concatenates "\nYou cannot" with the result of `you.has_spell(spell) ? "cast" : "memorise", neither of which are of type `string`, and so the `+` operation fails (for me at least). If one of the first two operands is explicitly a `string`, then it is able to use the built-in `+` operator to form a `string`.

(I'm using gcc version 7.3.0)

ebering: Oh, I see.

ostringstreaem is not `+`ed, it is `<<`ed, and has a method `.str()` to return a string. So you'd do something like

```
ostringstream description;

description << "Spells and stuff mp cost: " << 3;

return description.str()
```

NormalPerson7: Got it, thanks!

PR #936, opened 2019-01-05
Fix information leak for invisible monsters

ufshaikh:
Fixes bug #0011815 on Mantis.

If the player wielded certain unrands that do damage to nearby actors on
attack, and had allies nearby, then they got a prompt for confirmation
when attacking. This prompt happened even if the player could not see
the enemy. So the location of the enemy was revealed, in no time, by the
player's trying to move into the cell. This commit changes that: the
player trying to move into a cell with an unseen monster will attack
without a prompt.

NB: now characters walking around with one of these weapons equipped and
without SInv may inadvertently anger their allies or their god.

### DIFF
--- a/crawl-ref/source/melee-attack.cc
+++ b/crawl-ref/source/melee-attack.cc
@@ -109,7 +109,10 @@ bool melee_attack::handle_phase_attempted()
     if (attacker->is_player() && defender && defender->is_monster())
     {
         // Unrands with secondary effects that can harm nearby friendlies.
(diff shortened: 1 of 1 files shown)


ebering:
A crawl style convention, if you’re planning on continuing to submit PRs: when fixing a mantis ticket include the bug number in the commit in parens: (11815), this takes precedence over the commit title length.

ebering:
Thanks for the patch!

PR #938, opened 2019-01-08
Convert five spell-related enums to enum classes

NormalPerson7:
Following up on #927, convert five more enums to enum classes. Specifically:

* `enum spflag_type` becomes `enum class spflag` 
* `enum spschool_flag_type` becomes `enum class spschool`   
* `enum nothing_happens_when_type` becomes `enum class nothing_happens`   
* `enum miscast_source` becomes `enum class miscast_source`   
* `enum recall_t` becomes `enum class recall_t`.   

See individual commits for what each change involved; each commit message contains a detailed description of any special changes. Each commit contains one conversion and renaming, bundled together, since almost all of the lines changed by the conversion are subsequently changed by the renaming.

There are a couple of things I would like to do to make this better, but since the changes are mostly complete I am creating this PR for feedback purposes:
- [x] add a note to `docs/obsolete/cut_spells.txt` about the fact that enum values have changed since spells were cut
- [x] split the fourth commit into two parts: one dealing with the new struct, and a second dealing with the actual enum change—these were initially done in the opposite order so I made them 1 commit
- [x] change `spschool::RANDOM` to `spschool::random` for consistency with the rest of the enum class
- [x] resolve conflicts.

The result does compile on my machine; I'll let Travis check that more rigorously. I've done some very crude wizmode testing and everything seems to be working as intended—more specific testing around the areas where code logic has had to be changed is probably necessary.

I have reviewed my changes below to highlight some specific things I have questions on. 

Also feel free to suggest improvements to various things I have done here, not least including the new enum class names. I am trying to strike a balance between limiting the length of particular identifiers and the meaning of the type (when used for variable and argument type declarations for example) being clear.

This PR is up to date to the most recent commit at the time of most recently force-pushing, that is df2d5e0. I will force-push an update if any more conflicts arise in the meantime.

(The final commit does not have a commit message because its change is rather trivial.)

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -3030,7 +3030,7 @@ bool bolt::is_reflectable(const actor &whom) const
 
 bool bolt::is_big_cloud() const
 {
-    return get_spell_flags(origin_spell) & SPFLAG_CLOUD;
+    return testbits(get_spell_flags(origin_spell), spflag::cloud);
 }
 
 coord_def bolt::leg_source() const
--- a/crawl-ref/source/mapdef.cc
+++ b/crawl-ref/source/mapdef.cc
@@ -5241,14 +5241,14 @@ bool item_list::parse_single_spec(item_spec& result, string s)
         // spell: <include this spell>, owner:<name of owner>
         // None of these are required, but if you don't intend on using any
         // of them, use "any fixed theme book" instead.
-        spschool_flag_type disc1 = SPTYP_NONE;
-        spschool_flag_type disc2 = SPTYP_NONE;
+        spschool disc1 = spschool::none;
+        spschool disc2 = spschool::none;
 
         string st_disc1 = strip_tag_prefix(s, "disc:");
         if (!st_disc1.empty())
         {
             disc1 = school_by_name(st_disc1);
-            if (disc1 == SPTYP_NONE)
+            if (disc1 == spschool::none)
             {
                 error = make_stringf("Bad spell school: %s", st_disc1.c_str());
                 return false;
@@ -5259,14 +5259,14 @@ bool item_list::parse_single_spec(item_spec& result, string s)
         if (!st_disc2.empty())
         {
             disc2 = school_by_name(st_disc2);
-            if (disc2 == SPTYP_NONE)
+            if (disc2 == spschool::none)
             {
                 error = make_stringf("Bad spell school: %s", st_disc2.c_str());
                 return false;
             }
         }
 
-        if (disc1 == SPTYP_NONE && disc2 != 0)
+        if (disc1 == spschool::none && disc2 != spschool::none)
         {
             // Don't fail, just quietly swap. Any errors in disc1's syntax will
             // have been caught above, anyway.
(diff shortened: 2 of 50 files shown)


ebering:
When you're ready ping me and I can sweat this against qw if you don't have available compute power.

crawl-ref/source/spl-util.h
```diff
@@ -10,28 +10,30 @@
...
-DEF_BITFIELD(spschools_type, spschool_flag_type, 11);
-const int SPTYP_LAST_EXPONENT = spschools_type::last_exponent;
-COMPILE_CHECK(spschools_type::exponent(SPTYP_LAST_EXPONENT)
-              == SPTYP_LAST_SCHOOL);
+DEF_BITFIELD(spschools_type, spschool, 11);
+const int SPSCHOOL_LAST_EXPONENT = spschools_type::last_exponent;
+COMPILE_CHECK(spschools_type::exponent(SPSCHOOL_LAST_EXPONENT)
+              == spschool::LAST_SCHOOL);
+// Moved from mapdef.cc:5318, needed to ensure randbook spschools are short ints
+COMPILE_CHECK(static_cast<int>(spschool::LAST_SCHOOL) < SHRT_MAX);
```
NormalPerson7: I moved this `COMPILE_CHECK` over from an arbitrary place in `mapdef.cc` because I thought that having it in `mapdef.cc` and not right next to the enum class definition meant it could easily be missed when expanding this enum class.

Would you prefer it be here in `spl-util.h` (where it can easily be seen by anyone expanding the enum class) or in `mapdef.cc` (where its purpose is more obvious)?

(I have also marked the old position of this `COMPILE_CHECK` in this review.)

crawl-ref/source/spl-miscast.h
```diff
@@ -10,50 +10,58 @@
...
 #include "spl-util.h"
 
-enum nothing_happens_when_type
+enum class nothing_happens
 {
-    NH_DEFAULT,
-    NH_NEVER,
-    NH_ALWAYS,
+    DEFAULT,
+    NEVER,
```
NormalPerson7: Would using lowercase `default` here cause problems (since `default` is a keyword in C++ for use with `switch` statements)?

crawl-ref/source/mapdef.cc
```diff
@@ -5315,9 +5315,8 @@ bool item_list::parse_single_spec(item_spec& result, string s)
         const string owner = replace_all_of(strip_tag_prefix(s, "owner:"),
                                             "_", " ");
 
-        COMPILE_CHECK(SPTYP_LAST_SCHOOL < SHRT_MAX);
-        result.props[RANDBK_DISC1_KEY].get_short() = disc1;
```
NormalPerson7: (This is the old position of the `COMPILE_CHECK` which I moved to `spl-util.h`.)

crawl-ref/source/beam.cc
```diff
@@ -3030,7 +3030,7 @@ bool bolt::is_reflectable(const actor &whom) const
 
 bool bolt::is_big_cloud() const
 {
-    return get_spell_flags(origin_spell) & SPFLAG_CLOUD;
+    return testbits(get_spell_flags(origin_spell), spflag::cloud);
 }
```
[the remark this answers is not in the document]
stenellad: `static inline bool testbits(uint64_t flags, uint64_t test)
{
    return (flags & test) == test;
}`

Yes

crawl-ref/source/spl-util.h
ebering: I would prefer it here, so that it's not so easily missed. Might just be the GitHub interface but watch the wrap on the comment.

crawl-ref/source/spl-miscast.h
ebering: Yes, `default` is a reserved keyword and cannot be used as an identifier.

crawl-ref/source/spl-util.h
NormalPerson7: The comment is exactly 80 characters long.

ebering: That's what I get for eyeballing.

NormalPerson7:
Oops managed to miss a conflict.

NormalPerson7:
@ebering I am now ready for this to be put to qw, although I'll also do some crude manual testing on very specific code changes that qw may not catch. Sorry it took so long :)

ebering:
Ok, qw is away.

ebering:
When you rebased a new instance of SPFLAG had been added which remains unconverted, as Travis has pointed out. qw hasn't found any issues yet.

NormalPerson7:
Oh right I fixed conflicts but I never checked for new instances.

ebering:
qw is on iteration 177 now, and the only thing I've found are qw bugs, so I'm signing off on this.

It's late so I'm going to hold off on merging this, just in case, but I've marked it as approved and have high confidence that it can merge no problem if someone else is awake and looking for something to do.

PR #939, opened 2019-01-10
added lua spells.path function

mgdelmonte:
Adds a clua function to report the path taken by a spell.  Crawl's spell targeting mechanics are very complex, and without a function like this, proper Lua spell-casting assistance is hampered by the fact that it's impossible to know whether you can actually hit any given target, or what monsters will be hit by a spell.

Given a spell name and target x,y in player-centered coordinates, this function returns the actual path that the spell would take, in a string of form "x1,y1 x2,y2 x3,y3..." (also in player-centered coordinates).  This makes it possible to know whether the spell will hit its target, and (for bolt spells) what monsters it might hit along the way.


crawl-ref/source/l-spells.cc
```diff
@@ -97,6 +98,65 @@ LUAFN(l_spells_min_range)
...
+    if (!you.has_spell(spell))
+    {
+        lua_pushnil(ls);
+        return 1;
+    }
+    int power = calc_spell_power(spell, true);
+    int range = spell_range(spell, power);
+    zap_type zap = spell_to_zap(spell);
+    if( power <= 0 || range <= 0 )
+    {
```
NormalPerson7: A note on Crawl code style: `if` statements should be written like this: `if (condition_1 || condition_2)` with a space before the parentheses and no space between the parentheses and the conditions.

crawl-ref/source/l-spells.cc
```diff
@@ -97,6 +98,65 @@ LUAFN(l_spells_min_range)
...
+    beam.foe_info.dont_stop = true;
+    beam.ex_size = 0;
+    beam.aimed_at_spot = true;
+    beam.path_taken.clear();
+    beam.fire();
+
+    char buf[255];
+    int len = 0;
+    for (auto g : beam.path_taken) {
+        coord_def p = grid2player(g);
```
NormalPerson7: Braces should also be on their own line.

mgdelmonte: got it -- both style errors fixed

mgdelmonte:
Added optional source coordinates so you can evaluate whether a spell can be fired successfully from other places.

stenellad:
Out of curiosity, how did you go about testing this?

mgdelmonte:
Agreed that the return type would be more proper as a table.  I chose the string as a quickie solution; I wanted to see whether there are other objections to this kind of thing first.

As for testing, @stenellad, I combined this with a Lua script that tested a variety of spells with different geometries and monsters to see how it behaved, and it did exactly the right job, though I had to validate the spell by path by eye, so my testing was not exhaustive.  (To make it easy to see, though, the Lua script would set exclusions on the path returned by the function; after that, you can cast the spell for real and quickly see that the tracer path lines up with the exclusions.)

I didn't encounter any "really cast this spell" prompts, @ebering, and I think that's because this is using a `bolt` and not a `targeter`, and also because I set `is_targeting = true` on the bolt, which specifically prevents prompting because it treats the path as something that you're adjusting (see beam.cc:1148) -- so it's using the bolt in that same non-final phase where you're picking the target, and haven't said you want to cast the spell yet.

ebering:
I see, regarding the `bolt` versus `targeter` description. In principle a `bolt` with `is_targeting = true` should have no side effects, so if this does unearth strange bugs that would be good either way.

Looking forward to the version with a table return type. For future reference, if you want to float an idea and see if the team is interested you can drop in `##crawl-dev` on `irc.freenode.net` and ask, proof of concept code isn't required 😀.

ebering:
This is looking about ready to go. I'm not sure I'll have time in the balance of the weekend to land it, but it would speed things along if you could squash+rebase this into a single commit, and ensure that the commit has a good message when you do the squash. Thanks!

ebering:
Not sure what the merge commit is there for. For the commit message itself, the first letter of the title should be capitalized, and the body in paragraph form explaining the what and why of the commit. Your PR message (edited to reflect the actual state of the patch) would do nicely.

mgdelmonte:
Shoot I was trying to rebase and squash.  I'll fix it tomorrow.

NormalPerson7:
You can rebase by running `git rebase -i 96d62d8~1` and setting all commits except the first to `squash` instead of `pick`, in case you were unaware. :)

NormalPerson7:
Ah the empty merge commit might causes problem. You should be able to undo that with `git reset --hard 3a263ef` and then try again, although I'm not 100% sure that fixes it. If you're using the command line, you should be able to edit the commit message when the rebase is finished, but if it doesn't, you can run `git commit --amend`.

ebering:
If you can't get it sorted out, just a GitHub comment with a nicely edited version of what you want the final commit message to be is fine; I'll try to find some time to git-fu it.

ebering:
A plain `git rebase` onto master removed the merge commit, since all the ancestors of the merge commit were present and by default `git rebase` discards commits with 0 change. I've got this cleaned up and will land it soon.

crawl-ref/source/l-spells.cc
```diff
@@ -97,6 +98,69 @@ LUAFN(l_spells_min_range)
     PLUARET(number, spell_range(spell, 0));
 }
 
+
+/*** If this spell is aimed at (x,y), what path will it actually take?
```
ebering: You should indicate either here in the docstring or in the return type that `nil` is returned for spells that don't take a path.

ebering:
How would you like to be listed in CREDITS.txt

mgdelmonte:
Just `mgdelmonte` is fine -- thanks!

PR #940, opened 2019-01-11
Connect ends of slaughterhouse

stenellad:
https://i.imgur.com/cTi7H3w.png

Closes 11823

Add exits to float.des to avoid pockets.


ebering:
Having an entrance to a vault that doesn't lead to any of its encounters isn't particularly elegant. I just now pushed a change removing the teleport closet and making the inner door in that section clear.

PR #941, opened 2019-01-14
Fix holy word scroll (11821)

1Hyena:
When reading the holy word scroll, the player is now warned in
case there are friendly unholy creatures around. Also, another
warning dialogue is spawned if reading the scroll would put the
player under penance.

### DIFF
--- a/crawl-ref/source/item-use.cc
+++ b/crawl-ref/source/item-use.cc
@@ -2780,6 +2780,42 @@ void read(item_def* scroll)
         return;
     }
 
+    if (scroll->sub_type == SCR_HOLY_WORD && item_type_known(*scroll))
+    {
+        if (you_worship(GOD_YREDELEMNUL)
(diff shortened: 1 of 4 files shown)


ebering:
One last question, how would you like to appear in CREDITS.txt?

ebering:
Merged with some tweaks, let me know how you'd like to appear in credits.

1Hyena:
Thanks! By solving this task I learned quite a lot about the engine. While I can't promise periodic contributions, I do know that sometimes it can happen that I have free time and a desire to help crawl at the same time, so now that I know the drill it's very likely that we will talk again ;) As for the credits, you can refer to me by my full name (Erich Erstu).

1Hyena:
@ebering 
by the way, shouldnt https://crawl.develz.org/mantis/view.php?id=10686 also be marked done now?

PR #931, opened 2018-12-29
Beogh adjustments

bjobae:
This is my take on https://github.com/crawl/crawl/wiki/0.23-Development-and-Release-Plan#beogh-follower-menu-ebering. 

I don't know if there is any consensus on how to improve Beogh but I think splitting his recall ability and adding a list of named followers are useful additions. What do you think?

PR #944, opened 2019-01-16
Create a god extra pane

ebering:
From the plan, this builds on the UI part of #931 to create the god-extra pane suggested in the plan. It is almost complete. Remaining desires:

- [x] Webtiles support (required)
- [ ] Per-god string for the pane name to replace "Extra" (desired, but not a show stopper)

@bjobae I've given you write access to to my copy of the crawl repo if you want to continue working on this, push to ebering/crawl/god-extra

### DIFF
--- a/crawl-ref/source/decks.cc
+++ b/crawl-ref/source/decks.cc
@@ -548,9 +548,9 @@ static deck_type _choose_deck(const string title = "Draw")
     for (int i = FIRST_PLAYER_DECK; i <= LAST_PLAYER_DECK; i++)
     {
         ToggleableMenuEntry* me =
-            new ToggleableMenuEntry(deck_status((deck_type)i),
-                    deck_status((deck_type)i),
-                    MEL_ITEM, 1, _deck_hotkey((deck_type)i));
+            new ToggleableMenuEntry(deck_status(static_cast<deck_type>(i)),
+                    deck_status(static_cast<deck_type>(i)),
+                    MEL_ITEM, 1, _deck_hotkey(static_cast<deck_type>(i)));
         numbers[i] = i;
         me->data = &numbers[i];
         if (!deck_cards((deck_type)i))
--- a/crawl-ref/source/decks.cc
+++ b/crawl-ref/source/decks.cc
@@ -218,7 +218,7 @@ static const deck_archetype _cards_in_deck(deck_type deck)
     return {};
 }
 
-static const string _stack_contents()
+const string stack_contents()
 {
     const auto& stack = you.props[NEMELEX_STACK_KEY].get_vector();
 
@@ -227,7 +227,8 @@ static const string _stack_contents()
                 reverse_iterator<CrawlVector::const_iterator>(stack.end()),
                 reverse_iterator<CrawlVector::const_iterator>(stack.begin()),
               [](const CrawlStoreValue& card) { return card_name((card_type)card.get_int()); });
-    output += ".";
+    if (!stack.empty())
+        output += ".";
 
     return output;
 }
@@ -244,7 +245,7 @@ const string stack_top()
 const string deck_contents(deck_type deck)
 {
     if (deck == DECK_STACK)
-        return "Remaining cards: " + _stack_contents();
+        return "Remaining cards: " + stack_contents();
 
     string output = "It may contain the following cards: ";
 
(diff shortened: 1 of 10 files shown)


ebering:
If another devteam member who's more familiar with webtiles wants to take this up please chime in. I can probably figure it out eventually but it will be slow going to get todo point 1 done.

crawl-ref/source/decks.cc
```diff
@@ -547,8 +548,8 @@ static deck_type _choose_deck(const string title = "Draw")
     for (int i = FIRST_PLAYER_DECK; i <= LAST_PLAYER_DECK; i++)
     {
         ToggleableMenuEntry* me =
-            new ToggleableMenuEntry(_describe_deck((deck_type)i),
-                    _describe_deck((deck_type)i),
+            new ToggleableMenuEntry(deck_status((deck_type)i),
+                    deck_status((deck_type)i),
```
rawlins: could change these to `static_cast`

crawl-ref/source/describe-god.cc
```diff
@@ -548,6 +553,106 @@ static formatted_string _god_wrath_description(god_type which_god)
     return desc;
 }
 
+static formatted_string _beogh_extra_description()
+{
```
rawlins: would be nice to add something like "last seen on" to these descriptions, for anyone not currently in los. Maybe too involved for the initial iteration, though.

crawl-ref/source/describe-god.cc
```diff
@@ -548,6 +553,106 @@ static formatted_string _god_wrath_description(god_type which_god)
...
+static formatted_string _god_extra_description(god_type which_god)
+{
+    formatted_string desc;
+
+    switch (which_god)
+    {
+        case GOD_ASHENZARI:
+            if (have_passive(passive_t::bondage_skill_boost))
+            {
+                _add_par(desc, "Curse skill supports:");
```
rawlins: for some reason I found this heading kind of cryptic, maybe something like: `Ashenzari is supporting the following skills because of your curses:`

crawl-ref/source/describe-god.cc
```diff
@@ -1022,23 +1106,36 @@ static void build_partial_god_ui(god_type which_god, shared_ptr<ui::Popup>& popu
...
+            MORE_PREFIX "<w>Overview</w>|Powers|Wrath",
+            MORE_PREFIX "Overview|<w>Powers</w>|Wrath",
+            MORE_PREFIX "Overview|Powers|<w>Wrath</w>",
+            MORE_PREFIX "Overview|Powers|Wrath"
+        },
+        {
+            MORE_PREFIX "<w>Overview</w>|Powers|Wrath|Extra",
+            MORE_PREFIX "Overview|<w>Powers</w>|Wrath|Extra",
+            MORE_PREFIX "Overview|Powers|<w>Wrath</w>|Extra",
+            MORE_PREFIX "Overview|Powers|Wrath|<w>Extra</w>"
```
rawlins: we should eventually convert this to use a widget, aidainh wrote one for this kind of menu. This may actually be the easiest way to get it working on webtiles.

crawl-ref/source/describe-god.cc
```diff
@@ -1022,23 +1106,36 @@ static void build_partial_god_ui(god_type which_god, shared_ptr<ui::Popup>& popu
...
+        }
     };
 
-    for (int i = 0; i < 3; i++)
+    for (int i = 0; i < 4; i++)
     {
         const auto &desc = descs[i];
+        if (desc.empty()) continue;
```
rawlins: I prefer a linebreak here

crawl-ref/source/describe-god.cc
ebering: Converting it to a widget might also be the easiest way to make the pane's name depend on the god. I'm still not super familiar with the ui code in terms of the set of features available.

rawlins: I think the existence of this widget was more wishful thinking than reality -- maybe aidanh and I talked about it. But it doesn't seem to actually exist.

crawl-ref/source/describe-god.cc
rawlins: A minor oddity: possibly only WJC's extra screen should show up on the join menu UI, but right now only Ash's is prevented from showing up.

rawlins:
Changing the pane name would unfortunately be a real pain to do from the current state of things, this kind of UI element turns out to be hardcoded everywhere, including webtiles (in the html even).

ebering:
Wow, that's unfortunate. I guess it needs to wait until the rumored second big UI push. 

I think I'll attempt to fix up the Gozag issue before landing this, the code used for Ctrl+O can have the relevant logic placed in its own function if it doesn't exist already.

ebering:
Merged in 38971ea054..f2270e3573

PR #946, opened 2019-01-16
Don't let battlespheres be blinded (11827)

NormalPerson7:
Battlespheres are insubstantial magical constructs, and thus do not see
with eyes. It therefore makes no sense for them to be able to be
blinded. Therefore, add the M_UNBLINDABLE monster flag to their
mon-data.h entry.

11827 talks about potentially forbidding battlespheres from many of Ru's aura's effects, however I don't see a good reason why they shouldn't be able to be paralysed, confused or slowed, so I haven't looked at that in this PR.

There are probably several other monsters that should be immune to blinding for the same reasons: insubstantial with their sight not governed by the presence of eyes. One such example would be dancing weapons. What do you think about changing these monsters to also have M_UNBLINDABLE (this would be done in a separate PR).


ebering:
A bigger conversation reviewing unblindability should probably wait until after 0.23, but is something we can take a look at in the 0.24 cycle.

PR #947, opened 2019-01-16
Fix monsters not being woken when hit by animated trees (11819)

NormalPerson7:
This commit adds a call to behaviour_event to, among other things, wake
sleeping monsters, when they are attacked by an animated tree from
summon forest as long as the attack hits.

Reproducing this bug is difficult: I recommend the method Yermak talked about in his bug report: that is, place monsters behind a closed door, cast the spell, then open the door.


ebering:
This should definitely happen on hit, whether or not damage is done.

PR #948, opened 2019-01-16
C++ Cloud Generators

stenellad:
Edit 3-17-19: 
This patch has been finished/tested for months. Whenever the Crawl dev team feels like looking at it, they're free to.

C++ Cloud Generators!

Entry point is on_trigger at lm_fog.lua:209; after the level builder has
loaded everything, when it first calls the fog machines, it'll instead create
the C++ fog machines. The other fog-machine-related functions have been
commented out, but the markers will still be triggered and still fire, just
with no payload. This is done partly out of discomfort and partly in the
interest of save compatability.

The fog machines can run every time world_reacts() is triggered. Fog machines
in desolation are strange, and run with delay = 0; I've simulated this by
running them 20 times, and it's pretty close to the current effect.

Other fog-machine types defined in lm_fog.lua seem to be unused, other than
one vault which uses chained fog machines, hangedman_minor_magic. This
works as intended.

All parameters inside lm_fog.lua are re-used. Since it's easy to check, fog
machines won't fire if they're not visible (and their fog can't be visible),
with a distance check I guesstimated at 10.

Inspiration was taken from PleasingFungus' branch, cloud_gen, which is
referenced as a previous commit.

Synchronized generators (i.e. firing in specific order) would be a neat
addition, for future work.

Help wanted:

- [X] Test effects on webserver/desolation performance
- [X] Fix hangedman_minor_magic by implementing chained fog machines (in C++) 
- [X] Add other types of fog machines (presently unutilized)
- [X] Gut unused lua/clean up

I've tested these personally, but, recommended testing for reviewers:
Test desolation
Test hangedman_minor_magic
Test any other cloud-generator-heavy maps, such as Sprint 2

### DIFF
--- a/crawl-ref/source/cloud.cc
+++ b/crawl-ref/source/cloud.cc
@@ -599,6 +599,9 @@ void manage_clouds()
         _dissipate_cloud(cloud);
     }
 
+    for (CloudGenerator &generator : env.cloud_generators)
+        generator.run();
+
     update_cloud_knowledge();
 }
 
@@ -1517,7 +1520,9 @@ cloud_type cloud_name_to_type(const string &name)
 coord_def random_walk(coord_def start, int dist)
 {
     ASSERT(in_bounds(start));
-    ASSERT(dist >= 1);
+
+    if (dist <= 0)
+        return start;
 
     int moves_left = dist;
     coord_def pos = start;
@@ -1813,4 +1818,41 @@ void end_still_winds()
 {
     env.level_state &= ~LSTATE_STILL_WINDS;
     mpr("The air resumes its normal movements.");
+}
+
+
+// XXX: move CloudGenerator stuff into its own file
+
+/// Runs a cloud generator for a single turn.
+void CloudGenerator::run()
+{
+    aut_extant += you.elapsed_time;
+
+    delay -= you.elapsed_time;
+    while (delay <= 0)
+    {
+        const coord_def centre = random_walk(loc, walk_dist);
+        const pair<int, int> size_range = get_size_range();
+        const int size = random_range(size_range.first, size_range.second);
+
+        apply_area_cloud(
+            [this](coord_def where, int /*pow*/, int spreadrate,
+                   cloud_type /*type*/, const actor* agent, int excl_rad)
+            {
+                const int pow = random_range(pow_min, pow_max, pow_rolls);
+                place_cloud(type, where, pow, agent, spread_rate, excl_rad);
+                return 1;
+            }, centre, 0, size, type, 0, spread_rate, exclusion_radius);
+
+        delay += random_range(delay_min, delay_max);
+    }
+}
+
+pair<int, int> CloudGenerator::get_size_range() const
+{
+    /*
+    const int size_buildup_time = min(aut_extant, size_buildup_max);
+    const int size_buildup = size_buildup_max <= 0 ? 0 :
+        size_buildup_amnt * buildup_time / size_buildup_max;*/
+    return { size_min, size_max };
 }
--- a/crawl-ref/source/cloud.cc
+++ b/crawl-ref/source/cloud.cc
@@ -599,8 +599,7 @@ void manage_clouds()
         _dissipate_cloud(cloud);
     }
 
-    for (CloudGenerator &generator : env.cloud_generators)
-        generator.run();
+    manage_cloud_generators();
 
     update_cloud_knowledge();
 }
@@ -1520,9 +1519,7 @@ cloud_type cloud_name_to_type(const string &name)
 coord_def random_walk(coord_def start, int dist)
 {
     ASSERT(in_bounds(start));
-
-    if (dist <= 0)
-        return start;
+    ASSERT(dist >= 1);
 
     int moves_left = dist;
     coord_def pos = start;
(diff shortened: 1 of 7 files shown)


alexjurkiewicz:
Here's a test on CPO:

https://www.youtube.com/watch?v=gnescBHfei0

Please consider this a vote for merging the smallest MVP. Even just replacing desolation fog machines at the start. But I'm excited for the mapping possibilities of having non-killer fog machines.

crawl-ref/source/cloud.cc
```diff
@@ -1814,3 +1816,94 @@ void end_still_winds()
...
+
+
+void manage_cloud_generators() {
+    for (CloudGenerator &generator : env.cloud_generators)
+        generator.ran_this_turn = 0;
+    for (CloudGenerator &generator : env.cloud_generators)
+        generator.manage();
+}
+
+void CloudGenerator::manage()
```
stenellad: I had to change while (delay <= 0) to if (delay <= 0) for the chained loop, because it was infinite looping with ran_this_turn = 1 and delay > 0;

This leads me to believe I screwed up the loop. This function runs the cloud generators in order, unless they're part of a chained fog machine, in which case it will try to run all of them at the same time branching off of the first.

Expected behavior: what hangedman_minor_magic looks like right now. What actually happens: they just run like random cloud generators.

From what I can tell, normal cloud generators, assuming this code doesn't contain a memory leak, function as expected.

crawl-ref/source/cloud.cc
```diff
@@ -1814,3 +1816,94 @@ void end_still_winds()
...
+                [this](coord_def where, int /*pow*/, int spread,
+                    cloud_type /*type*/, const actor* agent, int excl_rad)
+            {
+                const int pow = random_range(pow_min, pow_max, pow_rolls);
+                place_cloud(type, where, pow, agent, spread, excl_rad);
+                return 1;
+            }, centre, 0, size, type, 0, spread, exclusion_radius);
+
+            delay += random_range(delay_min, delay_max);
+            if (delay <= 0 && delay_min == 0 && delay_max == 0)
```
stenellad: Specific case in desolation where clouds run as delay = 0

crawl-ref/source/cloud.cc
stenellad: "ran_this_turn" might be a mistake on my part due to not fully understanding the cloud code, but was my solution to running the cloud generators in the chained fog machine before/without knowledge of their own function call. This is probably the root of the problem

crawl-ref/source/cloud.cc
```diff
@@ -1814,3 +1816,94 @@ void end_still_winds()
...
+                    runs = 20;
+
+                while (delay <= 0)
+                {
+                    run(runs);
+                }
+            }
+            if (special == special_fog_type::chained) {
+                buildup_turns += you.time_taken;
+                if (slave_references.empty() && slaves.size() != 0) {
```
stenellad: The way I actually ascertained which generators are part of the group was by passing their marker IDs to C++, so this goes through every cloud generator in a loop and checks if it's in the group. Since this is two nested for loops of arbitrary size, I only want it to run once per level generation.

crawl-ref/source/cloud.cc
```diff
@@ -1814,3 +1816,94 @@ void end_still_winds()
...
+                    for (CloudGenerator &generator : env.cloud_generators)
+                        for (int i = 0; i < slaves.size(); i++)
+                            if (generator.marker_id == slaves[i] &&
+                                generator.marker_id != marker_id)
+                                slave_references.push_back(&generator);
+                }
+                if (!slave_references.empty()) {
```
stenellad: Uses the pointers from the previous loop to run the chained generator group. Not 100% I did this properly.

crawl-ref/source/cloud.cc
```diff
@@ -1814,3 +1816,105 @@ void end_still_winds()
     env.level_state &= ~LSTATE_STILL_WINDS;
     mpr("The air resumes its normal movements.");
 }
+
+void manage_cloud_generators() {
+    if (env.special_clouds == 1) {
```
NormalPerson7: Code style: all braces should go on their own line for function definitions, `if`s, `for`s, `enum class`es, etc. There are at least thirteen instances where this needs to be changed; I am only going to list the exceptions.

crawl-ref/source/cloud.cc
```diff
@@ -1814,3 +1816,105 @@ void end_still_winds()
...
+
+            apply_area_cloud(
+                [this](coord_def where, int /*pow*/, int spread,
+                    cloud_type /*type*/, const actor* agent, int excl_rad)
+            {
+                const int pow = random_range(pow_min, pow_max, pow_rolls);
+                place_cloud(type, where, pow, agent, spread, excl_rad);
+                return 1;
+            }, centre, 0, size, type, 0, spread, exclusion_radius);
+
```
NormalPerson7: This lambda expression is one place where braces on their own lines isn't required.

crawl-ref/source/cloud.h
```diff
@@ -60,3 +60,100 @@ const cloud_tile_info& cloud_type_tile_info(cloud_type type);
...
+    //Main loop for cloud generators
+    void manage();
+    //Actual "create a cloud please" code
+    void run(int runs);
+    //Whether this cloud generator has been run via a managed group
+    int touched = 0;
+
+    coord_def getLoc() const { return loc; }
+    cloud_type getType() const { return type; }
+
```
NormalPerson7: These short function declarations are also fine as-is.

crawl-ref/source/cloud.h
```diff
@@ -60,3 +60,100 @@ const cloud_tile_info& cloud_type_tile_info(cloud_type type);
...
+        string _marker_id, vector<string> _slaves)
+        :loc(_loc), type(_type), walk_dist(_walk_dist),
+        pow_min(_pow_min), pow_max(_pow_max), pow_rolls(_pow_rolls), kc(_kc),
+        size_min(_size_max), size_max(_size_max), spread_rate(_spread_rate), start_clouds(_start_clouds), exclusion_radius(_excl_rad),
+        size_buildup_amnt(_size_buildup_amnt), size_buildup_time(_size_buildup_time),
+        spread_buildup_amnt(_spread_buildup_amnt), spread_buildup_time(_spread_buildup_time),
+        buildup_turns(_buildup_turns), delay_min(_delay_min),
+        delay_max(_delay_max), special(_special),
+        marker_id(_marker_id), slaves(_slaves)
+    {
```
NormalPerson7: I'm not the greatest programmer but surely there's a better way to do this constructor...

crawl-ref/source/cloud.h
```diff
@@ -60,3 +60,100 @@ const cloud_tile_info& cloud_type_tile_info(cloud_type type);
...
+        int _start_clouds, int _excl_rad,
+        int _size_buildup_amnt, int _size_buildup_time,
+        int _spread_buildup_amnt, int _spread_buildup_time,
+        int _buildup_turns, int _delay_min,
+        int _delay_max, special_fog_type _special,
+        string _marker_id, vector<string> _slaves)
+        :loc(_loc), type(_type), walk_dist(_walk_dist),
+        pow_min(_pow_min), pow_max(_pow_max), pow_rolls(_pow_rolls), kc(_kc),
+        size_min(_size_max), size_max(_size_max), spread_rate(_spread_rate), start_clouds(_start_clouds), exclusion_radius(_excl_rad),
+        size_buildup_amnt(_size_buildup_amnt), size_buildup_time(_size_buildup_time),
```
NormalPerson7: There is a general rule in Crawl's code to avoid lines longer than 80 characters, for various reasons. There are at least 10 lines in this diff that are too long (not just here, but this is the most obvious example, at 134 characters).

crawl-ref/source/cloud.h
stenellad: I'm open to ideas.

rawlins:
I finally had a chance to spend some time trying to understand this PR & do some cloud profiling, and the end result was a9b13e7d511b and e98fbd949879, which address the desolation issue on the lua side. The fix is similar to a change that is in this PR: delay values of 0 were not being handled sensibly in lm_trig.lua, so they are now max'd at 1 there. With these commits I'm not sure of whether there is a need to move cloud generators into c++, though it of course is still possible (and ebering suggests that there may still be slow cloud issues elsewhere, so I won't close this PR). The slowdown in the case of desolation wasn't because of fog machines being in lua, but because the `do_fog` function (and therefore `apply_area_cloud`) was called 200 times per turn per fog machine in desolation.

ebering:
Per @rawlins comment; after those changes were pushed we stopped getting reports of "slow desolation" and "desolation player lagging the server". 

This was a good start but needs further work, and at present the performance demands of crawl don't justify the devteam effort. The submitter has indicated they think the patch is finished, however there are several crawl-dev conversations where various members point out desiderata before this is considered mergeable by the team.

In light of all the above I'm closing this.

PR #952, opened 2019-01-19
Indicate the damage done by spells with !s and "does no damage"

NormalPerson7:
This PR implements the "Consistent use of interpunctuation and damage" subsection of the UI improvements wiki page in part, by making direct damage spells display exclamation marks depending on the damage dealt, and if no damage is dealt, printing `"The %s hits %s but does no damage."` to the message log.

This does not implement this behaviour for clouds, duration effects, and weird effects such as torment, and I'm not sure how useful this would be.

I currently have it so that if a monster completely resists the damage of a spell through elemental resistances, `"%s completely resists"` is printed and `"but does no damage"` is not. Would you rather the `"but does no damage"` part of the message still be printed, and there be some amount of duplication in the message, or the "completely resists" clause be eliminated, or the current behaviour?

See https://github.com/crawl/crawl/wiki/User-Interface-Improvements#consistent-use-of-interpunctuation-and-damage.


crawl-ref/source/beam.cc
```diff
@@ -3237,15 +3230,9 @@ bool bolt::misses_player()
...
-            hit_verb = engulfs ? "engulfs" : "hits";
-
-        if (_test_beam_hit(real_tohit, dodge_more, pierce, defl, r))
-            mprf("The %s %s you!", name.c_str(), hit_verb.c_str());
-        else
+        if (!_test_beam_hit(real_tohit, dodge_more, pierce, defl, r))
             mprf("Helpless, you fail to dodge the %s.", name.c_str());
-
+        // Other hit messages are done by the caller.
         miss = false;
```
NormalPerson7: What would you prefer to be done about this helpless message? I currently have it so that it prints in addition to the `"The %s hits you!!!"` message, because it was necessary to move the hit messaging out of `bolt::misses_player` so that they could be printed after the bolt's damage had been calculated, but this helpless message relies on `_test_beam_hit` taking arguments that are variables local to the `bolt::misses_player` method, so it's not a simple task to move it.

ebering: In melee there is no message for when you could have dodged had you not been helpless, so you could just remove this message. It's nice flavour but it doesn't add much to know that you could have dodge that crystal spear had you not been paralyzed. It also seems to trigger while in tree form, which is a bit odd, as the player isn't *helpless* as a tree. Just a tree.

crawl-ref/source/beam.cc
```diff
@@ -4988,10 +4956,15 @@ void bolt::affect_monster(monster* mon)
         if (hit_verb.empty())
             hit_verb = engulfs ? "engulfs" : "hits";
 
-        mprf("The %s %s %s.",
+        // If the beam did no damage because of resistances,
+        // mons_adjust_flavoured below will print "%s completely resists", so
+        // no need to also say "does no damage" here.
+        mprf("The %s %s %s%s%s",
              name.c_str(),
```
NormalPerson7: Is this behaviour ok?

NormalPerson7:
I've applied your suggestion to remove the helpless message, fixed unbrace (hopefully), and squashed everything into one commit.

crawl-ref/source/attack.cc
```diff
@@ -1505,21 +1511,25 @@ bool attack::apply_damage_brand(const char *what)
...
+                defender->is_player()
+                ? make_stringf("You are electrocuted%s", punctuation)
+                : make_stringf("Lightning courses through %s%s",
+                               defender->name(DESC_THE).c_str(), punctuation);
             special_damage_flavour = BEAM_ELECTRICITY;
             defender->expose_to_element(BEAM_ELECTRICITY, 2);
         }
 
         break;
+    }
```
ebering: What are these braces for? I don't see a local variable for this case.

crawl-ref/source/beam.cc
```diff
@@ -1533,14 +1509,13 @@ int mons_adjust_flavoured(monster* mons, bolt &pbolt, int hurted,
...
-            simple_monster_message(*mons,
-                                       original == 0 ? " appears unharmed." :
-                                         hurted == 0 ? " completely resists." :
-                                   hurted < original ? " resists." :
-                                   hurted > original ? " writhes in agony!" :
-                                   " suffers an impossible damage scaling amount!");
+            simple_monster_message(*mons, hurted == 0 ? " completely resists." :
+                                    hurted < original ? " resists." :
+                                    hurted > original ? " writhes in agony!" :
+                               " suffers an impossible damage scaling amount!");
```
ebering: Any plausible modern compiler should be able to prove that this last case doesn't happen; better to delete it.

crawl-ref/source/player.cc
```diff
@@ -6652,7 +6653,8 @@ void player::splash_with_acid(const actor* evildoer, int acid_strength,
     const int dam = roll_dice(4, acid_strength);
     const int post_res_dam = resist_adjust_damage(&you, BEAM_ACID, dam);
 
-    mpr("You are splashed with acid!");
+    mprf("You are splashed with acid%s",
+         attack_strength_punctuation(post_res_dam).c_str());
     if (post_res_dam > 0)
     {
         mpr(hurt_msg ? hurt_msg : "The acid burns!");
```
ebering: This message should be removed and "but take no damage" added to the splash message.

PR #955, opened 2019-01-23
TAG_MAJOR_VERSION improvements

alexjurkiewicz:
Improve TAG_MAJOR_VERSION pragma coverage so you can compile crawl with both TAG_MAJOR_VERSION = 34 and 35 (although 35 requires additional changes to non-pragma-able files like art-data.txt, tiles.txt files and DES files)


alexjurkiewicz:
These additional patches allow compilation with `TAG_MAJOR_VERSION = 35`
https://github.com/alexjurkiewicz/crawl-ref/compare/patch-5c...alexjurkiewicz:patch-5b

crawl-ref/source/items.cc
```diff
@@ -3045,6 +3045,8 @@ static bool _similar_wands(const item_def& pickup_item,
 #if TAG_MAJOR_VERSION == 34
     // Not similar if wand in inventory is empty.
     return !is_known_empty_wand(inv_item);
+#else
+    return true; // XXX: is this correct?
```
alexjurkiewicz: Calling this out for review

crawl-ref/source/evoke.cc
```diff
@@ -235,8 +235,12 @@ static bool _evoke_horn_of_geryon()
         return false;
     }
 
+#if TAG_MAJOR_VERSION == 34
```
alexjurkiewicz: Calling out this construction for review. I've used it several times in `evoke.cc`. You can see my comment in the commit body of cee19d09c69062020d22862ac1ad74ded89edd2c for why it's left here in TAG 35.

crawl-ref/source/lookup-help.cc
```diff
@@ -358,9 +358,11 @@ static vector<string> _get_god_keys()
     for (int i = GOD_NO_GOD + 1; i < NUM_GODS; i++)
     {
         god_type which_god = static_cast<god_type>(i);
+#if TAG_MAJOR_VERSION == 34
```
alexjurkiewicz: should this function exist at all in TAG 35?

crawl-ref/source/items.cc
amalloy: I agree with this change.

crawl-ref/source/mon-death.cc
```diff
@@ -2307,16 +2311,20 @@ item_def* monster_die(monster& mons, killer_type killer,
 
                 if (have_passive(passive_t::mp_on_kill))
                 {
+#if TAG_MAJOR_VERSION == 34
```
alexjurkiewicz: the multiple `#ifdef` here is ugly, but I couldn't think of a way to do it with a single block, and it seemed brittle to duplicate the entire logic.

crawl-ref/source/religion.cc
```diff
@@ -654,13 +662,16 @@ void dec_penance(god_type god, int val)
...
         {
+#if TAG_MAJOR_VERSION == 34
             if (god == GOD_PAKELLAS)
             {
                 // Penance just ended w/o worshipping Pakellas;
                 // notify the player that MP regeneration will start again.
                 mprf(MSGCH_GOD, god, "You begin regenerating magic.");
             }
-            else if (god == GOD_HEPLIAKLQANA)
+            else
```
alexjurkiewicz: is this formatting OK?

crawl-ref/source/evoke.cc
amalloy: Seems plausible. It's kinda a bummer to leave all these constant 0s lying around, but I imagine that ifdef'ing out all the places that result is used would be even less pleasant.

alexjurkiewicz: Yeah, because the variable name changes (or at least it would be very misleading if it didn't change), you'd end up having to duplicate logic too, which I feel is just asking for someone to later come along and only update the logic in one place.

crawl-ref/source/evoke.cc
```diff
@@ -1573,18 +1607,21 @@ bool evoke_item(int slot, bool check_range)
     case OBJ_MISCELLANY:
         did_work = true; // easier to do it this way for misc items
 
-        if ((you.get_mutation_level(MUT_NO_ARTIFICE)
-             || player_under_penance(GOD_PAKELLAS))
-            && item.sub_type != MISC_ZIGGURAT)
+        if (item.sub_type != MISC_ZIGGURAT)
```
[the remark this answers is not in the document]
amalloy: You are correct that this diff does not change the semantics.

crawl-ref/source/lookup-help.cc
amalloy: Yes, definitely. It is used to power the `?/G` menu.

crawl-ref/source/mon-death.cc
amalloy: You could write:

```
mp_heal = 1 + random2(mons.get_experience_level() / 2);

#if ...
  switch(...) {
    case GOD_PAKELLAS: mp_heal = whatever;
      break;
    default:
      break;
}
#endif
```

crawl-ref/source/religion.cc
amalloy: LGTM.

crawl-ref/source/mon-death.cc
alexjurkiewicz: thanks, good idea. I even removed the switch

crawl-ref/source/viewchar.cc
```diff
@@ -66,7 +66,11 @@ static const char32_t dchar_table[NUM_CSET][NUM_DCHAR_TYPES] =
         // arch .. item_food
         '\\',  '}',  '~',  '8',  '{',  '{',  '{',  '}',  ')',  '[',  '/',  '%',
         // item_scroll .. item_amulet
-         '?',  '=',  '!',  '(',  ':',  '|',  '|',  '}',  '%',  '%',  '$',  '"',
+         '?',  '=',  '!',  '(',  ':',  '|',
+#if TAG_MAJOR_VERSION == 34
```
amalloy: Huh, I didn't realize there were character sets where we didn't use `\` for rods. I suggest adding a comment to this explaining that it was for rods.

crawl-ref/source/util/art-data.pl
```diff
@@ -483,16 +483,14 @@ sub process_line
     "base_type", "sub_type", "plus", "plus2", "COLOUR", "VALUE", "\n",
     "flags",
 
-    # Move FOG after FLY, and remove four copies of "unused", when
```
amalloy: Why is it okay to make these changes now, before the actual transition to 35?

crawl-ref/source/spl-cast.cc
```diff
@@ -1417,14 +1423,22 @@ spret your_spells(spell_type spell, int powc, bool allow_fail,
...
     {
+#if TAG_MAJOR_VERSION == 34
         const int surge = pakellas_surge_devices();
+#else
+        const int surge = 0;
+#endif
         powc = player_adjust_evoc_power(powc, surge);
         int mp_cost_of_wand = evoked_item->base_type == OBJ_WANDS
                               ? wand_mp_cost() : 0;
+#if TAG_MAJOR_VERSION == 34
```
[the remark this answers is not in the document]
alexjurkiewicz: `surge_power_wand` only prints a message (I guess the MP cost is paid elsewhere). But there's no need to calculate `mp_cost_of_wand` so I'll move that into the `#if`

crawl-ref/source/evoke.cc
alexjurkiewicz: That's definitely cleaner code (DRY at least), but given that once TAG is bumped the entire concept of surge will be removed, not sure it's worth my time. What do you think?

PR #959, opened 2019-01-25
Don't add unique annotations for an illusory unique (u/kitchen_ace)

nlavsky:
After commit eddfa9bf, if an illusionary unique created by phantom
mirror outlives the original, the corresponding unique annotation
won't be removed.
Fix this by not adding annotations for illusions.


ebering:
Thanks for the patch!

PR #963, opened 2019-01-26
feat: further distinguish skill drain <= 100 in status

Narretz:
This should make it easier for the player to judge how heavily he
is drained in the early game, and when using Ru's powers.

For monsters, the drain of a wraith is 25, and a hit from a draining
wand is a minimum of 75, but currently both give you the same status.

The same is true for Ru's Draw out power and Apocalypse: 30 vs 100,
but both give you the same status (and have the same "skill drain"
description in the god ability menu). For Ru it's quite important to
manage your skill drain, and introducing a new threshold should help
with that.

Unfortunately, there is no orange so I had to settle on brown.

I also considered using the old "You are drained" for < 50 and introduce a new wording for >= 50 but I couldn't think of anything that works.

I also thought about handling the difference in skill drain for Ru in the ability menu, by introducing a new "light skill drain" flag https://github.com/crawl/crawl/blob/dc03e01af6270c8ad31b42f997966e0998785fa6/crawl-ref/source/ability.cc#L112 but I the bitwise operators made me refrain. If this is issue is acknowleged and this should be the preferred way to handle it, I can work on it further.

### DIFF
--- a/crawl-ref/source/status.cc
+++ b/crawl-ref/source/status.cc
@@ -526,12 +526,19 @@ bool fill_status_info(int status, status_info& inf)
             inf.short_text   = "heavily drained";
             inf.long_text    = "Your life force is heavily drained.";
         }
+        else if (you.attribute[ATTR_XP_DRAIN] > 50)
+        {
+            inf.light_colour = BROWN;
+            inf.light_text   = "Drain";
+            inf.short_text   = "drained";
+            inf.long_text    = "Your life force is drained.";
+        }
         else if (you.attribute[ATTR_XP_DRAIN])
         {
             inf.light_colour = YELLOW;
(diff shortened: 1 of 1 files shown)


ebering:
I've gone ahead and merged the status light change. The comments about how to approach the second patch remain; if you have questions about doing this feel free to drop by ##crawl-dev.

Please let me know how you'd like to appear in CREDITS.txt. Thanks for the patch!

ebering:
I haven't yet heard from you how you'd like to appear in CREDITS.txt, I'm using your author field in your git commits. Send us a PR if you'd like to appear differently.

Narretz:
Hey, thanks for merging this. I must have read over the updates to this PR. Credits.txt representation is fine, thanks.

PR #965, opened 2019-01-27
Make yellow draconians' rAcid a true mutation (11381)

NormalPerson7:
Previously, yellow draconians' rAcid was hardcoded into the game and
they were given a fake mutation description in the species yaml file.
This had the problem that when transforming into statue or lich form,
the fake mutation description was greyed out on the A screen despite
continuing to function.

Therefore, add a new acid resistance mutation, MUT_ACID_RESISTANCE, and
give it to yellow draconians. Remove the special-casing for yellow
draconians' rAcid at the same time, and add a new minor version to
manually give yellow draconians in progress their rAcid mutation.

Additionally, add back "(rCorr)" to the mutation description on the A
screen. This was lost when species definitions were converted to yaml
files.

If it is wanted, this acid resistance mutation can now become a normally
available good mutation from any source of good mutations.
(I can do this in this PR if you'd like.)


NormalPerson7:
Oops. The mutation currently does nothing. I'll push a fix for that shortly.

crawl-ref/source/dat/species/draconian-yellow.yaml
```diff
@@ -31,11 +31,11 @@ levelup_stat_frequency: 4
 mutations:
   1:
     MUT_COLD_BLOODED: 1
+  1:
```
ebering: This `1:` is superfluous, mutations gained at XL are a YAML list.
```suggestion
```

PR #968, opened 2019-01-27
Seeded crawl

rawlins:
This branch implements improvements to seeding behavior, as well as optional dungeon pregeneration for (hopefully) stable seed->dungeon mappings.

Highlights:
* Separate RNG states for dungeon generation, one per branch, so that gameplay RNG is completely distinct. This means that even without a fixed dungeon generation order, D:1 will be the same given a seed. RAII interface for selecting an rng.
* Improvements to make seeding more stable cross-platform (tested on linux vs os x, but not yet on windows): before, vault choice was partly determined by how OS calls provided the .des file order, now, it should be stable (as long as the vault list is the same).
* Pregeneration, based more-or-less on https://github.com/belrogue/crawl/compare/master...belrogue:seeded_crawl. However, this is implemented in a way that should generate levels cleanly, without fake-entering them, and doesn't rely on wizmode calls. Also, there is a UI element that indicates to the user that generation is happening. (Not yet implemented for webtiles.)

Most pregen/seeding features are disabled or hidden for online play; separate game modes for that will come later. (You can see the seed of a dead character, though.)


rawlins:
Oh, I should note, since this gets asked a lot -- it isn't enough to just have separate RNG states for each branch, or sub-generators, or (insert fancy RNG trick here). The challenge is that vault generation is not at all independent  level-to-level, and for example if a vault has already placed, this could cause a later level to veto with completely different results (and that RNG will diverge after that point).  So, stable seed to dungeon mappings require that levels be generated in the same order. Without pregeneration enabled, a seed doesn't give a fully deterministic dungeon in case of shafts, entering branches in a different order, etc.

rawlins:
Missing things (not all of which I will do before merging this, or need to be done first imo):

- [ ] implement progress popup for webtiles
- [ ] add a menu option to play a daily seed
- [ ] implement staged pregeneration, so that it happens as-needed when you enter a new branch. This turns out to be technically nightmarish because of the way `load_level` is written, and I've 95% decided I can't do that yet in this PR.
- [ ] figure out how to allow special online seeded modes

PR #969, opened 2019-01-28
Fix 0011738: Setting a confirm_action for an ability generates a Y/N …

ghost:
…prompt every time the ability menu is opened

The confirm prompt was still being generated when
_check_ability_possible was called in "quiet" mode, though it shouldn't
have been.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -1420,21 +1420,24 @@ static bool _check_ability_possible(const ability_def& abil, bool quiet = false)
         return false;
     }
 
-    vector<text_pattern> &actions = Options.confirm_action;
-    if (!actions.empty())
+    if (!quiet) 
     {
-        const char* name = ability_name(abil.ability);
-        for (const text_pattern &action : actions)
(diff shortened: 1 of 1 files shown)


ebering:
Oh! And how would you like to appear in CREDITS.txt?

ebering:
Sorry, it seems I spoke too soon. Your patch has whitespace trailing at line ends, you can use the `util/checkwhite` utility in the source directory to clean these up.

crawl-ref/source/ability.cc
```diff
@@ -1420,21 +1420,24 @@ static bool _check_ability_possible(const ability_def& abil, bool quiet = false)
         return false;
     }
 
-    vector<text_pattern> &actions = Options.confirm_action;
-    if (!actions.empty())
+    if (!quiet) 
     {
```
NormalPerson7: Or you can manually fix the trailing whitespace - it's just a single space after this `if`.

ebering:
I'm adding you to CREDITS.txt as GenericPseudonym for now, if you'd like this changed, send us a PR!

PR #970, opened 2019-01-28
added more monster info to Lua

mgdelmonte:
-- AC, EV, MR, and MaxHP
-- optional "full description" option for desc()
-- defeat_mr(spell, evoked) gives your chance to defeat monster's MR

### DIFF
--- a/crawl-ref/source/l-moninf.cc
+++ b/crawl-ref/source/l-moninf.cc
@@ -16,6 +16,11 @@
 #include "spl-util.h"
 #include "stringutil.h"
 #include "transform.h"
+#include "math.h" // ceil
+#include "spl-zap.h" // calc_spell_power
+#include "evoke.h" // wand_mp_cost
+#include "god-abil.h" // pakellas_effective_hex_power
+#include "describe.h" // describe_info, get_monster_db_desc
 
 #define MONINF_METATABLE "monster.info"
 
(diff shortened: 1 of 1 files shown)


mgdelmonte:
The "full description" option to `desc()` does obviate the need for `max_hp`, `ac`, and `mr` -- those three could all be extracted from the full description.  (`ev` is *not* so easily extracted, due to the way it's handled.)  However, I think they could still be included just for ease of access.

It might also be a good idea to factor out the various calculations, so that they're not repeated whole in both the description code and the Lua version here.  But that's more of a design decision, so I've not gone ahead with it on my own.  Happy to do it, though, if it's desired by others.

ebering:
Sorry for the delay in reviewing this, I took a bit of a break from development while the 0.23 release tournament happened. I'll be taking a detailed look this week, but the commit message will need elaboration. Your comment on this PR is more along the lines of what we prefer in commit messages.

Regarding the broader design decision: if you wish to continue adding things to CLua's ability to access the monster_info C++ class my recommendation is that a refactor is appropriate. Specifically, both the description code and the Lua should simply reference the relevant mi struct, and the calculation of these values done when the monster info struct is created (when possible, spell success needs to be checked dynamically).

crawl-ref/source/l-moninf.cc
```diff
@@ -207,6 +212,94 @@ MIRES1(res_shock, MR_RES_ELEC)
...
+ * @treturn int MR level
+ * @function mr
+ */
+static int moninf_get_mr(lua_State *ls)
+{
+    MONINF(ls, 1, mi);
+    lua_pushnumber(ls, ceil(1.0*mi->res_magic()/MR_PIP));
+    return 1;
+}
+/*** Your probability of defeating the monster's MR with a given spell or zap.
```
ebering: At this place, and others function ends, add a newline before the next function starts.

PR #973, opened 2019-02-03
Update tutorial, fix tutorial/hints mode skill menu text (11851)

NormalPerson7:
Update a traps message that stated traps aren't always revealed, and a
berserk hunger message that stated you can't berserk while very hungry;
this restriction is now limited to near starving or worse.

Shorten the "Read skill descriptions" button description to "Skill
descriptions" to avoid this overlapping the text on the following line.

I have completely played through the tutorial; these were the only issues I found.

### DIFF
--- a/crawl-ref/source/dat/descript/tutorial.txt
+++ b/crawl-ref/source/dat/descript/tutorial.txt
@@ -371,9 +371,10 @@ of the screen. </localtiles>Rereading old messages with
 tutorial3 trap
 
 Traps can have a variety of unpleasant effects, such as alerting monsters or
-teleporting you into danger. They will often be hidden until you discover them
-by stepping on them. If there's no way around, you'll have to trigger the trap
-and deal with the consequences.
+teleporting you into danger. Some traps are permanent and will always be
+revealed as part of the map, but other traps can be triggered by exploring new
+tiles, and won't exist after being triggered. If there's no way around, you'll
+have to trigger the trap and deal with the consequences.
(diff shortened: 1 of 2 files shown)


crawl-ref/source/dat/descript/tutorial.txt
```diff
@@ -371,9 +371,9 @@ of the screen. </localtiles>Rereading old messages with
 tutorial3 trap
 
 Traps can have a variety of unpleasant effects, such as alerting monsters or
-teleporting you into danger. They will often be hidden until you discover them
-by stepping on them. If there's no way around, you'll have to trigger the trap
-and deal with the consequences.
+teleporting you into danger. They will always be revealed, but you can trigger
+their effects by exploring new tiles. If there's no way around, you'll have to
+trigger the trap and deal with the consequences.
 %%%%
```
rawlins: Thanks for going through the tutorial, super helpful! I would reword this slightly: `Some traps are permanent and will always be revealed as part of the map, but other traps can be triggered by exploring new tiles, and won't exist after being triggered. If there's no way around a permanent trap, you'll have to`...

Part of this rewording frames it as if trap effects are just temporary traps, which is probably confusing if one knows the gory details, but I think this might make a bit more sense to beginners.

NormalPerson7: Yea I thought I could probably do better with that phrasing. Thanks for the suggestion.

PR #977, opened 2019-02-05
Fix dancing weapon description when multiple monsters enter LOS (11843)

NormalPerson7:
When multiple enemies entered line of sight at once including at least
one dancing weapon, the message "There is." was being printed. This is
because there was an unnecessary check for enemies not being dancing
weapons in get_monster_equipment_desc. Remove this check, and add a
special case to the helper function _describe_monster_weapon to prevent
these messages being printed as "There is wielding a +4 trident."

These messages should now properly print as e.g.
"There is a +4 executioner's axe of electrocution."


NormalPerson7:
Weird how Travis errored on 9 checks... is Travis having issues?

PR #978, opened 2019-02-06
Add -branch-json and improve -playable-json

alexjurkiewicz:
Dump basic branch information as JSON.

The key use of this is knowing how to refer to branches when talking
about morgues. Depending on the branch, you might describe a place in
the following ways:

* the Temple
* a Sewer
* an Ice Cave
* Pandemonium
* level 1 of the Dungeon
* level 1 of the Depths
* level 1 of a Ziggurat

Without this data being accessible, figuring out the above
programatically required a hardcoded lookup table, which is duplicated
work and bitrots over time as the Dungeon's layout changes.

---

Also, put recommended jobs/species in `-playable-json` output.

### DIFF
--- a/crawl-ref/source/MSVC/crawl.vcxproj
+++ b/crawl-ref/source/MSVC/crawl.vcxproj
@@ -851,6 +851,7 @@ perl.exe "util/gen-cflg.pl" compflag.h "&lt;UNKNOWN&gt;" "&lt;UNKNOWN&gt;"
     <ClInclude Include="..\bloodspatter.h" />
     <ClInclude Include="..\book-data.h" />
     <ClInclude Include="..\book-type.h" />
+    <ClInclude Include="..\branch-data-json.h" />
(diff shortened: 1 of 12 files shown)


alexjurkiewicz:
For what it's worth, this feature would be used by CPO's "weekly challenge" feature. Currently, I pick a random species and random background each week. But I'd like to have more control over this by being able to know if a combo is recommended (or not). So I would be an immediate user of the improvement.

If anyone is interested in reviewing it, I'll update this PR to merge cleanly.

ebering:
I've taken a look at this and it looks good. The one concern I had is that playable-json is used by Sequell and other parts of crawl infrastructure, so changes shouldn't move fields.

If you resolve the merge conflicts I'll merge this, I don't think there are many?

ebering:
Actually never mind, I've done it (it was indeed a trivial conflict). Merge coming soon!

PR #979, opened 2019-02-06
Add additional functionality to monster list

damerell:
The monster list now provides information even if more than one monster of a given type is visible, can tell you more than one thing about monsters, and indicates many more statuses, most importantly whether monsters are wielding polearms or missile launchers.

The additional statuses have been added in an ordering that tries to follow the original; buffs like berserk first, then incapaciting conditions like paralysis, then maluses, then largely informational material like that a monster is invisible when you have SInv.

The exact list of additional statuses added was a bit arbitary but depended partly on what could be summarised shortly and what seemed too trivial but also high enough up the ordering it might push things youcare about off the screen.

### DIFF
--- a/crawl-ref/source/mon-info.cc
+++ b/crawl-ref/source/mon-info.cc
@@ -1948,6 +1948,13 @@ vector<monster_info_func> init_monster_info_funcs()
                 return newconditions && mi.is(MB_BRILLIANCE_AURA);
             }
         });
+// Not a great word but it'll turn up just after doing Enslave Soul
+    toret.push_back({"possessable", "possessable",
+            [](const monster_info &mi, bool newconditions)
+            {
+                return mi.is(MB_POSSESSABLE);
+            }
+        });
     toret.push_back({"insane", "insane",
             [](const monster_info &mi, bool newconditions)
             {
--- a/crawl-ref/source/mon-info.cc
+++ b/crawl-ref/source/mon-info.cc
@@ -1194,53 +1194,10 @@ bool monster_info::less_than(const monster_info& m1, const monster_info& m2,
 
 static string _verbose_info0(const monster_info& mi)
 {
-    if (mi.is(MB_BERSERK))
-        return "berserk";
-    if (mi.is(MB_INSANE))
-        return "insane";
-    if (mi.is(MB_INNER_FLAME))
-        return "inner flame";
-    if (mi.is(MB_DUMB))
-        return "stupefied";
-    if (mi.is(MB_PARALYSED))
-        return "paralysed";
-    if (mi.is(MB_CAUGHT))
-        return "caught";
-    if (mi.is(MB_WEBBED))
-        return "webbed";
-    if (mi.is(MB_PETRIFIED))
-        return "petrified";
-    if (mi.is(MB_PINNED))
-        return "pinned";
-    if (mi.is(MB_PETRIFYING))
-        return "petrifying";
-    if (mi.is(MB_MAD))
-        return "mad";
-    if (mi.is(MB_CONFUSED))
-        return "confused";
-    if (mi.is(MB_FLEEING))
-        return "fleeing";
-    if (mi.is(MB_DORMANT))
-        return "dormant";
-    if (mi.is(MB_SLEEPING))
-        return "sleeping";
-    if (mi.is(MB_UNAWARE))
-        return "unaware";
-    if (mi.is(MB_DAZED))
-        return "dazed";
-    if (mi.is(MB_MUTE))
-        return "mute";
-    if (mi.is(MB_BLIND))
-        return "blind";
-    // avoid jelly (wandering) (fellow slime)
-    if (mi.is(MB_WANDERING) && mi.attitude != ATT_STRICT_NEUTRAL)
-        return "wandering";
-    if (mi.is(MB_BURNING))
-        return "burning";
-    if (mi.is(MB_INVISIBLE))
-        return "invisible";
-
-    return "";
+    string desc=""; vector<monster_info> miv;
+    miv.push_back(mi);
+    mons_conditions_string(desc,miv,0,1,false);
+    return desc;
 }
 
 static string _verbose_info(const monster_info& mi)
@@ -1312,7 +1269,8 @@ void clear_monster_list_colours()
 }
 
 void monster_info::to_string(int count, string& desc, int& desc_colour,
-                             bool fullname, const char *adj) const
+                             bool fullname, const char *adj, 
+                             bool verbose) const
 {
     ostringstream out;
     _monster_list_colour_type colour_type = _NUM_MLC;
@@ -1334,7 +1292,7 @@ void monster_info::to_string(int count, string& desc, int& desc_colour,
     out << " av" << mons_avg_hp(type);
 #endif
 
-    if (count == 1)
+    if (count == 1 && verbose)
        out << _verbose_info(*this);
 
     // Friendliness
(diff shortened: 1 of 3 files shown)


damerell:
Possible problems with this PR:

Some statuses checked for can't (currently) be got by monsters. I intentionally didn't look into that; either the game will be changed so monsters can get those statuses (in which case it is best they be reported on), they will be removed altogether, or the reporting code is harmless. 

Some of the expressions used to describe statuses are long or otherwise clunky. The ordering of statuses also is a bit arbitary; I followed the rationale above, but there were still many wanting to go in about the same place.

I have done my best to follow vanilla's coding conventions, but probably missed a spot or two.

alexjurkiewicz:
Do you have any screenshots of this in action? What does it look like in particular when there is too much status to fit cleanly?

It could be also good to include distortion branded weapons, and holy wrath for undead species. But including some weapons feels somewhat like editorialising, eg "why does the game point out polearms but not giant spiked clubs?" There's a much clearer line between status and not-status.

damerell:
o   Saint Roka (slow)
 K   kobold demonologist
 3   smoke demon (summoned)
 3   neqoxec
 gggg  4 gnolls (3 polearms, 1 burning)

When there's too much to show cleanly, it's just truncated harmlessly at the righthand edge. I thought about doing something more, but it seemed to me generally preferable to see the start of the status that doesn't fit than nothing at all.

I picked polearms and launchers, to be frank, because I personally am prone to being annoyed by monsters being unexpectedly longer-ranged. One can certainly make a case for adding more unusual kinds of weapons.

crawl-ref/source/mon-info.cc
```diff
@@ -1825,6 +1783,502 @@ void get_monster_info(vector<monster_info>& mons)
...
+            if (info.test(mi[j], listall))
+            {
+                if (!listall)
+                {
+                    out << info.singular;
+                    desc=out.str();
+                    return;
+                }
+                else
+                {
```
alexjurkiewicz: crawl code style for single line blocks like here is to skip the braces. `util/unbrace` can automatically fix up all instances of this, but I don't remember the exact command

crawl-ref/source/mon-info.cc
```diff
@@ -1825,6 +1783,502 @@ void get_monster_info(vector<monster_info>& mons)
...
+            }
+	});
+    toret.push_back({"scary", "scary", 
+	    [](const monster_info &mi, bool newconditions)
+            { 
+                return newconditions && mi.is(MB_FEAR_INSPIRING); 
+            }
+	});
+    // Bah, these next four are really too long
+    toret.push_back({"chanting recall", "chanting recall",
```
alexjurkiewicz: "recalling" perhaps?

crawl-ref/source/mon-info.cc
damerell: I'm not sure if you mean "please resubmit making this change" (which I guess I'll have to figure out how to do) or "just FYI..." or "here's a note that if we accept this we'll have to change that".

crawl-ref/source/mon-info.cc
damerell: I considered and rejected "recalling" since it sounds like the monster's trying to remember where it put its keys. :-/

crawl-ref/source/mon-info.cc
ebering: It will help speed up an eventual acceptance if you make the change.

You can continue to push commits to your branch, GitHub will add them to this PR automatically. You can also force-push your branch (if you need to rebase &c.)

damerell: Done, I think. (I noticed util/unbrace unbraced something of the form if (foo) \n { \n this; that; the_other; } from which I infer that you'd prefer not to have multiple statements on one line like that).

rawlins: yeah, it's not explicitly part of the coding conventions but we do avoid that in all but a very few specific cases (e.g. `case X: something; break;`).

damerell:
I've added death knights' damage mirror to the set of reported statuses. I fear the fact that I missed that implies there may be some other important monster status also missed.

crawl-ref/source/mon-info.cc
```diff
@@ -1825,6 +1783,533 @@ void get_monster_info(vector<monster_info>& mons)
...
+                && mi.is(MB_SLOWED) && newconditions;
+            }
+        });
+    toret.push_back({"fast", "fast",
+            [](const monster_info &mi, bool newconditions)
+            {
+                return newconditions && mi.is(MB_HASTED)
+                && !mi.is(MB_BERSERK) && !mi.is(MB_SLOWED);
+            }
+        });
```
ebering: Strange that `fast+slow` is missing but `berserk+slow` is present.

crawl-ref/source/output.cc
```diff
@@ -1715,7 +1715,8 @@ static void _print_next_monster_desc(const vector<monster_info>& mons,
         {
             int desc_colour;
             string desc;
-            mons[start].to_string(count, desc, desc_colour, zombified);
+            mons_to_string_pane(desc,desc_colour,zombified,mons,start,count);
```
ebering: White space here, space out the commas; and delete the old line, instead of leaving it in a comment.

crawl-ref/source/mon-info.cc
```diff
@@ -1825,6 +1783,533 @@ void get_monster_info(vector<monster_info>& mons)
...
+{
+    mi[start].to_string(count, desc, desc_colour, fullname, nullptr, false);
+    mons_conditions_string(desc, mi, start, count, true);
+}
+
+void mons_conditions_string(string& desc, const vector<monster_info>& mi,
+                            int start, int count, bool listall)
+{
+    static vector<monster_info_func> monster_info_funcs =
+    init_monster_info_funcs();
```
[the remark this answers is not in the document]
damerell: I don't believe this is the case. The initialiser for a static variable in a function is only ever called once.

(Indeed, I've just sanity-checked myself by changing it such that a different word is returned for one status if your god is Ru, something that clearly could not be optimised out altogether. You can't make the status reported change by changing your religion.)

crawl-ref/source/mon-info.cc
damerell: I left out fast+slow because it's normal right now. Admittedly, it could go in as a warning the monster's speed is likely to change in the near future, but a concern I have been wildly inconsistent about is cluttering it up with less important information.

Berserk+slow tells me the monster still hits hard but isn't fast like a normal berserk monster. Arguably it could be rolled into "strong" by virtue of something like the rationale above about showing the status right now.

crawl-ref/source/mon-info.cc
```diff
@@ -1825,6 +1783,533 @@ void get_monster_info(vector<monster_info>& mons)
...
+    }
+    if (didany)
+        out << ")";
+    desc = out.str();
+}
+
+vector<monster_info_func> init_monster_info_funcs()
+{
+    vector<monster_info_func> toret; // well, I couldn't think of a good name
+    toret.push_back({"charmed", "charmed",
```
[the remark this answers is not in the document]
damerell: I intentionally didn't make them consistent. The expressions in monster_info::attributes are used in places where space is not constrained, so they can be chattier; space in the monster list is at a premium. Of course you may think consistency would be a greater virtue.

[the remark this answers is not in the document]
ebering: Attributes also appears in the message line during targeting, where space is at a premium. A consistent solution here that's brief (provided it isn't too brief) is preferred, as there are so many statuses they really don't need two names!

ebering:
I should have included these remarks in my review.

> When there's too much to show cleanly, it's just truncated harmlessly at the righthand edge. I thought about doing something more, but it seemed to me generally preferable to see the start of the status that doesn't fit than nothing at all.

My preference for overflow behavior is to pack as many statuses as can fit in the available width minus 3, and then if there's overflow end with `,…)`. The reason is for screen-reader users, for whom a partial status can be an obstacle.
 
> I picked polearms and launchers, to be frank, because I personally am prone to being annoyed by monsters being unexpectedly longer-ranged. One can certainly make a case for adding more unusual kinds of weapons.

Launchers and polearms are good to advertise for the exact reason given here, that they represent an added threat. I would like wands added to this list. (Just "wand" will suffice.)

crawl-ref/source/mon-info.cc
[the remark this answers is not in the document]
damerell: For some reason I can reply neither to the comment in which you request changes nor the one in which you suggest the addition of wands and addition of a (...) where truncated.

Thank you for the suggestion of adding wands (which I have done in Stoat Soup). However, I regret to say I'm unlikely to get around to changing the "build a vector of structs that include functions" pattern soon (or perhaps not at all). Of course I quite understand if you close this PR or leave it indefinitely on the offchance someone else does.

damerell:
ebering asked that I link to subsequent commits for this feature in Stoat Soup. I'm sorry if there's a "right" way to link to a commit here.

https://github.com/damerell/crawl/commit/063ee9f9cfab6028c1aa5c071475b6fc0e5d724a (add "wand" status (ebering))
https://github.com/damerell/crawl/commit/dc84e45775bccb5e0d06a29452ebd61532cd14bf Don't describe a monster as "swift, slow" - confusing
https://github.com/damerell/crawl/commit/09ee343497c9a1672b370d0419b8be5449cbd5e0 Typo fix in mon-info.cc.
https://github.com/damerell/crawl/commit/1b51726a1d16b89e2a1620b9503fad0aeecf9419 Add "mesmerising you" to list of monster statuses (ceb)
https://github.com/damerell/crawl/commit/d51f7d1946ba6a110a4301f94d30cbd4423617db Monster info pane reports on distortion

And
https://github.com/damerell/crawl/commit/842d55c91eba6bf077431270aa8d14716f0fa9c5 but only the change in mon-info.cc, if you agree with it (don't report that battlespheres are "wandering" because it's misleading).

alexjurkiewicz:
🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

PR #982, opened 2019-02-07
Make god gifts interrupt & clear runrest delays only (11825)

ufshaikh:

### DIFF
--- a/crawl-ref/source/religion.cc
+++ b/crawl-ref/source/religion.cc
@@ -2049,7 +2049,7 @@ bool do_god_gift(bool forced)
     }                           // End of gift giving.
 
     if (success)
-        you.running.stop();
+        stop_running(false);
 
 #if defined(DEBUG_DIAGNOSTICS) || defined(DEBUG_GIFTS)
     if (old_num_current_gifts < you.num_current_gifts[you.religion])


amalloy:
Please add a comment to the commit explaining why this is a good idea. I can guess from the name of your branch, but that won't be apparent once the change is merged.

amalloy:
Ah, I just noticed you give the number of a Mantis issue. It's still nice to explain why in the commit itself, so nobody has to go read the issue to understand the commit.

ufshaikh:
Apologies. I've rebased and added a description of the bug.

crawl-ref/source/religion.cc
```diff
@@ -2048,8 +2048,9 @@ bool do_god_gift(bool forced)
         }                       // switch (you.religion)
     }                           // End of gift giving.
 
-    if (success)
-        you.running.stop();
+    if (success && you.running.runmode != 0) // interrupt and clear runrest
```
amalloy: I think this should be using `stop_running` instead of reaching into `you.running` directly. `stop_running` also supports a `clear_delays` parameter, defaulting to true. It seems like if you set this to false instead, you might be able to call it unconditionally and avoid the `you.running.runmode != 0` check. Does that sound right to you?

ufshaikh: I think what you're describing will work too.

I used `you.running.stop` just because the original did, but yes, changing it to `stop_running` seems reasonable.

Calling `stop_running(false)` in `do_god_gift` fixes the bug as well. But it will also run checks (which in the case we're concerned with should never be passed) in `runrest::stop` to see if the screen should be redrawn, and to see if the user's Lua function on stopping running should be called. It's not clear to me whether it is better to check there or in `do_god_gift`.

(If the check remains in `do_god_gift` I'll change the condition to just check `success && you.running`, not `you.running.runmode != 0`; sorry that was a little goofy.)

In any case, let me know what you think and I'll push one or both changes.

amalloy: I think as long as we never actually call the user's Lua hooks unnecessarily, I'd rather the check be simple (i.e., just call `stop_running(false)`) even if this is slightly wasteful. If there's some chance we call the user's Lua hooks when we're not actually planning to stop running, then I'd prefer to avoid that even if it makes the check more complicated.

ufshaikh: I've pushed a new version; sorry such a small change turned into such a task.

PR #983, opened 2019-02-10
Expose MR and Stealth to Lua

alexjurkiewicz:
Two new functions return the number of pips for each of these
resistances.


amalloy:
Looks good to me, but I don't want to add a new "feature", obscure though it is, during tournament. Feel free to bump this PR after t to remind me to merge it.

PR #989, opened 2019-02-18
Support profile-guided optimisation (PGO)

alexjurkiewicz:
PGO produces (slightly) faster binaries. In the best case I've been able
to benchmark (re-running an idential short qw run) with and without PGO,
there is a 8% benefit. Real world benefit is likely to be much smaller,
of course.

TODO:
* Support GCC PGO
* Support Linux
* Use a broader test corpus, and run tests in parallel
* Improve the interface for creating a PGO build (currently it's a shell
  script, but could this be folded back into a Makefile target...?)


rawlins:
This PR has gone stale and given that it's still in draft state, I'm going to close it. Feel free to reopen of course.

PR #991, opened 2019-02-19
Fix Ozocubu's Refrigeration (11885)

NormalPerson7:
Since fb95405, the damage of both the player and monster versions of
Ozocubu's Refrigeration has been one third of what it should be,
specifically 3d((5 + power/10)/3), leading to a very underwhelming
level 6 spell.

This was because fb95405 converted the original dice_def that was used
to calculate the damage from this spell into a calcdice_calculator,
without adjusting for the fact that the calc_dice function divides the
number of sides on each die by the number of dice.

One way to fix this would be to use a dicedef_calculator instead;
however a calcdice_calculator is preferred for the fact that it randomly
rounds power/N instead of always using integer division (which creates
minor breakpoints). Therefore this commit instead fixes the damage by
tripling the inside of the dice roll bracket, leading to a
3d((15 + 3*power/10)/3) dice roll which is equivalent to the old
3d(5 + power/10).


amalloy:
I looked through the commit referenced and traced through the dice code. I agree with this PR's assessment: fb95405 accidentally cut damage to a third of what it previously was. 

Is a similar change needed for drain life? It looks to me like previously the damage for that was `3 + random2(7) + random2(pow)`, a maximum of 3+6+(pow-1)/2 damage, or around 9 + pow/2, averaged over two different-sized random2 calls. The new calculation, `calcdice_calculator<2, 4, 1, 2>`, seems to mean "a maximum of 4 + pow/2, averaged over two dice". This seems to be about 4 1/2 damage weaker on average than the old version, and probably not intentional.

NormalPerson7:
Given the commit message you're right, drain life is also affected.

NormalPerson7:
Actually no: drain life's formula was subsequently changed in ba00d9b to a way that used one dice only, and hence is not affected by the division in `calc_dice`.

amalloy:
Ah, well spotted. I only looked at the commit that changed those two, and not the current codebase. Silly me.

PR #993, opened 2019-02-26
Remove incorrect fake_mutations from vampire (11870)

josbraden:
Fixes bug #11870

Bug was introduced when species data was moved to yaml in 0.23


rawlins:
Thanks for the fix!

PR #994, opened 2019-03-01
Prevent allied Shambling Mangroves from constricting themselves (11877)

nlavsky:


ebering:
Thanks for the patch!

PR #995, opened 2019-03-02
Merge Piercer into the storm bow

kitchen-ace:
The storm bow is functionally the same as a +8 elec longbow; fairly
strong, but unremarkable. Piercer is a little better since it's a bow
user's only way to get a penetration weapon, but that's still not too
special.

Adding a penetration effect to the storm bow makes it much more
exciting, fits the electric/lightning theme, and puts it almost on par
with the other unrand bow Zephyr. This also makes Piercer obsolete.

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -539,6 +539,7 @@ BRAND:    SPWPN_VORPAL
 BOOL:     seeinv
 INSCRIP:  Acc+∞
 
+# TAG_MAJOR_VERSION == 34
 NAME:    longbow "Piercer"
 OBJ:     OBJ_WEAPONS/WPN_LONGBOW
 PLUS:    +7
@@ -547,6 +548,7 @@ TILE:    urand_piercer
 TILE_EQ: great_bow
 BRAND:   SPWPN_PENETRATION
 EV:      -2
+BOOL:    nogen
 
 # TAG_MAJOR_VERSION == 34
 ENUM:    BLOWGUN_ASSASSIN
@@ -654,6 +656,7 @@ COLOUR:   BLUE
 TILE:     urand_storm_bow
 TILE_EQ:  storm_bow
 BRAND:    SPWPN_ELECTROCUTION
+INSCRIP:  penet
 
 NAME:    large shield of Ignorance
 OBJ:     OBJ_ARMOUR/ARM_LARGE_SHIELD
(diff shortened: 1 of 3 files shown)


crawl-ref/source/art-data.txt
```diff
@@ -547,6 +547,7 @@ TILE:    urand_piercer
 TILE_EQ: great_bow
 BRAND:   SPWPN_PENETRATION
 EV:      -2
+BOOL:    nogen
```
ebering: The start of this entry should have `# TAG_MAJOR_VERSION == 34` to tag Piercer for removal.

crawl-ref/source/art-data.txt
```diff
@@ -647,13 +648,15 @@ TILE:     urand_cutlass
 TILE_EQ:  capt_cutlass
 BRAND:    SPWPN_SPEED
 
+ENUM:     STORM_BOW
```
ebering: This line is superfluous, enum names are automatically generated from `NAME` unless there's a need not to. It's better to not add one.

kitchen-ace:
Thanks ebering, I'll make those changes this weekend.

How is this for an updated description? (The general description for the elec brand would also still be present.)

"A longbow that imbues the arrows it fires with lightning, allowing them to pass through any targets they hit. It smells like a storm just past, or one about to begin."

ebering:
Something shorter, perhaps: "A longbow with the colour of dark rain clouds that fires arrows like lightning."

The information about its brands should be reflected automatically by `describe.cc`, which reminds me that you might need to add a check for the storm bow so that it gets the description both for piercing and for electric brands.

kitchen-ace:
Force pushed the changes. (I hope I did that right...)

I haven't done anything to make sure it gets the description for both brands. Is the best way to do this just a special case in `describe.cc`? (I was going to use Maxwell's Thermic Engine as an example, but it only gets the flaming brand description, which is maybe a bug.)

ebering:
Hm. Yes, a special case in `describe.cc` is the way to go; adding a separate commit to fix the bug with Maxwells to this PR would be appreciated.

ebering:
The bug fix can go in a separate PR on second thought, since it's not that big of a deal and the inscription communicates it well.

Are you in CREDITS.txt? If not, how would you like to appear?

kitchen-ace:
Thanks! I'll take a stab at editing describe.cc when I get a chance.

I'm not in the credits. I'd like "G. Bailey" if that works.

PR #999, opened 2019-03-03
Improve OTR's behaviour around allies and god conducts (11181, 11913)

NormalPerson7:
This pull request has now been updated to reflect the current consensus.

This pull request adds a warning for summoning non-rPois monsters while the Toxic status is active, lets OTR deal damage on the turn it is cast in addition to subsequent turns, and lets OTR trigger god conducts but only on the turn it is cast.

See the commit messages for more details, and the discussion below for what led to this solution.

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -4189,6 +4189,20 @@ void bolt::handle_stop_attack_prompt(monster* mon)
         beam_cancelled = true;
         finish_beam();
     }
+    // Handle enslaving monsters when OTR is up: give a prompt for attempting
+    // to enslave monsters that don't have rPois with Toxic status.
+    else if (flavour == BEAM_ENSLAVE && you.duration[DUR_TOXIC_RADIANCE]
+             && mon->res_poison() <= 0)
+    {
+        string verb = make_stringf("enslave %s", mon->name(DESC_THE).c_str());
+        if (otr_stop_summoning_prompt(true, verb,
+                                      mons_genus(mon->type) == MONS_ORC))
+        {
+            beam_cancelled = true;
+            finish_beam();
+            prompted = true;
+        }
+    }
 
     if (prompted)
     {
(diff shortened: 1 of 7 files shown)


kitchen-ace:
Hi NP7, thanks for working on this issue.

Regarding penance and OTR, there are a few situations I can think of where a player might put themselves under penance accidentally.

The first is with monsters appearing from out of LOS, into LOS when OTR is active. This could potentially happen with Beogh (conversion on sight, plus conversion from low HP), Jiyva, Gozag, the good gods (with neutral or peaceful monsters), or in rare cases with good gods and Oka after teleporting and having a summoned ally find their way back to you. 

In all those cases, using OTR seems tactically dubious even without penance, but I wonder if the potential for sudden penance is too strong regardless.

A similar issue could crop up with Jiyva and shapeshifters. It's never been clear to me as to whether shapeshifters' interaction with Jiyva is intentional, but it would make sense -- I would go so far as to reason that shapeshifters should always turn neutral to a player worshipping Jiyva, but that's a separate issue.

Lastly, a player could use OTR with Beogh to prevent getting orcish followers in situations where it would be undesirable. This is pretty extreme edge-case territory, and again possibly a separate issue, but it definitely comes up at times where the possibility of sudden orc allies is not desirable, e.g. when using immolation.

NormalPerson7:
I guess maybe the best solution to this that balances everything is to only trigger penance on actually casting the spell, and not subsequent turns? What do you think?

kitchen-ace:
>I guess maybe the best solution to this that balances everything is to only trigger penance on actually casting the spell, and not subsequent turns?

That sounds like a great solution to me.

NormalPerson7:
Hmm to implement this will require damaging monsters on the turn OTR is cast, and not just subsequent turns. Is there an issue with changing this? (Open question for anyone)

ebering:
I've been a bit less active the past few months, but I think that giving penance on cast and not subsequent actions (implemented by adding a turn of damage to OTR) is the good solution here. Clouds are still in a weird place, but you don't get penance if something walks into your cloud after you made it, it wasn't *your fault* the monster made a poor life choice.

NormalPerson7:
I will aim to implement the penance-on-cast fix for this conduct in subsequent days.

Regarding clouds, there's a lot I could say and I have a preferred solution for it as well, but I'll leave that for another time and another PR.

PR #1001, opened 2019-03-04
Use OR instead of equal for holiness check (11909)

josbraden:
Fixes bug 11909

Affects hell rats because they get the MH_EVIL flag applied as well as MH_DEMONIC, the bug would also affect any demonic monster with more than one holiness flag.

Probably just overlooked in c932079a4eb94c70e47aebf39e9aad417afd7f93 back in 2014 but I didn't do a lot of digging.

Thanks!

### DIFF
--- a/crawl-ref/source/mon-death.cc
+++ b/crawl-ref/source/mon-death.cc
@@ -1716,7 +1716,7 @@ static void _fire_kill_conducts(monster &mons, killer_type killer,
 
     mon_holy_type holiness = mons.holiness();
 
-    if (holiness == MH_DEMONIC)
+    if (holiness & MH_DEMONIC)
         did_kill_conduct(DID_KILL_DEMON, mons);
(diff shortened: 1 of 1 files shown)


amalloy:
I was about to merge this, but suddenly I'm not sure...because this is in an `else if` cascade, this change also affects other holiness combos involving `MH_DEMONIC`. 

The only example of this that I could find was demonspawn: they are `MH_NATURAL | MH_DEMONIC`. Before your change, they fell into the second clause (nondemonic, but natural, and evil), but now they fall into the first (demonic). This sounds like an improvement to me.

But what's up with hell rats? As far as I can see they are just `MH_DEMONIC`, nothing else. What made you think they have some other holiness?

rawlins:
I think anything with a drain attack gets `MH_EVIL` applied in `monster::holiness`: https://github.com/crawl/crawl/blob/master/crawl-ref/source/monster.cc#L3530

PR #1006, opened 2019-03-09
Start new wanderers with weapon in slot 'a'

joshgelbard:
After this commit, all new characters who start with a weapon wielded
will find that weapon in slot 'a'. This was already true for all
backgrounds except wanderer.

### DIFF
--- a/crawl-ref/source/ng-setup.cc
+++ b/crawl-ref/source/ng-setup.cc
@@ -545,6 +545,9 @@ static void _setup_generic(const newgame_def& ng)
         item_colour(item);  // set correct special and colour
     }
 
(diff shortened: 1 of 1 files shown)


amalloy:
This looks good to me, but I keep forgetting to test it when I have easy access to a compiler. Just wanted to comment here to apologise for not saying anything earlier. I assume you tested this yourself? If so, I'll just take your word for it and merge; if it breaks --Wn somehow we can roll it back.

joshgelbard:
No worries, not in a hurry. I ran ./crawl -test and also just made a bunch of characters, so I'm pretty confident this works.

PR #1007, opened 2019-03-09
Fix splash noises when dropping item into water

joshgelbard:
Added a missing `!`. Behavior was reverse of intended: instead of
swimming characters being the only ones exempt from item splash sounds,
they were the only ones making them.

### DIFF
--- a/crawl-ref/source/items.cc
+++ b/crawl-ref/source/items.cc
@@ -2614,7 +2614,7 @@ bool drop_item(int item_dropped, int quant_drop)
 
     // If you drop an item in as a merfolk, it is below the water line and
     // makes no noise falling.
-    if (silenced(you.pos()) || you.swimming())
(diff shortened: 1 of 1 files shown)


crawl-ref/source/items.cc
```diff
@@ -2614,7 +2614,7 @@ bool drop_item(int item_dropped, int quant_drop)
 
     // If you drop an item in as a merfolk, it is below the water line and
     // makes no noise falling.
-    if (silenced(you.pos()) || you.swimming())
```
joshgelbard: I removed this "are we silenced" check because that gets checked in prepare_message anyway.

ebering:
Thanks for the patch!

joshgelbard:
"Joshua Gelbard" would be great, thanks.

PR #1008, opened 2019-03-10
Reflect the Gnoll as a Jack of All Trades When Choosing Weapons

1Hyena:
Since gnolls are described in the wiki as hybrid characters that
should soften up targets with ranged weapons/spells before
engaging in melee, it makes sense for them to enter the dungeon
wielding any of the basic weapons, regardless of the profession.

To achieve this, we have refactored the weapon choosing dialogue
in the new game menu to provide gnolls with a mixed set of arms
to choose from. However, as the vertical space in this dialogue
is rather limited, we have implemented a mechanism that assigns
lower priority to restricted weapon choices that are not likely
to be preferred by the player anyway. As a result, restricted
weapon choices appear always in the end of the list.

While this may seem like a trivial change that has no practical
effect on the gameplay of a Gnoll character, it actually comes
with several significant implications. Firstly, gnoll fighters
and gladiators will have an option to enter the dungeon wielding
an upgraded ranged weapon. But secondly (and most importantly),
gnoll chaos knights can choose a ranged weapon for their initial
weapon, which will then receive the chaos brand, giving a whole
new dimension of fun to the Xom experience.

The latter is actually consequential enough to highlight the
Chaos Knight speciality as one of the recommended jobs for the
Gnoll species. With a ranged chaos brand weapon a Gnoll would
keep Xom entertained with the ranged attacks and switch to a
more powerful melee weapon for close combat. As a result, Xom
becomes just a little bit safer to amuse, giving the player that
crucial sense of fairness which is easily lost when playing as a
Chaos Knight.


ebering:
From a gameplay perspective, I am not in favor of merging this. As much as possible, backgrounds should not depend on the species of the player. In our current set up the differences in equipment between backgrounds are as small as possible subject to racial slot constraints.

Changing gnolls in this way would be a significant buff to them (as an already strong species), and distorts the inter-background balance. Even if allowing backgrounds to depend more strongly on species was desired, this is in my opinion too radical a departure.

1Hyena:
@ebering 
Then how about adding a sling to the list of available weapons for the chaos knight?

PR #1009, opened 2019-03-12
Traps cleanup

ufshaikh:
These are few commits cleaning up some minor things I noticed while digging around in traps.cc.

The last commit doesn't seem like an especially elegant way to do this to me---I'm open to suggestions.

### DIFF
--- a/crawl-ref/source/traps.cc
+++ b/crawl-ref/source/traps.cc
@@ -1373,9 +1373,6 @@ void roll_trap_effects()
  */
 void do_trap_effects()
 {
-    const level_id place = level_id::current();
-    const Branch &branch = branches[place.branch];
-
     // Try to shaft, teleport, or alarm the player.
     int roll = random2(3);
     switch (roll)
@@ -1383,9 +1380,7 @@ void do_trap_effects()
         case 0:
             // Don't shaft the player when we can't, and also when it would be into a
             // dangerous end.
-            if (is_valid_shaft_level()
-               && !(branch.branch_flags & brflag::dangerous_end
-                    && brdepth[place.branch] - place.depth == 1))
+            if (is_valid_shaft_effect_level())
             {
                 dprf("Attempting to shaft player.");
                 you.do_shaft();
--- a/crawl-ref/source/traps.cc
+++ b/crawl-ref/source/traps.cc
@@ -1320,7 +1320,7 @@ bool is_valid_shaft_effect_level()
     // Don't shaft the player when we can't, and also when it would be into a
     // dangerous end.
     return is_valid_shaft_level()
-           && !(testbits(branch.branch_flags, brflag::dangerous_end)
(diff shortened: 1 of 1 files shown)


crawl-ref/source/traps.cc
```diff
@@ -1320,7 +1320,7 @@ bool is_valid_shaft_effect_level()
     // Don't shaft the player when we can't, and also when it would be into a
     // dangerous end.
     return is_valid_shaft_level()
-           && !(testbits(branch.branch_flags, brflag::dangerous_end)
```
ufshaikh: Looking now at the git blame, I see that someone added in the use of the testbits function here with the idea that this use of bitwise & in a return statement wouldn't compile? (6e6263181c) I don't really know what they were thinking; it seemed to compile fine for me and now seems to be passing Travis' tests. But I thought I'd flag it since I'm not sure what they had in mind.

NormalPerson7: In this instance I believe it compiles fine, because it's converted to bool before the return statement acts. Return statements of the form `return a & b`, where `b` is of an enum class type, in a boolean function/method however don't compile, and this particular change was caught up in those sorts of fixes. What exactly is the issue with using `testbits` though?

FWIW I now use `static_cast`s to bool for return statement compile errors when converting enums.

ufshaikh: Oh, great!---thanks for explaining!

Why not use testbits here? Well of course I will defer to you more knowledgeable folks, but fwiw here was my reasoning:

- in the case we care about, where the test field is has a single bit turned on, using testbits doesn't change the behavior compared to a & b (or rather static_cast<bool> (a & b))
- using & to query a bit is common and readability isn't helped by calling testbits in addition
- in general, if some verbosity doesn't change the behavior or make the code more readable, it should be dropped

crawl-ref/source/traps.cc
```diff
@@ -1368,46 +1368,66 @@ void roll_trap_effects()
...
+        else
+            trap_effect = random_choose(TRAP_ALARM, TRAP_TELEPORT);
+    }
+    else
+    {
+        if (shaft_allowed)
+            trap_effect = random_choose(TRAP_SHAFT, TRAP_TELEPORT);
+        else
+            trap_effect = TRAP_TELEPORT;
+    }
```
ebering: This can be simplified by making a vector of available trap effects, pushing back traps and teleports and then using the `random_iter` template in random.h

crawl-ref/source/traps.cc
ebering: `&` is fine here.

crawl-ref/source/traps.cc
ufshaikh: Ah, sorry for letting that last change sit for so long, and thanks for wrapping it up for me!

ebering: No problem!

PR #1012, opened 2019-03-16
Give a Scroll of Brand Chaos Weapon to Unarmed Chaos Knights

1Hyena:
The aim of this improvement is to make the gameplay of a Chaos Knight
a bit more fun and fair for those who are actually seeking enjoyment
from the Xom experience, rather than doing it out of pure masochism.
Having played Crawl for countless hours as a Chaos Knight, it is of
course the most intense adventure one could get. It is fun and it is
fast, but lacks the sense of fairness. After all, the popular tactics
described on the Internet suggest you to throw away your chaos brand
weapon and reject Xom as soon as possible. Clearly, something is off
there, and as a professional game designer and developer, I hereby
propose a simple but effective solution.

When a Chaos Knight enters the dungeon unarmed, a gift from Xom will
be sitting in their inventory. It is a scroll of brand weapon that
will produce the chaos brand, if used on a non-chaos weapon. Such a
behavior is achieved by modifying the weapon rebranding methods, so
that scrolls gifted by Xom would produce the chaos brand as a first
priority. Should the weapon already have a chaos brand, the rebrand
would produce something else. This only affects scrolls specifically
gifted by Xom though. Scrolls randomly found from the dungeon remain
the same as they have always been.

In most of the cases, entering the dungeon unarmed can be considered
a disadvantage for the player. As an unarmed Chaos Knight, the start
is even more rough, because Xom gets bored fast if you are not using
a chaos branded weapon. However, should such a player survive until
they come across a worthy weapon, the Chaos Knight could utilize the
early gift from Xom and get a ranged chaos weapon such as the sling.

The ability to use a ranged chaos weapon gives that missing piece of
fairness to the gameplay of a Chaos Knight, mostly because it keeps
Xom amused while enabling the player to switch to a non-chaos melee
weapon during close combat. Such an advantage is important to have
because it provides an effective strategy for the Chaos Knight in a
way that still enforces role-play (sticking to the chaos weapon).

When applied, playing as a Chaos Knight will not only be fun, fast
and fair, but it will also be suitable for a wider range of players.
Considering the intricacies of Xom and the amount of effort that has
been put into the development of that particular deity, it would be
a shame if this potential remained in obscurity due to the lack of
variety in the strategies of the Chaos Knight.


MarvinPA:
Closing this, per ##crawl-dev discussions.

PR #1013, opened 2019-03-17
Add a menu shortcut to select last unequipped item

joshgelbard:
In equipment-related menus and prompts, make ';' (semicolon) an alias
for the letter of the player's most recently unequipped item. The goal
of this change is to reduce the amount of time the player spends
scanning through menus to remember which item is where. Examples:

* Wear-IDing a good cloak ('f') and dropping your old one: Wfd;<Enter>
* Wear-IDing a bad cloak ('f') and re-equipping the old: WfW;d;<Enter>
* Swapping to a utility ring ('g') for a certain enemy, then going back
to your usual ring: Pg<P;<

In those examples, we saved a little mental effort by not having to keep
track of the location in inventory of the old cloak and ring.

Notes/possible issues:

The semicolon key is losing existing, though undocumented, behavior as
one of several ways send a page-up key (besides '<', PGUP itself, and
inconsistently '-').

The fact that this alias is usable on the (d)rop menu is documented both
in crawl_manual.txt and on the d_ help screen, but for i, w, W, and P
it's manual-only. (Those other help screens don't yet exist.)

### DIFF
--- a/crawl-ref/docs/crawl_manual.rst
+++ b/crawl-ref/docs/crawl_manual.rst
@@ -2644,7 +2644,7 @@ will deselect it (except for ',' and '-', obviously).
 &amp;
   Select all carrion and inedible food.
 
-\+ or :
+:
   Select all books.
 
 /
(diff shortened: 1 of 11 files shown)


joshgelbard:
And there's the second commit fixing the menu +/- behavior as discussed on IRC.

If + is going to be pagedown everywhere, you'd expect - to be pageup everywhere, but it can't be because of menus like (w)ield. I don't know that there's a good fix for that besides just getting rid of them. For now, this should be at least a slight improvement.

PR #1014, opened 2019-03-19
Reduce troll starting mutations

alexjurkiewicz:
Old: 3 levels of claws, tough skin and fast metabolism at XL 1.

New:
```
XL      1   9  14
======+===+===+==
Claws | 2 | 3 |
Skin  | 1 | 2 | 3
Metab | 2 |   | 3
```

Even with these mutations coming later, trolls still have (and should
have) a very strong early game. However, it's hopefully a reduction from
nigh-invincible to something with at least a little uncertainty.

Claws upgrade to XL9 rather than later so there's no possibility of
avoiding XP to deal with hydrae.

### DIFF
--- a/crawl-ref/source/dat/species/troll.yaml
+++ b/crawl-ref/source/dat/species/troll.yaml
@@ -48,12 +48,18 @@ levelup_stats:
   - str
 mutations:
   1:
-    MUT_TOUGH_SKIN: 3
     MUT_REGENERATION: 1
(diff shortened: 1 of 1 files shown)


ebering:
It's an interesting idea, but I disagree with thee assessment that trolls need a nerf. The lack of armor options and low dodging apts mean that while they can tank through a lot of plinking a high roll or two from an ogre can crush a troll.

PR #1015, opened 2019-03-20
Don't assume some identified jewellery is boring

joshgelbard:
Make some autopickup-related lua code less eager to forget about
potentially-useful jewellery.

Old defaults for autopickup_exceptions would remove some items (e.g.
ring of poison resistance) from autopickup as soon as the subtype was
known, since you only ever need one. Now that it's possible to learn
what an rPois ring looks like by seeing a monster wear it, that behavior
occasionally leads to useful rings and amulets being left unnoticed on
the ground. To prevent that, don't assume such only-need-one items are
safe to ignore unless the player is actively carrying one around.

### DIFF
--- a/crawl-ref/source/dat/defaults/autopickup_exceptions.txt
+++ b/crawl-ref/source/dat/defaults/autopickup_exceptions.txt
@@ -70,11 +70,32 @@ ae += [^n]identified.*spellbook
 :   return it.stacks() or nil
 : end)
 
-# Excluding amulets as you only need one of each. (If you know the subtype
-# that means you already have one of it.). Some items may already be excluded
-# as bad_item, e.g. inaccuracy.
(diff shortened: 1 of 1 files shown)


ebering:
I'm a bit worried that people will get annoyed after intentionally dropping these items but not updating their auto pickup; however that annoyance is pure ui and lessens the gameplay negatives relative to the status quo so I think this should be good.

joshgelbard:
Yeah, even if it's "less wrong" I think it might be annoying. I'm working on a separate patch that makes it much easier to remove unwanted items from autopickup as you drop them -- if this PR gets left open, I can reference it when I submit the other and see if it makes more sense at that point.

ebering:
Thanks for the patch. Sorry for the delay but I've had some time for testing/reviewing this weekend. You might need to rebase/revise #1025 so that it applies cleanly and works.

PR #1024, opened 2019-03-30
Fix spurious message on moving into door with wall_jump_move (11940)

NormalPerson7:
When attempting to move into a closed door with the wall_jump_move
option set to true, the function wu_jian_can_wall_jump was being called,
which meant spurious messages describing why wall jumping was impossible
in the current situation were being printed to the message log.

To fix this, move the check for moving into a closed door (a situation
where it is not possible to wall jump by moving) to inside the
definition of bool can_wall_jump, before the function
wu_jian_can_wall_jump() is called.

### DIFF
--- a/crawl-ref/source/movement.cc
+++ b/crawl-ref/source/movement.cc
@@ -537,8 +537,13 @@ void move_player_action(coord_def move)
 
     const coord_def targ = you.pos() + move;
     string wall_jump_err;
-    bool can_wall_jump = Options.wall_jump_move &&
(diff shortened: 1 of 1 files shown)


rawlins:
I think the original order of checks was because `grd(targ)` will crash when the target is out of bounds, e.g. the edge of a level. So you may need to add an in_bounds check directly to the call, something like `(!in_bounds(targ) || !feat_is_closed_door(grd(targ)))` instead of the new conjunct maybe?

NormalPerson7:
Good spot. I've applied the fix you suggested.

PR #1026, opened 2019-04-02
Update Gyre and Gimble

aolowin:
Remove the -3 Dex and add the protection brand which is not currently
used by any unrand and compensates somewhat for locking out a shield.
The special "vorpal" brand halves the AC of the target. With this
effect damage is now slightly less vs. unarmoured targets and slightly
more vs. higher AC targets. Fsim shows per-turn damage somewhere
close to the Autumn Katana with the changes, although extra slaying tilts
things to the quick blades' favour.

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -1212,9 +1212,8 @@ TYPE:    pair of quick blades
 COLOUR:  ETC_ENCHANT
 TILE:    urand_gyre
 TILE_EQ: gyre
-PLUS:    5
-BRAND:   SPWPN_VORPAL
-DEX:     -3
+PLUS:    7
+BRAND:   SPWPN_PROTECTION
 
 ENUM:    ETHERIC_CAGE
 NAME:    Maxwell's etheric cage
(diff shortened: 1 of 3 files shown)


ebering:
I'm not sure about the new brand. Protection is flavorful (blocking with SWORDCHUCKS), but G&G are already distinctive because they're SWORDCHUCKS, so I'm not sure they really need more. I'd have to look back at the original commits to see about the Dex-3, but it can probably be removed. I don't have time to dig through the git logs at the moment.

aolowin:
The -3 Dex was originally just flavor - the chain getting in the way.

The main reasoning behind the the new "vorpal" brand (aside from the Jabberwocky reference) is that the low base quickblade damage means the existing piercing brand doesn't do much and G&G gets outclassed by a highly-enchanted (and single handed) quickblade. The new brand nerfs the damage a bit on low-AC targets (where it was already pretty high) and bumps it on high-AC targets where the extra attacks were just bouncing off.

The extra Protection brand just seemed flavorful so could stay or go.

MarvinPA:
I'm also not especially a fan of the proposed new brand - being less effective against very high-AC targets is the expected and intended behaviour of quick blades generally and I think it's fine for that to carry over to the unrand version. Calling the brand "vorpal" when that's already the internal name for a brand is also a bit messy maybe (although as a separate thing, I'd like to unify the descriptor for that brand so that it doesn't have different names per weapon anyway).

Removing the Dex-3 definitely sounds fine, and maybe giving it protection in exchange for just slightly increased enchantment would also be reasonable?

ebering:
In the future you don't need to revert/recommit. You can just force-push to your PR.

PR #1027, opened 2019-04-03
Warn the player that teleportation doesn't work in gauntlet

heuripedes:
Gauntlet prevents long distance translocations from happening (NoTele),
but the user won't know about this until they check the % screen,
read a scroll of teleportation or notice that teleportitis hasn't
triggered since they entered the branch.

### DIFF
--- a/crawl-ref/source/stairs.cc
+++ b/crawl-ref/source/stairs.cc
@@ -786,10 +786,10 @@ void floor_transition(dungeon_feature_type how,
                 mpr(branches[branch].entry_message);
             else if (branch != BRANCH_ABYSS) // too many messages...
                 mprf("Welcome to %s!", branches[branch].longname);
-
-            if (how == DNGN_ENTER_GAUNTLET)
-                _gauntlet_effect();
         }
+
+        if (branch == BRANCH_GAUNTLET)
+            _gauntlet_effect();
 
         const set<branch_type> boring_branch_exits = {
             BRANCH_TEMPLE,
--- a/crawl-ref/source/stairs.cc
+++ b/crawl-ref/source/stairs.cc
@@ -412,7 +412,7 @@ static void _gauntlet_effect()
         || you.wearing(EQ_RINGS, RING_TELEPORTATION, true)
         || you.scan_artefacts(ARTP_CAUSE_TELEPORTATION, true))
     {
-        mpr("Your feel stable.");
+        mpr("Your feel stable on this floor.");
     }
 }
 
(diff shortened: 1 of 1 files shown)


crawl-ref/source/stairs.cc
```diff
@@ -399,6 +400,22 @@ static void _rune_effect(dungeon_feature_type ftype)
...
+
+    mprf(MSGCH_WARN, "The nature of this place prevents teleportation.");
+
+    if (you.has_mutation(MUT_TELEPORT, true)
+        || you.wearing(EQ_RINGS, RING_TELEPORTATION, true)
+        || you.scan_artefacts(ARTP_CAUSE_TELEPORTATION, true))
+    {
+        mpr("Your feel stable.");
+    }
+}
```
NormalPerson7: Hmm I wonder: do you think this "You feel stable" message is necessary? I think it might just confuse people with -Tele items/teleportitis—they won't necessarily know why this message is being printed. Just putting it out there.

crawl-ref/source/stairs.cc
```diff
@@ -769,6 +786,9 @@ void floor_transition(dungeon_feature_type how,
                 mpr(branches[branch].entry_message);
             else if (branch != BRANCH_ABYSS) // too many messages...
                 mprf("Welcome to %s!", branches[branch].longname);
+
+            if (how == DNGN_ENTER_GAUNTLET)
+                _gauntlet_effect();
         }
```
NormalPerson7: I'd prefer this to be `if (branch == BRANCH_GAUNTLET)` I think.

There are two ways for the player to enter a gauntlet: through the portal, and from leaving the abyss. I think this message should be printed in both cases.

crawl-ref/source/stairs.cc
heuripedes: I believe they will know why the message is being printed because it comes immediately after the warning. Originally I used `You feel temporarily stable` but I thought it could be misinterpreted as something related to time/delay.

crawl-ref/source/stairs.cc
heuripedes: I'll implement it, thanks.

crawl-ref/source/stairs.cc
heuripedes: I have changed the message to reflect that it only affects the current floor.

ebering:
Thanks for the patch! How would you like to appear in `CREDITS.txt`?

heuripedes:
I'm ok with "Higor Eurípedes". Thanks for merging

PR #1030, opened 2019-04-09
Datify background definitions

NormalPerson7:
This PR follows on from #721 to convert all background definitions and the files `job-data.h` and `job-type.h` to being automatically generated from yaml files.

Things to do before this PR goes live:
[ ] Deal with job groups on the background selection screen, in the same way as species-groups.h deals with species groupings. The infrastructure on the python side is already there and included with the initial commit, but I haven't had time to make the adjustments on the C++ side to sort this out. The large blocks of commented out code are to do with this.
[ ] Update the documentation for creating backgrounds and add a README to dat/jobs/.

Optional things:
[ ] Convert `weapon_choice` to an enum class to allow simply `weapon_choice: ranged` in the yaml files instead of `weapon_choice: WCHOICE_RANGED`
[ ] Move `weapon_choice` out of job-data.h.
[ ] Move the struct definition out of job-data.h and replace it with a description of the struct.
[ ] Convert `skills` and `recommended_species` fields to use more logical strings rather than enums—which one would you prefer?


crawl-ref/source/dat/jobs/abyssal-knight.yaml
```diff
@@ -0,0 +1,25 @@
...
+#category: Zealot
+#category_priority: 90
+str: 4
+int: 4
+dex: 4
+equipment:
+  - "leather armour"
+weapon_choice: WCHOICE_PLAIN
+recommended_species:
+  - SP_HILL_ORC
```
alexjurkiewicz: Since species are data as well, it would be nice to refer to species by their name, eg `- Hill Orc` instead of exposing the enum.

crawl-ref/source/util/job-gen.py
```diff
@@ -0,0 +1,342 @@
...
+    #        raise ValueError('recommended_jobs must not be empty (or'
+    #                                            ' difficulty must be False)')
+
+        # Set attributes
+        self['enum'] = validate_string(s['enum'], 'enum', 'JOB_[A-Z_]+$')
+        self['name'] = validate_string(s['name'], 'name', '..+')
+        self['short_name'] = s.get('short_name', s['name'][:2])
+        # TODO: currently done by skill enum values; change to be strings?
+        self['skills'] = skills(s.get('skills', {}))
+        self['str'] = validate_int_range(s['str'], 'str', -15, 15)
```
alexjurkiewicz: This might be overreach, but maybe str+int+dex should be validated to equal 0 or 12.

NormalPerson7: I wondered about this. From the point of view of adding new backgrounds, where fork developers may choose to stray from this rule though, I thought it best to leave it.

crawl-ref/source/dat/jobs/abyssal-knight.yaml
NormalPerson7: Yep, that's on my todo list.

crawl-ref/source/util/gen_util.py
```diff
@@ -0,0 +1,83 @@
...
+
+# import argparse
+import os
+import sys
+import traceback
+import collections
+import re
+
+# import yaml  # pip install pyyaml
+
```
NormalPerson7: Currently some of these import statements are not required in this module, but there's a fair chance that if datification of data files continues to take place over coming versions, they may be necessary in this module. As a result, I've left them commented out. Would you prefer them to be removed entirely, or uncommented, or is commenting them out ok?

NormalPerson7:
I have now squashed the fixup commits and centralised some common functions from both species-gen and job-gen into a new file. I now have a few questions on how to proceed:
1. The name of gen_util.py contains an underscore instead of a hyphen because of Python's requirements on the names of imported modules. job-gen.py and species-gen.py both contain hyphens; this is inconsistent. Is this inconsistency a problem, and if it is, what would you prefer the name(s) of the files to be?
2. There are some functions that are common between job-gen and species-gen that I haven't yet centralised, because I didn't feel they would be applicable to other, future datifications. Would you prefer these also were centralised, perhaps under a new section in gen_util? (I'm leaning towards yes; I just wanted to clarify this first.)
3. I've currently decided to avoid straying from the enum names for the starting skills and recommended species fields, mostly on the advice of advil. Is this still the best approach? (See the struck off bullet point from the todo list above.)
4. Lastly, regarding the checking process, I've tried my best to ensure there are no errors, and from what I can tell there aren't, but of course I can't be sure. When it comes to aiding the checking process, I have two ideas that might help, but I'd like to know which one(s) you prefer / if you have any other ideas.
   - Idea 1. Add to the script a way to guarantee the output header files are limited to 80 character width, like most files in the Crawl codebase. This would make it somewhat easier to compare files, as hopefully this would reduce the number of lines that differ between the old version of species-data.h/job-data.h and the new version.
   - Idea 2. Write a script either to convert the original data file into yaml files (essentially the reverse of job-gen.py), or to intelligently compare the new version of job-data.h to the old version. This would probably take quite a bit of work to achieve, but in the first case could have the advantage of making future datifications easier and simpler (especially if it was ever desired to turn monster data into yaml) by being easily expandable.
@alexjurkiewicz @rawlins I'm pinging you because you've had the most involvement so far in this PR and also with the original PR, but anyone is welcome to comment.

Ozaq:
I just looked into this to see how to integrate the changes on my CMake PR. One thing I run across is what python version is targeted? 3.5, 3.6, 3.7? (I hope 2 is out due to it being EOL soonish).

Also a requirements.txt might be good to encode the PyYaml min version. Also I noted that with latest PyYaml you will get an easy to fix deprecation warning. (PyYaml 5.1 and later)

Some background about that warning:https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

alexjurkiewicz:
It's basically the same tech as what is used for species header file generation. The script should be compatible with both py2 & 3.

rawlins:
> I hope 2 is out due to it being EOL soonish

This is one of those things where we need to stick with what the servers have, and that means supporting python 2 for the foreseeable future (I think 2.7 at least). There's no reason to not write new code that won't work in python 3 though. (The webtiles server is not currently python 3 compatible...)

Ozaq:
> > I hope 2 is out due to it being EOL soonish
> 
> This is one of those things where we need to stick with what the servers have, and that means supporting python 2 for the foreseeable future (I think 2.7 at least). There's no reason to not write new code that won't work in python 3 though. (The webtiles server is not currently python 3 compatible...)

Hmm... right I was simply assuming tooling to be a different beast. I don't mind using a different python version for code generation, i.e. python3, than running the server as it is just switching an venv. But ofc others might think differently. I did not consider that this might be interpreted as a question to update the webserver to be python3 compatible or to go even further and drop Python2 support.

ebering:
For someone picking this up #1833 includes a delver yaml file.

rawlins:
This branch is incomplete and unfortunately a few years stale. I definitely think it could be resurrected but unless someone does pick it up, I may close this soon.

NormalPerson7:
I may look into reviving this PR in the near future, if there's still interest.

pooka109:
Considering that more things have been datified lately, I'm pretty sure there's still interest.

NormalPerson7:
I've rebased and updated all of the old commits from five years ago so this PR is up-to-date once again. I can't work out why checkwhite is failing so any help there would be appreciated. One more change I'd like to make is (in line with the monster YAML) to change all the enum values for spells, species and skills to use plaintext instead, so I'll do that next.

pooka109:
Checkwhite is failing because cinder-acolyte.yaml, delver.yaml, and reaver.yaml have CRLF line terminators (Windows-style) instead of LF line terminators (Unix-style) like the other YAML files. Running source/util/checkwhite on them (at least on this end) seems to fix that, though.

NormalPerson7:
Okay I have done the conversions to plaintext. This should make the YAML files easier to read, in a similar style to how dat/mons was recently done. This PR is ready for review again.

Note I haven't changed the `spells` field to use plaintext because some spells, (e.g. Ensorcelled Hibernation, Brom's Barrelling Boulder), have enum values (e.g. `SPELL_HIBERNATION`, `SPELL_BOULDER` resp.) that are different from their in-game names, and I figure it is easier to maintain and make clear this distinction if we use the enum values in the YAML.

pooka109:
Applied, after resyncing with current trunk. Thank you.

NormalPerson7:
Ok, hopefully everything works okay. I'll be looking out for issues in the next few days, as I'm sure you all will.

PR #1034, opened 2019-04-14
Webtiles: Disallow user login when account has been banned

floraline:
This PR changes webtiles login code to look at the user flags in the DGL database. If the banned flag is set, webtiles will not allow the user to log in. Previously, these flags were not used in the webtiles server.

I'm looking for feedback in the following areas:

1. Code review on the current changes
2. A user who holds a valid login token can bypass the normal login procedure. This would allow a banned user to login, as long as they hold on to a valid token.
3. I'm not aware of any existing script or other nice way of banning a user. Manually setting the account flag to 'banned' will not kick off any user who is currently logged in on an active connection.

Looking for solutions for 2 and 3 above.

### DIFF
--- a/crawl-ref/source/webserver/ws_handler.py
+++ b/crawl-ref/source/webserver/ws_handler.py
@@ -579,11 +579,14 @@ class CrawlWebSocket(tornado.websocket.WebSocketHandler):
         self.init_user(login_callback)
 
     def login(self, username, password):
-        real_username = userdb.user_passwd_match(username, password)
-        if real_username:
+        real_username, fail_reason = userdb.user_passwd_match(username, password)
+        if real_username and fail_reason is None:
             self.logger.info("User %s logging in from %s.",
                                         real_username, self.request.remote_ip)
             self.do_login(real_username)
+        elif fail_reason:
+            self.logger.warning("Failed login for user %s: %s", real_username, fail_reason)
+            self.send_message("login_fail", reason = fail_reason)
         else:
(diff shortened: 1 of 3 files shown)


crawl-ref/source/webserver/ws_handler.py
```diff
@@ -392,13 +392,17 @@ def do_login(self, username):
...
-        if real_username:
+        real_username, fail_reason = userdb.user_passwd_match(username, password)
+        if real_username and fail_reason is None:
             self.logger.info("User %s logged in.", real_username)
             self.do_login(real_username)
         else:
-            self.logger.warning("Failed login for user %s.", username)
-            self.send_message("login_fail")
+            if fail_reason:
+                self.logger.warning("Failed login for user %s: %s", real_username, fail_reason)
```
NormalPerson7: Python supports `elif` as else if, so you don't need an indented `if` block here.

rawlins:
Looking at this again, this code is probably mergeable as-is without the extra features (and those features could be added incrementally in trunk); could you unmark it as a draft?

(I recently needed to disable some griefer accounts on CAO and had to do it by changing the password to line noise, but something like this, even if it still needs to be accessed via dgl scripts, seems better. Though, is there a script to set the banned flag?)

floraline:
rebased to solve merge conflicts. It feels sloppy but the concept is there.

rawlins:
Thanks for this -- finally merged in 82aefee77902, and I also added a simple CLI in the linked commit.

PR #1035, opened 2019-04-14
Traps in full describe view

ufshaikh:
This is a minor UI tweak and a bit of code cleanup in full_describe_view (which handles the ctrl-x screen).

### DIFF
--- a/crawl-ref/source/directn.cc
+++ b/crawl-ref/source/directn.cc
@@ -569,8 +569,7 @@ void full_describe_view()
     for (radius_iterator ri(you.pos(),
                             you.xray_vision ? LOS_NONE : LOS_DEFAULT); ri; ++ri)
     {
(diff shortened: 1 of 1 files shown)


crawl-ref/source/directn.cc
```diff
@@ -557,7 +557,7 @@ class view_desc_proc
     void nextline() { cgotoxy(1, wherey() + 1); }
 };
 
-// Lists all the monsters and items currently in view by the player.
+// Lists monsters, items, and features (including traps) in the player's view.
```
amalloy: "Features" seems a little broad, since every possible tile is a "feature", including floor and wall. It's probably not worth explicitly listing every feature that's included, but maybe just say "and interesting features" or something?

ufshaikh: Sounds reasonable. I changed it to "some interesting features" since not all interesting features are listed, either; hopefully that strikes you as all right.

PR #1036, opened 2019-04-15
Add second brand to description for Thermic Engine and Storm Bow

kitchen-ace:
Duplicates the text description, but the only alternative I could come up with was special-casing the breaks in the switch statements which is probably worse?


ebering:
Hm, yeah, I don't see a good way to deduplicate this without a big refactor, so I'll merge it as-is for now. Thanks!

PR #1038, opened 2019-04-16
Fix local tiles monster panel showing summoner halo for all enemies

nlavsky:
This was happening because the packed_cell constructor was not
initializing `is_highlighted_summoner`.

halo_bug


MarvinPA:
Merged, thanks!

PR #1039, opened 2019-04-17
Use high-res app icon on all platforms

lightbulbjim:
Previously the 512x512 app icon was only used on MacOS, and all other
platforms used the 32x32 icon. This change is to use the 512x512 icon on
all platforms.

Note that I only have access to Linux, so that is all I've tested this on.

### DIFF
--- a/crawl-ref/source/tilesdl.cc
+++ b/crawl-ref/source/tilesdl.cc
@@ -351,11 +351,7 @@ bool TilesFramework::initialise()
     DATA_DIR_PATH
 #endif
 #endif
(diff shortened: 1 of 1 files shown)


lightbulbjim:
I've been able to test this on Windows (using MSYS2) and it works.

ebering:
Thanks for the patch! How would you like to be listed in CREDITS.txt?

lightbulbjim:
> Thanks for the patch! How would you like to be listed in CREDITS.txt?

No worries! Please list my name as "Chris Rendle-Short". Thanks!

PR #1040, opened 2019-04-17
The unrand staff of Battle

aolowin:
A staff of conjuration with an evokable 0-MP magic dart. It will
automatically spawn a battlesphere when monsters are in sight.

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -1326,11 +1326,12 @@ BRAND:    SPWPN_DISTORTION
 ENUM:     BATTLE
 NAME:     staff of Battle
 OBJ:      OBJ_WEAPONS/WPN_STAFF
+SH:       4
 PLUS:     +0
-COLOUR:   ETC_DISCO
-#TILE:     spwpn_staff_of_battle
+COLOUR:   ETC_JEWEL
+#TILE:     urand_staff_of_battle
 #TILE_EQ:  battle
-VALUE:    1000
+VALUE:    800
 
 # This entry must always be last.
 ENUM:   DUMMY2
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -1323,6 +1323,15 @@ TILE:     urand_rift
 TILE_EQ:  rift
 BRAND:    SPWPN_DISTORTION
 
+ENUM:     BATTLE
+NAME:     staff of Battle
+OBJ:      OBJ_WEAPONS/WPN_STAFF
+PLUS:     +0
+COLOUR:   ETC_DISCO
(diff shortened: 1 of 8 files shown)


crawl-ref/source/art-data.txt
```diff
@@ -1323,6 +1323,15 @@ TILE:     urand_rift
 TILE_EQ:  rift
 BRAND:    SPWPN_DISTORTION
 
+ENUM:     BATTLE
+NAME:     staff of Battle
+OBJ:      OBJ_WEAPONS/WPN_STAFF
+PLUS:     +0
+COLOUR:   ETC_DISCO
```
ebering: `ETC_DISCO` is special cased for disco_pan and won't animate in a weapon I don't think (I'd have to code dive).
`ETC_MAGIC` or `ETC_VEHUMET` are probably good choices.

crawl-ref/source/art-func.h
```diff
@@ -1426,3 +1426,35 @@ static void _ZHOR_world_reacts(item_def *item)
...
+    practise_casting(SPELL_MAGIC_DART, true);
+
+    return false;
+}
+
+static void _BATTLE_world_reacts(item_def *item)
+{
+    if (!find_battlesphere(&you)
+        && there_are_monsters_nearby(true, true, false)
+        && one_chance_in(3 * div_rand_round(BASELINE_DELAY, you.time_taken)))
```
ebering: I don't like the chance here. It encourages the user to kite/reset until the battle sphere appears. It should just be your ever present companion.

crawl-ref/source/item-prop.cc
```diff
@@ -1999,6 +1999,9 @@ bool item_skills(const item_def &item, set<skill_type> &skills)
         skills.insert(SK_EVOCATIONS);
     }
 
+    if (is_unrandom_artefact(item, UNRAND_BATTLE))
+        skills.insert(SK_CONJURATIONS);
```
ebering: Enhancer staves don't enable training of a skill. I know this one is a special case, since it brings a battlesphere, but the battle sphere isn't useful unless the player has other conj spells, in which case they'll be able to train conj just fine.

crawl-ref/source/art-func.h
```diff
@@ -1426,3 +1426,35 @@ static void _ZHOR_world_reacts(item_def *item)
...
+    {
+        your_spells(SPELL_BATTLESPHERE, 0, false);
```
ebering: `your_spells` bypasses god conduct checks (Trog). Trog should hate this, and Trog worshippers should be penannced for using it, since they'd get penance for training to use it.

ebering:
Thanks for this!

Sorry it took me so long to get to the merge; my real job has been outsized of late.

PR #1041, opened 2019-04-18

sdynet:
I think poison brand are enough for poison needles. And I'd also like to get rid of steel brands because I think they're nothing more than copying the top items.

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -1512,9 +1512,11 @@ static string _describe_ammo(const item_def &item)
                            "target, hitting an obstruction, or reaching its "
                            "maximum range.";
             break;
+#if TAG_MAJOR_VERSION == 34
         case SPMSL_STEEL:
             description += "It deals increased damage compared to normal ammo.";
             break;
+#endif
         case SPMSL_SILVER:
             description += "It deals substantially increased damage to chaotic "
                            "and magically transformed beings. It also inflicts "
--- a/crawl-ref/source/makeitem.cc
+++ b/crawl-ref/source/makeitem.cc
@@ -547,15 +547,11 @@ static special_missile_type _determine_missile_brand(const item_def& item,
     case MI_JAVELIN:
         rc = random_choose_weighted(30, SPMSL_RETURNING,
                                     32, SPMSL_PENETRATION,
-                                    32, SPMSL_POISONED,
-                                    21, SPMSL_STEEL,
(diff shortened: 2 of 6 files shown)


crawl-ref/source/makeitem.cc
```diff
@@ -647,6 +643,7 @@ bool is_missile_brand_ok(int type, int brand, bool strict)
     case SPMSL_EXPLODING:
         return type == MI_TOMAHAWK;
     case SPMSL_STEEL: // deliberate fall through
+		return false;
```
rawlins: I think this might be what checkwhite is finding (see the travis failure), extra whitespace here. In any case, you can run the `util/checkwhite` script to fix whitespace errors

crawl-ref/source/describe.cc
```diff
@@ -1512,9 +1512,11 @@ static string _describe_ammo(const item_def &item)
                            "target, hitting an obstruction, or reaching its "
                            "maximum range.";
             break;
+#if TAG_MAJOR_VERSION == 34
```
rawlins: In addition, I think this will need `#if TAG_MAJOR_VERSION == 34`s  in the relevant enum in `item-prop-enum.h` to remove the brand entirely on the major version increment.

crawl-ref/source/makeitem.cc
```diff
@@ -547,15 +547,11 @@ static special_missile_type _determine_missile_brand(const item_def& item,
...
-                                    32, SPMSL_POISONED,
-                                    21, SPMSL_STEEL,
                                     20, SPMSL_SILVER,
                                     nw, SPMSL_NORMAL);
         break;
     case MI_TOMAHAWK:
-        rc = random_choose_weighted(15, SPMSL_POISONED,
-                                    10, SPMSL_SILVER,
-                                    10, SPMSL_STEEL,
+        rc = random_choose_weighted(10, SPMSL_SILVER,
```
rawlins: These changes will lower the total weight of a branded tomahawk / javelin. Not sure if this is intentional; in fact it might be ok, but it also could make sense to distribute the weight over the remaining brands. (Some other dev might want to weight in.)

crawl-ref/source/makeitem.cc
NormalPerson7: (Specifically the issue here is using tabs instead of spaces.)

sdynet: I'll correct it and try again. Thank you for your advice.

PR #1042, opened 2019-04-18
Fix a couple Engulf-related bugs

joshgelbard:
Fix a bug where Engulf status wouldn't update on time if the player escaped via an instant action like Serpent's Lash movement or a retaliatory headbutt. (Instead of setting the remaining duration to one aut and waiting for the enchantment decrement step to finish it off, just clear it on the spot.)

Also remove buggy immunity to engulf. Escaping from being engulfed was supposed to have been conferring one turn of immunity, but the amount given has always been effectively zero. Since no one seems to have noticed, it's probably fine to just remove it.

handle_player_drowning() had a second copy of player::clear_far_engulf() living within it, so most of the diff is just fixing that.


ebering:
Thanks for the patch!

PR #1043, opened 2019-04-20
Two state vampires

ebering:
This branch simplifies vampires to two blood states: Alive and Bloodless (both match the current definitions of Alive and Bloodless). The player transitions between them using a 5 turn delay. Vampires are removed from the food system altogether, and made foodless but permitted to drink potions.

The reason for doing such a thing is detailed in the first commit message, the short version is that this revision is designed to remove blood micromanagement while preserving the "choose your undead state at strategic timescales" mechanic.

I've created a PR to collect devteam feedback and facilitate review, since this touches a lot of code.


ebering:
Based on play testing this has gone through some iterating.

At first, I tried an "automatic revivification" to alleviate tedium. However, it rapidly became apparent that state switching needs a true strategic cost.

The current state of this branch gives temporary frail for the Bloodless -> Alive switch, and stat drain for using bat form.

Making Bloodless and Alive balanced is an outstanding issue. The current sense is that bloodless is stronger than alive. Playtesters have suggested buffing alive in some way, my preference is to add a new drawback to Bloodless (No potion heal? No potions?) but I'm unsure of a good one.

PR #1044, opened 2019-04-22

sdynet:
They are removed to reduce inventory pressure.


sdynet:
The interval has not been adjusted.

PR #1045, opened 2019-04-22

sdynet:
They are removed to reduce inventory pressure.

### DIFF
--- a/crawl-ref/source/makeitem.cc
+++ b/crawl-ref/source/makeitem.cc
@@ -547,15 +547,11 @@ static special_missile_type _determine_missile_brand(const item_def& item,
     case MI_JAVELIN:
         rc = random_choose_weighted(30, SPMSL_RETURNING,
                                     32, SPMSL_PENETRATION,
-                                    32, SPMSL_POISONED,
-                                    21, SPMSL_STEEL,
                                     20, SPMSL_SILVER,
                                     nw, SPMSL_NORMAL);
         break;
     case MI_TOMAHAWK:
-        rc = random_choose_weighted(15, SPMSL_POISONED,
-                                    10, SPMSL_SILVER,
-                                    10, SPMSL_STEEL,
+        rc = random_choose_weighted(10, SPMSL_SILVER,
                                     12, SPMSL_DISPERSAL,
                                     28, SPMSL_RETURNING,
                                     15, SPMSL_EXPLODING,
@@ -635,7 +631,7 @@ bool is_missile_brand_ok(int type, int brand, bool strict)
     switch (brand)
     {
     case SPMSL_POISONED:
-        return type == MI_JAVELIN || type == MI_TOMAHAWK;
+        return false;
     case SPMSL_RETURNING:
         return type == MI_JAVELIN || type == MI_TOMAHAWK;
     case SPMSL_CHAOS:
@@ -647,6 +643,7 @@ bool is_missile_brand_ok(int type, int brand, bool strict)
     case SPMSL_EXPLODING:
         return type == MI_TOMAHAWK;
     case SPMSL_STEEL: // deliberate fall through
+        return false;
     case SPMSL_SILVER:
         return type == MI_JAVELIN || type == MI_TOMAHAWK;
     default: break;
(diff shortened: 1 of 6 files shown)


crawl-ref/source/makeitem.cc
```diff
@@ -647,6 +643,7 @@ bool is_missile_brand_ok(int type, int brand, bool strict)
     case SPMSL_EXPLODING:
         return type == MI_TOMAHAWK;
     case SPMSL_STEEL: // deliberate fall through
+	return false;
```
sdynet: I used one tab to adjust the spacing. But it caused a checkwhite=1 error. What's the problem?

NormalPerson7: You need to use four spaces for each tab. So in this case you want ```
case SPMSL_STEEL:
        return false;```
with eight spaces (two indents).

Also, a note about using git that might help you. If you're using the command line, you can make your fix and then run `git add -A; git commit --amend`, then `git push --force` to fix this without making a new PR or additional unnecessary commits. (It will update your latest commit and then push the updated commit to the PR, overwriting the outdated commit.)

sdynet:
```
 case SPMSL_STEEL:
         return false;
 ```
1. one tap change to eight space
2. use git command line
3. type `git add -A` and `git commit --amend`
right?

Comment (NormalPerson7) [reply crawl-ref/source/makeitem.cc]: Correct, then finally `git push --force`

Comment (sdynet) [reply crawl-ref/source/makeitem.cc]: Thank you very much for your kind explanation. I solved the problem.

File: crawl-ref/source/makeitem.cc [hunk 2]
```diff
@@ -647,6 +643,7 @@ bool is_missile_brand_ok(int type, int brand, bool strict)
     case SPMSL_EXPLODING:
         return type == MI_TOMAHAWK;
     case SPMSL_STEEL: // deliberate fall through
+        return false;
     case SPMSL_SILVER:
```
NormalPerson7: There's one more thing. You can condense the `SPMSL_POISONED` and `SPMSL_STEEL` cases together and place them inside `#if TAG_MAJOR_VERSION == 34` markers:
```
#if TAG_MAJOR_VERSION == 34
    case ...:
    case ...:
        return false;
#endif
```
You should be able to use the same method of `git commit --amend` to update this.

crawl-ref/source/shopping.cc
```diff
@@ -298,7 +298,9 @@ unsigned int item_value(item_def item, bool ident)
             case SPMSL_PARALYSIS:
             case SPMSL_PENETRATION:
             case SPMSL_SILVER:
+#if TAG_MAJOR_VERSION == 34
             case SPMSL_STEEL:
+#endif
             case SPMSL_DISPERSAL:
```
NormalPerson7: Please add the `#if TAG_MAJOR...` markers around the `SPMSL_POISONED` case as well. This also applies to the `switch-case`s in describe.cc and item-name.cc.

crawl-ref/source/makeitem.cc
NormalPerson7: No problem :)

NormalPerson7:
You are correct, sorry! You can disregard my comments then; you only need `#if TAG_MAJOR_VERSION` markers around the `SPMSL_STEEL` case in makeitem.cc (and remove the comment).

crawl-ref/source/makeitem.cc
sdynet: Something's wrong.
`$ git push --force
fatal: The current branch throwing has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin throwing


Administrator@ADMINIS-SDY MINGW64 /d/crawl (throwing)
$ git push --set-upstream origin throwing
remote: Permission to crawl/crawl.git denied to sdynet.
fatal: unable to access 'https://github.com/crawl/crawl.git/': The requested URL returned error: 403
`

sdynet:
I have a code to change, but I don't know how to apply it. I'll pr again.

PR #1047, opened 2019-04-24
Only generate one of Spider/Snake/Swamp/Shoals

alexjurkiewicz:
This is a pretty big change, I think there are some other things to do if this is merged:

- [ ] does the scoring algorithm need an update?
- [ ] should other branches roulette? In the past there has been discussion of elf/crypt and hell/pan
- [ ] should the difficulty of certain S branches be tweaked? Specifically making Swamp a little harder and Shoals a little easier.

--

* The first branch of these the player enters is typically a reasonable
  challenge, but the second is usually too easy. (There are exceptions,
  for example games where the player is poison immune and shoals
  generates.)
* Without the second branch, choice of where to go after getting first
  rune becomes much more difficult for the player. At the moment,
  completing the second branch is easier than anywhere else the player
  can go in the dungeon.

Alternate approaches:

* Make one of the S branch pairs much harder. The new harder branch
would overlap in difficulty with existing rune branches.
* Once you enter one S branch, close the other. Harder to implement, and
would probably result in a no-brainer decision where there is always an
optimal choice.

### DIFF
--- a/crawl-ref/source/branch.cc
+++ b/crawl-ref/source/branch.cc
@@ -105,14 +105,6 @@ static const branch_type danger_branch_order[] = {
 };
 COMPILE_CHECK(ARRAYSZ(danger_branch_order) == NUM_BRANCHES);
 
-static const int number_of_branch_swap_pairs = 2;
-
-static const branch_type swap_branches[number_of_branch_swap_pairs][2] =
-{
-    {BRANCH_SHOALS, BRANCH_SWAMP},
-    {BRANCH_SPIDER, BRANCH_SNAKE}
-};
-
 branch_iterator::branch_iterator(branch_iterator_type type) :
     iter_type(type), i(0)
 {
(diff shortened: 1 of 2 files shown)


doodoostink:
Commenting to say that I want as much branch roulette as possible. So many good changes are made to reduce tedium in DCSS and I find it so easy for the S branch and extended portions of the game to start feeling like a slog because of the numbers of enemies you end up needing to plow through (or run away from!). Additionally, with less total XP and branches in a given game, players could be forced into trying their hand at less favorable situations (exciting!) without the prospect of "throwing away" the run they're invested in.

acoolguy:
This is a pretty interesting change and look forward to playing it.  Thoughts on simply making the two S-branches two floors (entrance + rune floor)?

ruslan-tepelyan:
I started thinking recently about some ideas for Lair branch reform and came and found this PR. Here are some other potential approaches for consideration:

- Generate all 4 branches, shorten them, and only put runes in 2 of them. If you find the rune-filled branches early, you can technically avoid the others, but then you would miss out on XP, and some potential non-rune reward that would be placed in the non-runed branches.

- Generate all 4 branches, shorten them, and increase (some? all?) rune requirements by 2. This leads to a much more "standardized" mid-game, in that all won games would have to deal with all 4 branches at some point.

There are numerous advantages to shortening the branches and generating all 4 of them:

- Players would have an interesting choice by being able to interleave/defer their XP sources based on their abilities. For instance, a player with rPois might choose to do the first floors of Spider and Snake but hold off on the lower floor(s) until their other defenses get stronger, switch to a level or two of Swamp or Shoals, and keep going back and forth based on what they can handle. Currently, a "typical" character that has already cleared Lair:6 and Orc:2 but not yet gotten the Lair runes doesn't have much of a choice in terms of where to level.

- If all of the branches are shorter, a character may hope to get slightly stronger (or find the right resist/next spell) in a branch they can handle, but they won't level up so much that it would trivialize the branches they do second, third, and fourth.

- If certain branches are suspected to be easier than others, generating all 4 in every game would lead to very easily interpretable results about where players choose to go first, making it a lot more obvious which branches need to be tweaked and in which direction.

- If the branches are particularly short (entrance floor + rune floor, as suggested by acoolguy above), it becomes much easier to tweak branch difficulty just by adjusting the rune floors, without having to worry about anything else. Getting one typically easier entrance floor would be outweighed by a correspondingly harder rune floor.

PR #1050, opened 2019-04-28
Wands have an other behavior since the wand merge

syranez:
I have used texts from https://crawl.develz.org/trunk/changes.txt

### DIFF
--- a/crawl-ref/docs/crawl_manual.rst
+++ b/crawl-ref/docs/crawl_manual.rst
@@ -883,12 +883,10 @@ the 'q' command.
 ========================================
 
 Sometimes you will be lucky enough to find a stick which contains stored magical
-energies. Wands each have a certain amount of charges, and a wand will cease to
-function when its charges run out. You must identify a wand to find out how many
-uses it has left. This can be done with a scroll of identify; characters with a
-good Evocations skill may also deduce the number of charges simply upon evoking
-the wand. Evoking a wand without having fully identified the number of charges
-remaining will waste some charges.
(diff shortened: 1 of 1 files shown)


crawl-ref/docs/crawl_manual.rst
```diff
@@ -883,12 +883,10 @@ the 'q' command.
 ========================================
 
 Sometimes you will be lucky enough to find a stick which contains stored magical
-energies. Wands each have a certain amount of charges, and a wand will cease to
-function when its charges run out. You must identify a wand to find out how many
-uses it has left. This can be done with a scroll of identify; characters with a
-good Evocations skill may also deduce the number of charges simply upon evoking
```
NormalPerson7: Lines in the manual shouldn't be longer than 80 characters (as with all other areas of Crawl's code), and all four of these lines are slightly longer than that. Also, this sentence
```
Wands each have a certain amount of charges, which you immediately recognize as you pick it up.
```
doesn't make grammatical sense. Please rewrite it as
```
Wands each have a certain number of charges, which you immediately recognise when you pick them up.
```
Also note that, for historical reasons, Crawl follows Australian English spelling conventions, so "recognise" should be spelt with an "s".

syranez: Thank you.

I've made the necessary changes.

MarvinPA:
Merged, thanks! I don't think you're listed in CREDITS.txt yet - what name would you like to be credited under?

syranez:
I hardly did anything, so I do not need to be listed there... if you insist you can use "Michael Nowack".

Thanks for the merge!

PR #1051, opened 2019-04-29
Traps in stash search

ufshaikh:
The main effect here should be to allow stash searching to work for traps as well as for items. (Excluding web traps.) Some of these commits fix (what I take to be) bugs and clean up the code in stash.cc and stash.h as well.

I tried to keep things fairly fine-grained in terms of commit size, but I am happy to squash some together if it's desired.

### DIFF
--- a/crawl-ref/source/dat/database/help.txt
+++ b/crawl-ref/source/dat/database/help.txt
@@ -110,12 +110,13 @@ that contain the spell.
 You can search for artefact item properties (such as <w>/prevents.*teleport</w>)
 to find artefacts that have the property.
 
-You can search by some additional item properties: <w>artefact</w> or <w>artifact</w> will
-find identified artefacts, <w>ego</w> or <w>branded</w> will find non-artefacts with a
-brand and unidentified items which may be branded, <w>throwable</w> will find things
-you can throw, <w>dropped</w> will find items you have dropped, <w>in_shop</w> will find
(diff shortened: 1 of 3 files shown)


ufshaikh:
I'm going to add a justificatory note here. I said much of this in ##crawl-dev,
but I think it's good to have something more permanent. If someone picks this up
in the future, I think it would be natural to squash several of the commits in
the PR together; this can be a place to start for the overall commit message.


Add traps to stash search

Traps are tactical resources, and there can be reason to drag monsters back to
them, even across levels. For example, a teleport or dispersal trap can save a
scroll or substitute for the blink spell, and a shaft can be used in ways
similar to a down hatch. Currently the game does not expose trap locations
throughout the dungeon in a convenient way. The player can try to keep track
manually of traps as they discover them, or can scan levels looking for the
right kind of trap using X. Either approach is tedious and in practice likely to
be error-prone.

Here I solve these problems by exposing traps to the stash search screen reached
by using ctrl F. I also provide for collapse of duplicates in the search results
to be applied to traps.

(I believe that this behaviour was already intended and was at some point lost
(or perhaps not properly implemented in the first place): see for example
commits 23c5277317de657f8376d1b5f33144948a811a42 and
cbc0dd1c0e73770878e9a6df1b455e57995ef921, the ctrl F help screen's reference to
searching for features, and the logic error this change fixes in Stash::update.)

PR #1054, opened 2019-05-06
Remove weapon gifting from Trog

alexjurkiewicz:
There are a few reasons for this change:

1) Weapon gifting overlaps with Okawaru. It's not an exact overlap --
   Okawaru also gifts armour, and the more subtle stats of gifted
   weapons (enchantment & brand distribution) differ between the gods
   too. But it's nevertheless a very same-y god feature.
   a) Okawaru's thematic connection to item gifting is stronger than
      Trog's.
   b) Okawaru is a weaker god overall, and removing / nerfing weapon
      gifting would result in a very mediocre god.
2) Trog is a fantastically strong god for the early and mid game, but
   falls off a little as the game goes on. This change emphasises that
   power curve by cutting out a god feature with zero impact on the
   early game, and relatively little on the mid game.
   a) Even with this change, Trog will remain a very strong god, but
      perhaps less overwhelmingly strong at all points in the game.

### DIFF
--- a/crawl-ref/source/acquire.cc
+++ b/crawl-ref/source/acquire.cc
@@ -1188,18 +1188,6 @@ static string _why_reject(const item_def &item, int agent)
         return "Destroying unusable weapon or armour!";
     }
 
-    // Trog does not gift the Wrath of Trog, nor weapons of pain
-    // (which work together with Necromantic magic).
-    // nor fancy magic staffs (wucad mu, majin-bo)
-    if (agent == GOD_TROG
-        && (get_weapon_brand(item) == SPWPN_PAIN
-            || is_unrandom_artefact(item, UNRAND_TROG)
-            || is_unrandom_artefact(item, UNRAND_WUCAD_MU)
-            || is_unrandom_artefact(item, UNRAND_MAJIN)))
-    {
-        return "Destroying a weapon Trog hates!";
-    }
-
     // Pain brand is useless if you've sacrificed Necromacy.
     if (you.get_mutation_level(MUT_NO_NECROMANCY_MAGIC)
         && get_weapon_brand(item) == SPWPN_PAIN)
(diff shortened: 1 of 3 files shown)


aolowin:
I've been considering a patch where weapon gifts only happen if you burn a book, with better gifts for unidentified books. That would give a different flavor from Okawaru and force some choices if you plan to abandon Trog later on.

G3Kappa:
I agree with @aolowin, but I'd also like to see a stronger distinction between weapon gifts and armor gifts. In fact, I think gifts in general should reflect the various acquirement categories better.

- Okawaru gifts armor, weapons and ammunition
- Trog gifts weapons and ammunition
- Nemelex *used to* gift evocables

I would like to see an Okawaru that either focuses on armor or ranged fighting, while moving the weapon gifts exclusively to Trog and limiting them to being melee weapons. This would keep Trog a tempting midgame option without hurting its reputation as newbie-friendly. I would also remove projectile gifts from Trog, in order to put a stronger emphasis on its late-game (non-)viability and on the need to switch to a better god. If the weapon gifts are good enough, it might be worth it.

I also feel like there's space for a third god if Okawaru takes a strong decision one way or the other. If Okawaru becomes an armor/jewelry god, then there's space for a ranged god that gifts ranged weapons and projectiles.

But this is clearly out of scope for this PR. I'm just not not sure whether removing gifts is particularly fun; I'd rather add *more* gifts for that matter.

ebering:
Trog is one of the oldest gods in crawl, and there’s a lot of reverence for them. 

The gifts make finding a good weapon easier, and Trog’s role as a new player god means they shouldn’t go away. However G3Kappa’s suggestion that Trog gift only melee weapons is a good one, both for simplifying a recommended new player god and for lore. I think that’s a good direction to go, and will add on that Trog should have a restricted brand table: `none, vorpal, flame, antimagic`. A god of firey anger that hates magic shouldn’t be handing out freezing weapons. 

A new PR that does this would be welcome.

PR #1064, opened 2019-05-14
Added Lumpy Black Scales mutation which provides (2/5/8) AC and (2/3/4) ER

ByrelMitchell:
The goal here is to add another scales mutation that actually impacts your decision-making. Currently most scales mutations are simply varying amounts of power with little to no gameplay impact. One exception is Sturdy Frame (-2/-4/-6 ER), which both changes how you evaluate armours (scale mail is much more competitive with a robe of the Archmagi than leather armour is), and incentivises a playstyle with more spell casting.

This mutation should do the opposite. It should encourage players to transition to a heavier armour strategy with less spell casting, and should make armour evaluations at the high end different (especially for low-mid strength characters.)

Capping out at paying 4 ER for 4 AC over the typical scales mutation is about the same as upgrading to heavier armour tends to be in the mid and late game. (Upgrading from chain to plate costs 3 ER for 2 base AC, which will generally be 3-6 extra AC depending on investment and enchant armour scrolls. Upgrading from scale to chain costs 5 ER for 2 base AC.) I put the largest step at the first position because things tend to scale with ER^2. Going from 0->1 ER is a lot more negligible than going from 3->4. It's also a familiar progression, since most scales mutations follow the same pattern.

If this ends up being insufficiently impactful, it could be tweaked to match the Sturdy Frame mutation instead.


crawl-ref/source/mutation-data.h
```diff
@@ -1339,6 +1339,22 @@ static const mutation_def mut_data[] =
 },
 #endif
 
+{ MUT_LUMPY_BLACK_SCALES, 2, 3, mutflag::good, true,
+  "lumpy black scales",
+
+  {"You are partially covered in lumpy black scales. (AC +2, ER -2)",
```
amalloy: You want pluses here, not minuses, no?

crawl-ref/source/player.cc
```diff
@@ -5945,7 +5948,10 @@ int player::base_ac(int scale) const
     AC += get_mutation_level(MUT_ROUGH_BLACK_SCALES, mutation_activity_type::FULL)
           ? -100 + get_mutation_level(MUT_ROUGH_BLACK_SCALES, mutation_activity_type::FULL) * 300 : 0;
               // +2, +5, +8
-#endif
+#endif
```
amalloy: You accidentally changed the whitespace on this line.

MarvinPA:
I don't see this working as an good mutation compared to Sturdy Frame. Picking the heaviest armour available and getting a huge amount of AC is much less of an interesting thing to encourage than the skill tradeoffs of casting spells in some level of medium/heavy armour.

alexjurkiewicz:
I like the idea of an ER+ mutation. But maybe it shouldn't be the yin to
sturdy frame's yang. Maybe instead it should also encourage
spellcasting.... just in a different way to sturdy frame.

Maybe ER+ for Archmagi+? It would rule out any casting for plate+ users,
but other characters could tweak their build to get a benefit.

ByrelMitchell:
I guess it's not terribly interesting to trade off ER for AC in roughly that same proportions as swapping weights of armour would do. It's most interesting at the top and bottom with 'making plate be heavier' and 'making wearing nothing an interesting choice'. But the first isn't very interesting and it's not obvious that the second should be done at all. It's a little unintuitive.

Still though, I think any scale slot mutation should probably net positive defense in some situations. It's the theme of that tier of DS mutations. It would also be odd to give EV along with ER, since they effectively cancel each other out. That leaves either giving SH or some specialized defense (RMsl, Spirit, resistances, cloud immunity, etc.)

This could be the Alchemist's Hat mutation; give a pip of a pile of resists in exchange for ER?

ByrelMitchell:
There might be an interesting mutation around increasing ER but I haven't thought of it in the last year so I'm closing this PR.

alexjurkiewicz:
Re-reading this thread, I have one idea:
Magical Scales: You gain a little spell power by wearing heavy armour.

You'd gain x% spell power for each point of ER. X would scale with mutation
level.

ByrelMitchell:
@alexjurkiewicz I like the idea of that tradeoff, but I'm not sure it's a good idea to add a scales slot mutation that doesn't increase defense. That's part of what sets scales mutations apart from others.

Perhaps a 'Magical Resonance' along the lines of your suggestion would make a good additional Tier 2 mutation?

PR #1066, opened 2019-05-16
Remove sling bullet. And sling weapons's damage increase one.

sdynet:
Sling bullet is removed to reduce inventory pressure. As the stone became the only sling bullet, the damage from the slings increased by one.

### DIFF
--- a/crawl-ref/source/acquire.cc
+++ b/crawl-ref/source/acquire.cc
@@ -549,6 +549,7 @@ static int _acquirement_missile_subtype(bool /*divine*/, int & /*quantity*/)
 
     switch (skill)
     {
+	case SK_SLINGS:    result = MI_STONE; break;
     case SK_BOWS:      result = MI_ARROW; break;
     case SK_CROSSBOWS: result = MI_BOLT; break;
 
--- a/crawl-ref/source/acquire.cc
+++ b/crawl-ref/source/acquire.cc
@@ -549,7 +549,6 @@ static int _acquirement_missile_subtype(bool /*divine*/, int & /*quantity*/)
 
     switch (skill)
     {
-    case SK_SLINGS:    result = MI_SLING_BULLET; break;
     case SK_BOWS:      result = MI_ARROW; break;
     case SK_CROSSBOWS: result = MI_BOLT; break;
 
--- a/crawl-ref/source/acquire.cc
+++ b/crawl-ref/source/acquire.cc
@@ -549,7 +549,7 @@ static int _acquirement_missile_subtype(bool /*divine*/, int & /*quantity*/)
 
     switch (skill)
     {
-	case SK_SLINGS:    result = MI_STONE; break;
+    case SK_SLINGS:    result = MI_STONE; break;
     case SK_BOWS:      result = MI_ARROW; break;
     case SK_CROSSBOWS: result = MI_BOLT; break;
 
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -1094,7 +1094,6 @@ static void _append_weapon_stats(string &description, const item_def &item)
 
     const bool could_set_target = _could_set_training_target(item, true);
 
-    if (skill == SK_SLINGS)
     description += make_stringf(
     "\nBase accuracy: %+d  Base damage: %d  Base attack delay: %.1f"
     "\nThis weapon's minimum attack delay (%.1f) is reached at skill level %d.",
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -1095,12 +1095,6 @@ static void _append_weapon_stats(string &description, const item_def &item)
     const bool could_set_target = _could_set_training_target(item, true);
 
     if (skill == SK_SLINGS)
-    {
-        description += make_stringf("\nFiring bullets:    Base damage: %d",
-                                    base_dam +
-                                    ammo_type_damage(MI_SLING_BULLET));
-    }
-
     description += make_stringf(
     "\nBase accuracy: %+d  Base damage: %d  Base attack delay: %.1f"
     "\nThis weapon's minimum attack delay (%.1f) is reached at skill level %d.",
--- a/crawl-ref/source/god-blessing.cc
+++ b/crawl-ref/source/god-blessing.cc
@@ -170,7 +170,7 @@ void gift_ammo_to_orc(monster* orc, bool initial_gift)
     else
         ammo.sub_type = fires_ammo_type(*launcher);
 
-    if (ammo.sub_type = MI_STONE);
+    ammo.sub_type = MI_STONE;
 
     ammo.quantity = 30 + random2(10);
     if (initial_gift || !launcher)
--- a/crawl-ref/source/god-blessing.cc
+++ b/crawl-ref/source/god-blessing.cc
@@ -170,8 +170,7 @@ void gift_ammo_to_orc(monster* orc, bool initial_gift)
     else
         ammo.sub_type = fires_ammo_type(*launcher);
 
-    if (ammo.sub_type == MI_STONE)
-        ammo.sub_type = MI_SLING_BULLET; // ugly special case
+    if (ammo.sub_type = MI_STONE);
 
     ammo.quantity = 30 + random2(10);
     if (initial_gift || !launcher)
--- a/crawl-ref/source/item-prop.cc
+++ b/crawl-ref/source/item-prop.cc
@@ -616,10 +616,10 @@ static const weapon_def Weapon_prop[] =
         DAMV_NON_MELEE, 0, 0, 0, {}, },
 #endif
 
-    { WPN_HUNTING_SLING,     "hunting sling",       5,  2, 12,
+    { WPN_HUNTING_SLING,     "hunting sling",       6,  2, 12,
         SK_SLINGS,       SIZE_LITTLE, SIZE_LITTLE, MI_STONE,
         DAMV_NON_MELEE, 8, 10, 15, RANGED_BRANDS },
-    { WPN_FUSTIBALUS,        "fustibalus",          8, -1, 14,
+    { WPN_FUSTIBALUS,        "fustibalus",          9, -1, 14,
         SK_SLINGS,       SIZE_LITTLE, SIZE_LITTLE, MI_STONE,
         DAMV_NON_MELEE, 2, 2, 150, RANGED_BRANDS },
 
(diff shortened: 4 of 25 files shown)


crawl-ref/source/acquire.cc
```diff
@@ -549,7 +549,6 @@ static int _acquirement_missile_subtype(bool /*divine*/, int & /*quantity*/)
 
     switch (skill)
     {
-    case SK_SLINGS:    result = MI_SLING_BULLET; break;
     case SK_BOWS:      result = MI_ARROW; break;
```
NormalPerson7: You need to replace this with `result = MI_STONE` rather than removing the case altogether.

crawl-ref/source/describe.cc
```diff
@@ -1095,12 +1095,6 @@ static void _append_weapon_stats(string &description, const item_def &item)
     const bool could_set_target = _could_set_training_target(item, true);
 
     if (skill == SK_SLINGS)
-    {
-        description += make_stringf("\nFiring bullets:    Base damage: %d",
```
NormalPerson7: You need to remove the entire `if` statement here; in its current state this code does not work as intended.

crawl-ref/source/god-blessing.cc
```diff
@@ -170,8 +170,7 @@ void gift_ammo_to_orc(monster* orc, bool initial_gift)
     else
         ammo.sub_type = fires_ammo_type(*launcher);
 
-    if (ammo.sub_type == MI_STONE)
-        ammo.sub_type = MI_SLING_BULLET; // ugly special case
```
NormalPerson7: Also remove the `if` here please.

crawl-ref/source/item-prop.cc
```diff
@@ -793,6 +795,7 @@ const set<pair<object_class_type, int> > removed_items =
     { OBJ_FOOD,      FOOD_ROYAL_JELLY },
     { OBJ_FOOD,      FOOD_UNUSED },
     { OBJ_FOOD,      FOOD_FRUIT },
+	{ OBJ_MISSILES,  MI_SLING_BULLET },
 #endif
```
NormalPerson7: Please indent with four spaces instead of using tabs, for consistency. Tabs render differently on Github and will cause the checkwhite script to fail—this is the reason why Travis has failed.

crawl-ref/source/mon-gear.cc
```diff
@@ -1323,10 +1323,7 @@ static void _give_ammo(monster* mon, int level, bool mons_summoned)
             && (mon->type == MONS_JOSEPH
                 || mon->type == MONS_SATYR
                 || (mon->type == MONS_FAUN && one_chance_in(3))
-                || one_chance_in(15)))
```
NormalPerson7: Again please remove the entire `if` statement here.

crawl-ref/source/monster.cc
```diff
@@ -1955,8 +1955,7 @@ bool monster::pickup_missile(item_def &item, bool msg, bool force)
                 // Don't drop huge stacks for tiny stacks.
                 if (item.launched_by(*launch)
                     && (!miss->launched_by(*launch)
-                        || item.sub_type == MI_SLING_BULLET
-                           && miss->sub_type == MI_STONE
```
NormalPerson7: You can remove the MI_STONE line here as well; it will now be covered by the other checks in this `if` statement.

crawl-ref/source/newgame.cc
```diff
@@ -1670,7 +1670,7 @@ static void _construct_weapon_menu(const newgame_def& ng,
             if (is_ranged_weapon_type(wpn_type))
             {
                 text += " and ";
-                text += wpn_type == WPN_HUNTING_SLING ? ammo_name(MI_SLING_BULLET)
+                text += wpn_type == WPN_HUNTING_SLING ? ammo_name(MI_STONE)
                                                       : ammo_name(wpn_type);
                 text += "s";
```
NormalPerson7: You can remove the `condition ? true : false` switch here and replace it with simply `text += ammo_name(wpn_type)`.

crawl-ref/source/ng-setup.cc
```diff
@@ -166,7 +166,7 @@ item_def* newgame_make_item(object_class_type base,
     else if (item.base_type == OBJ_BOOKS && item.sub_type == BOOK_CHANGES)
         _autopickup_ammo(MI_ARROW);
     // You probably want to pick up both.
-    if (item.is_type(OBJ_MISSILES, MI_SLING_BULLET))
+    if (item.is_type(OBJ_MISSILES, MI_STONE))
         _autopickup_ammo(MI_STONE);
```
NormalPerson7: You can also remove this `if` statement altogether.

crawl-ref/source/makeitem.cc
```diff
@@ -640,7 +638,6 @@ static void _generate_missile_item(item_def& item, int force_type,
             random_choose_weighted(50, MI_STONE,
                                    20, MI_ARROW,
                                    12, MI_BOLT,
-                                   12, MI_SLING_BULLET,
                                    10, MI_DART,
```
NormalPerson7: Please give this 12 weight to stone generation (so that stones get total 62 weight), rather than removing it altogether. Removing it altogether causes the weight to effectively be distributed evenly across all ranged weapon types, which will cause more throwing weapons to generate than expected.

sdynet:
Thank you for your kind advice. I've solved all the problems. :)

NormalPerson7:
You need to push your changes to this PR with `git push`. If that doesn't work, try `git push --force`.

sdynet:
I was embarrassed again in the 'push --force' part, but I succeeded in doing push as a result of trying to follow your instructions.
Thank you, NormalPerson7, ebering. :)

crawl-ref/source/acquire.cc
```diff
@@ -549,7 +549,7 @@ static int _acquirement_missile_subtype(bool /*divine*/, int & /*quantity*/)
 
     switch (skill)
     {
-    case SK_SLINGS:    result = MI_SLING_BULLET; break;
+	case SK_SLINGS:    result = MI_STONE; break;
     case SK_BOWS:      result = MI_ARROW; break;
     case SK_CROSSBOWS: result = MI_BOLT; break;
```
NormalPerson7: From a technical perspective, this all looks fine, with the exception of this tab; as with before, please change it to four spaces. I'd also advise you set your text editor/IDE/whatever you use for editing to indent with spaces instead of tabs.

ebering:
A big thanks to @NormalPerson7 for the technical feedback on this PR.

If you're having trouble with whitespace, the script `util/checkwhite` in the source dir can fix your whitespace.

Regarding the gameplay changes; there's a bit more to think about here, in terms of comparing damage outputs using fsim. I will not have time to do this for a little while, but you could do it and add the results here. Run crawl in wizard mode and use the command `&F` to get data on the old effectiveness of stones and sling bullets, and the new effectiveness under this new system. Try against various targets (yak, stone giant, orb of fire).

Feel free to ask here or in ##crawl-dev if you get stuck.

sdynet:
I compared the previous version with the changed version.
Yak: https://drive.google.com/file/d/1O5ULgLiCdGBL44XTW8m5VvDYlWjDSeUW/view?usp=sharing
Death yak: https://drive.google.com/file/d/1OY4dLobEImitvfLt932fAO1xq1g26bEX/view?usp=sharing
Stone giant: https://drive.google.com/file/d/1zPjve4qoipgwvwCq-U6u8N2ra3uMUHKK/view?usp=sharing
Hell sentinel: https://drive.google.com/file/d/1Ci70PWOdYy3O9KBdUpsZ_Ei4Hp1St2mR/view?usp=sharing
OoF: https://drive.google.com/file/d/1WvWTKo4lm94kP66LBYvPKna5tHKe06rl/view?usp=sharing

I think it's a little weak. It needs a little bit of buff. But if I simply increase the damage, I wonder what's different from the throwable. So, It's half a joke, but what would it feel like to add exploding brand to the sling?
The damage from the explosion is 2d5, so it's not high. And it's very noisy. I think a little trimming would be okay, but is it too much of an adventure?

ebering:
I'm not sure I agree with the assessment that it's weak, the result is a decrease of .5 average effective damage at low skill, increasing to 1 at high skill. It moves the mean, but this damage is always available and not subject to ammo limitations. These limitations are less important late game, so there is a nerf to late game sling strength; but compared to the previous stone damage this is an early game buff, the result is a flattening of the sling power curve.

I don't think adding explosion branding is the way to go. I'm not sure about what to do about the numbers, I'll get another team member to comment.

sdynet:
I tried to delete these commits:
Revert "Clear mon_acting on _reset_game (11971)"
Revert "Revert "Clear mon_acting on _reset_game (11971)""

I ran 'git reset --hard slingstone~2', but nothing changed. How can I delete an unnecessary commits? Should I do PR again?

NormalPerson7:
I suggest you check the log of your local branch, e.g. `git log -5` to show the last five commits, to see if those commits are still there. If they are gone, you can run `git push --force` to force-push that update to the PR. If they are still there, I suggest you run `git rebase -i 1cb13bb~1` (`-i` gives you an interactive rebase) and set the commits you don't want to the "drop" option (they'll be on "pick" by default); then run `git push --force`.

sdynet:
You are my great teacher.

NormalPerson7:
I don't think you meant to do that...

crawl-ref/source/acquire.cc
NormalPerson7: The whitespace that needs fixing is here: please change the tab to four spaces.

sdynet: Uh... God, I was looking for the wrong place at all the wrong!

sdynet:
All right. It's time to move on to the next mountain. CHECKWHITE=1 problem wasn't the end. What's the problem with WEBTILES=1 FULLDEBUG=1?

NormalPerson7:
The issue is seed divergence, I think caused by changing the distribution of ammo generation (which can't be avoided). What I find more peculiar is the fact that only one of the FULLDEBUG builds failed, as opposed to all four. I don't think you need to do anything about it though.

sdynet:
Finally. It's over. Thank you. :)

theJollySin:
This PR is a fun idea. I like it.

I also thought "sling bullet" was a lame name to begin with.  Something like "lead weight" or "lead ball" would have made more sense.

Historically, yes, lead balls would do more damage than rocks.  Their weight helped, but playing around as a kid I think their real benefit is that they were perfectly round.

rawlins:
> The issue is seed divergence, I think caused by changing the distribution of ammo generation (which can't be avoided). What I find more peculiar is the fact that only one of the FULLDEBUG builds failed, as opposed to all four. I don't think you need to do anything about it though.

I didn't spot this conversation at the time, but the travis failure is divergence between runs on a random seed (`563868786` for posterity) that I doubt has anything to do with this particular PR; if the random case fails it's most likely finding a bug in trunk. I tried restarting the job, but something else is going weird now; in any case, the travis error is ignorable.

theJollySin:
Alternately, it might be fun if you could also throw sling bullets. The flavor text says they're just lead balls, right?

Hell, that sounds like a *great* thing to throw by hand.

ebering:
I'm sorry this has sat for so long; I would like to do a more thorough ranged-overhaul to remove all ammunition in 0.26. If this is something that interests you swing by ##crawl-dev and we can hash out a plan.

ebering:
Well I didn't remove all ammo in 0.26 but @PleasingFungus did now, so I'm closing this!

PleasingFungus:
0.26, 0.29... let's call the whole thing off.

PR #1067, opened 2019-05-17
Improve shaft messaging

alexjurkiewicz:
Always print the number of floors, so the message is consistent between falling one or multiple levels.

### DIFF
--- a/crawl-ref/source/stairs.cc
+++ b/crawl-ref/source/stairs.cc
@@ -540,13 +540,10 @@ static level_id _travel_destination(const dungeon_feature_type how,
                                     + shaft_dest.describe() + ".");
         }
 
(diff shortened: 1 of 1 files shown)


amalloy:
Why is that a useful consistency? It's not immediately obvious to me why "You fall through a shaft for 1 floor" is better than what you get now.

alexjurkiewicz:
I think players want to know how many levels they fell, even if it's just one floor. The current message for falling one floor doesn't say this, so unless you know the messaging is different for single floor drops, you might need to double check your source floor to know how far you fell.

ebering:
I agree with @alexjurkiewicz on this; having the number in the message line is helpful for a new player who doesn't know about the special case.

PR #1068, opened 2019-05-19
Fix Avatar Song displayed success rate

Doesnty:
Avatar song’s displayed success rate was incorrect, displaying 0% when the player could be vulnerable. This patch addresses the issue. 

*Editor’s note: the original pr message included unhelpful and unwelcome passive aggression.*

### DIFF
--- a/crawl-ref/source/mon-cast.cc
+++ b/crawl-ref/source/mon-cast.cc
@@ -1077,6 +1077,7 @@ static int _mons_power_hd_factor(spell_type spell, bool random)
             return 10 * ENCH_POW_FACTOR;
 
         case SPELL_SIREN_SONG:
+        case SPELL_AVATAR_SONG:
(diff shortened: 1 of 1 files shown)


ebering:
The patch is welcome but please keep your PR messages constructive and on topic.

Doesnty:
...That was a direct quote of a dev's response to me bringing up the issue. It is, in my humble opinion, unhelpful and unwelcome passive aggression toward someone pointing out a bug's existence.

To be more clear on the second now-deleted point: _siren_sing(), in mon-cast.cc, should probably be adjusted to actually use the song in question, and not just assume it's Siren Song instead of Avatar Song, for stylistic reasons. I do not have the desire to exert the effort to send you more than just a barebones fix though, as that would involve actually testing my code; effort that I'm not inclined to spend on a devteam that is incessantly hostile to its own userbase and willing to silence and slander people trying to call out this hostility.

PR #1071, opened 2019-05-23
Simplify missile description cases

cut1less:
Previously, descriptions for projectiles were based largely around
whether they had a damage value associated with them, under the logic
that "inherent missile damage => throwable missile". This made life easy
in dealing with the old needles and with regular launcher missiles.

However, it led to:
* the sling bullet description having lots of throwing information
  (despite sling bullets not being throwable) and
* the throwing net description lacking the fact that a net's delay is
  lowered with Throwing skill

With the changes to throwing missiles, we can now just focus on
throwable-ness to fix these issues and more.

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -1512,7 +1512,8 @@ static string _describe_ammo(const item_def &item)
     }
 
     const int dam = property(item, PWPN_DAMAGE);
-    if (can_throw)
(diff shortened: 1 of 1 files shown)


ebering:
Thanks for the patch! Are you listed in some fashion in CREDITS.txt? If not, how would you like to appear?

cut1less:
I'm not in CREDITS.txt; you can list me as "cut1less", if you'd like. And thanks for the quick review and merge!

PR #1072, opened 2019-05-24
Mention draining gaze's healing effect

NormalPerson7:
Draining gaze heals ghost moths and eyes of draining for the amount of
MP drained by the gaze. This commit adds a note to this effect to the
draining gaze spell description.

Note: orange crystal statues also cast this spell, but they don't get
healed from it because they cannot regenerate HP.


alexjurkiewicz:
Maybe the healing effect should be removed so the spell works consistently
across all monsters?

PR #1073, opened 2019-05-25
Don't allow the player to cancel monster spells with OTR (11990)

NormalPerson7:
The monster spell Monstrous Menagerie and monsters using scrolls of
summoning use the same functions in spl-summoning.cc as the player
versions of the spell; this is not the case for any other summoning
spells.

This was allowing the player to cancel casting a monster's monstrous
menagerie or using their scroll of summoning if they had OTR up. We add
special checks to these two functions to avoid calling
otr_stop_summoning_prompt with monster casters.

### DIFF
--- a/crawl-ref/source/spl-summoning.cc
+++ b/crawl-ref/source/spl-summoning.cc
@@ -279,7 +279,7 @@ spret cast_summon_ice_beast(int pow, god_type god, bool fail)
 
 spret cast_monstrous_menagerie(actor* caster, int pow, god_type god, bool fail)
 {
-    if (otr_stop_summoning_prompt())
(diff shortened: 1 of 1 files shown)


MarvinPA:
I don't think monster scrolls of summoning to use the same codepath as the player version, so it might just be the menagerie check that's needed?

NormalPerson7:
Oh right, is MON_SUM_SCROLL also used for player summoning scrolls then?

MarvinPA:
Yep, it's kinda confusingly named and implemented - MON_SUMM_SCROLL is a mon_summon_type enum, which is used in place of a spell_type enum for summons that aren't caused by a spell.

NormalPerson7:
Ah right, I see

PR #1074, opened 2019-05-26
Update vine stalker description (11979)

NormalPerson7:


NormalPerson7:
No idea why this failed...

ebering:
Seems to be a bug on malign gateways in the arena causing the travis failure.

PR #1075, opened 2019-05-30
Reduce monster glaciate damage by 20%

alexjurkiewicz:
On pandemonium lords during the orb run, glaciate is by far the most
common killer spell:

153 games for * (ckiller=pandemonium_lord ((d || depths || zot))
recent):
70x, 26x great icy blast, 9x great blast of fire, [...]

This is probably because the effective damage is higher than any other
spell in a pan lord's arsenal:
26 games for * (ckiller=pandemonium_lord ((d || depths || zot))
recent kaux=great_icy_blast): avg(dam)=56.69

Comparison: CBL 56, LCS 50, Firestorm 41. CBL gives at least a turn to
react, LCS has limited range, and Firestorm's damage is significantly
lower.

This reduction to glaciate damage will still give it a respectable 45
effective damage (80% of the above figure), which is still nothing to
sneeze at.

### DIFF
--- a/crawl-ref/source/spl-damage.cc
+++ b/crawl-ref/source/spl-damage.cc
@@ -2899,7 +2899,7 @@ spret cast_glaciate(actor *caster, int pow, coord_def aim, bool fail)
             beam.damage =
                 caster->is_player()
                     ? calc_dice(7, (66 + 3 * pow) / eff_range)
-                    : calc_dice(10, (54 + 3 * pow / 2) / eff_range);
+                    : calc_dice(8, (54 + 3 * pow / 2) / eff_range);
 
             if (actor_at(entry.first))
             {


NormalPerson7:
Wrt your data, I don't really think experimental data is the best way to determine that average. People tend to have more rF than rC on the orbrun, so that might partly explain the difference in damage.

alexjurkiewicz:
Yeah that's a good point!

alexjurkiewicz:
Looking at the most recent five deaths from glaciate vs fire storm, I don't really see evidence that differing amounts of rC+ vs rF+ or AC are responsible for glaciate damage being higher:

```
Glaciate deaths:
rC++  49AC 66dmg
rC++  15AC 59dmg
rC    33AC 64dmg
rC    61AC 42dmg
rC+++ 40AC 61dmg
Fire storm deaths:
rF++  22AC 41dmg
rF+   48AC 37dmg
rF++  73AC 42dmg
rF    46AC 59dmg
rF+++ 42AC 33dmg```

I found it hard to read the code which is why I didn't include that data.

Glaciate monster damage is in spl-damage.cc and is defined as `calc_dice(10, (54 + 3 * pow / 2) / eff_range)`. With HD of 20 => power 120 (based on logic in mon-cast.cc) and eff_range being 3-5 I get point damage of 10d46 to 10d78 which seems way too high so I might be missing something. Even with pow=0, the damage is 10d10 to 10d8 which still seems a little high.

Fire storm monster damage is defined in zap-data.h as `calcdice_calculator<8, 5, 1, 1>` which is 8d(5*1/1) = 40 max.

ebering:
You've misunderstood `calcdice_calculator`. A calcdice_calculator creates a damage deducer, which returns damage dice dependent on power, specifically the `calcdice_calculator<8,5,1,1>` is the die `calc_dice(8, 5+pow*1/1)`. In turn, `calc_dice` determines the appropriate dice sides to have n dice roll a given max damage.

I'm also not sure where the hd 20 -> spellpower 120 derivation came from. Monster spellpower is `hd * power factor`, and the power factor for both Firestorm and Glaciate is 12. So an HD 20 pan lord will cast those spells with power 240. Here's the deets on the corresponding damage dice (caveat: calc_dice uses div rand round, I rounded in the more likely direction)

|   | firestorm | glaciate point blank | glaciate max range |
| --- | --- | --- | --- |
| calc_dice | 8, 245 | 10, 138 | 10, 82 |
| die roll | 8 d 31 | 10 d 14 | 10 d 8 |
| mean | 128 | 75 | 45 |
| dev | 25.30 | 12.75 | 7.25 |

Glaciate is lower variance and its threat increases based on relative position.

Whatever the reason is that glaciate is responsible for more orb run kills than firestorm, it is not because it hits too hard relative to firestorm in a raw dice definition sense. I would be interested if you find a good explanation. It might be hard to get from morgues though, my hunch is that it has something to do with monster targeting logic.

PR #1078, opened 2019-06-05
Add missing loading images to webtiles.

Ash-R:
I heard someone mention they weren't ever seeing the newer title splash images on webtiles so I had a look at my test webtiles install and noticed they were completely missing. This fix seems to work at least tho perhaps it would've been better to refactor the makefile so the source folder is just scanned for images and client.html generated accordingly rather than having to explicitly add the images individually each time one is added. First time making a pull request here so apologies if I did anything incorrectly and if so please help me out with some tips. Thanks!


ebering:
This looks good; how would you like to appear in CREDITS.txt?

ebering:
Oh, for commit messages we usually leave off the trailing period for the subject line.

Ash-R:
Thanks for the tip and the credit... it'll be good motivation to try and contribute more so I feel like I deserve a spot in there :o
Most people seem to use real names so I'll go with that: "Ashley Elizabeth Raine"

PR #1080, opened 2019-06-09
Hepliaklqana Ancestor They/Them Pronouns

crystalsolenoid:
Now your ancestors that were "always neither, you're pretty sure" use "they" instead of "it"!

I think it makes the ancestor feel more like a person and less like a thing. It's not like your ancestor was a jelly or something. (And it's cool to be able to have my nonbinary friends help me beat up some gnolls without having to call them "it".)

The code to add "GENDER_NEUTRAL" to the gender list and the god ability was pretty simple, but I had to fix a lot of verb conjugations in the monster descriptions because they were all (reasonably-- the way singular they/them still conjugates verbs as plural is weird) hard coded as the singular conjugation.

This is my first time contributing to a project this big, git still really confuses me, and I'm pretty new to C++, so I have no idea what I'm doing right now. Please help me learn!

I tried my best to follow the conventions the existing code uses, but I haven't actually sat down and made sure I'm following the style guide yet. There are a couple places I want to refactor before this is 100% ready, such as monster::conj_verb and using the gender enums instead of hard coding the integers. I also would like to do a larger-scale test. Maybe temporarily change the gender of all the generic monsters in the dungeon in order to catch anything I missed? Would that be difficult to do?

Also, do I have to change anything to keep translations from breaking with these changes?

This was my intro project to get comfortable with Crawl's codebase, but if this is a change people would consider actually adding to the game, I'll put in the time to make it perfect.

AllyDescription

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -4044,9 +4044,12 @@ static string _monster_stat_description(const monster_info& mi)
         }
         _add_energy_to_string(
             speed, me.spell,
-            mi.is_actual_spellcaster() ? conjugate_verb("cast", plural) + " spells" :
-            mi.is_priest()             ? conjugate_verb("use", plural) + " invocations"
-                                       : conjugate_verb("use", plural) + " natural abilities", fast, slow);
+            mi.is_actual_spellcaster() ? conjugate_verb("cast", plural)
+                                        + " spells" :
+            mi.is_priest()             ? conjugate_verb("use", plural)
+                                        + " invocations"
+                                       : conjugate_verb("use", plural)
+                                        + " natural abilities", fast, slow);
         _add_energy_to_string(speed, me.special,
                               conjugate_verb("use", plural) + " special abilities",
                               fast, slow);
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -3931,16 +3931,19 @@ static string _monster_stat_description(const monster_info& mi)
     }
 
     const char* pronoun = mi.pronoun(PRONOUN_SUBJECTIVE);
+    const bool plural = mi.pronoun_plurality();
 
     if (mi.threat != MTHRT_UNDEF)
     {
-        result << uppercase_first(pronoun) << " looks "
-               << _get_threat_desc(mi.threat) << ".\n";
+        result << uppercase_first(pronoun)
+			   << " " << conjugate_verb("look", plural) << " " 
+			   << _get_threat_desc(mi.threat) << ".\n";
     }
 
     if (!resist_descriptions.empty())
     {
-        result << uppercase_first(pronoun) << " is "
+        result << uppercase_first(pronoun)
+			   << " " << conjugate_verb("are", plural) << " "
                << comma_separated_line(resist_descriptions.begin(),
                                        resist_descriptions.end(),
                                        "; and ", "; ")
@@ -3950,15 +3953,18 @@ static string _monster_stat_description(const monster_info& mi)
     // Is monster susceptible to anything? (On a new line.)
     if (!suscept.empty())
     {
-        result << uppercase_first(pronoun) << " is susceptible to "
+        result << uppercase_first(pronoun)
+			   << " " << conjugate_verb("are", plural) << " "
+			   << "susceptible to "
                << comma_separated_line(suscept.begin(), suscept.end())
                << ".\n";
     }
 
     if (mi.is(MB_CHAOTIC))
     {
-        result << uppercase_first(pronoun) << " is vulnerable to silver and"
-                                              " hated by Zin.\n";
+        result << uppercase_first(pronoun)
+			   << " " << conjugate_verb("are", plural) << " "
+			   << "vulnerable to silver and hated by Zin.\n";
     }
 
     if (mons_class_flag(mi.type, M_STATIONARY)
@@ -3970,8 +3976,9 @@ static string _monster_stat_description(const monster_info& mi)
     if (mons_class_flag(mi.type, M_COLD_BLOOD)
         && get_resist(resist, MR_RES_COLD) <= 0)
     {
-        result << uppercase_first(pronoun) << " is cold-blooded and may be "
-                                              "slowed by cold attacks.\n";
+        result << uppercase_first(pronoun)
+			   << " " << conjugate_verb("are", plural) << " "
+			   << "cold-blooded and may be slowed by cold attacks.\n";
     }
 
     // Seeing invisible.
@@ -3980,7 +3987,9 @@ static string _monster_stat_description(const monster_info& mi)
 
     // Echolocation, wolf noses, jellies, etc
     if (!mons_can_be_blinded(mi.type))
-        result << uppercase_first(pronoun) << " is immune to blinding.\n";
+        result << uppercase_first(pronoun)
+			   << " " << conjugate_verb("are", plural) << " "
+			   << "immune to blinding.\n";
     // XXX: could mention "immune to dazzling" here, but that's spammy, since
     // it's true of such a huge number of monsters. (undead, statues, plants).
     // Might be better to have some place where players can see holiness &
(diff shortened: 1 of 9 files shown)


MarvinPA:
I had starting working on implementing this myself recently but hadn't gotten very far at all yet, so this is definitely appreciated! At a glance, one minor coding convention thing is that Crawl uses 4 spaces to indent, rather than tabs - you can run util/checkwhite to check and fix those cases.

In terms of git workflow, for the purposes of a PR here it's fine to force push and rewrite history to make things neater (so you could rebase small fixes like d20c6fa rather than making it a separate commit). On merging everything will probably get squashed into a smaller number of commits so it's not a big deal either way, but it makes the PR easier to read in the meantime.

Changing the gender of monsters other than ancestors for additional testing would require a little more work - you can see how that's implemented currently by looking at M_MALE and M_FEMALE flags in mon-flags.h, so there'd need to be a new M_NEUTRAL and some additional handling/refactoring for that in a few places (mons_class_gender and monster::conj_verb in particular, I believe). It'd probably also be fine to leave that out for the purposes of this PR and just get ancestors working first.

Translations aren't really being actively worked on at the moment unfortunately, but they currently only cover descriptions (dat/descript/*.txt) anyway, so these changes wouldn't affect them.

crawl-ref/source/mon-info.cc
```diff
@@ -1841,3 +1841,12 @@ const char *monster_info::pronoun(pronoun_type variant) const
...
     return mons_pronoun(type, variant, true);
 }
+
+const bool monster_info::pronoun_plurality() const
+{
+	if (props.exists(MON_GENDER_KEY))
+	{
+		return props[MON_GENDER_KEY].get_int() == GENDER_NEUTRAL ? true : false;
+	}
+	return false;
```
MarvinPA: For conditionals like this with just a single line condition followed by a single line of of conditional code, you can skip the braces, for example:
```
if (props.exists(MON_GENDER_KEY))
    return props[MON_GENDER_KEY].get_int() == GENDER_NEUTRAL ? true : false;
```
Running util/unbrace should also catch these and fix them for you.

crawl-ref/source/mon-info.cc
```diff
@@ -1841,3 +1841,12 @@ const char *monster_info::pronoun(pronoun_type variant) const
     }
     return mons_pronoun(type, variant, true);
 }
+
+const bool monster_info::pronoun_plurality() const
+{
+	if (props.exists(MON_GENDER_KEY))
+	{
+		return props[MON_GENDER_KEY].get_int() == GENDER_NEUTRAL ? true : false;
```
MarvinPA: For conditionals like this with just a single line condition followed by a single line of of conditional code, you can skip the braces, for example:
```
if (props.exists(MON_GENDER_KEY))
    return props[MON_GENDER_KEY].get_int() == GENDER_NEUTRAL ? true : false;
```
Running util/unbrace should also catch these and fix them for you.

crawl-ref/source/describe.cc
```diff
@@ -4012,22 +4026,36 @@ static string _monster_stat_description(const monster_info& mi)
...
+        }
         _add_energy_to_string(
             speed, me.spell,
-            mi.is_actual_spellcaster() ? "casts spells" :
-            mi.is_priest()             ? "uses invocations"
-                                       : "uses natural abilities", fast, slow);
-        _add_energy_to_string(speed, me.special, "uses special abilities",
+            mi.is_actual_spellcaster() ? conjugate_verb("cast", plural) + " spells" :
+            mi.is_priest()             ? conjugate_verb("use", plural) + " invocations"
+                                       : conjugate_verb("use", plural) + " natural abilities", fast, slow);
```
crystalsolenoid: These three lines are too long now, right? I wasn't sure how to fix them but keep it readable.

MarvinPA: I think it'll end up a bit unwieldy no matter what, but the simplest thing is probably just to add a linebreak after each conjugate_verb(), eg:
```
mi.is_actual_spellcaster() ? conjugate_verb("cast", plural)
                             + " spells" :
[...]
```

MarvinPA:
Hmm, handling monster::conj_verb this way unfortunately doesn't work as intended. Most messages using conj_verb refer to the monster by name rather than by pronoun - as implemented this gives incorrect messages like "<ancestorname> are engulfed in flames."

I couldn't actually see any cases where monster::conj_verb would need to be conjugated as a plural (although I could easily have missed something), so it may be that this handling isn't needed and there just need to be checks elsewhere for the smaller number of messages that do use pronouns. The other notable case for that I spotted would be the description strings in directn.cc.

crystalsolenoid:
> The other notable case for that I spotted would be the description strings in directn.cc.
Okay, I'll work on fixing up directn.cc.

Is there a way to just undo a commit to put monster::conj_verb back to the way it was, or do I have to do it manually?

Thank you for all the feedback, by the way! :)

crystalsolenoid:
I'm noticing that I'm adding `conjugate_verb(some_verb, mi.pronoun_plurality())` all over the place. Would it make sense for the monster_info class to have a function that did that instead of repeating it over and over?

MarvinPA:
You can use "git revert \<commit hash\>" (so in this case 4a21fad) to undo that single commit. The other thing you can do is to use git rebase to remove that commit from the branch's history - I think the easiest way to do that is with "git rebase -i" which should give you a list of all your commits, and lets you pick and choose which commits to keep/edit/discard.

>
> I haven't tested it, but it looks like conjugate_verb("don't",plural) wouldn't work as intended? Does english.cc have a case for de-pluralizing contractions?

Looks like there isn't any handling for it, but I'd maybe reword that message to "They have not noticed you." since that would work correctly and is nice and concise too.

> 
> I'm noticing that I'm adding `conjugate_verb(some_verb, mi.pronoun_plurality())` all over the place. Would it make sense for the monster_info class to have a function that did that instead of repeating it over and over?

It might be fine as-is, since this way it's still pretty clear what's going on and why the plurality check is required (and having more functions that conjugate verbs in different ways could get confusing). I'm not certain on what best coding practice would really be there though, it could be fine either way.

Thanks for your work on this, messing with grammar is a pretty daunting first Crawl project to dive into!

crystalsolenoid:
I have finals and graduation this week, so if I disappear it's only temporary.
Other places I've found that might need looking at at some point:
- monster_info::wounds_description_sentence in mon-info.cc
- maybe line 1258 of delay.cc (but I don't know what it's doing so I'm not sure)
- line 104 of spl-goditem.cc (would never come up for the ancestor because you can't worship two gods)
I searched for "pronoun" in all the code and these were the only ones left I could find that would need it.

MarvinPA:
The delay.cc line is for the messages that warn you about monster equipment (eg "A goblin comes into view. It is wielding a dagger of distortion."), so would apply if extended to other monsters but wouldn't be applicable for ancestors.

q3cpma:
Would be nice if this was optional in some way (build time, for example) because this may look ridiculous to  some people, especially the not native english speakers.

alexjurkiewicz:
What part do you mean looks "ridiculous"? This PR is about improving grammar.

crystalsolenoid:
This content is already optional. Hep ancestors are always randomly generated as being either male or female. You have to use the Ancestor Identity ability to remember your ancestor as neither.

Just to clarify, non-unique monsters still all use "it" with this modification.

q3cpma:
The "ridiculous" part is when neutral masculine is the de facto standard in english, with singular they/them being a quite rare method being dug up for mainly political reasons. For non native speakers less experienced in english, this could be quite confusing.
>You have to use the Ancestor Identity ability to remember your ancestor as neither.
I see. Still ridiculous, since "neither" doesn't make sense when you refer to a known "thing". But, well, It's less invasive than I thought; shouldn't have jumped the gun.

MarvinPA:
Singular they is widely used - if you're unfamiliar with it then you'll quickly get the hang of it I'm sure, and I'm intending to extend its usage to some other uniques once this PR is merged which should help on that count.

Also consider that labelling peoples' pronouns as "ridiculous" isn't acceptable behaviour - it's worth explicitly noting here that any further comments along those lines won't be tolerated.

crystalsolenoid:
Okay, I changed all the little things that we discussed previously. I fixed the wounds description sentences, but I'm going to leave those other two files I found because they're irrelevant to the Hep ancestor.

I think this is finished now. I'm not 100% sure how to test this and double-check all the cases are working, but because it's just changing text and the cases I can see are working, I'm not too worried.

Anything else I need to do? Thanks again for all the support, I'm really excited to finally be contributing! 😄

MarvinPA:
I'll try and give it a full look through soon! Also, how would you like to be credited in CREDITS.txt?

crystalsolenoid:
Oh, hmm. I guess just as Quinten Konyn.

MarvinPA:
Merged, thanks again for the patches!

PR #766, opened 2018-05-25
Fix hints mode

dplusplus:
This PR contains several fixes for hints mode.

- add newlines
- add untag function
- remove redundant periods
- fix 'OBJ_MISSILES' hints (when you don' t have it)
- fix mentions about p)ray command
- fix typo

PR #1082, opened 2019-06-10
Fix hint for item description

dplusplus:
In #766 (and 2b1cce4), I made a mistake in the conditional expression. Sorry.


ebering:
Thanks, and not to worry. We all make mistakes sometimes; see 6d73c25c485060c63d18a87116c834595ccf27ae

PR #1089, opened 2019-06-21
Change some charm spells into a permanent buffs.

sdynet:
that some parts of the magic(charm, summoning, necromancy, etc)' require significantly more manipulation than others, which makes users easily tired. The goal of this PR is to improve the QoL for buff-spell users by lowering the fatigue level.

be written by Hellmonk: "Cast them once and they stay on forever, reserving some maximum MP. Max mp reserved is listed in the spell info screen. The formula is spell level/success rate^5. Right now this is evaluated as a double, then summed up for all active buff spells and rounded down; more complicated than just using an int for each of them but there's a little less breakpoint nonsense this way. This is recalculated every turn so you cannot cheat the system with equipment swaps. You have an ability to end all your buffs if you have any active, to restore your max (but not current) mp."

The spell to which this mechanism applies is Infusion, Song of Slaying, and Regeneration, Spectral Weapon, Darkness, Deflect Missiles, Ring of Flames. Permanent buff has improved in convenience but has become more powerful than necessary in the process. So the spell level and effect changed a little.
Song of Slaying is now a level 3 spell. The maximum stack is affected by the spell power.
Regeneration is is now a level 4 spell. Effect is affected by the spell power.
Spectral Weapon is now a level 5 spell. The damage sharing function of this spell has been remove. because, frequent buff swaps are expected from users who are concerned about the damage sharing.

[original PR: https://github.com/Hellmonk/hellcrawl/commit/82e5c05650abaeaf790ebdc8bcaf580c8233b5fa]


ebering:
While I appreciate the effort that's gone into this PR, I don't think a permanent buff system is the direction we want to take charms. 

I plan to look at Charms in 0.25 with an eye towards making the more noxious ones into items whose effect scales on spell skill, but I haven't fleshed out the design yet. I'm open to ideas and input.

PR #1090, opened 2019-06-27
Recompress all tiles with zopfli

alexjurkiewicz:
zopflipng --iterations=500 --filters=01234mepb --lossy_transparent

zopflipng is a thin wrapper around Google's zopfli, available from:
https://github.com/imagemin/zopflipng-bin.


alexjurkiewicz:
This reduces PNG size by about 800k in aggregate.

alexjurkiewicz:
Updated this PR so it only recompresses PNGs that are included in releases as-is (eg not combined into a sprite atlas).

aidanholm:
Thanks!

PR #1091, opened 2019-06-27
Allow travelling to altars, shops, and branches in ^O

wjchen-vlsi:
Currently, players need to use `Ctrl+O` to check discovered altars, but then have to use `Ctrl+F` to travel to the desired one.

This change makes it possible to travel to the nearest discovered altar within the Dungeon Overview
1. Within Dungeon Overview, press `_`.
2. Players will then be asked to enter the initial of a god. 
(`A/a` for Ashenzari except for TSO, which is replaced by `1`.)
3. Auto-travel to the nearest (level distance -wise) altar is then started.

This is done by modifying `G` such that `G_` can now be used to select the desired altar for travel, and `_` in dungeon overview is made a shortcut to `G_`.

However, this introduces UI inconsistencies now that `_` can only be pressed in dungeon overview while G only works in the main UI.

Therefore, `G` and `$` keypresses in dungeon overview are made shortcuts to G prompts and `^F "shop"`, respectively.

To do so, `StashTracker::search_stashes()` is modified to accept an optional predefined keyword. `StashTracker::search_stashes()` now directly return search results without asking what to search for when the keyword is specified.

### DIFF
--- a/crawl-ref/source/command.cc
+++ b/crawl-ref/source/command.cc
@@ -606,6 +606,11 @@ void show_interlevel_travel_depth_help()
     show_specific_help("interlevel-travel.depth.prompt");
 }
 
+void show_interlevel_travel_altar_help()
+{
+    show_specific_help("interlevel-travel.altar.prompt");
+}
+
 void show_stash_search_help()
 {
     show_specific_help("stash-search.prompt");
--- a/crawl-ref/source/dgn-overview.cc
+++ b/crawl-ref/source/dgn-overview.cc
@@ -18,6 +18,7 @@
 #include "feature.h"
 #include "files.h"
 #include "libutil.h"
+#include "macro.h"
 #include "menu.h"
 #include "message.h"
 #include "mon-poly.h"
@@ -55,12 +56,13 @@ static void _seen_altar(god_type god, const coord_def& pos);
 static void _seen_staircase(const coord_def& pos);
 static void _seen_shop(const coord_def& pos);
 static void _seen_portal(dungeon_feature_type feat, const coord_def& pos);
+static void _process_command(const char keypress);
 
 static string _get_branches(bool display);
 static string _get_altars(bool display);
 static string _get_shops(bool display);
 static string _get_portals();
-static string _get_notes();
+static string _get_notes(bool display);
 static string _print_altars_for_gods(const vector<god_type>& gods,
                                      bool print_unseen, bool display);
 static const string _get_coloured_level_annotation(level_id li);
@@ -219,7 +221,7 @@ string overview_description_string(bool display)
     disp += _get_altars(display);
     disp += _get_shops(display);
     disp += _get_portals();
-    disp += _get_notes();
+    disp += _get_notes(display);
 
     return disp.substr(0, disp.find_last_not_of('\n')+1);
 }
@@ -237,7 +239,7 @@ static string _get_seen_branches(bool display)
     disp += "\n<green>Branches:</green>";
     if (display)
     {
-        disp += " (use <white>G</white> to reach them and "
+        disp += " (press <white>G</white> to reach them and "
                 "<white>?/B</white> for more information)";
     }
     disp += "\n";
@@ -365,7 +367,7 @@ static string _get_altars(bool display)
     disp += "\n<green>Altars:</green>";
     if (display)
     {
-        disp += " (use <white>Ctrl-F \"altar\"</white> to reach them and "
+        disp += " (press <white>_</white> to reach them and "
                 "<white>?/G</white> for information about gods)";
     }
     disp += "\n";
@@ -468,7 +470,7 @@ static string _get_shops(bool display)
     {
         disp +="\n<green>Shops:</green>";
         if (display)
-            disp += " (use <white>Ctrl-F \"shop\"</white> to reach them - yellow denotes antique shop)";
+            disp += " (press <white>$</white> to reach them - yellow denotes antique shop)";
         disp += "\n";
     }
     last_id.depth = 10000;
@@ -531,7 +533,7 @@ static string _get_portals()
 }
 
 // Loop through each branch, printing stored notes.
-static string _get_notes()
+static string _get_notes(bool display)
 {
     string disp;
 
@@ -545,6 +547,9 @@ static string _get_notes()
 
     if (disp.empty())
         return disp;
+
+    if (display)
+        return "\n<green>Annotations:</green> (press <white>!</white> to annotate current level)\n" + disp;
     return "\n<green>Annotations:</green>\n" + disp;
 }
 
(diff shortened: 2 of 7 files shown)


crawl-ref/source/command.cc
```diff
@@ -596,6 +596,7 @@ void show_targeting_help()
     cols.add_formatted(1, targeting_help_2, true);
     show_keyhelp_menu(cols.formatted_lines());
 }
+
```
aidanholm: Extra whitespace here. It's a minor thing, but better to have whitespace fixups in a separate commit specifically for that purpose.

crawl-ref/source/travel.cc
```diff
@@ -2214,6 +2219,108 @@ static int _prompt_travel_branch(int prompt_flags)
...
+                line += make_stringf("(%c) %-14s ",
+                                     shortcut_letter,
+                                     altar_name.c_str());
+                ++linec;
+            }
+        }
+        if (!line.empty())
+            mpr(line);
+
+        mprf(MSGCH_PROMPT, "Which altar to go to? (? - help) ");
```
aidanholm: Should probably be "Go to which altar?"

crawl-ref/source/travel.cc
```diff
@@ -2214,6 +2219,108 @@ static int _prompt_travel_branch(int prompt_flags)
...
+    extern map<level_pos, god_type> altars_present;
+
+    if (altars_present.empty())
+        return altar_pos;
+
+    while (true)
+    {
+        clear_messages();
+
+        int linec = 0;
```
aidanholm: This should be `col` instead of `linec`; a fair amount of existing code uses `col` to mean "column".

crawl-ref/source/travel.cc
```diff
@@ -2214,6 +2219,108 @@ static int _prompt_travel_branch(int prompt_flags)
...
+
+            if (has_altar_been_seen)
+            {
+                if (linec == 4)
+                {
+                    linec = 0;
+                    mpr(line);
+                    line = "";
+                }
+                line += make_stringf("(%c) %-14s ",
```
aidanholm: This shouldn't be wrapped

crawl-ref/source/dgn-overview.cc
```diff
@@ -607,12 +607,48 @@ bool unnotice_feature(const level_pos &pos)
...
+    {
+        switch (ch)
+        {
+            case 'G': do_interlevel_travel(); return false;
+            case '_':
+            {
+                if (!altars_present.empty())
+                {
+                    bool skip_to_altar_travel = true;
+                    do_interlevel_travel(skip_to_altar_travel);
```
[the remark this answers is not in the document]
wjchen-vlsi: Nice, I didn't know that one exists.

crawl-ref/source/travel.cc
```diff
@@ -2214,6 +2219,108 @@ static int _prompt_travel_branch(int prompt_flags)
...
+                altar_name = "TSO";
+                shortcut_letter = '1';
+            }
+            else
+            {
+                altar_name = uppercase_first(god_name(god, false));
+                shortcut_letter = altar_name.at(0);
+            }
+
+            int altar_ldist = 1000;
```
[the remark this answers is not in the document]
wjchen-vlsi: I made it 1000 to make sure the first match will always be recorded. Upon closer inspection, I guess I didn't fully understand how `level_distance()` works back then. I have refactored the code and am using a different approach to do this now.

wjchen-vlsi:
I made the suggested changes.
Thanks for the code review, it means a lot to me.

crawl-ref/source/travel.cc
```diff
@@ -2214,6 +2219,126 @@ static int _prompt_travel_branch(int prompt_flags)
...
+        return level_pos();
+
+    level_pos nearest_altars[NUM_GODS];
+    int best_altar_dist = -1;
+    const level_id curr = level_id::current();
+
+    // Populate nearest_altars[] with nearest altars
+    for (const auto &entry : altars_present)
+    {
+        if (entry.second >= NUM_GODS)
```
aidanholm: Is this check necessary? Perhaps it should be an assert?

crawl-ref/source/travel.cc
```diff
@@ -2214,6 +2219,126 @@ static int _prompt_travel_branch(int prompt_flags)
...
+            continue;
+
+        if (nearest_altars[entry.second].is_valid())
```
aidanholm: I'd be inclined to rewrite this slightly; something more like this:
```c++
int dist = level_distance(curr, entry.first.id);
if (dist == -1)
    continue;

level_pos *best = &nearest_altars[entry.second];
int old_dist = best.is_valid() ? level_distance(curr, best) : INT_MAX;

if (dist < old_dist)
    *best = entry.first; 
```

This way, unreachable altars are immediately skipped via `continue`. Here, `INT_MAX` is a stand-in for infinite distance. This also makes the if condition a bit more readable.

crawl-ref/source/travel.cc
```diff
@@ -2214,6 +2219,126 @@ static int _prompt_travel_branch(int prompt_flags)
...
+        god_list.insert(god_list.end(), nt_god_list.begin(), nt_god_list.end());
+
+        // list gods in the same order as dgn-overview.cc lists them.
+        for (const god_type god : god_list)
+        {
+            if (!nearest_altars[god].is_valid())
+                continue;
+
+            // "The Shining One" is too long to keep the same G menu layout
+            altar_name  = (god == GOD_SHINING_ONE)? "TSO" : god_name(god, false);
```
aidanholm: No need for parentheses around the test condition:
```c++
altar_name  = god == GOD_SHINING_ONE ? "TSO" : god_name(god, false);
```
Also, `false` is the default for the second `god_name` parameter, so that can be safely omitted.

crawl-ref/source/travel.cc
wjchen-vlsi: As discussed on ##crawl-dev, this is necessary because of the faded altar (i.e., GOD_ECUMENICAL). Comment added to improve code clarity.

wjchen-vlsi:
Made the suggested changes and did the followings:

1. Merge those fixups into their original commits.
2. Addition of `_god_from_initial()` is now a separate commit.
3. Rebase on top of maser.

Thanks!

crawl-ref/source/dgn-overview.cc
```diff
@@ -607,12 +608,43 @@ bool unnotice_feature(const level_pos &pos)
...
+{
+public:
+    dgn_overview(const string& text = "") : formatted_scroller(FS_PREWRAPPED_TEXT, text) {};
+
+private:
+    bool process_key(int ch) override
+    {
+        switch (ch)
+        {
+            case 'G': do_interlevel_travel(); return false;
```
aidanholm: I only realized this now, but `do_interlevel_travel()` calls should be moved out as well, for the same reason; `ctrl-o`, `G`, `?` will currently show the overview screen underneath the travel help popup.

wjchen-vlsi: You are right, and it looks like `^O`, `_`, `?` behaves the same, too. (and therefore should be moved out as well).

wjchen-vlsi:
Should be good now.
Also, I refactored it a bit now that most conditional expressions won't be in `process_key()`.

aidanholm:
Merged! Thanks!

PR #1098, opened 2019-07-06
Add setting a waypoint to clua

marksg07:
As `travel.set_exclude` and `travel.waypoint_delta` exist as clua functions, I believe `travel.set_waypoint(int waynum, int x, int y)` is a pretty harmless and useful addition to the `travel` clua module. This function sets the position of waypoint #`waynum` to `(x, y)` in player-centered coordinates. In addition, this PR slightly refactors the current `TravelCache.add_waypoint` function to separate out the logic of actually setting the coordinates of a specific waypoint (which is now `TravelCache.set_waypoint`) from the logic of getting and parsing the player's input for which waypoint should be set. The `set_waypoint` clua function simply calls `travel_cache.set_waypoint`.

### DIFF
--- a/crawl-ref/source/l-travel.cc
+++ b/crawl-ref/source/l-travel.cc
@@ -26,7 +26,7 @@ LUAFN(l_set_exclude)
     s.y = luaL_safe_checkint(ls, 2);
     const coord_def p = player2grid(s);
     if (!in_bounds(p))
-        return luaL_error(ls, "Coordinates out of bounds: (%d, %d)", p.x, p.y);
+        return luaL_error(ls, "Coordinates out of bounds: (%d, %d)", s.x, s.y);
     // XXX: dedup w/_get_full_exclusion_radius()?
     int r = LOS_RADIUS;
     if (lua_gettop(ls) > 2)
@@ -48,7 +48,7 @@ LUAFN(l_del_exclude)
     s.y = luaL_safe_checkint(ls, 2);
     const coord_def p = player2grid(s);
     if (!in_bounds(p))
-        return luaL_error(ls, "Coordinates out of bounds: (%d, %d)", p.x, p.y);
+        return luaL_error(ls, "Coordinates out of bounds: (%d, %d)", s.x, s.y);
     del_exclude(p);
     return 0;
 }
@@ -129,7 +129,7 @@ LUAFN(l_set_waypoint)
     s.y = luaL_safe_checkint(ls, 3);
     const coord_def p = player2grid(s);
     if (!in_bounds(p))
-        return luaL_error(ls, "Coordinates out of bounds: (%d, %d)", p.x, p.y);
+        return luaL_error(ls, "Coordinates out of bounds: (%d, %d)", s.x, s.y);
     travel_cache.set_waypoint(waynum, p.x, p.y);
     return 0;
 }
--- a/crawl-ref/source/l-travel.cc
+++ b/crawl-ref/source/l-travel.cc
@@ -13,7 +13,7 @@
 #include "travel.h"
 
 /*** Set an exclusion.
- * Uses player centered coordinates
+ * Uses player-centered coordinates
  * @tparam int x
  * @tparam int y
  * @tparam[opt=LOS_RADIUS] int r
@@ -36,7 +36,7 @@ LUAFN(l_set_exclude)
 }
 
 /*** Remove an exclusion.
- * Uses player centered coordinates
+ * Uses player-centered coordinates
  * @tparam int x
  * @tparam int y
  * @function del_eclude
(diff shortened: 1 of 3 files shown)


crawl-ref/source/travel.cc
```diff
@@ -3962,8 +3962,13 @@ void TravelCache::add_waypoint(int x, int y)
...
     }
 
-    int waynum = keyin - '0';
+    set_waypoint(keyin - '0', x, y);
 
+}
+
+void TravelCache::set_waypoint(int waynum, int x, int y)
+{
+    // Assumes that the waynum input is a valid number between 0 and 9 inclusive.
```
rawlins: Might be good to instead of this comment to add an ASSERT, e.g. `ASSERT_RANGE(waynum, 0, 10);`. In this case, there's also a #define for the upper bound (in travel.h), so even better would be: `ASSERT_RANGE(waynum, 0, TRAVEL_WAYPOINT_COUNT);`

crawl-ref/source/l-travel.cc
```diff
@@ -112,6 +112,28 @@ LUAFN(l_waypoint_delta)
...
+ * @tparam int waynum
+ * @tparam int x
+ * @tparam int y
+ * @function set_waypoint
+ */
+LUAFN(l_set_waypoint)
+{
+    int waynum = luaL_safe_checkint(ls, 1);
+    if (waynum < 0 || waynum > 9)
+        return 0;
```
rawlins: possibly print an error of some kind here with `luaL_error`.

crawl-ref/source/l-travel.cc
```diff
@@ -112,6 +112,28 @@ LUAFN(l_waypoint_delta)
...
+    coord_def s;
+    s.x = luaL_safe_checkint(ls, 2);
+    s.y = luaL_safe_checkint(ls, 3);
+    const coord_def p = player2grid(s);
+    if (!in_bounds(p))
+        return 0;
```
rawlins: same here

amalloy:
This sets waypoints correctly, but still uses the old message "Waypoint 0 re-assigned to your current position" even though it's now possible to set waypoints to someplace other than your current location. I think I'm happy merging it as-is - people using the Lua console probably won't be too bothered by a wrong message - but if you think you can clean it up easily that would be better.

crawl-ref/source/l-travel.cc
marksg07: will do

marksg07:
@amalloy, same message appears when adding a waypoint while eXamining level map. How about something for both cases like "Waypoint N [re-]assigned to D:X."

crawl-ref/source/travel.cc
marksg07: done

crawl-ref/source/l-travel.cc
marksg07: done; also added luaL_errors to other bounds checks

marksg07:
made requested changes. The only waypoint message I changed is the one printed when you re-assign a waypoint from somewhere on a certain level to somewhere else on the same level, which was previously "Waypoint n re-assigned to your current position" and is now "Waypoint n re-assigned to D:X".

crawl-ref/source/l-travel.cc
```diff
@@ -112,6 +112,28 @@ LUAFN(l_waypoint_delta)
...
+{
+    int waynum = luaL_safe_checkint(ls, 1);
+    if (waynum < 0 || waynum >= TRAVEL_WAYPOINT_COUNT)
+        return luaL_error(ls, "Bad waypoint number: %d", waynum);
+    coord_def s;
+    s.x = luaL_safe_checkint(ls, 2);
+    s.y = luaL_safe_checkint(ls, 3);
+    const coord_def p = player2grid(s);
+    if (!in_bounds(p))
+        return luaL_error(ls, "Coordinates out of bounds: (%d, %d)", p.x, p.y);
```
amalloy: I like the idea of more useful error messages, but this makes it even easier to figure out absolute coordinates, right?

With this in my rcfile:

```
{
  function setwp()
    travel.set_waypoint(0, 200, 200)
  end
}
```

I can easily discover from the error message that my current absolute position is (59, 50):

```
Lua error: ...amalloy/src/dcss/crawl-ref/source/dat/dlua/macro.lua:31: [string
 "/home/amalloy/.crawl/init...."]:2: Coordinates out of bounds:
 (259, 250)
```

crawl-ref/source/l-travel.cc
```diff
@@ -112,6 +112,28 @@ LUAFN(l_waypoint_delta)
     return 2;
 }
 
+/*** Set a numbered waypoint.
+ * Uses player-centered coordinates.
```
amalloy: I agree with you that player-centered reads best, but its neighbours use "player centered". They should match, either by making this like the existing comments, or fixing the existing comments.

crawl-ref/source/l-travel.cc
marksg07: Whoops, didn't realize lua errors were visible but that makes a lot of sense upon further thought :) I'll change that to player-centered coords.

PR #1100, opened 2019-07-09
Add !cancellation to Zig loot table (EugeneJudo)

alexjurkiewicz:
"Do cancellation potions not spawn in zigs? They're very useful for surviving -wiz on pan floors when things get out of hand (brilliance doesn't help, or at least it still leaves things unreliable to cast.) My supply is running low, while it's pretty rare that I ever use them."

### DIFF
--- a/crawl-ref/source/dat/dlua/ziggurat.lua
+++ b/crawl-ref/source/dat/dlua/ziggurat.lua
@@ -587,9 +587,10 @@ local function ziggurat_create_loot_at(c)
                                   dgn.good_scrolls)
   local super_loot = dgn.item_spec("| no_pickup w:7000 /" ..
                                    "potion of experience no_pickup w:190 q:1 /" ..
-                                   "potion of mutation no_pickup w:190 /" ..
+                                   "potion of mutation no_pickup w:185 /" ..
                                    "potion of mutation no_pickup w:40 q:1 /" ..
                                    "ration no_pickup w:80 /" ..
+                                   "potion of cancellation q:5 no_pickup / " ..
                                    "potion of heal wounds q:5 no_pickup / " ..
                                    "potion of haste q:5 no_pickup / " ..
                                    dgn.good_scrolls)


crawl-ref/source/dat/dlua/ziggurat.lua
```diff
@@ -587,9 +587,10 @@ local function ziggurat_create_loot_at(c)
                                   dgn.good_scrolls)
   local super_loot = dgn.item_spec("| no_pickup w:7000 /" ..
                                    "potion of experience no_pickup w:190 q:1 /" ..
-                                   "potion of mutation no_pickup w:190 /" ..
+                                   "potion of mutation no_pickup w:185 /" ..
                                    "potion of mutation no_pickup w:40 q:1 /" ..
                                    "ration no_pickup w:80 /" ..
+                                   "potion of cancellation q:5 no_pickup / " ..
```
amalloy: The stuff in `super_loot` is all extremely rare, isn't it? I don't zig much but I wouldn't expect to see a pile of 5 !haste very often. Is !cancellation good enough to be spawned in the same way? I'd expect it to just appear naturally on occasion as either `*` or `|`, and for that to be "enough". If it never shows up in that way, it seems like we should fix `*` instead of fixing zig's `super_loot`. Have you confirmed whether it in fact never spawns?

alexjurkiewicz: Cancellation is pretty rare in the first zig, based on gammafunk's objstat data:

| Item | Level | Num | NumMin | NumMax | NumSD | NumHeldMons | NumPiles | PileQuant |
| -- | -- | -- | -- | -- | -- | -- | -- | -- |
| potion of cancellation | Zig | 0.36 | 0 | 4 | 0.64 | 0 | 0.33 | 1.08 |

I don't have objstat data for subsequent runs, but I guess the numbers stay pretty constant as I don't see there being major differences between the rate from `*` vs `|`.

Cancellation is a rare potion. From gammafunk's 3-rune data you can only find 4-8 in the connected dungeon (well everywhere but hell/pan/tomb/abyss/zig). Resistance is about equally rare and haste is about twice as common. I'm ambivalent about making it more common. Ultra-rare and highly-situational consumables have limited impact on the game outside weird corner cases like megazigs.

crawl-ref/source/dat/dlua/ziggurat.lua
```diff
@@ -587,9 +587,10 @@ local function ziggurat_create_loot_at(c)
                                   dgn.good_scrolls)
   local super_loot = dgn.item_spec("| no_pickup w:7000 /" ..
                                    "potion of experience no_pickup w:190 q:1 /" ..
-                                   "potion of mutation no_pickup w:190 /" ..
+                                   "potion of mutation no_pickup w:185 /" ..
```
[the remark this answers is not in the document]
alexjurkiewicz: #1101

ebering:
For loot table tweaks like this it would be helpful if you could run obj stat for comparison if its possible for you to leave one running for a bit. 

I have an obj stat running now to see the effect of this on Zig loot.

PR #1102, opened 2019-07-09
Check for known map bounds in the clua travel library

marksg07:
The set_exclude, del_exclude, and set_waypoint functions all previously ignored the bounds normally set on your cursor in map mode, i.e. you can only move your cursors to points within the bounding box of the known map. These bounds being ignored by the above functions meant both that clua could do something that couldn't be done manually, and that your absolute coordinates could be leaked by careful use of the above functions. This fix will prevent both from happening.

### DIFF
--- a/crawl-ref/source/l-travel.cc
+++ b/crawl-ref/source/l-travel.cc
@@ -8,46 +8,16 @@
 #include "branch.h"
 #include "cluautil.h"
 #include "coord.h"
-#include "env.h"
 #include "player.h"
 #include "terrain.h"
 #include "travel.h"
-
-static bool _in_map_bounds(int x, int y)
-{
-    // code to get bounds stolen from viewmap.cc
-    int min_x = GXM, max_x = 0, min_y = 0, max_y = 0;
-    bool found_y = false;
-
-    for (int j = 0; j < GYM; j++)
-        for (int i = 0; i < GXM; i++)
-        {
-            if (env.map_knowledge[i][j].known())
-            {
-                if (!found_y)
-                {
-                    found_y = true;
-                    min_y = j;
-                    if (y < min_y)
-                        return false;
-                }
-
-                max_y = j;
-
-                if (i < min_x)
-                    min_x = i;
-
-                if (i > max_x)
-                    max_x = i;
-            }
-        }
-
-    return y <= max_y && x >= min_x && x <= max_x;
-}
+#include "map-knowledge.h"
 
 static bool _in_map_bounds(const coord_def& p)
 {
-    return _in_map_bounds(p.x, p.y);
+    std::pair<coord_def, coord_def> b = known_map_bounds();
+    return p.x >= b.first.x && p.y >= b.first.y
+        && p.x <= b.second.x && p.y <= b.second.y;
 }
 
 /*** Set an exclusion.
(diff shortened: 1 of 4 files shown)


crawl-ref/source/l-travel.cc
```diff
@@ -8,10 +8,48 @@
...
 #include "cluautil.h"
 #include "coord.h"
+#include "env.h"
 #include "player.h"
 #include "terrain.h"
 #include "travel.h"
 
+static bool _in_map_bounds(int x, int y)
+{
+    // code to get bounds stolen from viewmap.cc
```
amalloy: Please extract this into a function shared by both uses instead of copying it. I'd suggest that it return a pair of coordinates forming a bounding box, as the existing code in `show_map` needs; then your usage here can just be a small wrapper around that that checks whether your target point is inside the box. It's slightly more expensive than what you're written now, since you won't be able to short-circuit, but that's not as important here as avoiding duplication.

marksg07: was going to do that originally but got bummed out because returning multiple values from a function sucks before c++17, will change

marksg07: Changed

crawl-ref/source/map-knowledge.h
```diff
@@ -41,3 +41,10 @@ map_feature get_cell_map_feature(const map_cell& cell);
 bool is_explore_horizon(const coord_def& c);
 
 void reautomap_level();
+
+/**
+ * @brief Get the bounding box of the known map.
+ *
+ * @return vector of [topleft coord, bottomright coord] of bbox.
+ */
+FixedVector<coord_def, 2> known_map_bounds();
```
amalloy: `pair<coord_def, coord_def>` seems like a more fitting type to me. Did you reject this intentionally because you think a `FixedVector<2>` is better? If so I'm happy to hear why.

crawl-ref/source/map-knowledge.cc
```diff
@@ -326,3 +326,31 @@ bool map_cell::update_cloud_state()
     // TODO: track decay & vanish appropriately (based on some worst case?)
     return false;
 }
+
+FixedVector<coord_def, 2> known_map_bounds() {
+    int min_x = GXM, max_x = 0, min_y = 0, max_y = 0;
```
amalloy: Do you know why they handle x and y differently here? It seems to me like they could be perfectly symmetric.

marksg07: It's because the outer loop is over y coordinates, so the y's we encounter are monotonically increasing. That means the first hit has to be the minimum and the last hit has to be the maximum.

crawl-ref/source/map-knowledge.h
marksg07: No reason, really; my other consideration was a tuple but they're annoying to work with before C++17. Either way seems fine to me.

marksg07: changed and pushed; thank you

crawl-ref/source/map-knowledge.h
```diff
@@ -41,3 +42,10 @@ map_feature get_cell_map_feature(const map_cell& cell);
 bool is_explore_horizon(const coord_def& c);
 
 void reautomap_level();
+
+/**
+ * @brief Get the bounding box of the known map.
+ *
+ * @return vector of [topleft coord, bottomright coord] of bbox.
```
amalloy: This should say pair, not vector (and use curly braces probably?)

marksg07: done

PR #1103, opened 2019-07-09
Allow running clua from map viewing mode; add clua functions to get information about map cursor

marksg07:
A clua function can be run by adding a level map keymap from a given key to ===functionName. When the function ends, we will still be in map mode unless the function kicked us out of it.

The new clua functions `crawl.in_map_mode`, `crawl.cursor_pos`, and `crawl.cursor_level` all give information specific to map mode. `cursor_level` only validly returns if in map mode, and `cursor_pos` only validly returns if in map mode and the cursor level is the same as the player's level.

### DIFF
--- a/crawl-ref/source/state.h
+++ b/crawl-ref/source/state.h
@@ -30,6 +30,15 @@ public:
     int      depth;
 };
 
+struct map_mode_state
+{
+    // Absolute position + level_id of the cursor
+    level_pos cursor_lpos;
+
+    // Stored level_id of the player character
+    level_id original_lid;
+};
+
 // Track various aspects of Crawl game state.
 struct game_state
 {
@@ -138,6 +147,12 @@ struct game_state
 
     // Should flushing a nonempty key buffer error or crash? Used for tests.
     bool nonempty_buffer_flush_errors;
+
+    // Whether we are in "map mode", i.e. currently in eXamine
+    bool in_map_mode;
+
+    // Info pertaining to map mode, if we are in map mode. If we aren't, value is undefined.
+    map_mode_state map_mode_info;
 
 protected:
     void reset_cmd_repeat();
--- a/crawl-ref/source/viewmap.cc
+++ b/crawl-ref/source/viewmap.cc
@@ -38,6 +38,7 @@
 #include "view.h"
 #include "viewchar.h"
 #include "viewgeom.h"
+#include "delay.h"
 
 #ifdef USE_TILE
 #endif
@@ -812,12 +813,19 @@ bool show_map(level_pos &lpos,
             c_input_reset(true);
 #ifdef USE_TILE_LOCAL
             const int key = getchm(KMC_LEVELMAP);
-            command_type cmd = key_to_command(key, KMC_LEVELMAP);
 #else
             const int key = unmangle_direction_keys(getchm(KMC_LEVELMAP),
                                                     KMC_LEVELMAP);
-            command_type cmd = key_to_command(key, KMC_LEVELMAP);
 #endif
+            command_type cmd;
+            if (is_userfunction(key))
+            {
+                run_macro(get_userfunction(key).c_str());
+                continue;
+            }
+            else
+                cmd = key_to_command(key, KMC_LEVELMAP);
+
             if (cmd < CMD_MIN_OVERMAP || cmd > CMD_MAX_OVERMAP)
                 cmd = CMD_NO_CMD;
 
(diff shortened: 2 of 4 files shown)


crawl-ref/source/state.h
```diff
@@ -139,6 +139,16 @@ struct game_state
     // Should flushing a nonempty key buffer error or crash? Used for tests.
     bool nonempty_buffer_flush_errors;
 
+    // Whether we are in "map mode", i.e. currently in eXamine
+    bool in_map_mode;
+
+    // If we are in map mode, absolute level pos of the cursor (position and branch id)
```
amalloy: Would it be cleaner to bundle these up into a struct? Also, I'm not super happy about these fields which are always set but might be meaningless/undefined if `in_map_mode` is false. We probably can't do anything much better, lacking std::optional, but if you can see an improvement I'd love to.

crawl-ref/source/viewmap.cc
```diff
@@ -1345,6 +1358,7 @@ bool show_map(level_pos &lpos,
     tiles.set_map_display(false);
 #endif
 
+    crawl_state.in_map_mode = false;
```
amalloy: Technically I think you should restore it to what it was previously, not to `false`. A lua function may cause us to enter map mode while already in map mode, somehow. Happily, this is what `unwind_bool` will do anyway, so this won't matter.

crawl-ref/source/state.h
marksg07: I do like the idea of having all the info about map mode be in one struct; I do think the boolean should be separate though. As for the undefinedness problem, this struct could be served through a getter, but I think that's somewhat overcomplicating things.

marksg07: structified the latter two variables, let me know what you think

crawl-ref/source/state.h
```diff
@@ -30,6 +30,16 @@ struct god_act_state
     int      depth;
 };
 
+struct map_mode_info_t
+{
+public:
```
amalloy: Leave out the `public` here, since it's the default. I see there's another struct in this file with an explicit `public`, but it's not the norm.

crawl-ref/source/state.h
```diff
@@ -30,6 +30,16 @@ struct god_act_state
...
+    // If we are in map mode, absolute level pos of the cursor (position and branch id)
```
amalloy: I think you can leave off the "If" preface in these comments. Really anytime you have a valid `map_mode_info` its members are valid; the comment in `game_state` explaining that we should sometimes treat that particular `map_mode_info` as invalid is sufficient.

crawl-ref/source/state.h
```diff
@@ -30,6 +30,16 @@ struct god_act_state
     int      depth;
 };
 
+struct map_mode_info_t
```
amalloy: We don't typically end our structs with `_t`.

crawl-ref/source/state.h
```diff
@@ -30,6 +30,16 @@ struct god_act_state
...
+{
+public:
+    // If we are in map mode, absolute level pos of the cursor (position and branch id)
+    level_pos cursor_lpos;
+
+    // If we are in map mode, the level id we started map mode from (i.e. your starting branch+depth).
```
amalloy: "starting" is a weird word to describe this, but I don't see anything obviously better. "The level the player is actually on" seems okay but it leaves murky the question of "the player character" or "the attention of the human being playing the game".

I'm okay leaving as-is if you don't have any better ideas either.

crawl-ref/source/state.h
```diff
@@ -139,6 +149,12 @@ struct game_state
     // Should flushing a nonempty key buffer error or crash? Used for tests.
     bool nonempty_buffer_flush_errors;
 
+    // Whether we are in "map mode", i.e. currently in eXamine
+    bool in_map_mode;
```
amalloy: This needs to be initialised in `game_state::game_state()`. There is no default initialisation for struct members, so leaving it uninitialised is undefined behaviour. I think `map_mode_info` itself can be left uninitialised, since you always write to it before reading from it, but it wouldn't hurt to initialise it anyway as a safeguard.

crawl-ref/source/state.h
marksg07: I think that "the level the player character is on" works as it's more accurate in the probably-impossible-but-maybe-someday case that we start map mode from somewhere that is not the player's position.

marksg07: (or specifically, the "player character" phrasing.)

PR #1108, opened 2019-07-13
Rework pain card's level 2 and cloud card

sdynet:
Level 2 pain card's old effect (summon flayed ghost) was an unsuitable effect on the deck of destruction. Also, flay effect is just another torment.
New effects: cast torment and bolt of draining. Torment does not damage the caster.

The old cloud card produces clouds that damage the enemy. This effect is more suitable for deck of destruction, so the card is moved to deck of destruction. It was also changed accordingly to create clouds under the enemy's feet, so that it could be used more aggressively. Instead, the size of clouds decreased from 3x3 to 1x1.

### DIFF
--- a/crawl-ref/source/decks.cc
+++ b/crawl-ref/source/decks.cc
@@ -89,7 +89,6 @@ deck_archetype deck_of_escape =
     { CARD_TOMB,       5 },
     { CARD_EXILE,      1 },
     { CARD_ELIXIR,     5 },
-    { CARD_CLOUD,      5 },
     { CARD_VELOCITY,   5 },
     { CARD_SHAFT,      5 },
 };
@@ -102,6 +101,7 @@ deck_archetype deck_of_destruction =
     { CARD_DEGEN,      3 },
     { CARD_WILD_MAGIC, 5 },
     { CARD_STORM,      5 },
+    { CARD_CLOUD,      5 },
 };
 
 deck_archetype deck_of_summoning =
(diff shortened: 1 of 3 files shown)


crawl-ref/source/decks.cc
```diff
@@ -1122,7 +1122,8 @@ static void _damaging_card(card_type card, int power,
 
     dist target;
     zap_type ztype = ZAP_DEBUGGING_RAY;
-    const zap_type painzaps[2] = { ZAP_AGONY, ZAP_BOLT_OF_DRAINING };
+    const zap_type painzaps[3] = { ZAP_AGONY, ZAP_BOLT_OF_DRAINING,
+                                   ZAP_BOLT_OF_DRAINING };
```
ebering: I think just tormenting is a fine Tier 3 effect for pain; no need to zap bolt of draining too.

crawl-ref/source/spl-goditem.cc
```diff
@@ -1190,6 +1190,7 @@ void torment_player(actor *attacker, torment_source_type taux)
     {
     case TORMENT_CARDS:
     case TORMENT_SPELL:
+    case TORMENT_CARDS_PAIN:
```
ebering:
```suggestion
    case TORMENT_CARD_PAIN:
```

`TORMENT_CARDS` shouldn't have been plural; but we don't have to repeat past mistakes.

crawl-ref/source/spl-goditem.cc
```diff
@@ -1232,8 +1233,8 @@ void torment_player(actor *attacker, torment_source_type taux)
 void torment_cell(coord_def where, actor *attacker, torment_source_type taux)
 {
     if (where == you.pos()
-        // The Sceptre of Torment doesn't affect the  wielder.
-        && !(attacker && attacker->is_player() && taux == TORMENT_SCEPTRE))
+        // The Sceptre of Torment doesn't affect the  wielder, also pain card.
```
ebering:
```suggestion
        // The Sceptre of Torment and pain card do not affect the user.
```

crawl-ref/source/spl-goditem.cc
```diff
@@ -1232,8 +1233,8 @@ void torment_player(actor *attacker, torment_source_type taux)
...
+        && !(attacker && attacker->is_player() && (taux == TORMENT_SCEPTRE || taux == TORMENT_CARDS_PAIN)))
```
ebering: This line is longer than 80 chars, it should be wrapped, like so

```suggestion
        && !(attacker && attacker->is_player() 
                && (taux == TORMENT_SCEPTRE || taux == TORMENT_CARDS_PAIN)))
```

PR #1109, opened 2019-07-14
Add crawl.get_target

marksg07:
This clua function opens the targeting dialog in the same way it is opened when you press x to examine. The user is allowed to move the cursor around (same as examine). If the user makes a selection and presses enter, this function returns the player-relative coordinates of the selection. Note that the target is valid even if it is past the bounds of the map, so there is no leak of world borders. If the user cancels the selection, this function returns nil.

### DIFF
--- a/crawl-ref/source/directn.cc
+++ b/crawl-ref/source/directn.cc
@@ -827,7 +827,7 @@ void full_describe_view()
 
 void do_look_around(const coord_def &whence)
 {
-    dist lmove = _look_around_target(whence);
+    dist lmove = _look_around_target(you.pos() + whence);
     if (lmove.isValid && lmove.isTarget && !lmove.isCancel
         && !crawl_state.arena_suspended)
     {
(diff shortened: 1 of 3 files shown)


crawl-ref/source/directn.cc
```diff
@@ -825,6 +826,25 @@ void full_describe_view()
...
+    if (lmove.isValid && lmove.isTarget && !lmove.isCancel
+        && !crawl_state.arena_suspended)
+    {
+        start_travel(lmove.target);
+    }
+}
+
+bool get_look_position(coord_def *c)
+{
+    dist lmove = _look_around_target(coord_def(0, 0));
```
amalloy: 0,0 here makes it sound like it will put the player's cursor at the origin. What if you started at the player's position instead, and then subtracted the player's position afterwards? That makes it clearer that we're using player-centered coordinates.

rawlins:
For commit messages, please format with max 72 chars width (vim `:set tw=72`). (We don't seem to worry about the full 50/72 rule though, just max 72 for title as well.)

might also be nice to describe a simple use case in the commit message and/or the docstring so that an rc scripter can see how they might use this (I can imagine one or more must exist, but I'm not sure exactly what you have in mind).

marksg07:
@rawlins @amalloy Fixed issues and added example, check it out.

PR #1110, opened 2019-07-14
Ensure stash update on being shafted (minqmay)

wjchen-vlsi:
Currently, if a player discovers an altar and get shafted by a trap at the same time, the altar will appear in dungeon overview but not in stash tracker.

This change ensures items, shops, and altars discovered at the moment the player gets shafted are correctly registered.


aidanholm:
Thanks!

PR #1113, opened 2019-07-15
Fix wizdump loading when rotted or Gozag follower.

CollinJ:
Recognize "HP:" and "MP:" as morgue line starts. These happen when
the character is rotted (see output.cc#L2173)
Recognize Gozag which doesn't have a piety meter, causing the God line
to only have 8 fields.

Sample morgue file: !lg collin38 502 -log


ebering:

CollinJ:
CollinJ is fine, thanks!

PR #1116, opened 2019-07-17
Change chance of triggering paralysis/pain bond to be turn-length inv…

ByrelMitchell:
…ariant

Previously, this formula gave you a significant higher chance of triggering a paralysis or painbond event per aut if you were taking shorter actions:

0 auts remaining on TIMER:
10 aut turn => 0.01 events/aut
5 aut turn => 0.02 events/aut

200 auts remaining on TIMER:
10 aut turn => 0.0033 events/aut
5 aut turn => 0.004 events/aut

This incentivizes doing fast actions whenever you have to wait at higher piety instead of simply resting. It's even optimal when enemies are out of sight, since a paralysis/pain bond event triggering when no enemies are in sight resets the relevant TIMER to 0.

The new formula makes the expected value of events proportional to the duration of the turn, and so eliminates all turn-length dependence except for rounding errors. Event rates are all set to the same as with 10 aut actions previously, regardless of action length. This is probably a slight nerf in general, but a significant nerf to quickblade Uskers.

### DIFF
--- a/crawl-ref/source/player-reacts.cc
+++ b/crawl-ref/source/player-reacts.cc
@@ -431,7 +431,7 @@ static void _handle_uskayaw_time(int time_taken)
     // timer down to a minimum of 0, at which point it becomes eligible to
     // trigger again.
     if (audience_timer == -1 || (you.piety >= piety_breakpoint(2)
-            && x_chance_in_y(time_taken, time_taken * 10 + audience_timer)))
+            && x_chance_in_y(time_taken, 100 + audience_timer)))
     {
         uskayaw_prepares_audience();
     }
@@ -439,7 +439,7 @@ static void _handle_uskayaw_time(int time_taken)
         you.props[USKAYAW_AUDIENCE_TIMER] = max(0, audience_timer - time_taken);
 
     if (bond_timer == -1 || (you.piety >= piety_breakpoint(3)
-            && x_chance_in_y(time_taken, time_taken * 10 + bond_timer)))
+            && x_chance_in_y(time_taken, 100 + bond_timer)))
     {
         uskayaw_bonds_audience();
     }


rawlins:
I was just reminded that I had looked at this while investigating mantis bug [11920](https://crawl.develz.org/mantis/view.php?id=11920) but only commented in ##crawl-dev. This change seems reasonable but I discovered that I didn't understand the originally-intended system well enough to merge it (and haven't had a chance to get back to it unfortunately). Here's the log of me being perplexed back in July or so:

```
11:03:47 <advil> hm, usk pain bond and solo time have at most (i.e. once the timer has run down) a 1 in 10 chance of retriggering after the first trigger, unless piety drops below the threshold and reraises, does that seem as intended?
11:04:00 <advil> it's kind of a weird mechanic, given that there's *also* a timer
11:07:50 <advil> I wonder if `x_chance_in_y(time_taken, time_taken * 10 + bond_timer)` should actually be x_chance_in_y(time_taken * 10, time_taken * 10 + bond_timer)
11:09:05 <advil> though time_taken might already be in auts, so idk
11:15:33 <advil> I guess this was changed in a recent PR but I don't understand the point of the old version either
11:16:55 <advil> %git 02ac524
11:16:55 <Cheibriados> Byrel Mitchell * 0.24-a0-505-g02ac524: Change chance of triggering paralysis/pain bond to be turn-length invariant (2 days ago, 1 file, 2+ 2-) https://github.com/crawl/crawl/commit/02ac52495da2
11:18:56 <advil> oh haha that's not merged, I see
11:21:29 <Lantell> Stable (0.23) branch on underhound.eu updated to: 0.23.1-82-g5129a83184
11:21:41 <advil> anyways, the cooldown is already like 30-50 turns for each of these
11:22:32 <advil> so why then give such a lot probability for retriggering
11:22:33 <advil> *low
11:27:25 <advil> also, is 30-50 turns even plausible given usk piety decay?
11:27:36 <advil> (cf https://crawl.develz.org/mantis/view.php?id=11920)
11:34:43 <advil> anyways, tldr my guess would be that there ought to be 1 in 1 chance of retriggering after the timer runs out, not 1 in 10
```

ByrelMitchell:
Alright, so I can walk through how this currently works, and then speculate on intent. I'm just going to use paralysis; pain bond works the same with a different piety threshold.

On Paralysis, the TIMER is set to 300-500 auts.

Every player action while Usk piety is above 3 stars, you have a TIMER-dependent and turn-length-dependent chance of triggering a paralysis attempt. [graphs](https://imgur.com/a/dvudVqp)

On a paralysis attempt:
  -if there are monsters in sight they are paralyzed and TIMER is reset to 300-500 auts. 
  -if there are no monsters in sight, TIMER is set to 0 auts (which means the average delay till next paralysis attempt is 40-200 auts, depending on player turn length.) 

Pain bond is identical, except it checks for 4 stars of piety.

I think the intent here is to implement a randomized timeout without any value to spoiler knowledge. If you simply make the timeout be 300-500, then a spoiled player would KNOW they aren't getting a reparalysis before 300 auts, and that they're guaranteed one when the timer gets to 480 auts and they breadswing. This would probably affect optimal play and reward watching the aut clock, which is not desirable at all.

Instead, the developer added on a layer of randomization; the 'TIMER' actually just defines the probability that you get a reparalysis at any given time-since-last-paralysis. That way a spoiled player doesn't have those sorts of hard thresholds to reward clock-watching; you could get a reparalysis the very next turn, or not get one for 1000 auts.

I think that's a good design. It's a bit complicated in implementation, but knowing those implementation details gives you essentially no advantage. However, they screwed up their attempt to normalize for player turn length by putting it in the denominator of the odds formula as well. My suggested fix keeps the same probability distribution at 10 aut turn lengths, but normalizes it per aut.  [graphs](https://imgur.com/a/kBiqxyS) (There's an argument that 7 aut turns ought to be kept the same instead of 10 auts; that would just require swapping the constant from 100 to 70.)

ebering:
Going through the PR backlog I've taken the time to understand this, and I think there's a much clearer way of explaining why the change is needed. Set the timer to 0, the chance of triggering is `x_chance_in_y(time_taken, time_taken * 10)` also known as `one_chance_in(10)`; so the roll in these lines is time-independent and is the same on a given turn. When the timer is positive, the chance is partly time dependent, the probability is equivalent to `one_chance_in(10 + timer / time_taken)` which is partially time dependent, converging to giving a fixed per-aut-probability as timer gets large.

Which is to say I'm hitting that merge button, thanks for the fix!

PR #1117, opened 2019-07-18
Fireproof trees

nlavsky:
This PR updates descriptions of trees to make it possible to distinguish flammable and non-flammable trees as well as permanent and summoned ones.


nlavsky:
Redid the 2nd commit: I didn't know it's possible to reuse an existing description for another dungeon feature, which is better than adding more code into describe.cc.

ebering:
Thanks for the patch!

PR #1118, opened 2019-07-19
Add guidelines for no_tele_into usage

alexjurkiewicz:
Discourage overuse and explicitly reject the idea all runed door / transporter vaults require it.

### DIFF
--- a/crawl-ref/docs/develop/levels/guidelines.md
+++ b/crawl-ref/docs/develop/levels/guidelines.md
@@ -3,6 +3,7 @@
 1. [Guidelines for D:1 arrival vaults](#guidelines-for-d1-arrival-vaults)
 2. [Guidelines for creating serial vaults](#guidelines-for-creating-serial-vaults)
 3. [Guidelines for creating ghost vaults](#guidelines-for-creating-ghost-vaults)
+4. [Guidelines for no_tele_into](#guidelines-for-no_tele_into)
 
 ## Guidelines for D:1 arrival vaults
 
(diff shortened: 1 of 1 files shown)


crawl-ref/docs/develop/levels/guidelines.md
```diff
@@ -186,3 +187,10 @@ CHANCE directly in the vault. For ghost vault rooms for Vaults, instead set the
 Other tags we generally require for ghost vaults are `no_tele_into` and
 `no_trap_gen`. If you use `allow_dup` in your vault, also use
 `luniq_player_ghost` to avoid multiple vault placement on the same level.
+
+
+## Guidelines for no_tele_into
+
+The `no_tele_into` KPROP prevents teleports landing you on the tagged locations. Don't overuse this. Prevent egregiously unfair situations only.
```
amalloy: Shaft/hatch destinations too.

crawl-ref/docs/develop/levels/guidelines.md
```diff
@@ -186,3 +187,10 @@ CHANCE directly in the vault. For ghost vault rooms for Vaults, instead set the
...
+
+You don't need to tag the interior of all runed door / transporter vaults. For example, no ghost vaults use this KPROP.
```
amalloy: Let's first mention the places where it is required: areas you can't escape from, e.g. teleport closets or islands surrounded by impassable terrain. Ideally of course you should avoid creating such locations at all, but if you do create them they must have an escape hatch out (if that makes sense thematically) or a `no_tele_into`.

If you first explain what it's for, your followup of "but don't overuse it in cases xyz" will make more sense.

PR #1119, opened 2019-07-23
New morgue section `screenshots`

giann:
Notes can have an associated screenshot. For now only user notes take
screenshots. They are printed out in the `screenshots` section of the
morgue.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -2266,7 +2266,7 @@ dump_message_count = 20
 
 dump_order  = header,hiscore,stats,misc,inventory,
 dump_order += skills,spells,overview,mutations,messages,screenshot,
-dump_order += monlist,kills,notes,skill_gains,action_counts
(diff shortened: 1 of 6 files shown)


giann:
- [ ] add ansi colors

ebering:
The limit of screenshots to one per turn is still desired. Looking through this it looks like you'll also need some save compatibility code: when you unmarshallString to load the screenshot this should be wrapped in a TAG_MINOR check, you can see the adjacent packed level code for a quick example.

I would like to add this to 0.25, if you have the time to make the changes great, otherwise I will try to get to it.

giann:
I'll try and look into it

giann:
Done. @ebering not sure I should have added a new tag for the version check. Let me know.

ebering:
Yes, any incremental shift in the binary format of the save file (which saving the screenshots requires) should be accompanied by a new tag. The CI is mad at you because you didn't wrap your if statement in a preprocessor, so when the tag-major-upgrade script is run it breaks the build. I've fixed that.

Thanks for the patch!

giann:
👍

PR #1120, opened 2019-07-24
Add threat level indicators for monsters in offline tiles

GnarledBeach:
Webtiles and console both have a monster list at the bottom-right of
the screen which lists all monsters in view color-coded by rough
threat level. Many players have expressed a desire for something
similar in offline tiles, to provide always-visible indicators of the
threat levels of all enemies in view. This commit adds translucent
colored backgrounds -- dark grey, light grey, yellow, and red --
to indicate trivial, easy, tough, and nasty monsters respectively.

Additionally, these indicators are configurable with a new option in
the options file, "tile_show_threat_levels". Each type of indicator
can be independently enabled or disabled by including or omitting it
in the option file line. For example, the following...

"tile_show_threat_levels = tough nasty"

...will show only yellow (tough) and red (nasty) threat indicators.

I have experimented with various styles of indicators, as shown in the
following thread:

https://crawl.develz.org/tavern/viewtopic.php?f=8&t=26439

I am open to changing back to a different style, or even making
multiple styles available and letting the player choose which style to
use, if any. This style is the best in my opinion, as it is very
noticeable while being subtle enough not to interfere with any other
monster or monster-associated graphics. I have playtested this feature
through a 6-rune run with no problems, technical nor cosmetic.


crawl-ref/source/tilepick.cc
```diff
@@ -1933,6 +1933,19 @@ tileidx_t tileidx_monster(const monster_info& mons)
         ch |= TILE_FLAG_GD_NEUTRAL;
     else if (mons.neutral())
         ch |= TILE_FLAG_NEUTRAL;
+    else if (mons.threat == MTHRT_TRIVIAL &&
```
alexjurkiewicz: crawl coding style is to use braces for each statement here, because the test takes more than one line:

```c
    else if (mons.threat == MTHRT_EASY &&
        Options.tile_show_threat_levels.find("easy") != string::npos)
    {
            ch |= TILE_FLAG_EASY;
    }
```

Adds a lot of extra lines here though...

alexjurkiewicz: https://github.com/crawl/crawl/blob/master/crawl-ref/docs/develop/coding_conventions.txt

GnarledBeach: Hmm. I hate switch() and never use it in code of my own, but given crawl's conventions, maybe it would be better, as to avoid the extra lines you refer to? Like so:

```
else
{
    switch (mons.threat)
    {
    case MTHRT_EASY:
        if (Options.tile_show_threat_levels.find("easy") != string::npos)
            ch |= TILE_FLAG_EASY;
        break;
    [etc.]
    }
}
```

Let me know if you think that would do, and if so I'll check it in.

GnarledBeach: Checked in the switch-case version a while back. Maybe I'm supposed to hit "Resolve conversation." Not sure whether it matters that much, but going to press it just in case. :)

aidanholm:
Thanks for the PR! Played a few games now with this, works pretty well.

Code looks good; could you please add webtiles support? You'll need to add the same logic in tiledgnbuf.cc to cell_renderer.js; the two are analagous. You'll also need to update enums.js with another call to `fg_flags.exclusive_flags.push` (I think); then you can just access `cell.fg.TRIVIAL` etc from cell_renderer.js.

Separately, please squash the five commits into one commit as well, since most of the commits are fixes, not functionality changes.

aidanholm:
Thanks for the PR; I've moved work on this to the `threat-levels` branch.

GnarledBeach:
Hi Aidan,

Interesting to hear from you again on this. I've had the older follow-up
(requesting consolidation of commits and support added to webtiles) sitting
in my inbox since you sent it, trying to decide whether to follow up...

You see, despite being a former software developer, I burned out on it and
really don't like to do it that much anymore. I was motivated to contribute
this feature because I badly wanted it. I didn't think webtiles needed it
because it already has the color-coded list at the bottom right, thus
already providing an always-visible indicator of apparent danger levels.
And I'm not much of a web developer or Javascript guy -- most of my
background and passion is in lower-level development, C and C++ being where
I am most comfortable. So I have been pretty ambivalent about following
up...apologies.

Cheers,
Caleb

aidanholm:
Hey Caleb,

No worries, that happens to all of us sooner or later. Thanks for all of the hard work you've already put into this feature; it should be merged soon!

Cheers,
Aidan

GnarledBeach:
Awesome, thanks for understanding and cool to hear that it will probably
still get merged!

Cheers,
Caleb

PR #1123, opened 2019-07-27
Update quotes.txt

rmcneive:
Just adding a couple literature quotes.


reaverb-Crawl:
Merged, thank you for the patch!

I have credited you as rmcneive in CREDITS.txt. Please inform me here if you would like to be credited differently.

PR #1124, opened 2019-08-01
Warn the player about leaving a ziggurat figurine behind (u/Quatzecoatl)

nlavsky:
Ziggurat figurines are usually buried below other loot and could be
easy to miss. This commit adds a confirmation prompt when exiting
Zig:27 without the figurine.

This change was suggested here:
    https://www.reddit.com/r/dcss/comments/al6bxh/

### DIFF
--- a/crawl-ref/source/main.cc
+++ b/crawl-ref/source/main.cc
@@ -1367,6 +1367,18 @@ static bool _prompt_stairs(dungeon_feature_type ygrd, bool down, bool shaft)
         }
     }
 
+    // Leaving ziggurat figurines behind.
+    if (ygrd == DNGN_EXIT_ZIGGURAT
+        && you.depth == brdepth[BRANCH_ZIGGURAT]
+        && find_floor_item(OBJ_MISCELLANY, MISC_ZIGGURAT))
+    {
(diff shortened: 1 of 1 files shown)


gmarks-coatue:
+1, if I'd seen this post I would have PR'ed this myself. I've personally lost a megazigger this way, so definitely a meaningful change.

marksg07:
+1, if I'd read this post 6 months ago it probably would've inspired me to make my first contribution. Having lost a megazigger this way myself, I can attest that this would be a meaningful change. @ebering, you're quoted in the above reddit thread 👀

reaverb-Crawl:
Thank you for the patch!

reaverb-Crawl:
I have updated CREDITS.txt to credit yo under "Nikolai Lavsky" . If you would like to be credited in different way, please tell me so I can change it.

Thanks again!

nlavsky:
Thank you for merging this PR! I'm already in the Credits.txt though ([line 238](https://github.com/crawl/crawl/blob/98e85ec6/crawl-ref/CREDITS.txt#L238)).

reaverb-Crawl:
Fixed the extraneous credit in 0b98641fa2. Thank you for pointing that out!

Issue #984, opened 2019-02-10
Mutation text names items I don't actually have.


KoenDG:
Platform: http://crawl.xtahua.com:8080/#play-dcss-0.23
So, the server in Paris.

Version: 0.23

Description: Here's a screenshot: https://i.imgur.com/itWEu0q.png

So I was testing a unidentified potion I picked up, it was a potion of mutation. Its text stated: "You feel your magical essence link to your wands."

The character is a Minotaur Fighter that started with a Flail. This is on the first level of the dungeon. The text speaks of wands but I have no wands. Does it mean something else or perhaps it's hardcoded?

rawlins:
That's somewhat confusingly worded, so perhaps could stand to be improved, but while you have the mutation, any wands that you find will draw on your MP.

KoenDG:
>any wands that you find will draw on your MP.

So anything I evoke will require MP?
Or what does that mean?

alexjurkiewicz:
Not anything you evoke, only wands. With this mutation, to evoke a wand
will cost 3 MP.

theJollySin:
I took a stab at making the warning for this mutation more clear:

https://github.com/crawl/crawl/pull/1143

My recommendation is to change the text to be more clear. Having said that, if the developers *want* the mystery to remain in the warning, we could close this ticket as "won't fix".

PR #1133, opened 2019-08-12
Rephrase MP-powered wands player alert (#984)

srvanmeter87:
...since the original wording was confusing to those without wands.


ebering:
Closed via 6aba940

ebering:

ebering:
I've added you as the name attached to your GitHub profile, if you would prefer not to appear or to appear under a different name please either let us know or just submit a PR.

PR #1134, opened 2019-08-14
additional insight on pyYaml

amcmullan:
Just went through the steps and used them to compile crawl. initially I ran into trouble with the version of python suggested here. Continued to get "no module named yaml" errors during compile, until installing the python3 pip package, and using 'pip install pyyaml' again. no issues, after that. I'm new here, but I wanted to help the next person who comes along. hopefully I'm not going out of bounds with this PR.

### DIFF
--- a/crawl-ref/INSTALL.txt
+++ b/crawl-ref/INSTALL.txt
@@ -317,7 +317,11 @@ from within the MSYS2 Shell.
 
   You can verify PyYAML is installed by running `python -m yaml`, which should
   give an error like "'yaml' is a package and cannot be directly imported"
-  (rather than "No module named yaml").
+  (rather than "No module named yaml"). If, after installing the above you still have 
+  issues, try installing python3 as well.
+  
+  pacman -S mingw64/mingw-w64-x86_64-python3-pip
+  pip install pyyaml
 
 * To get the Crawl source, follow the steps in the Source Code Checkout
   section above to clone Crawl into your MSYS2 home directory. We recommend


rawlins:
>hopefully I'm not going out of bounds with this PR.

No, it's fine, we've received several reports like this lately for windows. I do still want to double check them myself before merging when I get a chance, because our pyyaml dependency is supposed to work for either python 2 or 3 and something must have recently changed that is messing with this.

One small note is that we prefer commit messages that are at most 72 characters wide (in vim: `:set tw=72`). This can be fixed up by whoever merges this.

Goratrix:
Hi guys, I was one of the people reporting this some time ago (in ##crawl-dev on 7/31). The problem happens because the default python in MSYS2 runs version 3, and pip has separate installers for 2 and 3 (and those then install python packages for their respective versions only). The above solution (installing pip version 3) works fine, but if you want to make it even simpler, I found that pip doesn't need to be installed at all. It's possible to install PyYAML directly from the MSYS2 package manager: https://packages.msys2.org/package/mingw-w64-i686-python3-yaml using 'pacman -S mingw-w64-x86_64-python3-yaml'

So that whole section on installing pip and verifying could be replaced with just the following:

* To install PyYAML, run:

  pacman -S mingw-w64-x86_64-python3-yaml

amcmullan:
Thanks rawlins, I'll try to remember about commit message width.

That seems like a much cleaner solution Goratrix, nice find!

Goratrix:
Also after some googling, here are some notes and links on when MSYS2 did the change: https://github.com/msys2/MINGW-packages/issues/4950

PR #1135, opened 2019-08-15
Upgrade Wyrmbane's tile according to its enchantment

nlavsky:
As a followup to 1cd5daa3, this commit adds code to upgrade Wyrmbane's
tile as the lance grows more powerful.

It'll look like this:
wyrmbane_tiles


gammafunk:
Thanks for the implementation!

PR #1136, opened 2019-08-16
Expand access to floor items

ufshaikh:
This commit allows wielding weapons (and other items) and wearing armour directly from the floor. It also gives the ability to see and select from all items when quaffing, reading scrolls, and selecting the output of reading a scroll (e.g., ?branding) by pressing '*' in the selection menu. See commit message for more details.

This partially implements a [User Interface Implementable](https://github.com/crawl/crawl/wiki/User-Interface-Improvements#expand-floor-interaction). (Edit: now completed by later commits in the branch, modulo my dealing with full inventory in a way other than suggested there.)

This does not allow putting on jewellery or using evocables from the floor. (Edit: a later commit now deals with jewellery.) I'd think these to be natural next steps. This commit should make that easier though.

### DIFF
--- a/crawl-ref/source/delay.cc
+++ b/crawl-ref/source/delay.cc
@@ -767,7 +767,7 @@ void JewelleryOnDelay::finish()
 #ifdef USE_SOUND
     parse_sound(WEAR_JEWELLERY_SOUND);
 #endif
-    puton_ring(jewellery.link, false, false);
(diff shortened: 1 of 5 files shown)


ufshaikh:
Also, I suggest this not be merged until bug #0012046 is resolved somehow, since that can produce a crash and is much more likely to actually happen to people if the option to view all items is available.

PR #1137, opened 2019-08-20
fix mutations/vamp-attrs toggle

dplusplus:
When player is Vampire:
- Toggling works wrong. (console)
  - Hit 'A' -> '!', 'Mutations' is still highlighted, then, Hit '!', 'Blood properties' is highlighted.
- Didn't highlight mutations/vamp-attrs tab (webtile)


aidanholm:
Thanks for the fix!

PR #1141, opened 2019-08-27
Add missing feature tiles to the stash search menu

nlavsky:
Currently, the stash search menu doesn't show tiles for several features, including net traps, large runed doors, and permanent teleport traps:

missing_tiles

This PR fixes that by handling tile picking for these features separately.


aidanholm:
Thanks for the PR!

PR #1143, opened 2019-09-02
#984 making text more clear for wands mutation

theJollySin:
Trying to making the warning more clear for the wands mutation (issue #984 ).

I note that ticket has been open a while, so I assume people would like more clarity in the warning. The only other option I see is that we WANT the warning to be a bit mysterious, in which case I would suggest denying this Pull Request and closing #984 as "Won't Fix".

### DIFF
--- a/crawl-ref/source/mutation-data.h
+++ b/crawl-ref/source/mutation-data.h
@@ -1833,8 +1833,8 @@ static const mutation_def mut_data[] =
   "MP-powered wands",
 
   {"You expend magic power (3 MP) to strengthen your wands.", "", ""},
-  {"You feel your magical essence link to your wands.", "", ""},
-  {"Your magical essence is no longer linked to your wands.", "", ""},
+  {"You feel your magical essence would now be drained by using wands.", "", ""},
+  {"Your magical essence will no longer be drained by using wands.", "", ""},
 },
 
 { MUT_UNSKILLED, 0, 3, mutflag::bad, false,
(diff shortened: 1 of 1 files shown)


theJollySin:
Bump.
Any thoughts?  I know there are a lot of PRs open, and it's easy to lose them in the shuffle.

rawlins:
Looking back at this, the behavior of MP-powered wands was changed in 92d99f377 to be less surprising, I'm guessing maybe this specific rewording is targeted at the version of this mutation from before that commit? So the new behavior doesn't entail draining MP, but if the player has MP a wand will use it, and get a corresponding power boost.

theJollySin:
@rawlins Okay, I took a look at the commit you linked. It seems to me this mutation is now pretty much entirely positive. (Unless I'm missing something.) So, I have changed the re-wording to be more positive.  I think it's a good option. BUT we can close this pull request and keep the original, more mysterious, wording if the goal is to be less explicit.

crawl-ref/source/mutation-data.h
```diff
@@ -1833,8 +1833,8 @@ static const mutation_def mut_data[] =
   "MP-powered wands",
 
   {"You expend magic power (3 MP) to strengthen your wands.", "", ""},
-  {"You feel your magical essence link to your wands.", "", ""},
-  {"Your magical essence is no longer linked to your wands.", "", ""},
+  {"You feel your magical essence would now empower your wands.", "", ""},
+  {"Your magical essence will no longer be empower your wands.", "", ""},
```
oktalk: "Your magical essence will no longer empower your wands."

oktalk: Granitic change.

ebering:
Closed via 6aba940

PR #1145, opened 2019-09-05
Show scarves on player in tiles (11977)

kitchen-ace:
Scarves weren't being shown at all unless the player chose one through
the doll editor. And even then they were practically invisible, since
they were drawn behind the body (like cloaks) instead of in front
where I think they were meant to be.

The green scarf tile looks quite strange to me, but that's a separate
problem.


aidanholm:
Merged; thanks!

PR #1146, opened 2019-09-05
Improve some helmet tiles for draconians

kitchen-ace:
The draconian head looks funny with some helmet tiles. This gives them their own horns tile, and doesn't show the drac head when wearing a full helmet (which is only possible with the mask of the dragon, or the doll editor on local tiles).


kitchen-ace:
Done, thanks

Issue #1125, opened 2019-08-02
Outdated things in the help


Wuzzy2:
**PLATFORM:**
"local tiles"

**VERSION:**
0.23.2

**DESCRIPTION:**
I recently re-read the help and noticed a couple of things have become outdated. Probably someone forgot to update the help.

* Wands: The section about wands still claims that wands need to be identified with scrolls of identify. It also claims that “unidentified” wands take more than 1 use. There are some other false (outdated) claims in this section. Best to rewrite the section completely.
* Religion: This section mentions the “p”ray key for Fedhas but I think its no longer used since a while
* Items: The entry on books mentions the “r”ead key but books can not be read with “r” anymore

theJollySin:
Assign this one to me, and I'll take a crack at it?

rawlins:
@theJollySin we don't really bother with the "assignment" feature of github, but PRs are open and documentation ones usually do get merged fairly quickly, so definitely feel free to submit a PR that fixes these.

theJollySin:
@rawlins Actually, all of three of the above issues with the help menu have been fixed already.

https://github.com/crawl/crawl/blob/master/crawl-ref/docs/crawl_manual.rst

This ticket can be closed.

Wuzzy2:
OK, in the linked file, the first 2 issues are indeed fixed, but not the third one.

> : 	spellbooks 	(use 'r'ead and 'M'emorise and 'z'ap)

theJollySin:
@rawlins Ah! It looks like the changes to spellbooks were made in some places, but not others. I removed the "r" for "read spellbook" command from two places in the documentation.

There is a PR open here:

https://github.com/crawl/crawl/pull/1147

theJollySin:
I believe this issue can be closed, as it was solved by Pull Request number 1147.

PR #1147, opened 2019-09-08
#1125 spellbooks are no longer 'r'ead

theJollySin:
The "wands" and "religion" issues raised in ticket #1125 were already fixed.  And in some places the spellbooks problems were already fixed. But not in all places.

This is a minor documentation fix for spellbooks, since they are no longer "r"ead.


aidanholm:
Thanks for the fix!

PR #1148, opened 2019-09-10
Change gauntlet teleportation message (bugfix)

ufshaikh:
On entering the gauntlet the current message misleadingly suggests no
one can teleport, but the restriction is on the player only.


aidanholm:
Thanks!

PR #1149, opened 2019-09-11
Fix HP of summons created via the Phantom Mirror spell

nlavsky:
Because of an incorrect formula added in a559921a, Mara and rakshasa
clones had 200% hp of their original instead of 50%.

Regular monsters copied via `cast_phantom_mirror()` had 286% hp
and max hp instead of 35%.


aidanholm:
Lol; thanks for the fix!

PR #1151, opened 2019-09-17
Alpine Linux / musl support

alexjurkiewicz:


alexjurkiewicz:
Need to retest this before it's ready for merge

alexjurkiewicz:
Alpine Linux also doesn't install a `python` symlink, so #1269  is required

alexjurkiewicz:
After working on crawl in docker more I'm not going with Alpine as a base image, so this is no longer relevant.

aidanholm:
@alexjurkiewicz sorry for the delay in getting to this one; would the doc edits still be relevant?

alexjurkiewicz:
I've pulled out the non-Alpine improvements into #1275

Head-on-a-Stick:
Is there any chance the change described in https://github.com/crawl/crawl/pull/1151/commits/17844727410be6fc8dd6258b0868848569472a4f could be applied?

I've managed to get an Alpine package built by using a patch with that alteration but it would be nice if that wasn't needed.

rawlins:
a check for that case looks reasonable but I'm nervous about that particular `#if`, I would guess from just reading it that it's way too broad. @alexjurkiewicz any thoughts? This is the kind of thing that really ought to get a configure-style check for the header file, not a preprocessor check, but we don't have great support for that.

alexjurkiewicz:
I am terrible at make, pre processor and configure so not sure why you're
asking me 😅

It would be theoretically nice to support other libc than gnu and musl but
I don't see it realistically happening. Anyone building on such a system
will be more than capable of patching this. So from that perspective the
current logic of "Linux and not glibc" seems fine.

However people can't patch what they don't know about. Maybe the make file
should detect this scenario and print a warning that we think the libc is
musl and crash logging is nerfed.

draeath:
Please consider implementing just the MUSL execinfo fix! You're not the only one who wants to build this on Alpine :(

ref: 17844727410be6fc8dd6258b0868848569472a4f

PR #1154, opened 2019-09-22
Fix a crash when cloning an almost dead monster (geekosaur)

nlavsky:
A copy created via rakshasa's Phantom Mirror has only 35% of
the target's hp and max hp. If the original monster is almost dead
(1-2 HP), its copy will have 0 hp, which crashes the game.

Fix this by giving cloned monsters at least 1 hp. Also, it allows
rakshasas to clone lurking horrors, who always spawn with only 1 hp.

Thanks to geekosaur for finding what was causing two recent crashes on CUE.


aidanholm:
Thanks!

Issue #891, opened 2018-11-02
Crash if you have a typo in Arena mode


ghost:
**PLATFORM:**
Only tested it using local tiles on Windows. 

**VERSION:**
0.22.1

**DESCRIPTION:**
If you mistype something in Arena mode, it displays the message:

>Fatal error:
>Expected arena monster spec "xxx v yyy", but got "test"
>Hit any key to exit, ctrl-p for the full log.

Then exits the program, instead of going back to arena mode.

The-Fireplace:
I've tested this on local tiles for Linux, with the same results.

rawlins:
These issues should be fixed as of https://github.com/crawl/crawl/commit/33f6f7b652e4, which makes a number of improvements to error reporting in arena.

PR #1155, opened 2019-09-23
Fixes Issue #891 where Arena crashed if the monster name had a typo

createyourpersonalaccount:
Now it just informs the user of the error and asks again.

### DIFF
--- a/crawl-ref/source/arena.cc
+++ b/crawl-ref/source/arena.cc
@@ -1069,7 +1069,10 @@ namespace arena
             catch (const arena_error &error)
             {
                 write_error(error.what());
-                game_ended_with_error(error.what());
+                if(error.fatal)
+                  {
+                    game_ended_with_error(error.what());
+                  }
                 continue;
             }
             do_fight();


aidanholm:
Hi, thanks for the PR!

Fix looks good, but please reformat the code to be consistent with crawl's code style, like this:
```c++
if (error.fatal)
{
    game_ended_with_error(error.what());
}
```
There should be a space after `if` and braces should be directly under the `i` of `if`.

In addition, please adjust the commit message to use a more consistent format. Something like "Fix crawl exiting after arena fight spec typo (gh #891)"

You can make these changes by using `git commit --amend` and then force-pushing to your repository. I notice that you're working directly on your master branch; you may want to push from a separate branch so that you don't accidentally get any merge commits with upstream in your PR.

Thanks!

createyourpersonalaccount:
I think I did something wrong with the force push.

aidanholm:
The commit message has been updated, but the contents haven't changed. They might not have been included if you didn't stage them (`git add`) before amending the commit.

rawlins:
Can you clarify what this does on top of https://github.com/crawl/crawl/commit/33f6f7b652e4? Since that commit, afaik arena hasn't crashed on typos -- is there an error case that I missed?

createyourpersonalaccount:
It turns out my "fix" doesn't fix anything. You are right, it is already fixed.

PR #1156, opened 2019-09-26
Change spell chances in Kiku's second book

kitchen-ace:
Kiku's Recieve Corpses spell really works best with Animate Dead, which
makes is frustrating when you don't get it -- especially since you
might not get Animate Skeleton either.

This commit changes the spells in Kiku's second book to always give you
Animate Dead as well as Dispel Undead. The other spell options are also
changed, to make them a little more predictable:

* Agony or Exrcuciating Wounds (pure pain damage)
* Bolt of Draining or BVC (can be used behind undead allies)
* Simulacrum or Death Channel (high level necro allies)

### DIFF
--- a/crawl-ref/source/randbook.cc
+++ b/crawl-ref/source/randbook.cc
@@ -992,36 +992,12 @@ void make_book_kiku_gift(item_def &book, bool first)
     }
     else
     {
-        chosen_spells[0] = coinflip() ? SPELL_ANIMATE_DEAD : SPELL_SIMULACRUM;
+        chosen_spells[0] = SPELL_ANIMATE_DEAD;
         chosen_spells[1] = (you.species == SP_FELID || coinflip())
-            ? SPELL_BORGNJORS_VILE_CLUTCH : SPELL_EXCRUCIATING_WOUNDS;
-        chosen_spells[2] = random_choose(SPELL_BOLT_OF_DRAINING,
-                                         SPELL_AGONY,
-                                         SPELL_DEATH_CHANNEL);
-
-        spell_type extra_spell;
-        do
-        {
-            extra_spell = random_choose(SPELL_ANIMATE_DEAD,
-                                        SPELL_AGONY,
-                                        SPELL_BORGNJORS_VILE_CLUTCH,
-                                        SPELL_EXCRUCIATING_WOUNDS,
-                                        SPELL_BOLT_OF_DRAINING,
-                                        SPELL_SIMULACRUM,
-                                        SPELL_DEATH_CHANNEL);
-            if (you.species == SP_FELID
-                && extra_spell == SPELL_EXCRUCIATING_WOUNDS)
-            {
-                extra_spell = SPELL_NO_SPELL;
-            }
-
-            for (int i = 0; i < 3; i++)
-                if (extra_spell == chosen_spells[i])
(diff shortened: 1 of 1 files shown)


gammafunk:
Generally speaking, randomized god gifts are more interesting when they have more variability when that's possible and reasonable.

The first part of this commit forces animate dead on the basis of this being a good spell with Recieve Corpses, but this is equally true of Simulacrum. The latter is much more powerful and hence harder to cast, but it has the same kind of dependency on corpses. Some runs you'll get Animate Dead and will have an easily castable mass-corpse-ally spell, other runs you'll have one that's stronger but takes longer to get castable. And in 100% of runs you'll still have a lower-level spell that makes of the corpses that Kiku provides: either Animate Skeleton or Corpse Rot. If there were an actual problem along the lines of "Recieve Corpses not usable without Animate Dead", that would be a problem with the design of Recieve Corpses itself, but I don't think Kiku worshipers actually lack for uses of that ability if they fail to get Animate Dead.

The second part of this commit just reduces the variability of Kiku gifts without direct justification. Again it's better for replayability if randomized god gifts are kept more random, so that players have to adapt to the gifts they get.

kitchen-ace:
Thanks for the feedback.

>I don't think Kiku worshipers actually lack for uses of that ability if they fail to get Animate Dead.

I've had several discussions on various reddit threads with people about this, and it's pretty clear that many people see the possibility of not getting Animate Dead as a huge drawback, especially when it's also possible to not get Animate Skeleton. (I can try to find some of these threads if you are interested.) From my own experience I rarely follow Kiku when I don't already have AD for this reason. Getting some undead allies on demand is extremely helpful early on and one of the main reasons a player might follow Kiku.

Simulacrum is a massive skill investment compared to AD, and for most species not viable until much later in the game. Corpse rot is useful, but doesn't fill the same role (mechanically or thematically) at all.

As for the second part, I'm not going to object to it getting rejected if the randomness is seen as a positive.

gammafunk:
Kiku is a god that can and often does grants ways to make undead allies quite early, but it isn't a god that guarantees this--that would be a niche that Yred fills. People who refuse to worship Kiku unless they have Animate Dead are extremely likely to be underutilizing non-ally necromancy spells. Some of these spells may be less fun to use, but they way to address this is to rework those spells, not to make Kiku always gift the more popular spells.

There certainly isn't a power issue in terms of Kiku being weak if you don't get Animate Dead and/or Skeleton. The other spells Kiku grants are very strong, and some of those strong spells are corpse-using spells, at least one of which will make allies. Probably the best way to get at the general notion motivating this PR is to propose some rework of Corpse Rot, or propose some kind of corpse-using spell that completely replaces it. Preferably not an ally spell, since that is very well covered by Necromancy as it is.

kitchen-ace:
>People who refuse to worship Kiku unless they have Animate Dead are extremely likely to be underutilizing non-ally necromancy spells.

I disagree; I *eventually* use most or all of what I get from Kiku. It's just that Animate Dead is a single-school, level 4 spell, making it a lot easier to cast than many other necro spells, and one that also synergizes particularly well with Receive Corpses.

>Probably the best way to get at the general notion motivating this PR is to propose some rework of Corpse Rot, or propose some kind of corpse-using spell that completely replaces it.

I'll take this under consideration. Thanks for the comments.

PR #1158, opened 2019-09-27
Mark trove portals as single-use

### LINKED ISSUE #1126: Treasure trove description does not mention it's a one-time use portal
**PLATFORM:**
"local tiles"

**VERSION:**
0.23.2

**DESCRIPTION:**
The treasure trove is evil. Should you actually be able to gain entrance, you will learn it's actually a portal that will collapse after first entry. AFTER entering it, of course.

This has totally screwed me the first time I entered a treasure trove. I entered with a near-full inventory, not knowing it was actually a one-time portal (I was never warned), so I was forced to leave most treasure behind … Or exchange it for some of my current items. That was really frustrating!

I assumed I would be able to re-enter it again, like a branch.

The bug (IMO) is that the description of the treasure trove entry does not at all mention it that you can enter the treasure trove only once.

alexjurkiewicz:
Fixes #1126.

### DIFF
--- a/crawl-ref/source/dat/descript/features.txt
+++ b/crawl-ref/source/dat/descript/features.txt
@@ -704,7 +704,8 @@ A passage leading to a sewer. It is gradually rusting away.
 %%%%
 A portal to a secret trove of treasure
 
-A portal leading to a treasure trove.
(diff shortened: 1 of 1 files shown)


aidanholm:
Thanks!

PR #1159, opened 2019-09-27
Don't stop travel for changes from (un)melding

kitchen-ace:
If you are auto-travelling, and you lose stats from armour
melding or unmelding, don't stop auto-travel for that reason
alone. This is really only relevant for merfolk with artefact
boots, but the fix should apply to the general case.

### DIFF
--- a/crawl-ref/source/delay.cc
+++ b/crawl-ref/source/delay.cc
@@ -80,17 +80,6 @@
 #include "view.h"
 #include "xom.h"
 
-class interrupt_block
-{
-public:
-    interrupt_block() { ++interrupts_blocked; }
-    ~interrupt_block() { --interrupts_blocked; }
-
-    static bool blocked() { return interrupts_blocked > 0; }
-private:
-    static int interrupts_blocked;
-};
-
 int interrupt_block::interrupts_blocked = 0;
 
 static void _xom_check_corpse_waste();
--- a/crawl-ref/source/delay.h
+++ b/crawl-ref/source/delay.h
@@ -13,6 +13,25 @@
 #include "operation-types.h"
 #include "seen-context-type.h"
 
+class interrupt_block
+{
+public:
+    interrupt_block(bool block = true) {
+        m_block = block;
+        if (block)
+            ++interrupts_blocked;
+    }
+    ~interrupt_block() {
+        if (m_block)
+            --interrupts_blocked;
+    }
+
+    static bool blocked() { return interrupts_blocked > 0; }
+private:
+    bool m_block;
+    static int interrupts_blocked;
+};
+
 class monster;
 struct ait_hp_loss;
 
--- a/crawl-ref/source/player-equip.cc
+++ b/crawl-ref/source/player-equip.cc
@@ -162,6 +162,8 @@ void equip_effect(equipment_type slot, int item_slot, bool unmeld, bool msg)
     if (msg)
         _equip_use_warning(item);
 
+    const interrupt_block block_unmeld_interrupts(unmeld);
+
     if (slot == EQ_WEAPON)
         _equip_weapon_effect(item, msg, unmeld);
     else if (slot >= EQ_CLOAK && slot <= EQ_BODY_ARMOUR)
@@ -179,6 +181,8 @@ void unequip_effect(equipment_type slot, int item_slot, bool meld, bool msg)
         return;
 
     _assert_valid_slot(eq, slot);
+
+    const interrupt_block block_meld_interrupts(meld);
 
     if (slot == EQ_WEAPON)
         _unequip_weapon_effect(item, msg, meld);


kitchen-ace:
While fixing this bug I discovered a mess of problems related to equipment melding and stat zero, which I'll try to report and/or fix soon. I mention this now in case anyone is wondering about the possibility of stat zero after this patch -- it was already a problem before.

aidanholm:
Thanks for the PR! I'm not a fan of the extra parameter, however. It looks like you should be able to use `interrupt_block` to disable the interruption in `equip_effect` and `unequip_effect`, which would remove the need for the extra parameter.

In order for that to work, you'd need to make `interrupt_block` public, and add a `bool block` parameter to `interrupt_block`, so that you can conditionally block interrupts on the `meld` / `unmeld` parameter.

kitchen-ace:
Hi. I've been poking at this for a while now but can't figure out how to use `interrupt_block` to stop interrupts without adding an extra parameter to a function somewhere. I put `interrupt_block` delay.h and added the the bool, but that's as far as I'm able to get. You (or someone else) is free to fix this bug properly, otherwise I'm afraid I need some more guidance.

aidanholm:
No worries. If you currently have an `interrupt_block(bool block)` class, I believe the next step would be adding an instance at player-equip.cc:164, like so:
```c++
void equip_effect(equipment_type slot, int item_slot, bool unmeld, bool msg)
{
...
    if (msg)
        _equip_use_warning(item);

    interrupt_block(unmeld); // add this line

    if (slot == EQ_WEAPON)
        _equip_weapon_effect(item, msg, unmeld);
    else if (slot >= EQ_CLOAK && slot <= EQ_BODY_ARMOUR)
...
}
```
That should prevent the subsequent calls to `_equip_weapon_effect` etc. from causing any interrupts.
There's a similar line to add to `unequip_effect()`.

Hope this helps! also feel free to hop on ##crawl-dev

kitchen-ace:
Thanks. I'm going to reply here because it's easier to type a lot here than it is on IRC, I hope that's OK.

My understanding of C++ classes (which is not very strong) is that they can't be used the same way as functions? e.g. they can't have arguments or be called like you suggested.

I think the way `interrupt_block` is being used now is that when a new interrupt_block object is created, `interrupts_blocked` is incremented in that object, through the constructor. The class destructor decrements it. `interrupts_blocked` is static so all objects share the same value?

delay.cc:1340 essentially checks if `interrupts_blocked` is > 0, and if so, does not interrupt the current activity. 

So while I can create a new `interrupt_block` object in player-equip.cc, I can't do it conditionally after an `if(meld)` because when that if-block ends, the destructor is called. The only way I see around that is to duplicate the next if-block like this, but that seems like it's probably not the best solution:

    void equip_effect(equipment_type slot, int item_slot, bool unmeld, bool msg)
    {
    ...

        if (msg)
            _equip_use_warning(item);

        if (unmeld)
        {
            const interrupt_block block_unmeld_interrupts;

            if (slot == EQ_WEAPON)
                _equip_weapon_effect(item, msg, unmeld);
            ...
        }
        else
            if (slot == EQ_WEAPON) // duplicate code
                _equip_weapon_effect(item, msg, unmeld);
            ...
        }        

    }

aidanholm:
No worries; that's fine.

Your understanding of how `interrupt_block` works is correct. Your understanding of C++ classes is mostly correct, but I think you're stumbling over the syntax; classes can have constructors, and that notation passes a parameter to the constructor. I remember being completely flummoxed by this syntax as well. See e.g. `mouse_control` for another use of this pattern. `cursor_control` is another one that takes a boolean parameter.

The complicated and duplicated code is precisely why this method requires adding a boolean parameter to the constructor of `interrupt_block`. I think the implementation would be something like this:
```c++
class interrupt_block
{
public:
    interrupt_block(bool block = true) {
        m_block = block;
        if (block)
            ++interrupts_blocked;
    }
    ~interrupt_block() {
        if (block)
            --interrupts_blocked;
    }

    static bool blocked() { return interrupts_blocked > 0; }
private:
    bool m_block;
    static int interrupts_blocked;
};
```
So effectively, if `block` is false, the instance will do nothing. Hope this helps. Feel free to push work-in-progress as well.

kitchen-ace:
>classes can have constructors, and that notation passes a parameter to the constructor. 

Ah ha, that's what wasn't quite sinking in for me. Makes perfect sense now that you've spelled it out. Thanks a lot for walking me through this; the updated commit should work properly.

Two more questions, if you don't mind answering them (they're a bit broad, no need to go into a detailed reply):

1) What about doing it this way is preferable to adding an extra parameter in the `handle_stat_change` function? I gather it's preferable to avoid new parameters, but this way added a parameter to a class still.

2) Is `interrupt_block` a typical use for classes? It seems a bit unorthodox to me since the objects aren't really acted on at all, it's only the act of creating them that does anything. But I'm obviously not too experienced with classes in general.

aidanholm:
> Thanks a lot for walking me through this; the updated commit should work properly.

No problem! We've just hit feature freeze for 0.24, but we should be able to squeeze this one in.

> 1. What about doing it this way is preferable to adding an extra parameter in the `handle_stat_change` function? I gather it's preferable to avoid new parameters, but this way added a parameter to a class still.

Parameters are fine in general, but boolean parameters in particular should be avoided if it's not obvious from the context what they do, especially if there's multiple boolean parameters. Otherwise whenever you see `foo(true, true, false)` you need to check the function declaration. It's true that `interrupt_block` now has a boolean parameter, but it only has one and it's more obvious what it might do. Additionally, since this class is now much more useful, it can be used in more places; boolean parameters in commonly-used RAII classes like this are less mentally taxing because you tend to see the class a lot.

Another major advantage is that you don't have chains of functions that simply pass along a parameter to their callees. The equip code is already pretty bad in that respect, due to how (un)melding and (un)wielding are handled together, but it's best to not make a bad situation any worse. Long chains of functions that do this bind the shape of the functions tightly together, making refactoring and future changes extremely painful.

But the major advantage here is that the new code is focused purely on intent, not on implementation. Apart from the modifications to `interrupt_block`, which are logically distinct, your bugfix has two lines, and reading the code literally says "from here on, ignore interrupts if we're (un)melding"; it doesn't get much better than that!

In general, for this kind of stuff I highly recommend Fowler's Refactoring.

> 2. Is `interrupt_block` a typical use for classes? It seems a bit unorthodox to me since the objects aren't really acted on at all, it's only the act of creating them that does anything. But I'm obviously not too experienced with classes in general.

This use of RAII is relatively common in crawl. The really common one is `unwind_var<>`, which temporarily overrides the value of some variable; very useful. The main advantage is that the destructor always runs no matter how you exit the scope, even if an exception is thrown. E.g. for `unwind_var` the original value is always restored. I believe there's a `no_messages` RAII class, for example; that `bool showMsgs` parameter in the equip code could probably be cleaned up even more.

Since `interrupt_block` is now declared in a header file, I believe any reasonable C++ compiler will inline the constructor and destructor code, so the "object" essentially evaporates, and the assembly should just contain some increments and decrements of `interrupts_blocked`.

PR #1160, opened 2019-10-01
Fix incorrectly displayed success chance for confusion gaze

AdamPG:
The confusion gaze spell uses randomized spell power. To calculate the success chance of confusion gaze when used against the player, the average roll for spell power was used, leading to a mathematically incorrect chance being displayed. This commit corrects this error.

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -2897,11 +2897,14 @@ string get_skill_description(skill_type skill, bool need_title)
     return result;
 }
 
-/// How much power do we think the given monster casts this spell with?
-static int _hex_pow(const spell_type spell, const int hd)
+/** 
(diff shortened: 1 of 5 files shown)


AdamPG:
Playtesting with golden eyes indicates that this formula for success chance agrees with the current one for most values of MR. However, I have confirmed that the current formula incorrectly displays 0% chance when confusion is still possible. This formula correctly displays a 1% chance with the same amount of MR.

ebering:
Thanks for the patch. However, I went for a different solution: Confusion Gaze was the only monster spell with randomized spellpower, so I derandomized the spellpower instead of fixing the display.

AdamPG:
That solution makes more sense and in retrospect I should have done that. Thank you for the quick reply!

PR #1162, opened 2019-10-01
Fix being able to pick out barbs while incapacitated (12065)

Shalmezad:
Carefully pulling out barbs now requires that you can move (not paralyzed or petrified), aren't confused, aren't asleep, and not berserk.


ebering:

Shalmezad:
First name last name is fine (Richard Wardin). Hopefully will have some more patches coming this month.

PR #1164, opened 2019-10-02
Add base stats to species descriptions

alexjurkiewicz:
The need to explicitly mark newlines is a bit of a hack...

### DIFF
--- a/crawl-ref/source/newgame.cc
+++ b/crawl-ref/source/newgame.cc
@@ -1114,6 +1114,77 @@ public:
         m_sub_items->menu_id = m_choice_type == C_JOB ?
             "background-sub" : "species-sub";
         m_sub_items->descriptions = descriptions;
+        
+        auto stat = make_shared<Box>(Box::HORZ);
+        
+        auto attributes = make_shared<Box>(Box::HORZ);
+        auto attributes_str = make_shared<Text>(formatted_string("Base:", LIGHTBLUE));
+        attributes_str->margin = {0, 1, 0, 0};
+        attributes->add_child(attributes_str);
+        auto strength = make_shared<Box>(Box::HORZ);
+        strength->margin = {0, 1, 0, 0};
+        auto strength_str = make_shared<Text>(formatted_string("Str:", WHITE));
+        strength_str->margin = {0, 1, 0, 0};
+        strength->add_child(strength_str);
+        auto strength_num = make_shared<Text>("10");
+        strength->add_child(strength_num);
+        attributes->add_child(strength);
+        auto intelligence = make_shared<Box>(Box::HORZ);
+        intelligence->margin = {0, 1, 0, 0};
+        auto intelligence_str = make_shared<Text>(formatted_string("Int:", WHITE));
+        intelligence_str->margin = {0, 1, 0, 0};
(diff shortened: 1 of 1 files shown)


ebering:
The HP, MP, XP, and MR apts are all displayed in the `%` display aptitudes screen in character creation, which is where these descriptions get displayed. As these four numbers have a meaning different from base stats and each other, I think displaying them without a legend is too misleading.

Replacing them with some text that explains base stats are modified by background, and adding the background stat modifiers to the backgrounds (with an explanation of what they modify) would be grand.

One thing to watch out for: in `80x24` console the species descriptions have 2 lines; there are some extra lines on that screen so it might be ok. I'll check this version for overflows in a moment.

ebering:
Yes, adding this line causes some bad console formatting. @aidanholm has been doing some work with the new game menus recently, but I think using \n in the description DB isn't a good idea. The stats display should probably be a separate widget on the new game menu but Aidan will be able to say better.

aidanholm:
Yeah, the stats display should be a separate widget; the values displayed would be pulled from the species/background data crawl already has.

You could use a simple text widget, but I think a horizontal box with one text widget for each stat type is better; if the box is set to expand horizontally, each text widget will get an equal amount of space, and switching between displayed species won't cause the different fields to shift left and right.

alexjurkiewicz:
Alright, I'll have a go at implementing that. It sounds like a good way to
advertise all the work Aidan has put into the new UI code if I can sneak
this in for 0.24!

It sounds like there isn't any spare space on 80x24 display though, should
we just not display the line if there isn't enough space?

aidanholm:
> AIt sounds like a good way to advertise all the work Aidan has put into the new UI code if I can sneak this in for 0.24!

:)
> It sounds like there isn't any spare space on 80x24 display though, should we just not display the line if there isn't enough space?

The species/background part will get smaller to make the space needed.

Hiding it automatically could be tricky; widgets don't currently have a 'visible' property, although that could be added. There's also a `UIShowHide` class in menu.cc which supports this use case; it'd just have to be moved into ui.h, and maybe renamed.

Second problem is you need to handle screen resizes. You could change the visibility state in `UINewGameMenu::_allocate_region()`, based on the menu's region height, but that will probably force a relayout, since the visibility state might have changed. That'll probably cause an infinite loop unless you fix `UIShowHide::set_visible()` to only queue a layout if the visibility actually changes.

alexjurkiewicz:
Shrinking the species/job list sounds like the way to go then. I'll see if
doing so causes any scrolling, and if so if the scrollability is
discoverable to people (eg people not realising Mummy exists)

aidanholm:
There's no scrollbar on console; one could definitely be added. An alternative is to support displaying the scroll percentage somewhere on the screen.

ebering:
There should be space; the species list has a margin of 1 and the description has a margin of 1, so there's 2 blank lines between them. Reducing the top margin of the description should free up the line you need.

Another approach would be a toggle between "description" and "stats".

alexjurkiewicz:
Thanks for the info. Yes this is early work. I will get back to this next week

alexjurkiewicz:
Sorry, I am stumped by the UI code here. I'm going to stop work on this.

PR #1165, opened 2019-10-02
Increase Depth of cheibrodos_wizards_tower

AdamPG:
cheibrodos_wizards_tower potentially places wizards and orc sorcerers on D:7-12. While vaults containing out-of-depth monsters are fine in principle, this vault is particularly egregious for the following reasons: (1) The vault is open and thus easy to explore into and (2) The spells of these monsters (most notably paralysis and crystal spear) can lead to single-turn deaths. If generated on D:7, this vault at best leads to a floor skip and at worst leads to an exploration death. This commit moves this vault to D:12-15, placing these monsters after the entrance to Lair and closer to where wizards normally appear.

### DIFF
--- a/crawl-ref/source/dat/des/variable/float.des
+++ b/crawl-ref/source/dat/des/variable/float.des
@@ -3924,7 +3924,7 @@ ENDMAP
 
 NAME: cheibrodos_wizards_tower
 TAGS: no_item_gen no_monster_gen no_pool_fixup
-DEPTH: D:7-12, Elf
+DEPTH: D:12-15, Elf
 ORIENT: float
 : if you.branch() == "Elf" then
 MONS: deep elf sorcerer / deep elf annihilator


ebering:
From ##crawl-dev

`<+gammafunk> no one merge that as I will have to leave commentary later
                   when I have time`

gammafunk:
This vault is fine in its current state, although I will adjust it later to reduce the number of spawns in the earliest depths that it places, since placing that many 8-tier monsters at this point is a bit much.

Open layouts and placing out-of-depth monsters, including monsters with dangerous spells, are perfectly acceptable things to do in vaults. Forcing a player to sometimes avoid a floor is a great outcome, since it introduces challenge and non-linearity to exploration. "Unavoidable deaths" due to monster spells is not a consideration we make when designing vaults; that's a consideration made when designing those spells, and those spells are perfectly fine.

The only actual issue with this vault is that it's a bit spoilery to know that there's a dangerous threat, and we do like to minimize vault spoilers when possible. However having a small number of these vaults is certainly fine, and this vault is in a huge pool of vaults that don't place 8-tier monsters.

PR #1167, opened 2019-10-03
Clean up a function

ufshaikh:
Change the function is_habitable_feat to an equivalent formulation which is
simpler and idiomatic.


alexjurkiewicz:
This confused the heck out of me back when I converted species to data files. Nice.

PR #1169, opened 2019-10-05
Mantis Ticket 1583 - Cyclops have one eye

theJollySin:
This is in reference to Mantis 1583:

https://crawl.develz.org/mantis/view.php?id=1583

The problem with the visual is that it generates sentences like:

* "The cyclops covers its eyes."
* "The slime covers its eyes." 

Neither of which make sense.  We could change this visual to something like "The cyclops covers its face."  That's another solution. But... do slimes have faces? Do worms? Dart slugs?

I think a better solution would be to remove the offending visual gag entirely.

### DIFF
--- a/crawl-ref/source/dat/database/monspeak.txt
+++ b/crawl-ref/source/dat/database/monspeak.txt
@@ -639,8 +639,6 @@ VISUAL:@The_monster@ ponders the situation.
 @The_monster@ says, "Why is everything spinning?"
 
 @The_monster@ screams, "I'll kill you anyway!"
-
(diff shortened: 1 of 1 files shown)


aidanholm:
Merged manually at 36f0aeb843; thanks for the PR!

theJollySin:
> Merged manually at 36f0aeb

Do you guys do these merges manually so you can use 'git rebase' instead of 'git merge', is that the idea?  If so, there's no way for me as a non-maintainer to submit a PR that is based on 'rebase' through the GitHub PR system, right?

PR #1172, opened 2019-10-07
Indicate damage done by Ignite Poison

nlavsky:
This commit brings Ignite Poison in line with other damaging spells
that use exclamation marks to show the amount of damage done.

It is a followup for PR #952 and commits 4a3ee908 and 1f260a4b.


ebering:
Thanks for the patch!

PR #1176, opened 2019-10-10
Adjust OTR to be more turn length invariant.

ByrelMitchell:
Changed direct damage to use turn length in auts (instead of 10s of auts) for direct damage, and to scale odds of adding a poison stack linearly between 1 and 20 aut turn lengths (instead of 50% odds regardless of turn length.) This has the same results as previous calculation at 10 aut turns, but eliminates the incentive to 'quick-blade-swing' to add poison stacks rapidly.


crawl-ref/source/player-reacts.cc
```diff
@@ -799,9 +799,8 @@ static void _decrement_durations()
     }
 
     if (you.duration[DUR_TOXIC_RADIANCE])
-    {
         toxic_radiance_effect(&you, delay);
-    }
+
```
aidanholm: No need for the extra line here; one blank line between blocks is enough.

ByrelMitchell:
NP! Those should both be fixed now.

ByrelMitchell:
Changed it to handle really long turn lengths by potentially adding multiple stacks. See conversation with ebering in ##crawl-dev.

crawl-ref/source/spl-damage.cc
```diff
@@ -2699,7 +2699,7 @@ void toxic_radiance_effect(actor* agent, int mult, bool on_cast)
         if (agent->is_monster() && mons_aligned(agent, *ai))
             continue;
 
-        int dam = roll_dice(1, 1 + pow / 20) * mult;
+        int dam = roll_dice(1, 1 + pow / 20) * mult / 10;
```
ebering: This line isn't right, for delays less than 10 aut it will cause OTR to do zero damage. It should be

```suggestion
        int dam = roll_dice(1, 1 + pow / 20) * div_rand_round(mult, 10);
```

in order to match the old behavior in expected value

PR #1177, opened 2019-10-10
Fixed bug in OTR variance reduction commit. Now impossible to extend …

ByrelMitchell:
…duration by breadswinging.


ebering:
Could you trim the commit message title to 72 chars? 

And thanks for putting up with my backseat developing from mobile.

PR #1178, opened 2019-10-11
Details for newgame dolls

nlavsky:
This PR adds hair, claws, and horns to player dolls on the species selection menu.

The result will look like this: 
newgame_menu

Constructing all dolls twice - first time with `%texture none` and then with `%texture disabled-fg` - is a bit ugly. There is [`%repeat`](https://crawl.develz.org/wiki/doku.php?id=dcss:help:rltiles#repeat) command, but it doesn't seem to allow changing the texture. If there is a better way, please let me know.


aidanholm:
Thanks for the PR! I'm not sure how I missed all those little details before.. d'oh!

I couldn't see a better way to do that, so I added a feature where you can refer to already-defined enums. I've updated your commit to use that; manually merged at 653463b16e.

PR #1179, opened 2019-10-12
Fix manual typos

mwvdev:


ebering:
Could you make the commit message more specific?

`Fix manual typos` would suffice.

mwvdev:
@ebering Sure, I'll get right on that.

mwvdev:
@ebering Alright, should be changed now.

ebering:
Thanks for the patch!

mwvdev:
You're welcome!

PR #1185, opened 2019-10-18
Minor Text Fixes - Two Tickets

theJollySin:
This PR is actually two separate Mantis tickets:

https://crawl.develz.org/mantis/view.php?id=11907
https://crawl.develz.org/mantis/view.php?id=12093

They are both simple text fixes that I tried to solve as minimally as possible.

### DIFF
--- a/crawl-ref/source/dat/descript/unrand.txt
+++ b/crawl-ref/source/dat/descript/unrand.txt
@@ -66,7 +66,7 @@ enemies.
 %%%%
 sceptre of Torment
 
-An accursed golden sceptre, created by some unknown hellish power. When used in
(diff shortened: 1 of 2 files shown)


aidanholm:
Thanks again for all your work; these small fixes add up!

Minor nit: please use the imperative mood in commit messages, e.g:
- Fix Sceptre of Torment inconsistency (12093)
- Adjust barehanded attack prompt wording (11907)

theJollySin:
@aidanholm Will do in the future. I was just copying the text of the original Mantis ticket. But sure, that's easy.

aidanholm:
Thanks!

PR #1186, opened 2019-10-19
Draw shields on deep elf knights

kitchen-ace:
They used to have shields drawn on them in tiles; I don't know why this changed but I suspect it was a mistake. Here's how they looks after this patch:

Screenshot_2019-10-19_13-22-09-elfshield


aidanholm:
Thanks! looks much better now

PR #1187, opened 2019-10-19
Add art by Dynast as a splash screen

kitchen-ace:
This draconian was uploaded to mantis by Dynast for use as a splash
screen. With his approval I've prepared it into a commit.

Also change some tabs to spaces in the Makefile.


kitchen-ace:
BTW there are lots of tabs in the Makefile, but I only changed the ones relevant to the splash screen section.

kitchen-ace:
Updated the graphic with a more compressed version, using the same settings as in https://github.com/crawl/crawl/commit/9253d746ad11c3c17687860cefcd5c23c25c711d

aidanholm:
Thanks for the PR; while the artwork is fine, I'm not sure that this a good fit with the art style of the rest of the title images, or with the in-game tile art. For that reason, I'm ambivalent leaning "don't merge", but if another dev wants this in I won't complain.

alexjurkiewicz:
The other title images are pretty varied in their art style, compare
realism: https://github.com/crawl/crawl/blob/master/crawl-ref/source/dat/tiles/title_Cws_Minotauros.png
chibi https://github.com/crawl/crawl/blob/master/crawl-ref/source/dat/tiles/title_baconkid_gastronok.png
funny https://github.com/crawl/crawl/blob/master/crawl-ref/source/dat/tiles/title_ploomutoo_ijyb.png

rawlins:
> BTW there are lots of tabs in the Makefile, but I only changed the ones relevant to the splash screen section.

Art opinions aren't my thing (well, they can be, but no strong opinion here), but tabs should probably never be changed to spaces in a Makefile; the difference between them is syntactically important in the rule syntax (spaces and tabs are not equivalent in a Makefile) and most editors have things like expandtabs disabled because of this...

kitchen-ace:
>I'm not sure that this a good fit with the art style of the rest of the title images, or with the in-game tile art. 

I agree the style is fairly different than the hand-drawn/pixel art that is most of the art, but I think there's a lot of variation in the title screens especially, and that this style works well for a shiny, electrical black draconian.

>tabs should probably never be changed to spaces in a Makefile

It seemed to me that the tabs I changed were meant to be spaces. Maybe those are the only ones; I didn't realize until now the requirement for Makefiles to use tabs to distinguish between shell commands.

RojjaCebolla:
there's something alien/scifi/spooky about this draconian, not getting a strong "draconic" vibe here

ebering:
I'm gonna agree with the "too sci-fi aesthetic" take. Since no one else has merged this I'm going to close it on that basis.

PR #1188, opened 2019-10-20
Prevent torment from killing spriggan riders

kitchen-ace:
Torment shouldn't be able to kill spriggan riders or their mounts.


aidanholm:
Thanks!

PR #1189, opened 2019-10-21
Make fire spells better at burning trees

kitchen-ace:
Fireball could previously burn trees with all squares of its explosion,
not just the centre, but this was inadvertently changed in be81960.
This commit restores that behaviour.

It also allows the spells Fire Storm and Ignition to burn trees,
because it seemed strange to me that the most powerful fire spells
did nothing on trees. This allows a player to easily clear out large
amounts of trees at once, but by the time a player can cast these
spells, this is not usually all that relevant.

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -2490,7 +2490,10 @@ bool bolt::can_burn_trees() const
            || origin_spell == SPELL_BOLT_OF_FIRE
            || origin_spell == SPELL_BOLT_OF_MAGMA
            || origin_spell == SPELL_FIREBALL
-           || origin_spell == SPELL_INNER_FLAME;
+           || origin_spell == SPELL_FIRE_STORM
+           || origin_spell == SPELL_IGNITION
+           || origin_spell == SPELL_INNER_FLAME
+           || origin_spell == SPELL_STARBURST;
 }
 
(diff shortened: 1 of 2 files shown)


kitchen-ace:
PS Sorry for all the PRs right around the time you guys are preparing the new release/tournament.

kitchen-ace:
Going to fix this up in the next day or two. Do you want me to make Starburst burn trees as well? It doesn't currently.

ebering:
Yes, you should make starburst burn trees as well.

kitchen-ace:
I've force pushed a change that lets Starburst burn trees and prevents Ignition from burning trees under Fedhas.

Burning trees with Fire Storm was already docking you piety under Fedhas approrpiately. Fire Storm also wasn't burning trees made by Summon Forest.

For interaction between Ignition and Summon Forest, there are actually a number of cases where checking for temporary trees are not being done. I've been talking in ##crawl-dev about adding a new dungeon feature for temporary trees, so that the default interaction with normal trees doesn't apply; however, advil thinks it might be the wrong approach.

If desired, I can just add the check for temp trees now with Ignition and leave the other problems for a later commit.

kitchen-ace:
After talking with advil some more, I decided just to check for temporary trees when casting Ignition. The other changes can be made in a new PR.

PR #1191, opened 2019-10-22
Warn when exploding portal projectiles would hit you

kitchen-ace:
853fe2e fixed this for harming allies, but not yourself.

### DIFF
--- a/crawl-ref/source/throw.cc
+++ b/crawl-ref/source/throw.cc
@@ -854,6 +854,11 @@ bool throw_it(bolt &pbolt, int throw_2, dist *target)
                     monster *am = monster_at(*ai);
                     if (am)
                         cancelled = stop_attack_prompt(am, false, *ai);
+                    else if (*ai == you.pos())
+                    {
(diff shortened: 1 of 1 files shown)


kitchen-ace:
I'm not sure about the confirmation string. It works fine for this (produces "Damnation Bolt") but maybe a better comparison and/or projectile name should be used.

crawl-ref/source/throw.cc
```diff
@@ -854,6 +854,13 @@ bool throw_it(bolt &pbolt, int throw_2, dist *target)
                     monster *am = monster_at(*ai);
                     if (am)
                         cancelled = stop_attack_prompt(am, false, *ai);
+                    else if (*ai == you.pos())
+                    {
+                        string prompt = make_stringf("That %s is likely to hit you. Continue anyway?",
```
aidanholm: Maybe just "That is likely to hit you. Continue anyway?" is sufficient?

kitchen-ace: Sure, that works. Updated.

aidanholm:
Thanks!

PR #1194, opened 2019-11-03
Remove haste potion from FoWn start (12115)

theJollySin:
This PR relates to Mantis ticket 12115:

https://crawl.develz.org/mantis/view.php?id=12115

Essentially, Formicid Wanderers can start with haste potions, but that is a useless item for them.  We already have already have several exclusions so that various types of wanderers don't start with items that would be useless for them (`crawl-ref/source/ng-wanderer.cc`).

It appears this items was an oversight on the useless item exclusions.  Luckily, it was super easy to fix.


aidanholm:
Thanks for the fix!

PR #1195, opened 2019-11-04
Add some more chaos brand effects

alexjurkiewicz:
* Common effects added: might, agility
* Rare effects added: infestation, unravelling, chain of chaos
* Cloning can sometimes make neutral/friendly copies of monsters
* Level 0 miscasts removed, and miscast overall chance reduced

### DIFF
--- a/crawl-ref/source/attack.cc
+++ b/crawl-ref/source/attack.cc
@@ -610,8 +610,10 @@ static const vector<chaos_effect> chaos_effects = {
             if (!clone)
                 return false;
 
-            const bool obvious_effect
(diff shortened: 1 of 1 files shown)


crawl-ref/source/attack.cc
```diff
@@ -683,6 +664,8 @@ static const vector<chaos_effect> chaos_effects = {
         },
     },
     { "hasting", 10, _is_chaos_slowable, BEAM_HASTE },
+    { "mighting", 1000, nullptr, BEAM_MIGHT },
```
ebering: Looks like you have a left over test probability here

PR #1198, opened 2019-11-11
Prevent overtraining skill targets

nirrattner:
Currently, skill targets are not always respected when a single experience gain is large enough to push the the skill level to and beyond a skill target. This is fairly easy to reproduce when starting a new game and training only one skill with a skill target of 0.1 -- after the first experience gain, the skill will likely jump beyond 0.1 (perhaps 0.2, 0.3, etc. depending on the gain) before turning off.

This change will better ensure that skill targets are respected and not overtrained. It's a minor change with an arguably negligible impact on gameplay, but some players, like myself, prefer to train to "round" numbers ending in .0 or .5 and are put off when the skills are trained beyond the set targets.

Regarding the approach, I initially tried to take a more clever strategy of calculating the skill points required to reach the target, and using that as a ceiling for [`skill_inc`](https://github.com/nirrattner/crawl/blob/4073cf80fb89cbf380539eab6ea495c3f1859f43/crawl-ref/source/skills.cc#L1268), and that worked well except for incorporating the Ashenzari skill boost. So instead of putting together that more complex approach, I opted for the brute-force iteration of adding experience one-at-a-time while checking each time to see if the target were hit. I felt like this would be an easier way to handle the Ashenzari skill boost and that it might also be easier to maintain if any other edge cases were introduced to skill calculations. Of course the tradeoff here is some performance.

Still, I'm open to suggestions for other approaches if this one isn't desirable. Also looking for as much scrutiny as possible for the code in case there are other edge cases I didn't consider or if the code style could be improved.

Thank you for your time!

### DIFF
--- a/crawl-ref/source/evoke.cc
+++ b/crawl-ref/source/evoke.cc
@@ -529,6 +529,25 @@ int manual_slot_for_skill(skill_type skill)
     return slot;
 }
 
+int get_all_manual_charges_for_skill(skill_type skill)
+{
+    int charges = 0;
+
+    FixedVector<item_def,ENDOFPACK>::const_pointer iter = you.inv.begin();
+    for (;iter!=you.inv.end(); ++iter)
+    {
+        if (iter->base_type != OBJ_BOOKS || iter->sub_type != BOOK_MANUAL)
+            continue;
+
+        if (iter->skill != skill || iter->skill_points == 0)
+            continue;
+
+        charges += iter->skill_points;
+    }
+
+    return charges;
+}
+
 bool skill_has_manual(skill_type skill)
 {
     return manual_slot_for_skill(skill) != -1;
--- a/crawl-ref/source/evoke.h
+++ b/crawl-ref/source/evoke.h
@@ -6,6 +6,7 @@
 #pragma once
 
 int manual_slot_for_skill(skill_type skill);
+int get_all_manual_charges_for_skill(skill_type skill);
 bool skill_has_manual(skill_type skill);
 void finish_manual(int slot);
 void get_all_manual_charges(vector<int> &charges);
--- a/crawl-ref/source/skills.cc
+++ b/crawl-ref/source/skills.cc
@@ -1022,7 +1022,7 @@ static void _train_skills(int exp, const int cost, const bool simu)
         skill_type sk = SK_NONE;
         if (!skip_first_phase)
             sk = static_cast<skill_type>(random_choose_weighted(sk_exp));
-        if (is_invalid_skill(sk))
+        if (is_invalid_skill(sk) || !you.train[sk])
             sk = static_cast<skill_type>(random_choose_weighted(you.training));
         if (!is_invalid_skill(sk))
         {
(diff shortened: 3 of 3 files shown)


alexjurkiewicz:
I'm mildly worried about the effect on server CPU. Max XP from one monster is in the low tens of thousands.

nirrattner:
Seems like a fair concern, I'll try to put together the approach that uses some maths as opposed to guess-and-check 👍

rawlins:
Yes, I think CPU is potentially an issue; IIRC gnoll more or less uses this technique and was extremely slow compared to regular training, but it didn't matter too much outside of wizmode because skill gain is spread out evenly for gn. But we didn't reimplement everything using the gn code for exactly this reason.

nirrattner:
Because the Ashenzari skill boost is a function of the [skill level](https://github.com/crawl/crawl/blob/4073cf80fb89cbf380539eab6ea495c3f1859f43/crawl-ref/source/god-passive.cc#L1015), it's effectively a step-wise function of the corresponding skill points, so I couldn't find a good way to incorporate it comprehensively into the calculation without some guess-and-check iteration, so for now I'm leaving its effects it out of the calculations of the new experience currently being gained. Regardless, even with the omission, this change should greatly reduce the chance of overtraining for characters without the Ashenzari skill boost (barring cross-training effects) and still somewhat reduce the chance of over-training even for characters with the skill boost.

nirrattner:
Hey, sorry to be a pest, not sure what the normal timelines for a PR like this is, but are there any other comments or thoughts on the change? Thanks!

rawlins:
Sorry to take so long for someone to look at this, we're stretched relatively thin for PR review; I've had a look and in principle the approach of finding the diff relative to the target and subtracting it out seems like a pretty clean approach -- not too worried about cpu from anything I see in the code right now.

However, it doesn't work in all cases. It seems to work great if you have one skill target and just that skill enabled for training, but I'm guessing it's focused on that case? Here's a few things I tried that didn't work right:

* When I have two skills training, and a skill target set for one but not the other, and get a huge burst of xp, I get overtraining on the targeted one (same behavior as in current trunk), and the remainder in the second one. For example, on a new HOFi+rapier, switch to manual mode and enable both armour and shields (initial values 3.2 for each), and set a skill target for armour of 3.3, no skill target for shield. Then create and kill an orc knight (in case you didn't figure this out, wizmode `X` + `ctrl-k` makes the killing part easy for testing things like this). When I do this I end up with 6.3 skill in each, which matches current trunk.

* When I have two skills training, and skill targets on each, and get a huge burst of xp (more than is needed for both targets) I get some overtraining on the first one, and then it stops on the second one exactly at the target and makes me select a skill. (Where order is probably determined somewhat arbitrarily for each calculation.) With the same setup+steps as above except a 3.3 target for both skills, the skill menu pops up with 3.3 reached in armour, and 3.6 in shields. This one is actually different than trunk, where I end up with the skill menu popped up asking me to select a skill when there is 6.3 in each.

I haven't yet tried to dig into what's happening in this cases, but I would like to incorporate a version of this patch into the game. Do you know if you'll have time to work on this further?

nirrattner:
Thanks for looking into this! Good catch on those failed cases. It looks like they are fixed by adding a `reset_training` call if the training target is met so that the `you.training` values are updated accordingly for the [loop condition](https://github.com/crawl/crawl/blob/323f10f0e0ee2d8fe1d2e4aa7d66d940a62d45f1/crawl-ref/source/skills.cc#L995) and so generally the remaining experience is distributed correctly.

ebering:
Thanks for the patch! Sorry this sat in limbo for so long, we don't really have a formal handoff procedure for PRs if one of us doesn't have a time to look at things.

How would you like to appear in `CREDITS.txt`

ebering:
I've added you to the credits was `nirrattner` let me know if you would like to appear differently.

nirrattner:
That's great, thanks so much!

PR #1199, opened 2019-11-13
Replaced strange quotes with real ones

theJollySin:
The quotes used widely throughout the game frequently didn't have the actual quotes themselves: `"`.

Instead they used underscores as quotes: like `_this_` instead of `"this"`.

I feel like this was a decision made by a software engineer, because it just looks horrible from the perspective of actual English grammar. And it was the first thing my wife noticed about the game when I showed her webtiles.

So, I fixed it. And updated the grammar to match typical English.

I noticed something else odd in there: the quotes for Lindwurm and Manticore are not actually in English.  That's odd.

### DIFF
--- a/crawl-ref/source/dat/descript/quotes.txt
+++ b/crawl-ref/source/dat/descript/quotes.txt
@@ -11,7 +11,7 @@ his will: wounded, he reared up at his enemy and beat the hero down with his
 hooves. Peleus received the resounding blows on helmet and shield, and
 defending his upper arms, and controlling the weapon he held out, with one blow
 through the arm he pierced the bi-formed breast.”
-    -Ovid, _Metamorphoses_, XII, 330. 8 AD.
+    -Ovid, “Metamorphoses”, XII, 330. 8 AD.
 %%%%
 __d_suffix
 
@@ -30,7 +30,7 @@ __r_suffix
 
 HAMLET [Drawing his sword.]: How now! a rat? Dead, for a ducat, dead [Stabs
 through the arras.]
-    -William Shakespeare, _The Tragedy of Hamlet, Prince of Denmark_, III, 4.
+    -William Shakespeare, “The Tragedy of Hamlet, Prince of Denmark”, III, 4.
 1603.
 %%%%
 __cap-D_suffix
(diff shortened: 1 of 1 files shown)


rawlins:
>I feel like this was a decision made by a software engineer, because it just looks horrible from the perspective of actual English grammar.

This isn't some crazy software engineering thing, but rather [MLA](https://www.mla.org/) style guidelines for titles where italics aren't available for titles (not quotes): https://style.mla.org/underscore-instead-of-italics/

I'm afraid if you want to be prescriptive about what is or isn't good English grammar there are few better authorities than the MLA...

theJollySin:
@rawlins I mean, okay.  If this is on purpose following a style guide, I'll live with it.  I just think it's ugly.  And I'm not sure why you can't just use quotes.  BUT, your call.

alexjurkiewicz:
Quotes seem more common for this purpose than underscores amongst people
who don't speak markdown

rawlins:
The MLA style guideline precedes markdown by quite a long time, it originates from the typewriter era. But part of the issue that is confusing me is that none of these are quotes, they are titles -- that's why the MLA guideline says they should be italicized. I don't have strong feelings about keeping this style (I do actually kind of agree that it looks ugly), but I also do think that putting quotes around titles is not the way to go.

theJollySin:
@rawlins I feel like, in the modern computer era, a nice solution would be to use bold or italics.  I assumed that wasn't an option.  But MLA style guide or no, I'm a native English speaker and have never seen that before.  If you don't like the quotes, I suppose I would suggest bold.  Because, functionally, there is no difference between those underscores and quotes, except one is more familiar to a wider audience.

Your call!

aidanholm:
We would certainly use italics if we could. Unfortunately the current set of text rendering primitives are very primitive indeed. Italics or bold are not an option at present. One of my pipe dreams is to use pango for text rendering, but that will be a lot of work.

rawlins:
It occurs to me that in webtiles, italics for _text_ might be achievable in javascript.

theJollySin:
This PR was tacitly rejected by the dev who responded (@rawlins ).  So I will just close it.  At this point, the result seems clear.

No hard feelings, I just don't like loose ends.

PR #1201, opened 2019-11-13
Make the Warlock's Mirror's reflection chance scale as intended

AdamPG:
Commentary about the omnireflection property granted by the Warlock's Mirror (see line 3099 of beam.cc) indicates that it is intended to reflect 50% of ench-type effects at 20 displayed SH; however, the the actual chance is 33% due to a mathematical error. This commit changes the chance to the intended 50% at 20 SH.


ebering:
Thanks for the patch; how would you like to appear in CREDITS.txt?

AdamPG:
Thanks! AdamPG is fine.

PR #1207, opened 2019-11-19
Use yaml.safe_load instead of load

imarko:
PyYAML's yaml.load is insecure and it is deprecated in 5.1. Some
distributions even disable it, making it difficult to build DCSS on
them. Use the recommended safe_load instead.


ebering:

imarko:
> Thanks for the patch! How would you like to appear in CREDITS.txt?

"István Markó" or "Istvan Marko" if we are keeping it ASCII. Much appreciated :)

PR #1208, opened 2019-11-19
Adjust Trog weapon gifting (ebering)

alexjurkiewicz:
Only gift melee weapons with a restricted set of brands (plain, vorpal,
flaming, antimagic).

This also slightly modifies the minimum enchantment of gifts from Trog
and Okawaru. Previously the minimum was 0 or 1 (random per-item), now it
is always 0.

### DIFF
--- a/crawl-ref/source/acquire.cc
+++ b/crawl-ref/source/acquire.cc
@@ -1125,21 +1125,34 @@ static bool _brand_already_seen(const item_def &item)
  * Take a newly-generated acquirement item, and adjust its brand if we don't
  * like it.
  *
- * Specifically, if we think the brand is too weak (for non-divine gifts), or
- * sometimes if we've seen the brand before.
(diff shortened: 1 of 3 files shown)


crawl-ref/source/acquire.cc
```diff
@@ -1181,7 +1204,7 @@ static string _why_reject(const item_def &item, int agent)
     // (which work together with Necromantic magic).
     // nor fancy magic staffs (wucad mu, majin-bo)
     if (agent == GOD_TROG
-        && (get_weapon_brand(item) == SPWPN_PAIN
+        && (is_range_weapon(item)
```
alexjurkiewicz: is it worth adding a sanity check for the brand type here as well?

alexjurkiewicz:
Issues: allows randarts with disallowed brands, and still gifts ammunition

PR #1209, opened 2019-11-22
Add tiles and plumbing to display monster poison level as a debuff icon.

ByrelMitchell:
Currently has 3 tiles for 1, 2 or 3, 4 stacks of poison.


crawl-ref/source/tile-flags.h
```diff
@@ -48,6 +48,8 @@ enum tile_flags ENUM_INT64
     TILE_FLAG_PINNED       = 0x200000000000000ULL,
     TILE_FLAG_VILE_CLUTCH  = 0x400000000000000ULL,
     TILE_FLAG_POSSESSABLE  = 0x800000000000000ULL,
+    TILE_FLAG_MORE_POISON  = 0x1000000000000000ULL,
+    TILE_FLAG_MAX_POISON   = 0x2000000000000000ULL,
```
aidanholm: You only need to add a single flag, since there are four possible states (no poison, and poison 1-3), and four states can be represented with two bits. You'd need to change the rest of the code that sets and checks these flags.

An alternative that I'd accept is just to leave a comment above these saying that if we run out of space, these two can be combined.

crawl-ref/source/mon-info.cc
```diff
@@ -1399,6 +1405,11 @@ vector<string> monster_info::attributes() const
 
     if (is(MB_POISONED))
         v.emplace_back("poisoned");
+    else if (is(MB_MORE_POISONED))
+        v.emplace_back("very poisoned");
+    else if (is(MB_MAX_POISONED))
+        v.emplace_back("extremely poisoned");
```
aidanholm: Grammar is a little odd here, but I can't think of a good alternative :/

crawl-ref/source/tile-flags.h
ByrelMitchell: I did the alternative you suggested for now.

crawl-ref/source/mon-info.cc
ByrelMitchell: I'm not sure; if we expand this to show all 4 levels, perhaps it should be poisoned, very poisoned, extremely poisoned, completely poisoned? 

'completely poisoned' sounds weird to my ears though.

aidanholm:
Merged manually at e5692877c645 after fixing braces and optimizing images.

PR #1211, opened 2019-11-24
The Elemental Staff give spell power boost on wearer

sdynet:
It has improved its performance to become a more useful item for users of elemental schools.

I thought about adding a function to reduce costs by one, limited to elemental schools. But I didn't implement this idea because I thought it wouldn't be acceptable to the dev-team.


ebering:
Thank you for the patch; I've been focusing my energy on positional magic, but now that's merged so I'll take a look at this soon.

PR #1214, opened 2019-11-25
Add threat level indicators for monsters

ByrelMitchell:
Webtiles and console both have a monster list at the bottom-right of
the screen which lists all monsters in view color-coded by rough
threat level. Many players have expressed a desire for something
similar in offline tiles, to provide always-visible indicators of the
threat levels of all enemies in view. This commit adds translucent
colored backgrounds -- dark grey, light grey, yellow, and red --
to indicate trivial, easy, tough, and nasty monsters respectively.

Additionally, these indicators are configurable with a new option in
the options file, "tile_show_threat_levels" with valid values being any set of 'trivial', 'easy', 'tough' and 'nasty'. Each type of indicator
can be independently enabled or disabled by including or omitting it
in the option file line. For example, the following...

"tile_show_threat_levels = tough nasty"

...will show only yellow (tough) and red (nasty) threat indicators.

This patch was mostly written by GnarledBeach in PR #1120. I picked it up and added webtiles support, plus squashed some trivial commits.


aidanholm:
Thanks for the PR; I've moved work on this to the `threat-levels` branch.

PR #1216, opened 2019-11-28
Remove an inventory hotkey for corpses and skeletons

nlavsky:
Although the player cannot pickup carrion anymore, the pickup menu
still displays a hotkey for it. The hotkey already does nothing
because corpses and skeletons are stationary objects (62bc8482).


aidanholm:
Thanks for the fix!

PR #1217, opened 2019-12-01
Add missing SUBST to gammafunk_ghost_icy

jbuck3:
Fixes an ugly green tile when placed in Lair:
ghost_vault_bad_tile_cropped


gammafunk:
Thanks for the fix!

PR #1219, opened 2019-12-02
Fix initial god ability slot assignment in local tiles

jbuck3:
Ability slots need to be claimed before setting initial piety. This is because if a player's starting piety is at least 1*, the local tiles version will trigger a redraw of the ability tab, a side effect of which is to fix any available but unassigned abilities using the default logic (starting at f) rather than the god ability rules (a-e + X).


ebering:
Thanks for the patch!

PR #1220, opened 2019-12-02
Fix edge cases for destroying Summon Forest's trees

kitchen-ace:
Tornado and forest fires were not checking whether a tree was created
by Summon Forest.


ebering:
Merged via 105ea2

PR #1221, opened 2019-12-03
Make Death Cobs and Deathcaps friendly under Fedhas

kitchen-ace:
This was changed in 215b21a but the problems with multi-holiness
monsters don't seem to be an issue with these two. And now that Fedhas
no longer dislikes necromancy, it seems an appropriate time to change
it back.


kitchen-ace:
I originally thought this was part of the Fedhas revision, which shows you how often I played with Fedhas before they were revised.

gammafunk:
This isn't a good change, and the issues kate mentioned in 215b21a do apply in the general sense; having multiple-holiness monsters introduces complications we'd rather not have. 
A better direction would be to remove the multiple-holiness aspect from monsters entirely.

Even the game-play of this change is worse, since specifically makes one spell of a non-friendly monster (curse toes) behave very oddly without adding a special case. It's perfectly fine if fedhas regards plant creatures that have been made undead through the powers of necromancy to be beyond the god's control, even if the god doesn't hate necromancy specifically. The original holiness simplifications in that commit certainly had nothing to do with whether gods hated necromancy.

gammafunk:
advil pointed out to me that currently toes do summon a mix of friendly and non-friendly monsters, which I had forgotten. So a special case for fedhas will be necessary regardless so that the curse toe spell doesn't summon player-friendly summons at all.

kitchen-ace:
Fair enough, thanks for the feedback.

PR #1223, opened 2019-12-05
Let the player renounce religion while silenced

alexjurkiewicz:
Divine abilities are blocked by silence, but renouncing your religion
isn't really a divine ability.


ebering:
I can't imagine *when* a player would want to do this, but I agree that we should let them.

PR #1230, opened 2019-12-14
Increase Quicksilver Dragon Armour ER to 8 (was 7)

alexjurkiewicz:
QDA is extremely rare and can't be enchanted, so it can afford to be
quite powerful. But its stats are extremely attractive when comparing
against similar AC (IDA: 11 vs 7 ER) or similar ER (Swmp DA: 7 AC vs 9).

Increase its ER by one. Another option could be to tweak its AC, but
no other armour has 8 ER so this is probably a more interesting choice.

### DIFF
--- a/crawl-ref/source/item-prop.cc
+++ b/crawl-ref/source/item-prop.cc
@@ -157,7 +157,7 @@ static const armour_def Armour_prop[] =
         ARMF_RES_STEAM),
     DRAGON_ARMOUR(ACID,        "acid",                    6,  -50,  400,
         ARMF_RES_CORR),
-    DRAGON_ARMOUR(QUICKSILVER, "quicksilver",             9,  -70,  600,
+    DRAGON_ARMOUR(QUICKSILVER, "quicksilver",             9,  -80,  600,
         ARMF_RES_MAGIC),
     DRAGON_ARMOUR(SWAMP,       "swamp",                   7,  -70,  500,
         ARMF_RES_POISON),


marksg07:
I think your comparison to SDA is really a negative point for your argument. SDA is much more common than QDA and can be enchanted to +7 (?). This makes QDA a pretty bad lategame armor in comparison unless you really need the MR. Therefore, I think this nerf is unwarranted.

ebering:
For context: ER is used quadratically in armour penalty formulas in crawl, so this change would make QDA 30% more difficult to use (skill wise) than WDA. Since skill cost is also quadratic, the ultimate increase in strategic cost of this one point of ER becomes about 70%.

The decision of WDA vs QDA vs IDA depends on the strategic state of the game: what enchants are available and what is available to be enchanted. This change would make QDA less of a no brainer in the limited situation of (+0 WDA vs QDA vs +0 IDA and no enchants available yet). As enchants become available the choice becomes between a higher plus WDA or IDA and QDA. In this later situation, the non-enchantability of QDA becomes a much greater liability in comparison to a WDA: at 20 armour skill a +2 WDA gives the same AC contribution as QDA. So for a character looking at armours with an ER of 7 or less, QDA is an intermediate choice among several, and there are strategic factors that lead to good decisions.

For QDA vs IDA the calculus goes the other way, with QDA being 6x cheaper than IDA for "equivalent" ER penalties. A maximally enchanted IDA at high armour skill gives only 50% more AC (26 vs 17) than the QDA. Here the decision depends on the character and skilling: is it dependent on magic, does it have other defenses that allow it to take the AC hit and use spells, or would it rather ignore higher level spells in favor of more AC. Percentages are deceptive here: there are few sources of AC other than items, skill training gives only 1/22 base ac. So a player willing to use heavier armour is faced with a choice when presented a QDA: give up AC from enchant and have an easier time of spells and dodging, or keep the AC for the heavier weight. The higher the ER of QDA, the less dramatic this choice is.

So, I think QDA is in a good place ER wise. Going to close; if you strongly disagree we can chat it out in ##crawl-dev, maybe another team member will take up the torch.

PR #1232, opened 2019-12-17
Rework the Pain card's level 2

sdynet:
Level 2 pain card's old effect (summon flayed ghost) was an unsuitable effect on the deck of destruction. Also, flay effect is just another torment.
New effects: cast torment. Torment does not damage the caster.

### DIFF
--- a/crawl-ref/source/decks.cc
+++ b/crawl-ref/source/decks.cc
@@ -1143,27 +1143,7 @@ static void _damaging_card(card_type card, int power,
     case CARD_PAIN:
         if (power_level == 2)
         {
-            mpr(prompt);
-
-            if (monster *ghost = _friendly(MONS_FLAYED_GHOST, 3))
-            {
-                apply_visible_monsters([&, ghost](monster& mons)
-                {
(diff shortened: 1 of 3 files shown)


sdynet:
I tried to change the Level 2 effect to only cast torment, but it couldn't. I don't know the cause....

alexjurkiewicz:
You need to change the switch statement in `card_effect` to call a special function for the pain card. Look at the code for `CARD_TORMENT` as a starting point.

sdynet:
void card_effect(card_type which_card,
                 bool dealt,
                 bool punishment, bool tell_card)
{
    const char *participle = dealt ? "dealt" : "drawn";
    const int power = _card_power(punishment);

    dprf("Card power: %d", power);

    if (tell_card)
    {
        // These card types will usually give this message in the targeting
        // prompt, and the cases where they don't are handled specially.
        if (which_card != CARD_VITRIOL
            && which_card != CARD_ORB)

I deleted 'which_card != CARD_PAIN'. But nothing has changed. Isn't it here?

PR #1233, opened 2019-12-17
Improve the Degeneration card

sdynet:
This card has an interesting and tactical effect. Although powerful, it is the least predictable of all the cards with Wild Magic card. It also has the potential to pose the most lethal danger from any wrong result of the cards. It is also dangerous for HD to change into a lower but more dangerous creature, but what is more deadly is that the enemy recovers all their HP with the change.

By adding new features, I want to give players the option to try a challenge rather than run away from more dangerous change enemies.  Also, This feature destroys the enemy directly, so it would be more appropriate for the name of the deck of destruction.

New feature: reduce an enemy's HP by (card's power level / 5 * 100)%. 20% for level 1, 40% for level 2.

original: (https://github.com/b-crawl/bcrawl/commit/e3360951ca5c6c7e31ed38316197fb9f4d8ac4c2)

### DIFF
--- a/crawl-ref/source/decks.cc
+++ b/crawl-ref/source/decks.cc
@@ -1630,6 +1630,7 @@ static void _degeneration_card(int power)
                if (mons.can_polymorph())
                {
                    mons.polymorph(PPT_LESS);
+                   mons.hit_points -= mons.hit_points * power_level / 5;
                    mons.malmutate("");
                }
                else


ebering:
The potential to pose lethal danger from a bad draw of the cards is an important part of Nemelex's balance. The card already uses the `PPT_LESS` polymorph type, which attempts to polymorph the monster into one with a lower HD, and malmutates them, which applies a flat debuff to the monster ( -8 AC and -25% effective HD for accuracy, spellpower, and any HD scaling effect). The second, malmutation, already does give the player some chance at trying to fight a challenge roll.

There is in both the plutonium sword and this PR an issue, but the issue isn't with these items/effects. The issue is with polymorph using HD and monsters at a given HD not being particularly uniform in threat level. As a result, certain monsters are a particularly bad idea to polymorph, but this information isn't revealed to the player. If you want to address polymorph issues, I suggest thinking of a way to reveal some information about the danger level of polymorphing a given monster.

PR #1235, opened 2019-12-19
Remove the Shaft card and Rework the Exile card

sdynet:
Shaft: This card has a similar effect as an exile card. They both send their enemies to other places. I don't think it's necessary to have two cards with roles like similar.
Also, In the last floor(and Tomb with Ziggurat), which is the most dangerous section for the player, this card has no effect and instead becomes a deadly trap card that wastes one turn. Of course, you can't use exile cards in Abyss, but there are far fewer places that are ineffective and there is a very low probability that you can get this card.

Exile: Remove the function of self banishing. This effect is too powerful and dangerous. The risk of self banishing is gone, so the chance of this card appearing will be a little higher.

### DIFF
--- a/crawl-ref/source/decks.cc
+++ b/crawl-ref/source/decks.cc
@@ -87,7 +87,7 @@ typedef map<card_type, int> deck_archetype;
 deck_archetype deck_of_escape =
 {
     { CARD_TOMB,       5 },
-    { CARD_EXILE,      2 },
+    { CARD_EXILE,      3 },
     { CARD_ELIXIR,     5 },
     { CARD_CLOUD,      5 },
     { CARD_VELOCITY,   5 },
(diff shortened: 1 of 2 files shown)


marksg07:
Suggestion: make the weight of the exile card the same as the other cards rather than just slightly higher; its effect is now near the other cards in power rather than being radically worse or better depending on the situation.

sdynet:
However, this card ignores magic resistance. So it is too powerful to give this card the same chance as any other card. It seems not bad to increase to 3 instead of 5.

marksg07:
Taking a look at the other escape cards, you're right -- they are all at best situational besides elixir. I do agree with increasing to 3 due to the fact that there's a chance exile can do literally nothing even with enemies on screen (there is no minimum on the number of exiled).

ebering:
This looks good, but it needs save compatibility code to remove Shaft cards from players stacked decks (or replace them with Exile cards).

ebering:
This is a small enough tweak that if you have time to look at the save compatibility we can merge it for 0.25; however I will be working on tournament preparations so won't have time to implement it myself most likely.

PR #1239, opened 2019-12-23
Fix macOS OS detection with catch2

alexjurkiewicz:
Catch2 erroneously defines TARGET_CPU_ARM on macOS, which causes
platform.h to mis-detect the target OS as Nintendo DS.

Fix this by changing the order of OS detection.

There are a couple of alternate approaches for this fix:
* Fix Catch2 so that TARGET_CPU_ARM (and possibly other TARGET_CPU_*
  defines) is not erroneously defined on macOS
* Remove Nintendo DS support from the project

### DIFF
--- a/crawl-ref/source/platform.h
+++ b/crawl-ref/source/platform.h
@@ -212,13 +212,6 @@
 #endif
 
 #if !defined (OS_DETECTED)
-#if defined (TARGET_CPU_ARM)
-#define OS_DETECTED
-#define TARGET_OS_NDSFIRMWARE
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
For the record, the files `catch2-tests/*.cc` seem to have all TARGET_CPU_ variables defined somehow. Not just TARGET_CPU_ARM.

reaverb-Crawl:
This commit was included in the catch 2 coverage pull request, and is already in that branch.

PR #1241, opened 2019-12-24
Change USK Line Pass's cost

sdynet:
Uskayaw's core is to gather piety and use the Grand Finale, Solo Times, and Pain Bonds. Players must suppress the use of piety to use Solo Time and Pain Bond frequently. With such a structure, Stomp and Line Pass meet the requirements despite their low star power and are difficult to use immediately. Due to these difficulties, line pass are only used in very limited situations. Therefore I hope that by altering the piety cost of the line pass, this power can be a tactical option on par with other powers.

Now the Line Pass is zero the piety cost and gets exhaustion time.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -616,7 +616,7 @@ static const ability_def Ability_List[] =
     { ABIL_USKAYAW_STOMP, "Stomp",
         3, 0, 100, generic_cost::fixed(20), {fail_basis::invo}, abflag::none },
     { ABIL_USKAYAW_LINE_PASS, "Line Pass",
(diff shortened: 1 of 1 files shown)


ebering:
Unfortunately, Exh as a cost is more limiting; it's also a generic cost that we are trying to use less in crawl. Usk's non-traditional piety means we really want to cost Usk's abilities on piety; multiple line-passes is a tactical option that should be available to an Usk player.

sdynet:
Well, I see. But I still think solo time is overwhelmingly the right choice then line pass, so I think line pass should be more beneficial. Maybe In a different way than this proposal. :)

PR #1242, opened 2019-12-27
Rework the Plutonium Sword

sdynet:
The effect of this weapon sometimes changed the enemy into a much stronger one, which agony the user. The effect is fun, but not particularly useful. The new effect causes a radiation explosion that damages enemies. This effect is powerful but generates a lot of noise, so the benefits and penalties seem reasonable.

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -611,13 +611,14 @@ EV:       4
 STEALTH:  1
 
 NAME:    plutonium sword
+INSCRIP: irradiate,
 OBJ:     OBJ_WEAPONS/WPN_TRIPLE_SWORD
 FB_BRAND: SPWPN_CHAOS
 PLUS:    +11
 COLOUR:  ETC_MUTAGENIC
 TILE:    urand_plutonium
 TILE_EQ: plutonium_sword
-BOOL:    mutate, chaotic
+BOOL:    chaotic, skip_ego
 STEALTH: -1
 VALUE:   1000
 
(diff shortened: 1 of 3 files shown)


semitonal:
I think the plutonium sword is a strong enough weapon that having to deal with an occasional polymorph is fine as a rare downside - noisy explosions don't seem quite as interesting to me.

sdynet:
Polymorph is not a fine downside but a fatal one. The effect of change an enemy into a random target in a roguelike game is a very threatening effect. This can only be useful in a few exceptional circumstances. In the case of Wand, it can be used from afar, and the player can decide what to attack and when to attack. For cards, as its power increases, it can greatly weaken the enemy. But this weapon cannot determine what to attack and when to want to attack, nor can it be used far away, nor can it guarantee to clearly weaken the enemy. Sometimes that gives players a very terrible gift like this screenshot[https://gall.dcinside.com/rlike/279977]. This change also completely recovers the enemy's HP by nullifying the damage it has inflicted on the enemy. In the CWZ community, the weapon treats the player as a trap weapon that tempts the player with high enchants.

ebering:
The plutonium sword inflicts transmutations miscasts, which have a difficult to reason about chance of polymorph. However, the resulting polymorph effect is either PPT_SAME or PPT_LESS; within fuzzing it tries to get a monster with lower HD. The other tmut miscasts are straight debuffs on a monster. I think that if a rework is to happen, it should keep the transmutation flavor but decouple from the miscast system; possibly making the poly-effect PPT_LESS always.

It's not clear to me a change is needed, in the [unrand review](https://github.com/crawl/crawl/wiki/Unrand-Review) rubric we ask that unrands be either good or interesting; the definition of interesting is "An interesting unrand is one that has drawbacks but in such a way that its use requires a change in play style to fit the item. This change might or might not be desired by a given player, but at least some play styles should be interested or tempted by the unrand.". In my opinion the plutonium sword is squarely in the interesting category. If you'd like to revise it, the best way forward would be decoupling it from the miscast system, but keeping the malmutation *and* polymorph flavor.

PR #1245, opened 2019-12-30
Add code coverage tracking

alexjurkiewicz:
Add code coverage tracking

Supports both GCC and LLVM toolchains.

Bluffer's guide:
1. make catch2-tests
2. util/coverage

More explanation has been added to testing.md (which reworks
testing.txt) and `util/coverage --help`.

### DIFF
--- a/crawl-ref/docs/develop/arena.txt
+++ b/crawl-ref/docs/develop/arena.txt
@@ -1,30 +1,4 @@
-A.   Unit tests
-===============
-
-Crawl has a set of unit tests in the source/test/ directory, with each lua
-file being a separate unit test. To use unit tests, you must compile Crawl
-either with the macro DEBUG_DIAGNOSTICS defined (the debug make target) or
-with both the macros DEBUG and DEBUG_TESTS defined (the wizard make target,
-plus -DDEBUG_TESTS). You can then do
-
-    crawl -test
-
-to run the unit tests. To run paricular unit tests, you can do
-
-    crawl -test foo
-
-and that will run all unit tests which have "foo" in their file name.
-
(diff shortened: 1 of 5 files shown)


alexjurkiewicz:
llvm-cov:
Screen Shot 2019-12-30 at 4 15 52 pm
Screen Shot 2019-12-30 at 4 16 02 pm
Screen Shot 2019-12-30 at 4 16 08 pm

lcov:
Screen Shot 2019-12-30 at 4 19 43 pm
Screen Shot 2019-12-30 at 4 19 51 pm
Screen Shot 2019-12-30 at 4 19 58 pm


gcovr:
Screen Shot 2019-12-30 at 4 20 56 pm
Screen Shot 2019-12-30 at 4 21 02 pm
Screen Shot 2019-12-30 at 4 21 05 pm

aidanholm:
Awesome! :sunglasses: 

Minor nit: there's some extra whitespace changes to the Makefile; i'd prefer they were made as a separate commit; it makes reviewing and rebasing easier.

Separately, I'd prefer that COVERAGE be enabled by default on all debug builds, and would even prefer not exposing a COVERAGE option at all unless there's a good reason. In general I think keeping the number of build options to a minimum is best.

alexjurkiewicz:
> Separately, I'd prefer that COVERAGE be enabled by default on all debug builds, and would even prefer not exposing a COVERAGE option at all unless there's a good reason. In general I think keeping the number of build options to a minimum is best.

makes sense. Compilation is only ~10% slower

reaverb-Crawl:
Merging into master after testing locally. Thanks for the code!

PR #1247, opened 2020-01-01
Improve macOS Makefile support

alexjurkiewicz:
macOS provides BSD sed which lacks -r. Allow the user to override the
sed binary name so they can specify eg make sed=gsed.

macOS also provides BSD find which lacks -o as an alias for -or. Use -or
instead as it's also supported by GNU find.


alexjurkiewicz:
It would be good to have this upstream to help build windows binaries on my Mac.

PR #1248, opened 2020-01-01
Fix an information leak about Mara and rakshasa clones

nlavsky:
This PR fixes an information leak about Mara/rakshasa clones by making all illusions inner-flameable.


ebering:
Thanks for the patch!

PR #1249, opened 2020-01-02
Encode nested colours in formatted_string

### LINKED ISSUE #1224: Strange '</lightgreen>' message output in tiles mode
webtiles

**DCSS VERSION:** 
0.25-a0-258-g6d7e32a812

**DESCRIPTION:**
Some colored text in webtiles has strange closing html tags after it.
e.g.: Found a glowing spear.&lt;/lightgreen&gt;

**EXTRA FILES:**
*Please include any morgue files, crashlogs, save files here, or extra
screenshots here. If this is a crash bug, the crashlog and save file will be
extremely helpful; the location of these is sytem-specific but if you ask
we can help you find them.*

*To upload a save file directly to github, compress it to a .zip archive, and
then github will accept the upload. (It will not take a .cs file directly.)*

*If this is an online game, see [these instructions](https://crawl.develz.org/wiki/doku.php?id=mantis:savebackup)
for more information on how to make a save backup; you can provide the
resulting link here.*

aidanholm:
This attempts to make formatted_string a closer fit to the textual
colour string format; nested colours are now encoded, and parsing a
colour string and then converting to a string again should in many cases
now be a no-op.

parse_string_to_multiple() is also changed so that the colour nesting is
preserved across newlines, i.e. the colour stack is unwound and then
rewound at each newline.

Not sure if anyone else has a preferred unit test style, e.g. BDD? Catch2 does support that, but I stuck with the existing style from `test_english.cc`.

Fixes https://crawl.develz.org/mantis/view.php?id=12143
Fixes #1224

### DIFF
--- a/crawl-ref/source/Makefile.obj
+++ b/crawl-ref/source/Makefile.obj
@@ -310,6 +310,7 @@ catch2-tests/test_coordit.o \
 catch2-tests/test_describe.o \
 catch2-tests/test_english.o \
 catch2-tests/test_files.o \
+catch2-tests/test_format.o \
 catch2-tests/test_items.o \
 catch2-tests/test_mon-util.o \
 catch2-tests/test_ng-init-branches.o \
--- a/crawl-ref/source/catch2-tests/test_format.cc
+++ b/crawl-ref/source/catch2-tests/test_format.cc
@@ -61,3 +61,23 @@ TEST_CASE( "formatted_string::parse_string_to_multiple() preserves nesting over 
         REQUIRE( out[1].to_colour_string() == "<red>1</red>" );
     }
 }
+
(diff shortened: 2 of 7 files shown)


aidanholm:
There's currently a bug in this on console / webtiles builds where colours get very deeply nested, resulting in ~2-5 seconds of lag before the messages window will open. This doesn't affect local tiles for some reason.

rawlins:
This PR is a couple years stale, and I know I've refactored the code it would need to merge on top of a couple times in the interim (though still not to do what this PR does); I also fixed both of the mentioned bugs in simpler ways. I'm going to close it as stale given that aidanh isn't around these days. If he were around of course I have no objection to him picking this back up / reopening.

PR #1251, opened 2020-01-03
Re-add Regeneration spell with tweaks

alexjurkiewicz:
The regeneration spell returns with changes:
* Doubled regen bonus
* HP rot on cast
    * This additionally prevents use by Gargoyles, statue form, and
      demonspawn with rot immunity.
* No special interaction with Trog's Hand

Because this variant of the spell is far more situational the book
changes were partially kept -- Ozocubu's Armour stays in the book of
battle, but Regeneration remains in the Book of Necromancy.

This partially reverts 7e26dd5d75866b5f609e078cc4b528d09f51ae44.


sdynet:
What do you think about get degeneration instead of rot? Maybe it's similar to Borgnjor's Revivification.

alexjurkiewicz:
```
02:37:42     +gammafunk | I think one issue with it is that you want to
                        | cast this spell more and more towards the end of
                        | the game
02:37:57     +gammafunk | so it's a low-level spell you'd be using a lot
                        | more in endgame
02:39:22     +gammafunk | not just due to stockpiling !curing (you can tell
                        | alex doesn't play console :), where you could be
                        | using more of that throughout the earlier, more
                        | dangerous parts of game and might actually end up
                        | with not that much due to this spell
02:39:47     +gammafunk | but also because the rot cost is proportionally
                        | less and less of your mhp
02:40:51     +gammafunk | it's not a terrible proposal, but it sort of
                        | feels like it's probably wanting to be a higher
                        | level spell to some degree
02:41:18     +gammafunk | I'm not particularly attached to regen and would
                        | love to see a more decisive charms reform, which
                        | re-adding this spell might get in the way of to
                        | some degree
02:41:27     +gammafunk | we might be able to just find more interesting
                        | effects+tradeoffs
02:41:37     +gammafunk | but that said, we don't have a definitive charms
                        | reform proposal
```

sdynet:
I see. Let's find a better offer :)

sdynet:
How about this?
Opinion1: very fast regen. Instead, they get more damage(+20%) in a short duration when the duration is over.
Opinion2: lose half of HP. It then regenerates at an extremely fast rate up to twice the HP it lost.

PR #1252, opened 2020-01-04
Combine three types of bullets into one

sdynet:
Integrate sling bullet, arrow, and bolt into bullet. Each of these bullets has a different appearance, but the given role is the same, and there is no tactical difference between the three. The difference between them is not in themselves, but in weapons that use them.


sdynet:
I added a new tile called bullet. I want to pick this up and make the arrow tile fire if it pulls the bow. What is the cause of this bug? I don't know where the problem is.
[](https://drive.google.com/open?id=1-EK74lTNi7JGkfq8X16iR_GWaDrBzbX6)

ebering:
This (and your sling bullets) PR are steps in the right direction but don't seem like the best way to change ranged. I think if you are serious about improving ranged combat in crawl you should go further:

> Remove ammunition altogether. 

This is a buff to ranged, especially in the early game. I think the best solution is to make the range of ranged weapons scale with skill (with a similar scaling to weapon delay), starting with range 2 and scaling up to LOS range. I've been meaning to flesh these ideas out more, but at present this is the extent of my sketch.

Other team members might disagree with me, it would be best to check in in ##crawl-dev before doing a lot of work on the topic, but an ammoless change would definitely be worth setting up an experimental branch for.

sdynet:
I'm completely in the same direction as you are.

PR #1253, opened 2020-01-05
Webtiles improvements

alexjurkiewicz:
Re-implement features from webtiles-changes that CPO makes use of. Ultimate goal is to move CPO off webtiles-changes.

Summary of changes:

* Fix autologin cookie handling
* New features for game definitions:
    * Environment variables
    * `-dir` support
    * Milestone path (previously there was a separate global list, now you can specify per-game)
* Dynamic game definitions
    * Load defs from `games.d/*.yml`
    * Dynamic reload (while server is running)
* Move repo-provided templates to eg `banner.html.default` and allow server-specific `banner.html` override
* Allow editing RC files when they don't already exist (previously servers had to create blank rc files for every game for every user)
* Allow disabling HTTP access logs
* Update README


alexjurkiewicz:
22:42:00     +gammafunk | if updatig the trunk binary is an aim, surely you just do that via a symlink
                        | update?
22:42:07     +gammafunk | *updating
22:42:19     +gammafunk | I guess I'm not clear what you mean by updating the trunk binary
22:43:39     +gammafunk | you seem to have on specific kind of update in mind, but I was pointing out that
                        | if certain paths have changed in an existing game's config, any running games
                        | with that config will have out of date and potentially broken config info
22:44:41     +gammafunk | obviously the admin can just use this feature carefully and not try to update
                        | paths using this feature (instead using a full webtiles restart), however it's
                        | probably good to document this aspect and possibly warn the admin
22:44:50     +gammafunk | maybe advil or |amethyst can comment more on this though
22:46:00     +gammafunk | but I don't quite understand your comment about updating trunk binary paths with
                        | this feature, since if that's all you're changing and you want to let existing
                        | processes run with the old config params etc, you don't need a feature like
                        | this, you just use symlinks
22:46:19     +gammafunk | obv that functionality is useful for adding new games dynamically
22:47:21     +gammafunk | and yeah, regarding webtiles-changes, I authored much of the dynamic config
                        | reload stuff, and that was very much untested; I think my work was mainly
                        | reloading the main server config data, not the game data
22:47:30     +gammafunk | I seem to recall that edlothiol implemented that stuff

alexjurkiewicz:
I'm running this on CPO without issue, and I've tested live updates to the games config 👍

crawl-ref/source/webserver/ws_handler.py
```diff
@@ -616,8 +643,12 @@ def go_lobby(self):
...
     def get_rc(self, game_id):
         if game_id not in config.games: return
-        with open(self.rcfile_path(game_id), 'r') as f:
-            contents = f.read()
+        path = self.rcfile_path(game_id)
+        if os.path.isfile(path):
+            with open(path, 'r') as f:
+                contents = f.read()
+        else:
+            contents = ''
```
aidanholm: There's a race condition here; one alternative is this:
```python3
try:
    with open(path, 'r') as f:
        contents = f.read()
except OSError:
    contents = ''
```

alexjurkiewicz:
This has been sitting on CPO without issue for a few weeks now. I'm wondering what the next step is.

Do you want it to run on CPO for longer?
Or is it just a matter of someone finding review time?

ebering:
For me I'm happy having seen it run on CPO; it's a question of review time (I took 5 weeks on arrival vaults, its a busy time for me).

If you can talk to another server admin and verify the "seamless deploy" (@floraline perhaps) it'll be easier to merge with a lighter review. (Breaking every server other than CPO is something I'd prefer not to do.)

floraline:
What do I need to do?

crawl-ref/source/webserver/ws_handler.py
```diff
@@ -616,11 +643,17 @@ def go_lobby(self):
...
+        try:
+            with open(path, 'r') as f:
+                contents = f.read()
+        # IOError for py2, OSError for py3
+        except (OSError, IOError):
+            contents = ''
         self.send_message("rcfile_contents", contents = contents)
 
     def set_rc(self, game_id, contents):
+        game_id = str(game_id) # protect against JS submitting bad data
```
aidanholm: sorry, i don't understand this; could you describe in more detail? is it about preventing js sending non utf-8 data?

crawl-ref/source/webserver/server.py
```diff
@@ -250,11 +275,14 @@ def check_config():
...
-        if len(sockets) > 0:
-            ioloop.start() # We'll wait until all crawl processes have ended.
-
-    remove_pidfile()
+        logging.info("Received keyboard interrupt, shutting down gracefully.")
+        try:
+            for server in servers: server.stop()
+            shutdown()
+            ioloop.add_timeout(time.time() + 2, ioloop.stop)
+            if len(sockets) > 0:
```
aidanholm: `if sockets:`

crawl-ref/source/webserver/config.py
```diff
@@ -59,8 +65,20 @@
...
+#
 # morgue_url is for a publicly available URL to access morgue_path
+#
+# milestone_path is the path milestones are written (used for tracking active
+# game actions in the lobby list). If not set, it can be calculated from
+# dir_path. If that is also not set, there is no default. You just need to
+# specify the path to the "milestones" file, the "milestones-seeded" file next
+# to it will be automatically discovered.
 games = OrderedDict([
     ("dcss-web-trunk", dict(
```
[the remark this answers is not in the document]
alexjurkiewicz: All existing webtiles servers have all their games config stored in this way (rather than as yaml). I figured we should show how that works rather than completely deprecating it and leaving it undocumented.

crawl-ref/source/webserver/server.py
```diff
@@ -203,14 +209,31 @@ def check_config():
...
         success = False
+    logging.debug("check_config finishing success:%s", success)
     return success
 
+
+def usr1_handler(signum, frame):
+    assert signum == signal.SIGUSR1
+    # Support configs without load_games defined
+    if not hasattr(config, 'load_games'):
+        return
```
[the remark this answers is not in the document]
alexjurkiewicz: Existing webtiles servers using their existing config.py

crawl-ref/source/webserver/terminal.py
```diff
@@ -11,7 +11,22 @@
...
+                 id_header,
+                 logger,
+                 io_loop,
+                 termsize,
+                 env_vars, # type: Dict[str, str]
+                 game_cwd, # type: Optional[str]
+                 ):
+        """
+        Args:
+            command: array of command to run, eg [cmd, args, ...]
```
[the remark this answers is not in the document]
alexjurkiewicz: That's not very Pythonic, is it?

crawl-ref/source/webserver/ws_handler.py
alexjurkiewicz: I think the issue was that if you have a `game_id` of `0.23` it can get interpreted as a number in some parts of the client-side JS. But I'm not sure why I added this protection only to `set_rc` and not also to `get_rc` (or even setting it higher up, wherever this value is deserialised.

I'll remove it, if the problem re-appears I can add it as a new PR.

(I think the error case happens when creating RC files when you don't currently have an RC file, which only CPO runs into because the other DGL-enabled servers explicitly create blank RC files on user creation.)

alexjurkiewicz:
I'd like to give this a once-over myself before this gets finally merged, but I think I've addressed all of @aidanholm 's feedback

crawl-ref/source/webserver/process_handler.py
```diff
@@ -539,6 +539,9 @@ def _base_call(self):
         if "options" in game:
             call += game["options"]
 
+        if "dir_path" in game:
```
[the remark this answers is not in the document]
alexjurkiewicz: How does it sound to document the first entry in `crawl-ref/source/webserver/games.d/base.yaml`?

crawl-ref/source/webserver/server.py
```diff
@@ -186,13 +188,17 @@ def init_logging(logging_config):
     logging.getLogger().addFilter(TornadoFilter())
     logging.addLevelName(logging.DEBUG, "DEBG")
     logging.addLevelName(logging.WARNING, "WARN")
+    if config.logging_config.get("access_logs") != True:
+        logging.getLogger("tornado.access").setLevel(logging.ERROR)
```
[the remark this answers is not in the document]
alexjurkiewicz: actually, the default here should be for access logs to be enabled (match current behaviour). I'll update the test.

crawl-ref/source/webserver/server.py
```diff
@@ -186,13 +188,17 @@ def init_logging(logging_config):
...
 
 
 def check_config():
+    logging.debug("check_config starting")
```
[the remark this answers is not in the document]
alexjurkiewicz: removed this and the "check_config finished" message, was from development

crawl-ref/source/webserver/server.py
```diff
@@ -203,14 +209,31 @@ def check_config():
...
 
+
+def usr1_handler(signum, frame):
+    assert signum == signal.SIGUSR1
+    # Support configs without load_games defined
+    if not hasattr(config, 'load_games'):
+        return
+    logging.info("Received USR1, reloading config.")
+    try:
+        config.load_games()
```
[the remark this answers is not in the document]
alexjurkiewicz: Each game that gets loaded prints a message, which seems sufficient.

crawl-ref/source/webserver/terminal.py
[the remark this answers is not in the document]
alexjurkiewicz: I guess it is Unixic, though. Updated.

alexjurkiewicz:
Alright, I'm happy with this now 👍

aidanholm:
looks good to me

crawl-ref/source/webserver/games.d/base.yaml
```diff
@@ -0,0 +1,95 @@
...
+    # Directory where ttyrec files for active games are written to
+    inprogress_path: ./rcs/running
+    # Directory where ttyrec files are stored for the individual user
+    ttyrec_path: ./rcs/ttyrecs/%n
+    # Directory where sockets are placed. These sockets are used for
+    # communication between the webtiles server and DCSS executable.
+    socket_path: ./rcs
+    # Static content used by the game (eg spritesheets, game HTML). DCSS builds
+    # this as the web/ directory.
+    client_path: ./webserver/game_data/
```
aidanholm: all paths need to document what they are relative to

crawl-ref/source/webserver/games.d/base.yaml
```diff
@@ -0,0 +1,95 @@
...
+    ttyrec_path: ./rcs/ttyrecs/%n
+    socket_path: ./rcs
+    client_path: ./webserver/game_data/
+    # dir_path: .
+    # cwd: .
+    morgue_url: None
+    # milestone_path: ./rcs/milestones
+    send_json_options: True
+    options:
+      - -sprint
```
aidanholm: there's a lot of duplication here; could we add a `defaults` field, so we don't have to repeat the same stuff over and over for each game item?

Also, note that the paths for some servers don't seem to fit this scheme, e.g. cao has paths like `rcfile_path = "%%CHROOT_RCFILESDIR%%/crawl-0.16/"`; the defaults field should permit substituting in the game id, similar to the `%n` variable:

```yaml
defaults:
  rcfile_path = "%%CHROOT_RCFILESDIR%%/%{id}/"
  ...
```

crawl-ref/source/webserver/games.d/base.yaml
```diff
@@ -0,0 +1,95 @@
...
+    rcfile_path: ./rcs/
+    macro_path: ./rcs/
+    morgue_path: ./rcs/%n
+    inprogress_path: ./rcs/running
+    ttyrec_path: ./rcs/ttyrecs/%n
+    socket_path: ./rcs
+    client_path: ./webserver/game_data/
+    # dir_path: .
+    # cwd: .
+    morgue_url: None
```
aidanholm: please make this relative to the server if it's not an absolute url

crawl-ref/source/webserver/games.d/base.yaml
```diff
@@ -0,0 +1,95 @@
...
+    send_json_options: True
+    # Array of extra options to add to the DCSS command.
+    options:
+      - -seed
+    # Map of extra environment variables to set when executing the DCSS command.
+    # All env vars from the webtiles server environment are automatically
+    # inherited.
+    # env:
+    #   LANG: en_US.UTF8
+  - id: sprint-web-trunk
```
aidanholm: please use the common naming conventions, `spr-git` or `sprint-git` for sprint trunk, `tut-git` for tutorial trunk, etc.

crawl-ref/source/webserver/games.d/base.yaml
```diff
@@ -0,0 +1,95 @@
...
+    # isn't told where their dump was written -- but it's still written.
+    morgue_url: None
+    # The path to the game's "milestones" file. If you specify dir_path this can
+    # be inferred automatically (since it's $dir_path/milestones), but otherwise
+    # you should specify this explicitly. The milestones file is read to
+    # populate the 'milestones' column in the HTML interface. If you don't
+    # specify a milestones path, you'll just see nothing specified there.
+    # (Note: DCSS writes both "milestones" and "milestones-seeded", we'll
+    # automatically pick up both if you specify the former's path.)
+    # milestone_path: ./rcs/milestones
```
aidanholm: it's not clear visually that there's an option here; perhaps recomment the entire block? e.g:
```
# # desc...
# option = foo
```

crawl-ref/source/webserver/games.d/base.yaml
```diff
@@ -0,0 +1,95 @@
...
+    # Legacy option. Leave this alone. (Or submit a PR to remove it.)
+    send_json_options: True
```
aidanholm: can we just delete this? is there any reason to keep it?

crawl-ref/source/webserver/games.d/base.yaml
alexjurkiewicz: I'd rather not. This duplication exists today, and this pr already has a lot of unrelated changes.

crawl-ref/source/webserver/games.d/base.yaml
alexjurkiewicz: I haven't changed the behaviour of how this option is interpreted.

crawl-ref/source/webserver/games.d/base.yaml
alexjurkiewicz: As per above, don't want more change in this PR.

crawl-ref/source/webserver/games.d/base.yaml
alexjurkiewicz: These use the existing names in master: https://github.com/crawl/crawl/blob/master/crawl-ref/source/webserver/config.py#L64

crawl-ref/source/webserver/games.d/base.yaml
alexjurkiewicz: Actually, I think this might be used by some things like robotfindskitten. I'll leave the comment as-is though, since all that is pretty "legacy".

crawl-ref/source/webserver/games.d/base.yaml
alexjurkiewicz: done

crawl-ref/source/webserver/games.d/base.yaml
aidanholm: ah, I was looking that the names that were used in the dgamelaunch-config repo, that's all right then

crawl-ref/source/webserver/games.d/base.yaml
aidanholm: re more changes, fair enough; I'll try some other time to figure out what the hell that thing is doing and rip it out

crawl-ref/source/webserver/games.d/base.yaml
aidanholm: i meant in general, but we can hold off due to the size of this pr

crawl-ref/source/webserver/games.d/base.yaml
aidanholm: fair enough

alexjurkiewicz:
@rawlins did you have any comments on this PR?

crawl-ref/source/webserver/server.py
```diff
@@ -266,12 +270,27 @@ def ensure_tornado_current():
...
             " to at least version 4.")
 
+def usr1_handler(signum, frame):
+    assert signum == signal.SIGUSR1
+    # Support configs without load_games defined
+    if not hasattr(config, 'load_games'):
+        return
+    logging.info("Received USR1, reloading config.")
+    try:
+        config.load_games()
```
rawlins, line 280: would be better to trigger `load_games` via `IOLoop.add_callback_from_signal`, rather than calling it directly.

(I'm not 100% sure that this is in the patched tornado 2.4, I seem to remember it needing a compat hack for that case.)

crawl-ref/source/webserver/config.py
```diff
@@ -72,53 +85,166 @@
...
-        client_path = "./webserver/game_data/",
-        morgue_url = None,
+        # milestone_path = "./rcs/milestones",
         send_json_options = True,
-        options = ["-tutorial"])),
+        # env = {"LANG": "en_US.UTF8"},
+        )),
 ])
 
+def load_games():
```
rawlins, line 97: because of the complexity of dgamelaunch setups and how the handle config.py, and for the sake of easing transition, my instinct is to not put code like this in a module named `config.py`. One idea might be to introduce a new module entirely, e.g. `server_config.py` for servers that have moved to the new setup?

crawl-ref/source/webserver/process_handler.py
```diff
@@ -797,7 +802,7 @@ def _ttyrec_id_header(self): # type: () -> bytes
                  clrscr)
 
     def _on_process_end(self):
-        self.logger.info("Crawl terminated.")
+        self.logger.debug("Crawl PID %s terminated.", self.process.pid)
```
rawlins, line 805: not sure why this one is switched to `debug`?

crawl-ref/source/webserver/server.py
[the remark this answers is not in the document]
rawlins: see note above re maybe using a different module name for code like this

crawl-ref/source/webserver/util.py
```diff
@@ -28,6 +28,10 @@ def __init__(self, root_dir):
 
     def load(self, name, parent_path=None):
         name = self.resolve_path(name, parent_path=parent_path)
+        # Fall back to template.html.default if template.html does not exist
```
rawlins, line 31: as discussed briefly in ##crawl-dev I want to have a look at how this interacts with dgamelaunch-config's handling of templates (which are copied to the running directory in a somewhat ad-hoc way), but I haven't had a chance to do this yet.

crawl-ref/source/webserver/util.py
```diff
@@ -132,13 +138,24 @@ def send_email(to_address, subject, body_plaintext, body_html):
...
+    # Returns an error string describing the problem, or None.
+    # NOTE: a blank email address validates successfully.
     if not address: return None
     if " " in address: return "Email address can't contain a space"
     if "@" not in address: return "Expected email address to contain the @ symbol"
     if not re.match(r"[^@]+@[^@]+\.[^@]+", address): return "Invalid email address"
     if len(address) >= 80: return "Email address can't be more than 80 characters"
     return None
+
+def humanise_bytes(num):
```
rawlins, line 153: this looks basically fine but here's an alternative version of this that I wrote for the scoring code which does the math slightly differently(it's 2-space indent sry):

```
def fmt_byte_size(b, precision=1):
  if b < 1000:
    return str(b) + 'B'
  b = b / 1000.0
  fmt = "%%.%df" % precision
  for suffix in ['KB', 'MB', 'GB', 'TB']:
    if int(b / 1000) == 0 or suffix == 'TB':
      return (fmt % b) + suffix
    b = b / 1000.0
```

If one were to be really pedantic, my code should probably use `kB` etc and yours should say `kibibytes` etc.

crawl-ref/source/webserver/ws_handler.py
```diff
@@ -658,8 +687,13 @@ def go_admin(self):
...
         if game_id not in config.games: return
-        with open(self.rcfile_path(game_id), 'r') as f:
-            contents = f.read()
+        path = self.rcfile_path(game_id)
+        try:
+            with open(path, 'r') as f:
+                contents = f.read()
+        # IOError for py2, OSError for py3
+        except (OSError, IOError):
+            contents = ''
```
rawlins, line 696: should maybe log an error if this happens? Or do some servers not initialize rc files?

crawl-ref/source/webserver/README.md
```diff
@@ -0,0 +1,85 @@
...
+* [Running the server for testing purposes](#running-the-server-for-testing-purposes)
+* [Running a production server](#running-a-production-server)
+* [Contributing](#contributing)
+
+## Prerequisites
+
+To run the server, you need:
+
+* Linux or macOS (other platforms are untested)
+* Python 2.6 or newer (Python 3.6+ is preferred)
```
rawlins, line 21: I'm surprised the server works with python 2.6 but even if it does, we should definitely say python 2.7 here as the lower bound.

crawl-ref/source/webserver/README.md
```diff
@@ -0,0 +1,85 @@
...
+the options are commented or should be self-evident. Suggestions:
+
+* Set uid and gid to a non-privileged user
+* Enable logging to a file in `logging_config`
+* If required, write a script that initializes  user-specific data, like copying
+  a default rc file if the user doesn't yet have one. You can have the script be
+  run on every login by setting `init_player_program`. There is an example
+  script in `util/webtiles-init-player.sh`, but you will probably want to
+  customise it.
+
```
rawlins, line 68: Could mention https://github.com/crawl/dgamelaunch-config here (this could be updated independent of this PR too).

crawl-ref/source/webserver/server.py
```diff
@@ -232,6 +232,9 @@ def init_logging(logging_config):
         logging.getLogger().addFilter(TornadoFilter())
     logging.addLevelName(logging.DEBUG, "DEBG")
     logging.addLevelName(logging.WARNING, "WARN")
+    # Not a simple boolean check because the default (if unset/None) is True
+    if config.logging_config.get("access_logs") == False:
+        logging.getLogger("tornado.access").setLevel(logging.ERROR)
```
rawlins, line 237: trunk server.py `init_logging` already has some code that does this, should recognize the two. This was always supposed to be suppressed by default but until recently the suppression code only worked in tornado 2.

crawl-ref/source/webserver/ws_handler.py
rawlins: ok, this comment was largely a byproduct of not reading the PR by commit.

crawl-ref/source/webserver/process_handler.py
alexjurkiewicz: this prints every time a player ends their game -- it is log spam IMO

crawl-ref/source/webserver/util.py
alexjurkiewicz: Sure. I believe it won't cause problems. The existing paths remain as "server-defined" locations, so dgamelaunch-config's logic will still be correct.

crawl-ref/source/webserver/ws_handler.py
alexjurkiewicz: CPO doesn't. All dgamelaunch-config servers do, but with this change they won't have to create 10+ 0-byte files per user sign-up 😊

crawl-ref/source/webserver/config.py
alexjurkiewicz: That makes sense. I've split to a new file.

crawl-ref/source/webserver/server.py
alexjurkiewicz: Copied the logic from the existing signal handler

crawl-ref/source/webserver/util.py
rawlins: Right now, most instances of dgamelaunch-config copy (with overwriting) `client.html` and `game_links.html`, and ignore everything else in that directory. The most recent version (which no one is running yet) copies everything else with an `.html` extension as well but with --no-clobber.

example of typical current setup: https://github.com/neilmoore/dgamelaunch-config/blob/cbro/chroot/sbin/install-trunk.sh#L68
most recent: https://github.com/crawl/dgamelaunch-config/blob/master/chroot/sbin/install-trunk.sh#L68

AFAIK `client.html` should never actually be customized by a server owner, at least in a way that doesn't allow for automatic updates, because this can get modified when changing the webtiles ui -- so I think using this `.default` logic for it is actually wrong (i.e. if it changes in trunk it always needs to get automatically updated). It's possible that someone might customize `game_links.html` (do you know?), though that also seems iffy in the present setup to me. I think my preference would be to not use `.default` versions (i.e. rename them back to the .html version) of these two files so that the current dgamelaunch-config logic will continue to work, and if this logic is needed for game_links.html, phase it in later. And never use it for `client.html`.

The `.default` logic seems ok for the other templates -- it'll require an update to the most recent dgamelaunch to copy those files as well (maybe without `--no-clobber`but that's fine, if someone's already using that they'll be on the ball enough to deal with it. Existing servers using dgamelaunch-config won't need to immediately do anything because they have already copied (and possibly customized) versions in their chroot.

aidanholm:
I'd like to cherry-pick some of the smaller changes that are self-contained and unrelated to the main thrust of this PR:
 - f2b645c23 and 8d42377: create RC files on demand
 - 7bbbf849d and cd57e460: unit tests for `util.py`
 - 36a00ba: webtiles gitignore
 - 117120ca: javascript string/number fix
 - 5b9b1c88: auth.py test coverage
 - 167b7ebfc: removal of useless valueerror catch

Secondly, while I approve of the conscientious use of commit message credits, I think not amending the original commits also clutters the git history; in the case of the unit test fixups this actually introduces broken builds, so I'd prefer all these to be squashed.

rawlins:
That's fine with me, though the whole thing is probably close to merging, I just still need to do some testing. (And yes, some rebasing is probably good before merging, but I'm guessing @alexjurkiewicz simply didn't do that yet so I could see the changes relevant to my comments.)

Also, there's still a failing test on py2.7 that looks like a real failure.

crawl-ref/source/webserver/util.py
alexjurkiewicz: Changing the logic to be per-file seems very difficult and brittle (you'd have to hardcode this difference in template loading behaviour). Would you be alright with me adding a `{# comment #}` to the top of client.html saying "don't modify this file, webtiles server admins"?

game_links isn't modified by any server, yet. But I think it will be in the near future for CPO (changing formatting of trunk vs stable vs old stable vs fork).

alexjurkiewicz:
Cherry picking the easier-to-merge stuff seems like a great idea from my side 👍 .

crawl-ref/source/webserver/util.py
rawlins: > Changing the logic to be per-file seems very difficult and brittle (you'd have to hardcode this difference in template loading behaviour).

I don't think it's so crazy to hardcode that there are some specific templates that are not overrideable by the server, but maybe what would be simplest for now for compatibility would be to rename `client.html.default` and `game_links.html.default` back to the plain `.html` form and not provide a `.default` version of them in the repository. (I'm not sure if it's intentional, but this is already how `game.html` is handled, which should also not be overrideable and is loaded via this same function.)

alexjurkiewicz: That's a clever solution!

rawlins:
Merged starting with fc79a7346936f4! A few minor further compatibility fixes needed for dgamelaunch-config, nothing too major.

For a dgamelaunch-config server to actually use yml config will take a bit more work in that repository, as the `games.d` directory isn't handled well by it right now. I will try to figure out the minimal needed changes before 0.25 release or so.

PR #1259, opened 2020-01-16
Add Github Actions CI

alexjurkiewicz:
I initially implemented this for code coverage support, which is
included. But using GA for all CI tasks has several advantages over
Travis:

* Simpler configuration language.
* Easier composability of steps from third parties. The steps can be
  published as Docker images or Node modules and are fully versioned.
* Better integration with GitHub (one status check per build job, rather
  than one for the whole CI system).

Requires `CODECOV_TOKEN` set up as a GitHub repo secret (from
codecov.io).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,157 @@
+name: Build
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  linting:
+    name: Code Linting
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Unbrace
+        run: util/unbrace -n
+        working-directory: crawl-ref/source
+      - name: Checkwhite
+        run: util/checkwhite -n
+        working-directory: crawl-ref/source
+  build_linux:
+    name: "${{ matrix.compiler }} ${{ matrix.build_opts }} ${{ matrix.debug }}"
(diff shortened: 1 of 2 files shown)


alexjurkiewicz:
You can see examples of the codecov report here:
https://github.com/alexjurkiewicz/crawl-ref/pull/5
https://codecov.io/gh/alexjurkiewicz/crawl-ref/src/github-actions/crawl-ref/source/english.cc#L27

alexjurkiewicz:
This isn't intended to supplant or replace Travis CI in a single pull request. The PR adds similar build coverage as Travis implements now, but there are a few shortcomings:

* No native support for notifying of broken builds (except for to the committer via email). IRC support (or support for any messaging service eg Discord) would need an external system listening for GitHub's [`CheckSuiteEvent` webhook](https://developer.github.com/v3/activity/events/types/#checksuiteevent). I'm writing a little function to do that [here](https://github.com/alexjurkiewicz/gh-checks-to-irc), if it's required.
* I didn't copy some of the more esoteric Travis build configurations over. They would be easy to add if required.

SeveMonahan:
Merged this for the code coverage support. I could see either moving the travis configuration to github actions CI or moving the code coverage to travis in the future, particularly if keeping two continuous integration files updated proves to be a burden.

PR #1260, opened 2020-01-16
Remove the per-weapon-type vorpal brand adjectives

emilazy:
These mostly served to confuse players and hide the relation between
the vorpal brands.


emilazy:
(sorry for the trillion pushes; in my defence I wrote way too much of this without a compiler)

PR #1266, opened 2020-01-27
Raise Troll Maces & Flails aptitude to +2

pperezvillacastin:
This is intended to provide a replacement for Ogres old big club
aptitude and playstyle in Trolls.

This way Ogre stays more of a versatile fat human, while Trolls get an
alternative melee style that is not so clearly inferior to Unarmed
Combat. It also helps differentiate their relationship to big clubs.

Compared to unarmed combat, big clubs give Trolls a road that requires
less XP investment, and that rewards them with the possiblity of brands
and using god gifts - but punishes them by removing the possibility of
using shields like they tipically do with UC.

Right now, big clubs are usually not really worth pursuing for Ogres,
and simply inferior to UC for Trolls. This should help bring the
playstyle back for those that enjoyed it.

### DIFF
--- a/crawl-ref/source/dat/species/troll.yaml
+++ b/crawl-ref/source/dat/species/troll.yaml
@@ -13,7 +13,7 @@ aptitudes:
   short_blades: -2
   long_blades: -2
   axes: -2
-  maces_and_flails: -1
+  maces_and_flails: +2
   polearms: -2
   staves: -2
   slings: -4


ebering:
I'm going to close this because I disagree with your motivating analysis that M&F is strictly inferior to UC on a Troll.

The number of skill points needed for a giant club to reach minimum delay: 12,843; giant clubs have a base damage of 20 and a min delay of 0.7. Spending these skill points on UC instead gives 19 UC, which gives a base UC damage of 3 + 19 + 6 = 28 and a delay of 0.65 (random rounds between 0.7 and 0.6). Unarmed has a higher base damage here (by 8), but denies the player access to brands, enchantments, and the flexibility of using other M&F class weapons (if a cool unrand drops, for example). A +0 giant club with a multiplicative brand will be comparable to UC with equivalent skill investment.

In your proposed change, the skill points needed for M&F 18 becomes 7,737. Spending this on UC instead gives 15 UC, which gives a base UC damage of 24 and a delay of 0.73. The results are that UC and an unbranded +0 giant clubs are closer together in power level; once the giant club gains a brand it becomes better. It doesn't reach no-brainer levels when compared to UC, but this change doesn't move in the right direction.

pperezvillacastin:
What about shields though?

They are cheap and very valuable on races with bad defenses. The Giant Club route forces your to forfeit them. UC doesn't.

Doing this comparison without accounting for shields seems silly to me.

Might as well just remove giant clubs at this point. They are just baiting players into a bad decision by forfeiting shields on races that desperately need some sort of defense, without offering that much in return.

alexjurkiewicz:
Why don't you compare the damage and bonus defenses a troll gets from
investing a certain amount of SP into m&f/uc with and without putting some
into shields? The same way ebering did.

PR #1268, opened 2020-01-29
Fix several monster spells descriptions

nlavsky:
This PR introduces a new type of marker for database entries, `[[foo]]`, and fixes several descriptions of monster spells.


ebering:
Thanks for the desc patches!

PR #1250, opened 2020-01-03
Doc updates

alexjurkiewicz:

PR #1269, opened 2020-01-29
Make species-gen.py explicitly depend on python3

alexjurkiewicz:
Rather than "python".

Currently this script works with both major versions, so `python` was
specified to use whatever version is installed. But we are starting to
see OSes that install Python 3 only with a python3 name and no python
shortcut. For example, macOS Catalina (10.15), and possibly msys mingw
on Windows.

This new behaviour follows recommendations from the Python project (PEP
394), so is likely to become more prevalent over time.

---

Older webtiles servers (CAO, CDO) may not have Python 3 already
installed, so will need some manual intervention.

If getting Python 3 on these servers proves too difficult, the build process could be changed to initialise a Python virtualenv and run the command within that (to guarantee the existence of a `python` binary). But this would add complexity to everyone else's build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,8 +92,7 @@ jobs:
       - name: Install Dependencies
         run: |
           brew install ccache
-          sudo easy_install pip
-          sudo pip install PyYAML
+          sudo pip3 install PyYAML
       - name: Add ccache to PATH
         run: echo "::add-path::/usr/local/opt/ccache/libexec"
       - name: Cache compilation
--- a/crawl-ref/INSTALL.md
+++ b/crawl-ref/INSTALL.md
@@ -72,7 +72,7 @@ These instructions may work for other DPKG-based distros.
 
 ```sh
 sudo apt install build-essential libncursesw5-dev bison flex liblua5.1-0-dev \
-libsqlite3-dev libz-dev pkg-config python-yaml binutils-gold
+libsqlite3-dev libz-dev pkg-config python3-yaml binutils-gold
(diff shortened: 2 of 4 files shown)


Goratrix:
Regarding MSYS2 on Windows, as I mentioned [here](https://github.com/crawl/crawl/pull/1250) they went the opposite direction and now use only 'python' as default everywhere instead of the versioned shortcuts. But I guess it's aliased, so python3 should work regardless.

ebering:
I don't have admin access to CAO and CDO, so I don't know if they have `python3`

I'm not sure it's the pythonic approach, but it might be better from a crawl infrastructure standpoint to have the makefile explicitly call `$(PYTHON) species_gen.py` and deal with setting `$(PYTHON)`

@rawlins or @gammafunk are definitely the ones to weigh in here.

alexjurkiewicz:
> I'm not sure it's the pythonic approach, but it might be better from a crawl infrastructure standpoint to have the makefile explicitly call $(PYTHON) species_gen.py and deal with setting $(PYTHON)

That seems good to me. I'll send a patch

alexjurkiewicz:
Rebased and squashed. Thanks for the idea.

ebering:
It looks like this is failing the check because the macOS environment has python3 but the pip dependency install is getting yaml for python2? I'm not 100% sure. Don't have time for a local build to investigate atm.

alexjurkiewicz:
Yep, that looks like it. The CI runner has a bunch of useful software installed by default. Should be fixed... 🤞

PR #1271, opened 2020-02-02
Clarified scroll only summons allied creatures

theJollySin:
Recently, a player new to DCSS pointed out she/he didn't know if a "scroll of summoning" would summon friendly or dangerous creatures:

https://old.reddit.com/r/dcss/comments/exa6zm/observations_from_a_dcss_newbie/

Which is a fair point.

I added a single word to clarify that this scroll only ever summons allies.


SeveMonahan:
Thanks for the contribution!

PR #1272, opened 2020-02-03
Cigotuvi's Embrace, unrand body armour

PleasingFungus:
Cigotuvi's Embrace is a unique body armour (currently based on ring
mail) that automatically sucks up corpses within LOS and turns them
into temporary AC. This is primarily intended to provide a fun
press-your-luck mechanic - how long can you keep fighting and keep
your corpse armour on before you have to rest? It also comes as
something of a 'hunger' conduct (since you can't eat those corpses)
and effectively turns off most forms of necromancy, which can be a
plus or a minus, depending on whether you're a necromancer or your
enemies are.

At present, the Embrace starts as +4 rN+ rRot armour (the rRot is for
flavour). It goes to +8 with one corpse harvested, which takes about
100 turns to fully slough off. Four corpses give total ench +12 and
take 200 turns to fully go away, sixteen corpses give total ench +20 and
take 400 turns to fully rot away, etc. All numbers are very much open
to discussion.

The original Embrace was removed because it couldn't compete with
necromancy as a spell; as an unrand it might be more feasible to
balance and make fun.

Some levers to tweak:
- Base type. Ring mail seemed like a reasonable type - most characters can use it if they really want to, and Armour skill is at least a little relevant - but I'm not wedded to it. Going lighter is probably more interesting than going heavier, since the no-corpses conduct is usually a bit more relevant for casters than melee characters.
- Base enchantment. My goal was that the base case should be weak but not terrible, and +4 rN ring mail seems about right.
- The AC from one corpse. My goal was for this to get the armour to be good but not amazing; +8 rN ring seemed pretty decent.
- AC scaling from further corpses. Goal was diminishing returns but still some impact.
- The duration. I wanted players to be able to plausibly carry some AC between fights if they kept moving.
- The decaying-AC-over-time effect. The current approach (directly tying AC to remaining duration) means there's no hidden information for the player to track, but it does make the effect harder to reason about in some respects. 1 corpse is 4 ac for a 25 turns, then 3 ac for the next 25, etc. We could instead have the AC only slowly stack up until the whole thing falls off at once; that seems like its own flavour of mess, though...

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -1388,7 +1388,6 @@ COLOUR:   ETC_UNHOLY
 # XXX TODO: tiles
 LIFE:     1
 INSCRIP:  rRot
-BOOL:     evil
 
 # This entry must always be last.
 ENUM:   DUMMY2
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -1388,6 +1388,7 @@ COLOUR:   ETC_UNHOLY
 # XXX TODO: tiles
 LIFE:     1
 INSCRIP:  rRot
+BOOL:     evil
 
 # This entry must always be last.
 ENUM:   DUMMY2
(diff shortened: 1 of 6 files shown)


sdynet:
Cool

PR #1277, opened 2020-02-09
Rename shields to reduce confusion

PleasingFungus:
Having the name of a category (shields) also be the name of an item in
that category (shields, large shields) has always been annoying and
confusing. ("Oh, I'm using a long sword and a shield." "What kind of shield?") Let's use more specific words instead.

- large shield -> pavise, a sort of large shield often used by crossbowmen.
- shield -> hoplon, a medium-sized shield used by hoplites.

https://github.com/crawl/crawl/commit/8322e1f2f0018ea2bc97ec5bdc8da818606cd6da has more details.


ebering:
I'm ok with hoplon but have you considered that this is an opportunity to introduce **triple shield** as a new name in place of large shield. (For Lore parity with swords).

PleasingFungus:
lore is just so important

PleasingFungus:
Rebased and merged manually.

PR #1279, opened 2020-02-10
Elyvilon rework

sdynet:
Overview
Elyvilon is a god with a theme of healing and protection. This theme is currently well managed but unfortunately too loyal to the theme image, the Elyvilon play style is dull and boring. So I'm not going to deviate from the frame of the theme, but I have try to change Elyvilon in the most dynamic way possible.

Main Changes:
1. Lesser Healing, Greater Healing, Purification, Divine Vigour, Divine Protection has been remove.
2. The new power purifies you and makes holy clouds under your feet. You heal slowly here.
3. If you succeed in pacify the enemy, recover hp / mp. Also, you can now swap places with pacified enemies and are not attacked by them.
4. Elyvilon sometimes disarm your enemies(passive effect).
5. You can transform yourself into a holy creature that is protected from the Elyvilon. Its has strong unarmed attacks and high defenses.

For information on all changes, please check this post: https://crawl.develz.org/tavern/viewtopic.php?f=8&t=26791&p=343380#p343380

Feedback wanted:
I deliberately curbed the cost of mana so that non-Warrior jobs could serve Elyvilon. Because, I wonder if various combinations are possible. Perhaps it is too strong. But balance can be controlled. What I'm paying attention to is whether it's fun to play this. Is this god interesting enough to serve until you wins the game?


ebering:
This is one of those times where checking in with the devteam earlier is a good plan. This PR has had a lot of work put into it, but isn't a direction we'd like to take Elyvilon in. I'm going to close it.

Currently Elyvilon is compatible with a wide variety of play styles. In this redesign making the 5* ability a "holy transmutation" makes Ely highly attractive to UC builds (with or without transmutations skill) and not very attractive to other styles (what good is this form for them other than defenses?), since the cost of the form must be pretty high considering the UC bonus.

Adding a positional component to Ely's abilities could be interesting, but it's hard to think of one that is on theme and makes positioning matter, other than requiring the player to stay still in order to receive the healing, which is basically trading immediate healing for "regen in place" and not that compelling unless it's dramatic.

PR #1286, opened 2020-02-14
Potion of agility -> launching

PleasingFungus:
Potions of agility were not an interesting item. You plopped them
on when you were going to get in a big fight and they made you take
slightly less damage. It wasn't something that you used differently
from !might or !brill or arguably !haste, just another buff to stack
on the pile and clog up your inventory.

Instead, replace !agility with !launching. !launching increases
ranged accuracy somewhat and makes ranged attacks pierce, for a short
duration. This differentiates itself from other buff potions in a
few ways. It's most useful against groups, rather than single tough
enemies, and past the early game (once inventory burden hits), it's
only really useful for ranged characters and melee chars who invested
in Throwing. That means that most characters won't feel obliged to
carry it.

This effect didn't feel insanely strong in testing, but if it does
end up too strong, we can easily tweak the to-hit, duration, and
chance of penetration (perhaps a roll per-attack?). If it's too
weak, we could increase accuracy much more.


alexjurkiewicz:
Does this idea obsolete the potion of camouflage?

PleasingFungus:
Yes, it's intended to be an alternative proposal (and i think it's probably a bit more interesting). Option 3 is removing !agi without replacement.

ebering:
`!stab` got the merge instead

PR #1290, opened 2020-02-20
Modifies five vaults in large.des to comply with current practices.

ByrelMitchell:
The current arrival vault standard is for arrival vaults to have multiple exits, a hatch, or a large enough internal space for relevant tactics. I went through large.des picking out clear offenders (using the simplified rubric that anywhere a single gnoll could spawn should still leave you with more options than 'tab and hope you win.')

elethiomel_arrival_infinity_welcomes_careful_drivers: This vault has an issue if a gnoll spawns in the closet just outside the starting room. I added an escape hatch to the entrance room, so you have the option of backing up to there and going to the next floor.

erik_arrival_gehennom: This vault explictly places a goblin and kobold (with standard weapon generation) right outside the starting room. Many starts are incapable of fighting a goblin and kobold with a bad weapon spawn in this layout. I added a second exit to the room just north of the starting stairs. At least that way you have a pillar to work with if you can't handle that fight.

kilobyte_arrival_exit_the_forest: This vault is effectively a single long corridor with monster-gen on; if a gnoll spawns and you can't fight him, you're screwed. I added a path through the woods in the northwest so you have a large pillar to work with.

arrival_zaba_pyramid: Each of the possible starting rooms are single-exit, and monster-gen is on. So it's possible for a gnoll to spawn right in front of the door to your room, and completely block you in. I expanded the small rooms between entry-rooms, and added a second exit to each of them.

minmay_arrival_shoal_huts: This vault is another case with single-exit entry rooms which can be blocked by an unlucky gnoll spawn. I added a hatch to each entry room as an additional option.

### DIFF
--- a/crawl-ref/source/dat/des/arrival/large.des
+++ b/crawl-ref/source/dat/des/arrival/large.des
@@ -44,6 +44,8 @@ ENDMAP
 NAME:    elethiomel_arrival_infinity_welcomes_careful_drivers
 TAGS:    arrival no_rotate
 ORIENT:  northwest
+KMASK:   a = no_monster_gen
+SUBST:   a : .
 SUBST:   ' : . x ':40, '=x.
 SUBST:   c : cxxx
 MAP
@@ -51,11 +53,11 @@ xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 xxxxxxx....xxxxxxxxxxxxxxx....xx..xxx.''xxxxxx
 xxxxx''........xx...''.........''...'..xx...xx
 xxx'....ccccc...xxx.....ccccc......'......x..@
-x....cccc...cccc.....cccc...cccc.......''..xxx
-x..ccc.........ccc+ccc.........ccc....''..xxxx
-x'.c......{......+.+......T......+.........xxx
-x'.ccc.........ccc+ccc.........ccc...''...xxxx
-x....cccc...cccc.....cccc...cccc......'..'.xxx
+x....ccccaaacccc.....ccccaaacccc.......''..xxx
+x..cccaaaaaaaaaccc+cccaaaaaaaaaccc....''..xxxx
+x'.caaaaaa{aaaaaa+a+aaaaaaTaaaaaa+.........xxx
+x'.cccaaaaaaaaaccc+cccaaaaaaaaaccc...''...xxxx
+x....ccccaaacccc.....ccccaaacccc......'..'.xxx
 x.......ccccc......'....ccccc......'.....xxxxx
 x'................'..xx........'....'....xxxxx
 x.''..xxx...''.....'..xx....'''..''.xxx''.xxxx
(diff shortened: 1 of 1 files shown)


crawl-ref/source/dat/des/arrival/large.des
```diff
@@ -53,7 +53,7 @@ xxxxx''........xx...''.........''...'..xx...xx
 xxx'....ccccc...xxx.....ccccc......'......x..@
 x....cccc...cccc.....cccc...cccc.......''..xxx
 x..ccc.........ccc+ccc.........ccc....''..xxxx
-x'.c......{......+.+......T......+.........xxx
+x'.c...>..{......+.+......T......+.........xxx
```
ebering: This is probably better fixed by adding KPROP: no_monster_gen to the ∞ part of the vault (the two stone rooms here).

crawl-ref/source/dat/des/arrival/large.des
```diff
@@ -230,8 +230,8 @@ x.....l.cc+ccccccccccc..l....x
 x.....l.cc..........cc..l....x
 x.....l.ccc.........cc..l....x
 x.....ll.cc.....2.3.cc.ll....x
```
ebering:
```suggestion
x.....ll.cc.....G.G.cc.ll....x
```
This vault is one we wouldn't accept these days (big spoiler and has monsters). Hit the remove button on this, or if you don't think that's appropriate turn the 2 and 3 into statues and remove the rat and oozes and put some gaps in the lava.

crawl-ref/source/dat/des/arrival/large.des
```diff
@@ -332,8 +332,8 @@ FTILE:  .t1 = floor_moss
 KMASK:  Ww = no_monster_gen
 MAP
 ttttttttttttttttttttttttttttttttttttttttttt
-tttttttttttt1ttttttwwtttttttttttttttttttttt
-tttt...1ttttt..tttwwwttt.tttttttt....ttttt
+ttttttttt1.t1ttttttwwtttttttttttttttttttttt
+tttt...1..t....tttwwwttt.tttttttt....ttttt
```
ebering: This is a good change, I think this vault wold also benefit from a second exit path past the river.

crawl-ref/source/dat/des/arrival/large.des
```diff
@@ -476,8 +476,8 @@ MAP
 .................@.................
 .cccccc...cccccc...cccccc...cccccc.
 .c....c...c....c...c....c...c....c.
-.c.(..c...c.{..c...c..[.c...c..>.c.
-.c....cc+cc....cc+cc....cc+cc....c.
+.c.(..cc+cc.{..cc+cc..[.cc+cc..>.c.
+.c....+...+....+...+....+...+....c.
```
ebering: I think instead of re-shaping these small rooms, a `KPROP: no_monster_gen` for the first part of this pyramid (the entry rooms and their anterooms) is a better solution.

ByrelMitchell:
Implemented those changes. I removed the Gehennom vault instead of improving it because it's just weird to have an 'homage' vault that doesn't resemble anything in its inspiration.

PR #1291, opened 2020-02-21
Prevent swapping to regen when resting

ufshaikh:
The main effect of these commits is to make troll leather armour and artefact armour with regen+ behave like "regen in requiring you to have full hp before getting the bonus to your regen rate. This should prevent the pattern of swapping back and forth between TLA and your "real" armour in order to regen faster with TLA while resting. This is fairly boring and repetitive gameplay, and at the cost of tedium allows one to get a part of the armour's benefit without committing to using the armour as armour.

### DIFF
--- a/crawl-ref/source/player-equip.cc
+++ b/crawl-ref/source/player-equip.cc
@@ -127,6 +127,7 @@ static void _equip_jewellery_effect(item_def &item, bool unmeld,
 static void _unequip_jewellery_effect(item_def &item, bool mesg, bool meld,
                                       equipment_type slot);
 static void _equip_use_warning(const item_def& item);
+static void _equip_regeneration_item(const item_def& item);
 
 static void _assert_valid_slot(equipment_type eq, equipment_type slot)
 {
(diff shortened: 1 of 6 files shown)


semitonal:
This definitely sounds like a good change to me in principle. As implemented, it doesn't look like regeneration from artefact armour is currently handled properly, however - it'll never become active in _update_equipment_attunement() unless you're already at full HP when you equpi it. If you have multiple pieces of +Regen armour equipped, there'd also need to be handling so that removing one piece doesn't cancel your regeneration from other pieces.

ufshaikh:
Hi Kate, thanks for taking a look.

> it'll never become active in _update_equipment_attunement() unless you're already at full HP when you equip it.

It seems to pass some basic testing: if I wizmode up a TLA, get hurt, put on the TLA, and rest, then I see the message about attunement print when my hp tops off; and checking before and after completing resting, I see the prop go from 0 to 1 and my regen rate in the % screen bump up. Have I got the case you're worried about right?

In case it's not obvious, note that _update_equipment_attunement is run each turn (unless you have MUT_NO_REGENERATION), not on equipping the armour.

> If you have multiple pieces of +Regen armour equipped, there'd also need to be handling so that removing one piece doesn't cancel your regeneration from other pieces.

I'm afraid I systematically assumed that you can only have the regen+ property on body armour, which looking again seems wrong. So I will have to complicate a few things, including what you mention. Sorry about this.

semitonal:
> It seems to pass some basic testing: if I wizmode up a TLA, get hurt, put on the TLA, and rest, then I see the message about attunement print when my hp tops off; and checking before and after completing resting, I see the prop go from 0 to 1 and my regen rate in the % screen bump up. Have I got the case you're worried about right?

I believe you're right that troll leather works, but randart armour of other types can get the +Regen property, so for example if you find a shield/plate mail/pair of boots with ARTP_REGENERATION, those won't activate unless you're at full heath when equipping them.

ufshaikh:
Oh, I see what you mean.

I've just pushed another version that should fix these problems, hopefully without creating more. I can rebase to clear away the old approach when you want, but I left it for now in case that makes it easier to review.

ebering:
Sorry for the delay in getting to merging this. I made a few tweaks and squashed things. Thanks for the patch!

PR #1292, opened 2020-02-21
Disallow -tele on artefact weapons and jewellery

ufshaikh:
-tele is usually not a meaningful bad artefact property on weapons and
jewellery, since swapping away from those items is so quick.
Furthermore, putting -tele on quick-swapping items allows the player to
explore with a -tele item equipped and swap to their "real" equipment
for fights, thereby avoiding teleport traps and teleportitis at marginal
cost. This is frustratingly uninteresting gameplay. And it allows the
player to neutralize the gameplay benefits of teleport traps with little
risk and no real investment or decision-making regarding their
character's loadout.

A simple solution is to prevent -tele from spawning on any weapons or
jewellery. We could also try to make sure that on jewellery and weapons
-tele always pairs with other properties that make swapping less
trivial (*contam, fragile, harm, etc.), but I am skeptical that the
complexity is worth the trouble.

I also considered changing the behaviour of -tele to stop translocations
only from an internal source. Then wearing -tele would not affect
teleporting unless the player could already choose whether or not to
teleport, so there is no reason to swap to -tele to explore. However
this is additional "player-facing" complexity, in that there is another
somewhat fiddly rule for the player to learn; even very careful players
don't need to know about rules governing placement of artefact
properties. Furthermore it does not address the first issue mentioned,
that weapon and jewellery -tele is rarely an interesting drawback.
Finally this approach takes away an arguably interesting strategic
choice that players are making, to wear -tele on armour to prevent
teleport traps in particular places (esp. Zot).

### DIFF
--- a/crawl-ref/source/artefact.cc
+++ b/crawl-ref/source/artefact.cc
@@ -561,7 +561,9 @@ static bool _artp_can_go_on_item(artefact_prop_type prop, const item_def &item,
             // and obv we shouldn't generate contradictory props
         case ARTP_PREVENT_TELEPORTATION:
             return !extant_props[ARTP_BLINK]
-                   && !extant_props[ARTP_CAUSE_TELEPORTATION];
+                   && !extant_props[ARTP_CAUSE_TELEPORTATION]
+                   && item_class != OBJ_WEAPONS
+                   && item_class != OBJ_JEWELLERY;
             // no contradictory props
         case ARTP_BLINK:
             return !extant_props[ARTP_PREVENT_TELEPORTATION];


theJollySin:
Hasn't this been a much-loved game feature for like a decade?

Why remove a feature that is clearly there on purpose?

ufshaikh:
I didn't check to see how long -tele has been on weapons and jewellery. I also don't know if whoever made it that way intended for items with the property to be used as a (partial) toggle for the trap mechanic. I suspect that they simply didn't anticipate the consequence, not that it was intentional, but that's just a guess.

I'm surprised to hear that you think -tele on weapons, and the swapping possibilities it raises, has been a much-loved feature. My experience is that the swapping strategy has been mostly ignored until the recent trap changes made the possibility of getting teleport trapped in Zot more salient. It's not as if I've made a point of tracking players' attitudes about it, though.

As for why change the mechanic... well, for the reasons in the commit message. Without hearing more about why you don't find the reasoning compelling I'm not sure how to answer this.

PR #1293, opened 2020-02-23
WIP: Add server chroot migration scripts

aidanholm:
This adds some scripts for migrating aging server chroots to newer ubuntu images.

### DIFF
--- a/crawl-ref/source/util/server/install-ubuntu-chroot
+++ b/crawl-ref/source/util/server/install-ubuntu-chroot
@@ -1,0 +1,130 @@
+#!/bin/bash
+#---help---
+# Usage: install-ubuntu-chroot -d destination-directory
+#
+# This script downloads and installs a recent version of Ubuntu to a chroot.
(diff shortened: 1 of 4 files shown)


crawl-ref/source/util/server/install-ubuntu-chroot
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+#---help---
+# Usage: install-ubuntu-chroot -d destination-directory
+#
+# This script downloads and installs a recent version of Ubuntu to a chroot.
+#---help---
+
+set -euo pipefail
+
+die() {
```
alexjurkiewicz: you could pull these commands out into a `common.sh` and source them for all scripts `source ./common.sh`

aidanholm: I could, but I don't necessarily think it's worth it; these scripts are supposed to be relatively self contained

PR #1298, opened 2020-02-26
Put runes back on default autopickup (Lici the Crawler)

ufshaikh:
Fixes a syntax error in 458a23e4a.


semitonal:
Oops, good catch. Thanks!

PR #1300, opened 2020-02-29
Random Item Strings Cleanup

theJollySin:
I wanted to add a couple of new strings to the random item names files:

* rand_all.txt
* rand_arm.txt
* rand_wpn.txt

But when I looked at our data files they weren't alphabetized.  So the primary goal of this PR is just to alphabetize these three files.  In the process I deleted some duplicate entries that we had, presumably because they were hard to see before alphabetization.

The Git diff on the first commit looks large, because Git is not terribly clever about tracking lines that have been re-ordered.

I then also added some new strings into each file.  Not many, mostly in the realm of medieval jobs and ancient philosophies.

### DIFF
--- a/crawl-ref/source/dat/database/rand_all.txt
+++ b/crawl-ref/source/dat/database/rand_all.txt
@@ -18,11 +18,11 @@
 ######################################################
 _jewellery_animal_
 
+Death Scarab
+
 Firefly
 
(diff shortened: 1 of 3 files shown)


ebering:
Closed via 16f5e22

PR #1301, opened 2020-03-01
Shortened versions of scarf tiles (#12210)

Goratrix:
Some of the current scarf tiles are too long and obstruct the body armor
underneath. This makes all of them the same length, which is
2 pixels shorter than the current shortest ones.


Goratrix:
updated with multiple commits by mistake, sorry. will try to submit these properly later.

PR #1302, opened 2020-03-01
Shortened versions of scarf tiles (#12210)

Goratrix:
Some of the current scarf tiles are too long and obstruct the body armor
underneath. This commit makes all of them the same length, which is
2 pixels shorter than the current shortest ones.


Goratrix:
I just noticed that the deleted transparent pixels have RGB white instead of black, will update them.

ebering:
Thanks for the updated tiles! How would you like to appear in the CREDITS?

Goratrix:
Thank you for the merge! "Goratrix" is fine for the credits, thanks :)

PR #1310, opened 2020-03-08
Remove wand of enslave/para. Also, simplify wand of random.

sdynet:
There are too many wands in the game, so I'm going to organize it. The goal of cleanup is to unite the wands with similar roles into one. The immediate effect of damaging the enemy is already being faithfully work by the other wands. I will remove the hex wands, and at the same time leave only the non-destructive effects on wand of random effects.

[List of random effects that change]
add effect: enslavement
remove effects: throw flame, throw frost, iceblast, fireball, bolt of draining, venom bolt


sdynet:
One more thing. I changed the starting items that Artificer with random's wand(15), flame's wand(15) and iceblast's wand(5). This is a temporary measure, so if you have a better opinion, please change it. :)

theJollySin:
I only every find a use for three wand types: Acid, Clouds, and Paralyze.

In particular, Wand of Paralyze is a major crutch in bad situations for most Stab-centric builds.

By preference, I absolutely LOVE the Paralyze wand, but "Wand of Flame" could go.  The damage is hardly ever high enough that I want to use it.  Until you get very high evocations, and untrained slingshot is significantly higher damage, as are untrained javelins and hand crossbows.

sdynet:
me too. I don't think flame's wand is a funny wand. I only use it to generate smoke.

ByrelMitchell:
Wand of flame is very strong in the early game, and as a steam generator through the midgame. It absolutely wrecks electric eels, and is useful through shoals/swamp.

Maybe there's some other reason to remove it, but I don't think it's at all useless.

ebering (changes requested):
I like the concept of the random effects change: consolidate it to a random buff or debuff, and make it the only hex wand. Things that need to be done still:

- [x] Random effects is no longer evil, so this should be updated in the appropriate place
- [ ] Random effects is now chaotic, which also needs updating (so Zin hates it but the other good gods don't)
- [ ] The effect list is now coherent enough that the description can be updated to say something about what will happen. You can either list all of the effects or just say it will "debuff or buff the target".

I'm less convinced that iceblast is the appropriate replacement for the artificer background. With the change to Random Effects, the artificer now gets a debilitating wand and a damage wand to start. It is probably appropriate to just remove the item outright and increase the wand charges to 20 each for the two starting wands.

Ash-R:
Garbage. Evocables gutted again. Do you just not train hexes or play chars that make heavy use of wands? I can't believe you want to get rid of enslave. The proposed change to random effects is completely against it's very design and I wonder if you realise that...
People seem to miss the entire point of wand of flame too.

ebering:
@Ash-R I'm gonna ask if you comment on PRs you refrain from calling the suggestions of other community contributors "garbage" and being demeaning and hostile in your feedback. You can say "I don't like this, I disagree with this direction, I think this fails to meet the goals" without being insulting or callous.

ebering:
We've hit feature freeze for 0.25 but I am still interested in the requested changes in some form. They won't get merged until after the 0.25 tournament (if they reach a good state), so I've added the 0.26 milestone. Feel free to continue working in the interim time, and don't hesitate to drop by ##crawl-dev if you have questions!

ebering:
@sdynet 0.26 trunk development is in full swing now, and looking at evocations is part of the plan. Do you plan to continue working on this PR?

sdynet:
Sure. Is there anything I can help you with?

ebering:
The [changes I requested in the review](https://github.com/crawl/crawl/pull/1310#pullrequestreview-404565936) are a good place to start.

It would be in a new pull request, but I would like to do something about wand of polymorph this cycle. Polymorph is cool and fun, but suffers because it's very opaque: when you use it its unclear what you'll get. I don't have any specific ideas yet, but if you have one I'd love to see it.

sdynet:
I see. Also, That's a pleasant challenge. I'll think about it.

sdynet:
New polymorph wand idea: Turn the target into a pig or a sheep forever(MR check). But with a non-low probability(33 - Evo skill%), the target turns into a more powerful monster(Same effect as current version).

I tried to change it to a function that even beginners can understand. What do you think?

ebering:
That's a step in the right direction. However, the old effect still has its problems; reducing its chance isn't enough.

One possibility is to have polymorph simply use the spell `SPELL_PORKOLATOR`, which is what Kirke uses, to always turn things into a pig.

I worry that a single monster is too narrow. Player polymorph can turn the player into a bat, a mushroom, a wisp, a pig, a tree or a dragon. Perhaps a similarly short list for monsters would work, scaling the available forms based on how strong the monster is?

alexjurkiewicz:
> scaling the available forms based on how strong the monster is?

One of the problems with poly was that it's hard to tell how strong the monster is, see the early deaths to Grinder the Hellion or Gastronok the shard strike. Maybe if poly used the same monsters always, but scaled their HP based on the source monster....

PR #1312, opened 2020-03-09
Added room for tactics to vaults in simple.des

ByrelMitchell:
This file is also intended as a tutorial of sorts for reading vaults, so I attempted to keep changes I made as syntactically simple as possible. This was most impactful in some of the large vaults (eg erik_arrival_triangle_huge) which needed the first few rooms marked no_monster_gen; I made the entire vault no_monster_gen rather than introduce a KMASK in such cases to retain their syntactic simplicity.

Vault list:
minmay_arrival_wavy_path: Single path, single exit vault. I added a second exit with a door.

matt_arrival_spirals: Two of the three starting locations could have you trapped by a randomly generated monster. Adding no_monster_gen allows the player to get to the main corridor, with multiple exits, prior to finding monsters.

lemuel_arrival_funnel: The island in the center of this vault had a 1:8 chance of being removed. In that case, this vauls was single-path, single-exit. I removed that possibility from the SUBST list.

minmay_arrival_pointy: Single path, single exit. I added a corridor between the points, which should allow you some tactical flexibility.

dpeg_arrival_radiant: This had a change to generate a single exit; I changed it to guaranteeing two. I also fixed what appeared to be a typo making the top-right path have two chances to be closed off, instead of one like the rest of the paths.

dpeg_arrival_tiny_i/j/k: These have the potential (or guaranteed for k) for the left end of the path to be blocked off by mapgen (making it single exit.) I added an escape hatch in case that happened.

dpeg_arrival_windows_large_door: Single-exit room. I added a door to one of the back corners (which corner will be randomized by vault-flipping.)

erik_arrival_triangle_small/medium/large/huge: All of these have the potential for a monster to be generated in one of the first 3 rooms, before you have any interconnectivity or options. I marked these vaults as no_monster_gen; possibly they should be moved out of this file and a KMASK applied to just the first 3 rooms.

minmay_arrival_plants_in_alcoves: Single path, single exit. I added an island made of plants/fungus/bushes/trees to the center.

noname_arrival_rooms_in_the_neighbourhood: This one was single-exit rooms, and was missing the SHUFFLE statement to make it possible to spawn in each room. I added a second exit to each room and the SHUFFLE statement, and moved it to small.des since the header was too long for simple.des.

onia_arrival_mini_spirals: This one had 4 paths with no interconnection and a single exit. I added a door connecting two of the paths and marked it as no_monster_gen to avoid a monster spawning in one of the initial spirals.

zelgadis_glass_arrival_small: This has guaranteed islands outside of the spiral. I added no_monster_gen to avoid monsters spawning and blocking the initial spiral.

dpeg_arrival_snail: Single path, single exit. I added a door to allow the player to generate a pillar in the center of the spiral while maintaining the current auto-explore spiral path.

ncdulo_halloftheorcs: Single exit room into the rest of the vault with monster_gen on. I added a second exit to the room (facing the same direction to give the player a similar first time experience.)

tarquinn_simple_flora_water_a: Single path, single exit. I removed the tree that kept you from walking around it.

tarquinn_simple_flora_water_c: Each side is single path, single exit unless pool_fixup creates a shallow water path across. I added a second exit to each side (and eliminated a meaningless A in the second subst.)

tarquinn_simple_flora_water_e: Single exit room. I added two more exits.

saegor_arrival_simple_d: Single exit room, and no handy way to add an exit without changing the aesthetics. I added an escape hatch instead.

scummos_arrival_gate: Single path into a single exit room. I added pillars of shallow water and lava to the sides of the path.

lightli_molten_gateway: Single exit room, and no handy way to add an exit without changing the aesthetics. I added an escape hatch instead.

mrwooster_arrival_statue_2/3: Open alcove design. I added a couple statues to the center to give the player something to work with tactically.

wheals_arrival_linesprint: Single path, single exit. I added a door connecting two parts of the 'line-sprint'.

wheals_arrival_shading: Room with a single exit. I added a couple islands.

lightli_arrival_flooded_house: Single path, single exit. I added a second exit.

giann_wishing_wells: This vault had a single exit with a plain antechamber; not much you could do aside from retreat to one of the sides. I added a couple more entrances to the antechamber.

psy_arrival_conway: This vault actually had two shallow water exits (one toward the top and one toward the bottom) which could be used for pretty interesting tactics. However, it's not clear without fully exploring which of the 4 alcoves are dead ends and which have a path, which makes this spoilery. I added a third path across which ensures that no matter which way you run, you'll find a path before you have to commit to an alcove.

psy_arrival_dodecagon: This vault had a 50% chance to be a single exit room. I added a second guaranteed exit.

psy_arrival_heroes: This was a single exit room. I added a couple more hero statues which could be used as pillars or chokepoints as needed.

wander_arrival_desolate: This vault is essentially just empty space. Odds are that this will be fine, since D:1 mapgen will probably put multiple entrances to it, but that's dependent on the current choice of map generator for D:1. Just in case, I added an escape hatch. I also applied the 'desolate' floor tiling to the features (arch, fountains, dungeon-exit, and escape hatch.)

wander_arrival_chamber: Single exit room. I added two more exits.

### DIFF
--- a/crawl-ref/source/dat/des/arrival/simple.des
+++ b/crawl-ref/source/dat/des/arrival/simple.des
@@ -60,14 +60,15 @@ ENDMAP
 NAME:    minmay_arrival_wavy_path
 TAGS:    arrival no_monster_gen
 ORIENT:  float
+NSUBST:  { = 1:{ / .
 MAP
 xxxxxxxxxxxxxxxxxxxxxx
-xx.{xxxxxxxxxxxxxxxxxx
-x....xx....xxxx....xxx
-x....x......xx......xx
+xxxxxxxxxxxxxxxxxxxxxx
+@..xxxx....xxxx....xxx
+@...xx......xx......xx
 xx..xx..xx..xx..xx..xx
-xx..xx..xx..xx..xx..xx
-xx..xx..xx..xx..xx..xx
+xx..xx..x.{{.x..xx..xx
+xx..xx..x.{{.x..xx..xx
 xx..xx..xx..xx..xx..xx
 xx......xx......xx...@
 xxx....xxxx....xxxx..@
@@ -77,7 +78,7 @@ ENDMAP
 
 ##############################################################################
 NAME:    matt_arrival_spirals
-TAGS:    arrival
+TAGS:    arrival no_monster_gen
 ORIENT:  float
 SHUFFLE: {[(
 SUBST:   [=., (=.
@@ -173,7 +174,7 @@ ENDMAP
 NAME:   lemuel_arrival_funnel
 TAGS:   arrival no_monster_gen
 ORIENT: float
-SUBST:  c : ccccwlx.
+SUBST:  c : ccccwlx
 MAP
 xxxxxxxxxxxxxxx
 x......{......x
@@ -207,9 +208,9 @@ xx.xx                 xx.xx
  xx..xx             xx..xx
   x...xx           xx...x
   xx...xx         xx...xx
-   x....xx       xx....x
-   xx....xx     xx....xx
-    x.....xx   xx.....x
+   x....xxxxxxxxxxx....x
+   xx....xx.....xx....xx
+    x......xxxxx......x
     xx.....xx xx.....xx
      x......xxx......x
      xx......x......xx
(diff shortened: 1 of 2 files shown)


crawl-ref/source/dat/des/arrival/simple.des
```diff
@@ -2270,8 +2249,8 @@ MAP
    ..xx...........xx..
    ..xpx.........xpx..
    ..xppx.......xppx..
-   ..xxxxxxxqxxxxxxx..
-   .......xqqqx.......
+   ..xxxxxxx.xxxxxxx..
+   .......x.q.x.......
    ........xqx........
```
ebering:
```suggestion
   .......x...x.......
   ........x.x........
```

Go ahead and make the second guaranteed exit symmetric.

ByrelMitchell:
> `mrwooster_arrival_statue` looks like it could also be problematic depending on how the level generator connects the entrance.
> 
> This looks overall good, I have a few small tweaks. Thanks for reading through the arrival vaults.

Alright, I fixed those up. I shifted the statues in the first mrwooster_arrival_statue out a tile so they can serve as one-tile pillars (since that vault is no_monster_gen, that doesn't introduce the potential for monsters to be generated behind the statues.)  The rest I just followed your requests.

PR #1313, opened 2020-03-13
Apply AC to Ramparts damage before monster speed effect (Flugkiller)

ByrelMitchell:
Frozen ramparts does damage after every monster action. Previously, it would compensate for the length of the action and then apply AC. That effectively increased the AC of fast monsters and reduced the AC of slow monsters. (Juggernauts, for instance, take almost no damage when moving due to their considerable AC, but would take a big chunk after an attack action (which is 30 auts and still only had AC applied once.)

This changes the formula so that the AC is applied before the monster speed normalization.


ByrelMitchell:
Irrelevant now that ramparts doesn't check AC

PR #1316, opened 2020-03-14
A tweak to reactivate layout_cellular_growth

Skrybe:
layout_cellular_growth was disabled way back in 0.15 with a note that it sometimes takes a long time to finish generating a level.  I've bumped up the minimum size in grid_opts from 2 to 3, which noticeably speeds up level generation.  A better solution would be to optimize the layout generator so that using smaller values won't slow things down, but at least this way the layout can be utilized, rather than sitting totally unused.


ebering:
Thanks for the patch and testing parameters to get ones that run quickly. The todo mentions an optimization that would be able to run this layout quickly, if you're up to it that would be a great contribution!

PR #1321, opened 2020-03-19
British Spelling Fixes

klorpa:


rawlins:
Our spelling convention is actually supposed to be Australian, not British, because Linley Hanzell is Australian. There are only a few minor differences (and I'm not really an expert), the only obvious one I see is that `labor` is ok I think.

Also, I have to mention, we don't enforce these in code (i.e. it is not part of the [coding conventions](https://github.com/crawl/crawl/blob/master/crawl-ref/docs/develop/coding_conventions.txt)) and I'm not personally going to start :-). Doesn't seem like that should be a bar against merging a patch like this though.

ebering:
As a general workflow we don't use merge commits, a devteam member will rebase your patch when we're ready for it. We also prefer that commit titles be short phrases instead of camel case.

ebering:
How would you like to appear in CREDITS.txt? I've listed you as klorpa for now, please let me know if you'd prefer something else.

PR #1323, opened 2020-03-20
Replace trap tile for magic-mapped traps (12170)

Goratrix:
After the trap rework, magic-mapped mechanical traps (regardless
of actual trap type) show as Arrow Trap tile. This replaces
the arrow trap tile with a new 'unknown trap' tile. The tile is
the current Pressure Plate tile overlaid with the questionmark
taken from the unknown portal tile.


Goratrix:
oops, the tile lost transparency for some reason when I saved it, will update.

Goratrix:
And yes, I do realize this doesn't fix the underlying issue, because the true trap type is still visible in x-v, so the traps aren't really 'unknown'. I suppose the only real fix would be to split DNGN_TRAP_MECHANICAL into separate traps, but I'm sure that's not worth it, due to how little these traps are used in the game.

Alternatively, could it be possible to somehow extempt these traps from the magic map revealing, so that they would only show as 'unknown mechanical trap' also in x-v?

aidanholm:
Sorry for the slow response on this one. This is something I'd rather have actually fixed, rather than papering over the issue. The issue you mention is one of several; because all traps are under `DNGN_TRAP_MECHANICAL`, you can't lookup mechanical traps, even though they have differing mechanisms, e.g. net traps, and even though the descriptions are actually already there!

The tile is great though; I'll plonk that into the unused tiles folder and maybe we'll find a use for it in the future.

Goratrix:
Yes, feel free to close this PR and use the tile elsewhere, no problem :)

aidanholm:
Merged fix for the original issue at 0.25-a0-824-gc5703f2, and your tile in at 0.25-a0-828-ga6db794; thanks again!

PR #1325, opened 2020-03-21
Allow autotravel to shafts (12196)

nlavsky:
Also, fix autotravel trying to path around shafts and ignoring shorter routes.


aidanholm:
thanks for the fix :)

PR #1326, opened 2020-03-21
Add zombie hornet tile

Goratrix:


aidanholm:
thanks, merged at 7a61802294

PR #1328, opened 2020-03-22
Correct meliai and hornet spectral and simulacrum

Goratrix:
To properly derive the bee spectral and simulacrum tiles
from the new meliai and hornet zombies.


Goratrix:
Just a note on why this is needed: the spectrals and simulacra are not derived directly from the monster, but via the zombie list (note [here](https://github.com/crawl/crawl/blob/master/crawl-ref/source/tilepick.cc#L1215)). Previously, bees and hornets and meliai all derived into bee zombie, which derived into bee simulacrum and spectral. After 4257648 and 7a61802, both of them have their own zombie, but because the zombies were not added to the spectral and simulacra lists, hornets and meliai now result in the generic small spectral and simulacrum (both are humanoid shaped) instead of the correct bee.

aidanholm:
thanks for the fix!

PR #1329, opened 2020-03-23
Add nightstalker to kobolds.

rjcrjc:
Kobolds are generally quite a bland species that overlap a lot of with existing species, but they haven't been removed yet, presumably in part for flavor reasons. Let's give them a reason to exist that makes them potentially more interesting to play. I've seen a lot of support for adding Nightstalker to kobold so I decided to implement this.

Some of the support I've seen for this idea also suggests removing nightstalker from demonspawn, but I haven't done that at the moment. I don't have any replacement mutation ideas so for now I'm leaving nightstalker on demonspawn as well.

Notes:
* Species description for non-english left unchanged.


ebering:
This is a good kicking off point for a conversation. Since species is such an impactful part of crawl, any major change tends to sit for a bit until the devteam has time to ponder it. There's a definite desire to do something about Kobolds, and this is an option. We've hit feature freeze for 0.25 but I'll give this a hard look in 0.26.

If this is a direction we decide to pursue this PR will need some changes:
- Remove nightstalker from Demonspawn, its ok that it doesn't have a replacement
- Remove the outdated non-English descriptions. A correct description in a foreign language is better than an incorrect one.

crawl-ref/source/dat/species/kobold.yaml
```diff
@@ -33,6 +33,11 @@ levelup_stat_frequency: 5
 mutations:
   1:
     MUT_CARNIVOROUS: 1
+    MUT_NIGHTSTALKER: 1
```
ebering, line 36: Having night stalker ramp up tilts the power balance to the lategame which is already true of so much of crawl. I would be interested in trying this with night stalker 3 from XL 1, or maybe the ramp up being much shorter.
```suggestion
    MUT_NIGHTSTALKER: 3
 ```
```

crawl-ref/source/dat/species/kobold.yaml
```diff
@@ -33,6 +33,11 @@ levelup_stat_frequency: 5
...
+  7:
+    MUT_NIGHTSTALKER: 1
+  13:
+    MUT_NIGHTSTALKER: 1
```
ebering, line 40:
```suggestion
```

crawl-ref/source/dat/descript/species.txt
```diff
@@ -79,7 +79,7 @@ quite fragile, but they learn to fly swiftly later on.
 %%%%
 Kobold
 
-Kobolds are small, mischievous and a little fragile, but multi-talented.
+Kobolds are small, mischievous and a little fragile, but multi-talented. Shadows envelop them as they grow in power.
```
ebering, line 82:
```suggestion
Kobolds are small, enveloped in shadow and a little fragile, but multi-talented.
```

ebering:
@rjcrjc Do you plan to pick this PR back up during the 0.26 season or should I include making the changes in my planning/time allotment? It would be very helpful if you could continue its development!

ByrelMitchell:
If @rjcrjc isn't interested in continuing this PR, I could pick it up. It's interesting to me.

ebering:
@ByrelMitchell If rjcrjc doesn't resurface in a week or so feel free to take this up!

rjcrjc:
Sorry for kind of disappearing, I got distracted with work and had been meaning to do this but never got around to finishing it up. I should have some time tonight or Thursday to handle these changes since what's left is pretty minor.

RojjaCebolla:
the mut gives some free Stealth+ as a side effect, right?  seems like that should be trimmed out now, reduced LOS is plenty powerful, the species size factor boosts stealth, and they have a good apt already

also thanks for making this PR!  would love kobolds to be more interesting

rjcrjc:
> the mut gives some free Stealth+ as a side effect, right? seems like that should be trimmed out now, reduced LOS is plenty powerful, the species size factor boosts stealth, and they have a good apt already
This was part of my original motivation of giving levels of nightstalker on level up. Playing around on D1-D2 with nightstalker 3 is definitely pretty excessive (heck, a brigand start has 6 pips of stealth on D1). 

I think reducing it to a single pip of stealth is pretty reasonable though, which matches octopode D1 stealth.

ebering:
Thank you for the updated patch! This will probably get some further discussion in ##crawl-dev, and I'll see if we can get an experimental set up for play-testing, but that might take a few days.

PleasingFungus:
Note: the high crossbows apt is quite antisynergistic with Nightstalker, since you don't have many turns to shoot at melee enemies with impunity. I don't think that's a problem, though - if anything, I'd be interested in raising all ranged weapons' apts further, specifically because they play differently with nightstalker.

RojjaCebolla:
> Note: the high crossbows apt is quite antisynergistic with Nightstalker, since you don't have many turns to shoot at melee enemies with impunity. I don't think that's a problem, though - if anything, I'd be interested in raising all ranged weapons' apts further, specifically because they play differently with nightstalker.

newbies often use apts as a signal of how to play the species, though...

PleasingFungus:
True! I don't think it's necessarily a bad thing to push newbies toward ranged weapons here, though - they can learn in play what the downsides of not having higher LOS with ranged is (if they don't realize immediately), and having high apts means that they won't actually bad that bad at it! Does that seem reasonable to you?

PleasingFungus:
Recommended backgrounds are also helpful here.

ByrelMitchell:
Frankly, crossbows will still be plenty strong, so if newbies decide to play them that way they'll get a unique and strong experience. 👍

ebering:
The experimental has been running for a bit, and I haven't seen any major concerns. NS3 from the start is strong, but with normal move speed and Kobold apts it doesn't seem outrageous.

I agree with the "good ranged apts aren't an issue here" camp, for the same reason: they play differently with night stalker. No more getting your few crossbow shots in before you're in range of a nasty short range spell and all that.

Thank you for the work on this patch! It's merge time.

PR #1332, opened 2020-03-30
Check holiness before casting dispel undead

ufshaikh:
Commit 3b7ae84e64 removed SPELL_DISPEL_UNDEAD from monsters and replaced
it with SPELL_DISPEL_UNDEAD_RANGE, but forgot to update the case in
_ms_waste_of_time accordingly. This resulted in monsters attempting to
cast dispel undead on non-undead creatures; before they never thought
that was worth trying. This commit restores the check.


aidanholm:
nice find! thanks for the fix

PR #1335, opened 2020-04-01
Update god-abil.cc

### LINKED ISSUE #1333: Does Vehumet support/gift Absolute Zero?
I think they should, but they currently don't.

AndrewGYork:
Add "Absolute Zero" to the list of spells Vehumet supports (and therefore, can gift, right?). If correct, this should close #1333


gammafunk:
Thanks, I had this fix in a local branch when I was play testing, but forgot to push it to trunk.

PR #1336, opened 2020-04-01
Fix stray commas in chardump notes for dungeon features

nlavsky:
After c225cfcd, notes for dungeon features end in a colon instead
of a full stop.

For example, "60174 | Depths:2 | Found a sacrificial altar of Ru,"


gammafunk:
Thanks for the fix.

PR #1337, opened 2020-04-02
Fix stash search menu not showing tiles for some dungeon features

nlavsky:
This PR fixes the stash search menu no longer displaying tiles for some dungeon features:
searching_for_hat
Also, it fixes the ?/F lookup-help.


aidanholm:
thanks for the fix!

PR #1342, opened 2020-04-06
Allows followers of Qazlal to walk on lava

petercordia:
Walking on lava has an epic feel to it, and fits thematically with
Qazlal. Followers of Qazlal get the ability at *** piety, and it
functions identically to Beogh's water walking. Nothing has been
removed from Qazlal as I don't anticipate the ability being
particularly useful.

This commit also fixes a logic bugg in player_likes_water in player.cc, which formerly only recognised merfolk as permanently liking water when they were in iceform.
I know that that should ideally be done in a separate commit, but I couldn't figure out how to make 2 different commits and I didn't want to leave the bug in the code.

### DIFF
--- a/crawl-ref/source/dat/descript/gods.txt
+++ b/crawl-ref/source/dat/descript/gods.txt
@@ -375,8 +375,8 @@ blocks attacks, and (for the particularly devout) deflects incoming
 projectiles. Qazlal allows followers to incite nature against their foes,
 causing a localised natural disturbance or, for particularly devout followers,
 a more widespread disaster. Followers of Qazlal eventually gain temporary
-resistances after taking damage, and can give life to clouds, turning them into
-allied elementals.
+resistances after taking damage, learn to walk on lava, and can give life to
+clouds, turning them into allied elementals.
 %%%%
 Ru powers
 
--- a/crawl-ref/source/god-passive.cc
+++ b/crawl-ref/source/god-passive.cc
@@ -368,6 +368,7 @@ static const vector<god_passive> god_passives[] =
         {  0, passive_t::cloud_immunity, "and your divine allies are ADV immune to clouds" },
         {  1, passive_t::storm_shield,
               "generate elemental clouds to protect yourself" },
+        {  3, passive_t::lava_walk, "walk on lava" },
         {  4, passive_t::upgraded_storm_shield,
               "Your chances to be struck by projectiles are NOW reduced" },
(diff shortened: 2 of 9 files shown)


petercordia:
I should add, this is my first pull request.
I'd like to learn how everything works so I can work on other projects in the future.

petercordia:
I just found a mistake which crept in whilst merging with master, (the wrong line got deleted,) and I don't know how to fix it

RojjaCebolla:
I assume the intent is to help the player walk through their own Disaster Area?  Temporarily causing the player to fly when they use the ability, might help with keeping lava in vaults as a seriously difficult terrain.

petercordia:
No, as I wrote in the commit and pull request, the main intent is to give players some way to lava-walk. Because it is thematically cool.
Qazlal is the most thematic god to give the ability to, since Qazlal and Vehumet are the gods of elemental destruction and of those 2 Vehumet is more interested in spells.
Not getting boxed in by your own Disaster Area is a happy side effect.

I'm running a poll in the tavern, and there seems to be some support for this change: https://crawl.develz.org/tavern/viewtopic.php?f=8&t=26849&p=343882#p343882

petercordia:
Whilst playtesting I occasionally got the message "Infinite lua loop detected, aborting."
I was in wiz-mode with time frozen (&E) and killing stuff with an axe and Disaster Area.
I don't know whether this is a known 'bug' or something somehow induced by my code change.

ebering:
Welcome to contributing to crawl! I've seen you around `##crawl-dev`, which is the central place where the devteam discusses development.

When it comes to gameplay changes, the justifications you've provided are ones that the devteam isn't particularly moved by. We don't use "it would be cool" or player polls as criteria. We look at gameplay benefit based on how it changes play.

In this case, @RojjaCebolla points out that the gameplay benefit of this change is to allow a Qaz follower to pass through their own Disaster Area. 

Wizmode freeze time does strange things to crawl, I wouldn't worry about that.

ebering:
Following up on my previous comment, the gameplay benefit of allowing Q followers to pass through their own Disaster Area must be weighed against: the codebase penalty of increased complexity and the possible tedium issues with cheezing lava using lava walk. The latter is already a problem with player flight (which is plentiful) and I don't think adding more sources of this problem is the right way to go.

If DA really is too weak with its occasional movement restrictions, RojjaCebolla 's suggested fix is the better way to go, but I'm not convinced of that.

PR #1343, opened 2020-04-08
make merfolk treat water the same way as ground for pathfinding purpo…

petercordia:
Formerly, merfolk would avoid travelling through water because 'their boots would meld'. In water merfolk gain bonus EV and speed (equivalent to 4 boots of running), so this was a bit silly.

a poll showed support to even make merfolk prefer travelling through water, but I don't think I can do so without introducing bugs.
The forum discussion (and poll) is here: https://crawl.develz.org/tavern/viewtopic.php?f=8&t=26848&p=343894#p343894

This pull request might conflict with my other pull request (https://github.com/crawl/crawl/pull/1342) because I've fixed the same bug in both. (Hopefully in the same manner)

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -610,9 +610,9 @@ bool player_in_connected_branch()
 
 bool player_likes_water(bool permanently)
 {
-    return !permanently && you.can_water_walk()
-           || (species_likes_water(you.species) || !permanently)
-               && form_likes_water();
+    return species_likes_water(you.species)
+           || !permanently && you.can_water_walk()
+           || !permanently && form_likes_water();
 }
 
 /**
--- a/crawl-ref/source/travel.cc
+++ b/crawl-ref/source/travel.cc
@@ -214,11 +214,8 @@ static inline bool _is_safe_cloud(const coord_def& c)
 static inline int _feature_traverse_cost(dungeon_feature_type feature)
 {
     if (feat_is_closed_door(feature)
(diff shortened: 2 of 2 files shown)


ebering:
Thanks for the patch! I've just merged it with some tweaks, the reasons follow. But most importantly: how would you like to appear in the CREDITS.txt?

Some style remarks: for commit messages it's preferred that the message title start with a capital letter, and that the commit body's explanation (if the title was not sufficient) cover all the bases. In this case, you should mention the `player_likes_water` change.

However, there's no bug in the function's logic. The function asks if the player *currently* likes water, the parameter `permanently` asks if this will change absent player input (e.g. form expiration or piety decay). However, if a merfolk is in a form that can't swim, they don't like water.

petercordia:
> Thanks for the patch! I've just merged it with some tweaks, the reasons follow. But most importantly: how would you like to appear in the CREDITS.txt?
> 
> Some style remarks: for commit messages it's preferred that the message title start with a capital letter, and that the commit body's explanation (if the title was not sufficient) cover all the bases. In this case, you should mention the `player_likes_water` change.
> 
> However, there's no bug in the function's logic. The function asks if the player _currently_ likes water, the parameter `permanently` asks if this will change absent player input (e.g. form expiration or piety decay). However, if a merfolk is in a form that can't swim, they don't like water.

You can list me in the credits as Peter Gerlagh

Glad I could help, even if I didn't understand 'player_likes_water' :)

PR #1348, opened 2020-04-12
Always save macros

alexjurkiewicz:
We don't need to ask the player this. It's a simpler mental model for
macro changes to be always saved, rather than sometimes asking the
player (during normal session end, but not eg webtiles tab close /
dgamelaunch hangup).


RojjaCebolla:
this is especially nice when, eg, mashing Esc at the end of a game to get through all the prompts/windows, and accidentally not saving your Macros

PR #1351, opened 2020-04-12
Send DCSS processes SIGABRT rather than SIGTERM

alexjurkiewicz:
Both signals do the same thing now (controlled crash), but this is
hopefully the first step in changing SIGTERM to handle the same way as
SIGHUP (cleanly shut down).

It was a bit off a toss-up between using SIGQUIT and SIGABRT. QUIT means
"quit program" while ABRT means "abort program", so both are probably
suitable. But dgamelaunch-config uses ABRT already:
https://github.com/crawl/dgamelaunch-config/blob/master/utils/webtiles
so we'll follow convention.

### DIFF
--- a/crawl-ref/source/webserver/process_handler.py
+++ b/crawl-ref/source/webserver/process_handler.py
@@ -487,7 +487,7 @@ class CrawlProcessHandlerBase(object):
     def kill(self):
         if self.process:
             self.logger.info("Killing crawl process after SIGHUP did nothing.")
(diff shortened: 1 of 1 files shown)


rawlins:
> It was a bit off a toss-up between using SIGQUIT and SIGABRT. 

crawl is ignoring SIGQUIT entirely so that's another reason to use SIGABRT for now. SIGQUIT might be better eventually but it doesn't seem like a problem to merge this so as to get the migration away from SIGTERM going.

PR #1352, opened 2020-04-13
New Hep+Dith shared overflow altar

riverfiend:
I wrote a vault! It places a hep altar and a dith altar. In front of the altar to hep are clouds of mist with a player ancestor locked up, in front of the altar to dith are clouds of dark smoke with a shadow locked up.


ebering:
Thanks for the vault! How would you like to appear in the CREDITS.txt?

riverfiend:
> Thanks for the vault! How would you like to appear in the CREDITS.txt?
Just as riverfiend is fine, thanks!

PR #1354, opened 2020-04-14
Fix the x-mode description of walls covered in ice and blood

nlavsky:
Currently, if icy walls made by Frozen Ramparts are covered in blood,
then they have only the second part of a covering description:
for example, "A rock wall, and spattered with blood."


aidanholm:
Thanks for the fix!

aidanholm:
Thanks for the fix!

PR #1356, opened 2020-04-17
Free the crawlies

nlavsky:
Currently, `nicolae_crawler_farm` has a few demonic crawlers trapped behind lava,
because they can no longer cling to walls.


ebering:
Thanks for the fix!

PR #1359, opened 2020-04-17
Remove the Deflect Missiles spell

emilazy:
See 4c9e86bff6355764059e0a1ba96d981712c12a01 for rationale.


ebering:
Thanks for the submission. This looks good as a gameplay change; however the PR needs some save compatibility code to remove Deflect Missiles from existing games. The save compatibility code is in `tags.cc`, for a recent example see 0889b22a 688896f and for handling player ghosts: 618c0ec23 d0e028f1 c8cd6a5d6a4.

Detailed documentation on the save compatibility is in [docs/develop/save_compatibility.txt](https://github.com/crawl/crawl/blob/master/crawl-ref/docs/develop/save_compatibility.txt) in the repository.

emilazy:
The spell is added to both [`removed_spells`](https://github.com/crawl/crawl/pull/1359/commits/4c9e86bff6355764059e0a1ba96d981712c12a01#diff-1640204164db01d6db0d8af416d0f8e9R2335) and [`_ms_waste_of_time`](https://github.com/crawl/crawl/pull/1359/commits/4c9e86bff6355764059e0a1ba96d981712c12a01#diff-d26960db6504e49ee629fb63f2deb1a9R8028), which are the only save compatibility considerations I'm aware of for removing spells. It'd be possible to forcibly remove it from players and ghosts on load so that transferred saves just silently lose the spell rather than having a stub lying around; I'm not aware of this having been previously done for simple spell removals, but can implement it if it's desired.

emilazy:
Force-pushed to fix a typo (Ozocubu's Armor → Ozocubu's Armour).

ebering:
You're correct. Digging in the save compat code for the more complex changes of positional magic has poisoned my mind.

ebering:
Thanks for the patch!

PR #1361, opened 2020-04-18
Upload release builds to GitHub releases

alexjurkiewicz:
Whenever a GitHub release is created, GitHub Actions will add the
following builds:

* Windows (32-bit) console & tiles zips, plus combined install
* macOS console & tiles apps

Also make the following changes:

* Disable some unnecessary Clang builds
* Rename windows package Makefile targets
* Use advzip on macOS packages

PR #1365, opened 2020-04-18
WIndows installer package build cleanup

Goratrix:
1. The installer package build (unlike the zip build) currently
doesn't delete the build-win folder after it finishes
and the folder needs to be deleted manually.
There may be reason that I'm missing, but I can't think of any,
as the zip build deletes the folder, only the installer build doesn't.

2. The final win packages are actually named 'stone_soup...',
so update gitignore accordingly.

### DIFF
--- a/crawl-ref/.gitignore
+++ b/crawl-ref/.gitignore
@@ -186,7 +186,7 @@ makefile.dep
 /source/debian/files
 
 # Windows packaging
-/source/crawl-win
+/source/stone_soup
 
 # Windows thumbnail files
 Thumbs.db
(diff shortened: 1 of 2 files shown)


alexjurkiewicz:
Funny we are both looking at this area of the code. #1361 also touches the same code

Goratrix:
I see :) feel free to add this to your PR if it's not yet there. I just got annoyed by having to delete the stuff manually.

alexjurkiewicz:
Good idea! I've merged it in.

Goratrix:
Closing, has been merged into 1361. Thanks!

PR #1366, opened 2020-04-18
Remove the Controlled Blink spell

emilazy:
See f2075c271285faae80920bb9d334553625e42ce6 for rationale.


petercordia:
What is the appropriate place to discuss whether this would improve crawl?
CB is a strong utility spell, and those are part of what makes high-level magic fun (to me).

crawl-ref/source/book-data.h
```diff
@@ -192,8 +191,8 @@ static const vector<spell_type> spellbook_templates[] =
 {   // Book of the Warp
     SPELL_GRAVITAS,
     SPELL_PORTAL_PROJECTILE,
+    SPELL_GOLUBRIAS_PASSAGE,
```
ebering, line 194: Warp is fine as a 4 spell book in my opinion.
```suggestion
```

PR #1367, opened 2020-04-18
Build system portability improvements

emilazy:
This should obsolete [the current patch used in NixOS/nixpkgs](https://github.com/NixOS/nixpkgs/blob/09c4736405414ce33b08043f69d50171617c6116/pkgs/games/crawl/crawl_purify.patch) and allow Crawl to build out of the box on more systems.


ebering:
@rawlins made this comment in `##crawl-dev` on irc.freenode.net (our development channel): 

```
15:19:19 <advil> my opinion remains the same as last time someone tried this, pkg-config should not be enabled by default on mac
15:36:48 <advil> e.g. the pkg-config thing has come up before and there are reasons why it is that way
```

I haven't gone digging into the logs for his elaboration so I'll leave it to him to elaborate.

Have you considered joining us in `##crawl-dev`? If you're planning to continue to contribute large high-code-quality PRs (please do) it's a good place to check in so your effort isn't wasted. Continuing to interact via GitHub is fine but the central place where the devteam discusses is that irc channel.

aidanholm:
thanks for the pr! I've merged the first two sqlite-related commits. Re other commits;
- advil / @rawlins has plans for something related to pkg-config on mac
- `pkg-config --cflags-only-I sdl2` gives `-I/usr/include/SDL2` on arch for me, and all the SDL headers  are in that directory, so i'm not sure the sdl2-include-fixing commit is correct
- the font stuff is probably a net improvement but it looks like it'd break for users who are specifying `OUR_PROPORTIONAL_FONT`

rawlins:
This is now merged to the extent it's going to be.
* pkg-config stuff was handled in [61de542](https://github.com/crawl/crawl/commit/61de54256c8e86c3ee86a36efe9190598c8fee49), sqlite commits were merged a while ago.
* The fontconfig patch was wrong, because it didn't handle the case correctly where dejavu isn't installed at all (this should trigger using the contrib font, but instead would use Verdana). It also had some false premises -- this script was already searching fontconfig directories correctly via `fc-list`, the hardcoded directories are just a fallback. I have taken a look at the current nixpkgs patch and pulled the only necessary change from there (the `-L`). As a side-note, for anyone watching, I'm pretty sure the change in that patch to the hardcoded directories is unnecessary (or if it is, it's a sign of something else being misconfigured). See d6ef553a8bdd and the comments that that commit added.
* The SDL2 header changes may be "right" for mac under some configurations (and either way works for the contrib SDL2 if it's used), but unfortunately they are not general, so can't be merged. See https://discourse.libsdl.org/t/why-i-get-an-error-when-including-sdl-image/36207 for some recent discussion about this. Unless you go wholesale into the somewhat odd SDL2 config scripts, it does not seem that there's an easy platform-independent solution that we can adopt right now, so we will keep the version that works on most linux distros and continue to leave any needed changes to downstream packagers.

Issue #1364, opened 2020-04-18
Maybe break save compatibility sometime this decade


emilazy:
Save compatibility hasn't been broken [since 2012](https://github.com/crawl/crawl/blob/19abc6cbe401640f82ef30f66498c605bcd7f2b7/crawl-ref/docs/develop/history/major-tag-history.md); there are probably 8 year olds playing Crawl who have never seen a compatibility break. Currently, just running the `tag-major-upgrade` script removes ~10k lines from the source code:

```
emily@renko ~/s/c/c/source> util/tag-major-upgrade -t 35
…
emily@renko ~/s/c/c/source> git diff --unified=0 --minimal | wc -l
13268
```

[...]

PR #1368, opened 2020-04-19
TAG_MAJOR_VERSION 35 CI support

alexjurkiewicz:
Inspired by #1364.

- [x] Add util/tag-35-upgrade.py
    - [x] Update species files
    - [x] Update rltiles files
    - [x] Update DES files(?)
    - [x] other places to update TBC...
- [x] Add CI job to build TAG_MAJOR_VERSION=35


alexjurkiewicz:
This is looking good!

alexjurkiewicz:
You can see the three builds with TAG_MAJOR_VERSION 35 here:

https://github.com/crawl/crawl/pull/1368/checks?check_run_id=601270761
https://github.com/crawl/crawl/pull/1368/checks?check_run_id=601270772
https://github.com/crawl/crawl/pull/1368/checks?check_run_id=601270789

I jammed the next-tag logic into the existing CI matrix which makes it a little unwieldy but prevents having to duplicate build logic again. It might be cleaner to do it separately... open to thoughts

crawl-ref/source/util/tag-35-upgrade.py
```diff
@@ -0,0 +1,387 @@
...
+import os
+from typing import Sequence, Dict
+
+logging.basicConfig(level=logging.DEBUG)
+
+NEW_TAG = 35
+
+
+def _edit_file(
+    file_name: str, delete: Sequence[str] = [], replace: Dict[str, str] = {}
```
aidanholm, line 25: default parameters in python should not be mutable

crawl-ref/source/util/tag-35-upgrade.py
```diff
@@ -0,0 +1,387 @@
...
+):
```
aidanholm, line 26: This should probably have `-> None`

crawl-ref/source/util/tag-35-upgrade.py
```diff
@@ -0,0 +1,387 @@
...
+        "mummy",
+        "naga",
+        "octopode",
+        "ogre",
+        "spriggan",
+        "tengu",
+        "troll",
+        "vampire",
+        "vine-stalker",
+    ):
```
aidanholm, line 131: maybe use `os.walk`?

crawl-ref/source/util/tag-35-upgrade.py
```diff
@@ -0,0 +1,387 @@
...
+            "# Remove lamp when TAG_MAJOR_VERSION > 34\n",
+            "misc_lamp MISC_LAMP_OF_FIRE\n",
+            "misc_lamp_inert MISC_LAMP_OF_FIRE_INERT\n",
+            "# Remove stone when TAG_MAJOR_VERSION > 34\n",
+            "misc_stone MISC_STONE_OF_TREMORS\n",
+            "misc_stone_inert MISC_STONE_OF_TREMORS_INERT\n",
+            "# Remove when TAG_MAJOR_VERSION > 34\n",
+            "misc_lantern MISC_LANTERN_OF_SHADOWS\n",
+            "# Remove when TAG_MAJOR_VERSION > 34\n",
+            "misc_sack MISC_SACK_OF_SPIDERS\n",
```
aidanholm, line 303: I think the better approach might be to remove everything between `# REMOVE WHEN TAG_MAJOR_VERSION > 34 {` and `# }` line comments. Same for everything else that uses # comments, e.g. des files. These files don't support any sort of ifdef mechanism, so you shouldn't need to actually evaluate the condition; a textual match should suffice

crawl-ref/source/util/tag-35-upgrade.py
alexjurkiewicz: It lacks gnoll or any new species that will get added in future.

crawl-ref/source/util/tag-35-upgrade.py
alexjurkiewicz: That makes sense, but I'd rather not rewrite this all now unless you think it's really important :\

alexjurkiewicz:
Addressed review comments!

crawl-ref/source/util/tag-35-upgrade.py
alexjurkiewicz: I did this 😊

crawl-ref/source/util/tag-35-upgrade.py
alexjurkiewicz: I did it! The comment format is standardised now

crawl-ref/source/util/art-data.pl
```diff
@@ -533,6 +534,19 @@ sub process_line
...
+#     "{", "BRAND", "AC", "EV", "STR", "INT", "DEX", "\n",
+#     "FIRE", "COLD", "ELEC", "POISON", "LIFE", "MAGIC", "\n",
+#     "SEEINV", "INV", "FLY", "FOG", "BLINK", "BERSERK",  "NOISES", "\n",
+#     "NOSPELL", "RND_TELE", "NOTELEP", "ANGRY", "\n",
+#     "MUTATE", "SLAY", "CURSE", "STEALTH", "MP", "\n",
+#     "BASE_DELAY", "HP", "CLARITY", "BASE_ACC", "BASE_DAM", "\n",
+#     "RMSL", "REGEN", "NO_UPGRADE", "RCORR", "\n",
+#     "RMUT", "CORRODE", "DRAIN", "SLOW", "FRAGILE", "\n",
+#     "SH", "HARM", "\n",
+#     "}",
```
aidanholm, line 548: Is commenting out this range intentional?

alexjurkiewicz:
The tag 35 specific stuff is commented out, otherwise it would be active in
current code. The Python upgrade script takes care of uncommenting it.

aidanholm:
merged, thanks again! (manually becauses i wanted to fix a python nit)

PR #1369, opened 2020-04-19
Change level annotations from Dungeon Overview (Ctrl-O) - New annotate

petercordia:
this fork changes the way annotate works to allow any dungeon floor to be annotated from anywhere via "shift-O !"
It is still under development, ie still has issues which need to be resolved before it can be merged in.
But I thought it might still be useful to make it publicly available.

It does compile, and doesn't crash, as far as I know.


ebering:
FYI, you can mark a pull request as a "draft" in GitHub.

If you have specific issues and point them out in the comments other contributors and developers can suggest fixes.

petercordia:
All of the issues 8 I'm aware of are labelled with XXX or TODO

The ones which I will find hardest to solve myself:
- [ ] convert level_id into string (lines 124-126 on travel.cc)
- [ ] either move more functions into annotate.cc, or move everything into dgn-overview.cc, so that everything dealing with level_annotations can be together

crawl-ref/source/annotate.h
```diff
@@ -0,0 +1,17 @@
...
+ * @brief Annotation functionality
+**/
+
+#pragma once
+
+#include <vector>
+
+void annotate_level();
+void do_annotate(level_id li);
+//static int _prompt_annotate_branch(); // don't need need this?
```
aidanholm, line 12: You don't (usually) want to declare static functions in header files, because they are only usable from the `cc` file in which they are defined. This is usually intentional; if you want code from other files to be able to call a function, it shouldn't be static.

crawl-ref/source/dgn-overview.h
```diff
@@ -37,12 +37,15 @@ string get_level_annotation(level_id li = level_id::current(),
...
 
 class writer;
 class reader;
 void marshallUniqueAnnotations(writer& outf);
 void unmarshallUniqueAnnotations(reader& inf);
 bool connected_branch_can_exist(branch_type br);
+
+// XXX I botched this in to get crawl to compile, but don't actually understand
+// on a deep level what it does.
+typedef map<level_id, string> annotation_map_type;
```
aidanholm, line 50: This line defines a type called `annotation_map_type`, which will be available to every file that includes `dgn-overview.h`. Currently only `dgn-overview.cc` uses this type, so adding this line is not a good idea; ideally, we'd like to keep the data structure used to store annotations (the `map<level_id, string>` bit, local to the individual file, not every file that includes the corresponding header.

The next line declares a variable called `level_annotations` located "somewhere else"; again, since this is in `dgn-overview.h`, this is accessible to every file that includes this header, and any file can use this global variable directly. Again, including this line is not a good idea because it exposes the internal data structure unnecessarily.

Instead, these definitions should go at the top of `annotate.cc`, where you've added them commented out.

crawl-ref/source/annotate.cc
```diff
@@ -0,0 +1,240 @@
...
+#include "traps.h"
+#include "travel.h"
+#include "unicode.h"
+#include "unwind.h"
+#include "view.h"
+
+// XXX The same 2 lines occur in gdn-overview.cc, so that this variable-type
+// is defined twice
+//typedef map<level_id, string> annotation_map_type;
+//annotation_map_type level_annotations;
```
aidanholm, line 66: These should be uncommented, and you should add `extern` to the `level_annotations` line. Then you can remove the stuff in `dgn-overview.h`.

I'm guessing that you aded the stuff in `dgn-overview.h` because this code here failed with a linker error, e.g. "duplicate symbol: level_annotations". That's because the `annotations_map_type level_annotations` line in two different files will declare two different variables with the same name. If you add `extern` to one line, that line will then reference the variable that is now defined in only one place.

crawl-ref/source/annotate.cc
```diff
@@ -0,0 +1,240 @@
...
+            {
+                if (toupper_safe(keyin) == branches[br].travel_shortcut)
+                    // It is possible to annotate a not-generated branch.
+                    return br;
+            }
+            // Otherwise cancel
+            return ID_CANCEL;
+        }
+    }
+}
```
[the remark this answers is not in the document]
petercordia: I belief it is impossible to exit the switch statement without going for one of the cases, because I included "default". Though I just noticed that _case control(P)_ has the possibility of ending with **break** rather than **return**, so I do indeed need to fix something.

petercordia:
I think this is ready for implementation now, though if I understand it correctly I should merge the 3 pull requests and deal with conflicts with Master

petercordia:
there are 2 remaining issues:
there's
`    // XXX I'd be grateful if someone could refactor this into something that's
    // not black magic. The functionality doesn't have the be the exact same,
    // so long as the list is of reasonable length.
    // these 3 lines make a vector containing all shown branches.
    // the current implementation shows all known branches.
    vector<branch_type> brs;
    for (branch_iterator it; it; ++it)
        if (is_known_branch_id((**it).id))
            brs.push_back(it->id);`
where there's some black magic which works but I don't expect everyone reading it could figure out how it works.

and there's the help message, which I've implemented as 
`    if (branch == ID_HELP) // XXX: this help message could be improved.
    {
        clear_messages();
        mpr("press the letter corresponding to the branch of the level you want to annotate, or");
        mpr("? for help, . or enter to annotate the level you're currently in, < and > to annotate");
        mpr("the levels upstairs and downstairs respectively, and control-P for the parent branch");
        do_annotate();
        return;
    }`
which is a bit ad-hoc

aidanholm:
For the first one, I suggest moving `_get_branches` from travel.cc to branch.cc, and using that. You'll need to rename it to `get_branches` (remove the leading underscore), and add a corresponding declaration for the function in branch.h. This would be best done as a separate commit that comes before your changes adding new annotation functionality. If you really feel like going the extra mile, adding a test case for `get_branches` would be awesome. Feel free to jump on ##crawl-dev or comment here if you get stuck with any of that.

Re help message, that should be implemented as a popup, similar to interlevel travel (G).

crawl-ref/source/dgn-overview.cc
```diff
@@ -1094,3 +1064,164 @@ bool connected_branch_can_exist(branch_type br)
...
+        annotate_level(lid);
+        return;
+    }
+
+    if (branch == ID_DOWN)
+    {
+        lid = find_down_level(lid);
+        annotate_level(lid);
+        return;
+    }
```
aidanholm, line 1181: having these keybinds isn't very useful because 1) they only work before you've selected a branch, unlike in interlevel travel, where they only work *after* you've selected a branch, and 2) you can't press > and < multiple times 

the travel code uses _travel_depth_keyfilter() to accomplish this

crawl-ref/source/dgn-overview.cc
```diff
@@ -1094,3 +1064,164 @@ bool connected_branch_can_exist(branch_type br)
...
+            // Otherwise cancel
+            return ID_CANCEL;
+        }
+    }
+}
+
+void do_annotate()
+{
+    level_id lid  = level_id::current();
+    int branch = _prompt_annotate_branch(lid);
```
aidanholm, line 1144: these should be const if possible

crawl-ref/source/dgn-overview.cc
```diff
@@ -1094,3 +1064,164 @@ bool connected_branch_can_exist(branch_type br)
...
+    int depth;
```
aidanholm, line 1145: `depth` shouldn't be declared until it's ready to be used

crawl-ref/source/dgn-overview.cc
```diff
@@ -1094,3 +1064,164 @@ bool connected_branch_can_exist(branch_type br)
...
+        return;
+    }
+
+    int max_depth = branches[branch].numlevels;
+    // Handle one-level branches by not prompting.
+    if (max_depth==1)
+        depth = 1;
+
+    clear_messages();
+    const string prompt = make_stringf ("What level of %s? ", branches[branch].longname);
```
aidanholm, line 1189: no whitespace after `make_stringf`

crawl-ref/source/dgn-overview.cc
```diff
@@ -1094,3 +1064,164 @@ bool connected_branch_can_exist(branch_type br)
...
+    depth = prompt_for_quantity(prompt.c_str());
+    if (depth > 0 && depth <= max_depth)
+    {
+        branch_type br = branch_type(branch);
+        annotate_level(level_id(br,depth));
```
aidanholm, line 1194: space after comma

crawl-ref/source/dgn-overview.cc
```diff
@@ -1094,3 +1064,164 @@ bool connected_branch_can_exist(branch_type br)
...
+    }
+    else
+    {
+        mpr("that's not a valid depth");
```
aidanholm, line 1198: capitalize That's, and end with a full stop please

crawl-ref/source/dgn-overview.cc
```diff
@@ -1094,3 +1064,164 @@ bool connected_branch_can_exist(branch_type br)
 
     return true;
 }
+
+static int _prompt_annotate_branch(level_id lid)
+{
+    // XXX I'd be grateful if someone could refactor this into something that's
+    // not black magic. The functionality doesn't have the be the exact same,
+    // so long as the list is of reasonable length.
```
aidanholm, line 1072: which is the black magic bit? the branch_iterator stuff? or the prompt message printing?

crawl-ref/source/dgn-overview.cc
```diff
@@ -1094,3 +1064,164 @@ bool connected_branch_can_exist(branch_type br)
...
+    // these 3 lines make a vector containing all shown branches.
+    // the current implementation shows all known branches.
```
aidanholm, line 1074: the code for this is reasonbly readable, so i think this comment is probably not necessary

crawl-ref/source/dgn-overview.cc
petercordia: Apparently you understand what's happening.
I don't understand `branch_iterator`, `**it`, and `it->id`

crawl-ref/source/dgn-overview.cc
```diff
@@ -1094,3 +1064,164 @@ bool connected_branch_can_exist(branch_type br)
...
+    else
+    {
+        mpr("that's not a valid depth");
+        return;
+    }
+}
+
+void annotate_level(level_id li)
+{
+    string old = get_level_annotation(li, true, true);
```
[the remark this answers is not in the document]
petercordia: Will do, but what is the function?

crawl-ref/source/dgn-overview.cc
aidanholm: er, i think `(**it).id` and `it.id` are equivalent, so please change the former one to the latter.
`branch_iterator` is a helper class that runs through the list of all branches, `it` is an instance of this iterator that represents the current position in the list of all branches, and `it->id` gets you the branch id (`enum branch_type`) for that position.

crawl-ref/source/dgn-overview.cc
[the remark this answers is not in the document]
aidanholm: A lack of const means it's mutable. Generally we try to keep things const if they won't be modified, as it reduces cognitive load when reading.

crawl-ref/source/dgn-overview.cc
petercordia: I tried compiling with it.id instead of (**it).id, and it threw an error: `‘class branch_iterator’ has no member named ‘id’`

crawl-ref/source/dgn-overview.cc
petercordia: lid can't be const. I'll make branch constant since that's what you want

crawl-ref/source/dgn-overview.cc
petercordia: currently, when you're standing on a stair, you can annotate the other side with > , I think. There is a slightly obnoxious menu that you have to press through even if you want to annotate the current level, but overall I like this functionality. the < and > shortcuts are intended to keep the old functionality. This will also be helpful if anyone is attached to the current functionality.

I don't care much for control-p as I have implemented it. I just thought, maybe someone will find it useful. Probably it would be more useful if it just annotated the branch entrance (for tournament challenges etc).

crawl-ref/source/dgn-overview.cc
aidanholm: ah, sorry, meant to say `it->id` not `it.id`

petercordia:
one feature which could be improved: when you're annotating a floor via control-X, you look at your current position whilst writing the annotation, before going back to the place where you're supposed to be looking

there are a bunch of warnings whilst compiling, but I don't think any of those are my fault.

crawl-ref/source/dgn-overview.cc
petercordia: it works

crawl-ref/source/dgn-overview.cc
petercordia: I tried it, and we really do need that space

crawl-ref/source/dgn-overview.cc
```diff
@@ -1094,3 +1062,156 @@ bool connected_branch_can_exist(branch_type br)
...
+        depth = 1;
+    else
+    {
+        clear_messages();
+        const string prompt = make_stringf ("What level of %s? ", branches[branch].longname);
+        depth = prompt_for_quantity(prompt.c_str());
+    }
+    if (depth > 0 && depth <= max_depth)
+    {
+        branch_type br = branch_type(branch);
```
aidanholm, line 1186: super nit: this should probably be const as well

crawl-ref/source/dgn-overview.h
```diff
@@ -37,8 +37,8 @@ string get_level_annotation(level_id li = level_id::current(),
 
 bool level_annotation_has(string str, level_id li = level_id::current());
 
-void annotate_level();
-void do_annotate(level_id& li);
+void do_annotate();
+void annotate_level(level_id li);
```
aidanholm, line 41: Can we flip these? it'll make the diff more readable

petercordia: done

petercordia:
all comments dealt with afaik, and rebased to master

aidanholm:
two bugs I've found:
 - on d:1, pressing < will offer to annotate d:-1
 - choosing d then pressing backspace will result in an upside-down question mark in the input

petercordia:
I was actually aware of the first issue. Since annotating D:-1 does nothing I didn't see it as a problem. I understand it looks buggy, so I should fix it.
It took me a while to figure out, but the second issue also occurs in regular crawl, but only if you put something like `bindkey = [g] CMD_PICKUP_QUANTITY` in your rc file. (Currently most players will never trigger CMD_PICKUP_QUANTITY.) I'll see whether I can fix prompt_for_quantity to not have this issue.

petercordia:
I realise I could switch from prompt_for_quantity to prompt_for_int, which would probably make for better code because prompt_for_quantity actually calls prompt_for_int after assigning the 1st digit. However, I don't understand the code of prompt_for_int, which makes using it significantly more work.

aidanholm:
I've added an issue for the backspace bug at https://crawl.develz.org/mantis/view.php?id=12259; please revert the check for CK_BKSP, as it's incomplete, and i'd rather this be fixed properly.

Yeah, just using prompt_for_int in your code is better. nonneg should be true if you don't want negative numbers (this is the case), and so -1 will be returned on cancel / bad input.

petercordia:
Done. I've also rebased to master.
I have only done a little testing with prompt_for_int.

ebering:
Aidan has been handling the review of this PR so I'll leave that to him but I believe we'll be able to land this for the 0.25 release. Thank you for your persistence and for implementing this desired feature!

aidanholm:
This code looks good, and mostly ready for merge, but there were some last minute adjustments requested in IRC (2020-07-07):
```
10:57:53 <elliptic> hm, reading the commit description it sounds like that PR changes the behavior if you just use ! in normal play and makes annotating the current level more complicated?
10:58:47 <elliptic> I'm all in favor of making annotating other levels easier (I use the level map to do it quite frequently) but I don't think we should make the most common use of it more involved
10:58:57 <elliptic> (I should actually compile to see what the commit does though)
11:07:09 <aidanh> that is exactly what it does
11:07:26 <aidanh> you now need !. or !<enter> to annotate the current level, instead of just !
11:07:40 <elliptic> yeah, I just compiled it
11:07:57 <elliptic> I guess it is just one extra keypress and you are writing stuff anyway
11:08:54 <elliptic> and the ./</> interface does feel fairly natural
11:09:09 <elliptic> so it seems fine to me
11:09:21 <aidanh> one concern that just occurred to me is whether the menu forces messages out of view that the user might have wanted to refer to when writing
11:13:19 <elliptic> yeah, possibly the list of branches doesn't need to be displayed all the time?
11:15:14 <elliptic> e.g. display the branch codes in the ? menu instead
11:15:16 <advil> why not if you start typing and use any character but <>. it starts annotating the current level
11:16:28 <elliptic> well, currently that won't work given how the annotating other levels by branch code works
11:17:30 <advil> oh does it allow branch letters too?
11:17:54 <elliptic> it brings up the usual G list of branch letters yeah
11:18:25 <advil> maybe something like !! to get the proposed new behavior then, with ! + anything else its current behavior?
11:19:09 <advil> it does seem like a useful feature to add, the X[]! is actually pretty hard to find for most players
11:20:56 <Pinkbeast> IDK if you care what I think but while I'd hate to see the default behaviour of ! change, every time I want X[]! I have to reason that there must be a way to do it and what keystrokes would it logically use
11:22:56 <advil> and you're even a console player, so you know what the difference between X and x is (which is more obscure on tiles)
11:23:41 <Pinkbeast> I had some incomprehensible conversations with tiles players before I knew that was different
11:24:33 <elliptic> yeah, having something that isn't X[]! or XG<blah>! would be very nice
11:25:35 <elliptic> advil: I honestly think it's fine if you have to do !.<message> instead of !<message>, my only minor complaint is it spewing the list of branch codes out each time
11:26:04 <aidanh> spewing is consistent with travel though
11:26:08 <elliptic> and in particular I like removing the weirdness with current behavior depending on whether you are on top of stairs
11:26:22 <elliptic> well, G is mostly used for interlevel travel
11:26:29 <elliptic> ! mostly isn't used for interlevel stuff
11:26:32 <Pinkbeast> !! sounded good to me for the new behaviour
11:27:12 <elliptic> well, I like the !> and !< parts of the new behavior too
11:27:21 <elliptic> I woludn't want to have to do !!> or whatever
11:27:52 <elliptic> basically I like !. !> !< being what you do for the current level and the two adjacent levels
11:27:57 <advil> !! for spew?
11:35:41 <Pinkbeast> I guess at least some people now do '!!' for "add warning to current level" :-/
11:38:01 <ebering> Since ! previously brought up a text field (with an occasional stair prompt) in principle any keystroke following ! might be intended as part of the message
11:38:46 <aidanh> i don't think we have to keep key-for-key compatibility if there's a better way available
11:39:02 <ebering> I was just typingn that but not as eloquently
11:39:06 <aidanh> (says the dev who wanted to change scrolls from ? to : so we could have a single key for 'help')
11:39:21 <aidanh> :)
11:40:27 <ebering> so a short prompt saying ./</>/SOMEKEYFORSPEW
```

petercordia:
Sounds reasonable. Would backspace be a good key for the spew? or else Tab or Spacebar or #?

ebering:
I think that `!` is probably the best key for spew, coming from that conversation, since this is going to break keystroke compatibility no matter what and we do have an idiom of "repeat the key for an extended version of the command"

petercordia:
I have implemented the request to only show the spew after a 2nd !, but there is some kind of bug. The message after !! only shows after the next key-press.

I have rebased to master

There were also 2 image files I needed to add. Don't quite know why that was necessary, but I assume it was due to an update of master.

ebering:
Rebase does not mysteriously create files not in the index. I'm not sure what on your system is creating these temporary image files, but they don't belong in the repository.

crawl-ref/source/dgn-overview.cc
```diff
@@ -1093,3 +1062,164 @@ bool connected_branch_can_exist(branch_type br)
...
+            return ID_CANCEL;
+        }
+    }
+}
+
+void do_annotate()
+{
+    const level_id lid  = level_id::current();
+    const int branch = _prompt_annotate_branch(lid);
+
```
rawlins, line 1148: Given the way int is used here, I might recommend the following paradigm rather than the if sequences:

```
if (branch < 0)
{
    ASSERT(branch >= FIRST_ANNOTATION_POSSIBILITY && branch <= LAST_ANNOTATION_POSSIBILITY); // you would need to add these to the enum (and probably name them differently if you adopt my enum rename suggestion in a different comment)
    AnnotationPossibilities a = static_cast<AnnotationPossibilities>(branch);
    switch(a)
    {
    case ID_CANCEL:
        ...
    }
}
else
{
    ...
}
```

The reason is that the switch will check that every enum value is used so it provides a compile-time guard in case someone later changes the enum.

It might also be worth adding some ASSERTions to validate the positive case, i.e. at least that `branch < NUM_BRANCHES`

If this were still to be done with an `if` sequence, I would still suggest using `if ... else if` rather than unconnected `if`s, and adding the validation checks.

crawl-ref/source/dgn-overview.cc
```diff
@@ -25,11 +26,27 @@
 #include "religion.h"
 #include "scroller.h"
 #include "stairs.h"
+#include "store.h" //for level_id()
 #include "stringutil.h"
 #include "terrain.h"
 #include "travel.h"
 #include "unicode.h"
 
+enum AnnotationPossibilities
```
rawlins, line 35: this is very cosmetic but I would not choose this enum name, as it's not very descriptive, and we also don't tend to use camel case for enums. What about `nonbranch_annotations`?

crawl-ref/source/dgn-overview.cc
```diff
@@ -25,11 +26,27 @@
...
+{
+    // Annotate the dungeon floor the player character is currently on
+    ID_HERE     = -100,
+
+    // Annotate one level up
+    ID_UP       = -99,
```
rawlins, line 41: The sequencing here is odd, it looks like maybe the intent is to be able to meaningfully compare ID_HERE, UP, and DOWN using int comparison, but I would basically recommend against ever trying to build that directly into enum values. (For one thing, if they ever become an enum class it's impossible.) Of course if you reorder these to put ID_UP first I wouldn't even have commented on this :-)

crawl-ref/source/dgn-overview.cc
```diff
@@ -1093,3 +1062,164 @@ bool connected_branch_can_exist(branch_type br)
...
+            line = "";
+        }
+        line += make_stringf("(%c) %-14s ",
+                             branches[br].travel_shortcut,
+                             branches[br].shortname);
+        ++linec;
+    }
+    if (!line.empty())
+        mpr(line);
+    return;
```
rawlins, line 1091: I think you'll need some kind of refresh call here to get this to show. I haven't tested so I'm not sure exactly which is needed but maybe `flush_prev_message();`.

crawl-ref/source/dgn-overview.cc
rawlins: or something like `annotation_menu_commands`

petercordia:
I've tried to resolve the conflict in help.txt, which originated in the removal of butchering. I'm still not fluent with Git, but I hope I did it correctly.

ebering:
This did indeed look ready to go, through the commit history needed some cleanup using `git rebase --interactive`. I merged it manually.

Thanks again for your patience with the iteration process; as devteam time waxes and wanes things can be left sitting for a while.

PR #1370, opened 2020-04-19
Fix gitignore for windows packages

Goratrix:
Change to wildcard matching, also add the exe, which is already excluded
elsewhere with *.exe, but should be also here for consistency.

### DIFF
--- a/crawl-ref/.gitignore
+++ b/crawl-ref/.gitignore
@@ -186,7 +186,8 @@ makefile.dep
 /source/debian/files
 
 # Windows packaging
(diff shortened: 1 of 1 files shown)


Goratrix:
Just to clarify the intent, originally the gitignore had _/source/crawl-win_ which didn't work at all, maybe it was a relic from the past when the packages had different names. I changed it with 363036d to _/source/stone_soup_ but didn't test it and didn't realize it won't match the pattern because there is a folder name. So it needs to have the wildcards and extensions to match correctly.

PR #1373, opened 2020-04-24
Preferred food and menu_colour rules cleanup

nlavsky:
This PR removes a few *leftovers* from when potions of blood and more types of
food for carnivores/herbivores existed.

Also, it removes `menu_colour` rules for decks of cards and the lantern of shadows.


ebering:
Commit message of the version tbh. Thanks for the patch!

PR #1377, opened 2020-04-27
Expose player::get_noise_perception() through clua.

Implojin:
This commit adds a clua bind for you.noise_perception().

This information has been available to players visually through the noise bar,
but it wasn't exposed through clua. Noise information can be helpful for
scripted decisions re: resting and retreating, so here's a bind for it.


ebering:
Thanks for the patch!

PR #1378, opened 2020-04-30
Tweak some option defaults

alexjurkiewicz:
I recently watched a few players play their first games, and they all made the same errors:

* Running out of food as MiBe (due to auto butcher leaving chunks behind)
* Auto exploring at low HP

The show_more change is a little more speculative, I saw it cause a problem in one game where a newbie was surrounded and getting attacked multiple times. They didn't even realise the `--MORE--` prompt was showing up, and started getting frustrated that the game was ignoring their commands. They figured out pressing spacebar made the problem go away, but didn't understand the underlying principle.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -955,7 +955,7 @@ rest_wait_percent = 100
         it will stop resting when this percent of maximum HP or MP is refilled.
         Resting after this point will still rest up to 100%.
 
-explore_auto_rest = false
+explore_auto_rest = true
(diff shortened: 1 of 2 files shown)


aidanholm:
LGTM, but could you update the options guide too? thanks

ebering:
We rely on the message log for a lot of despoiling transparency, I would prefer that the `show_more` default remain the same and the issue with new players interactions with the message log handled a different way.

The other two changes look good, seconding the caveat that options_guide.txt should be updated to match.

PR #1381, opened 2020-05-01
Tweak phial of floods description (Flugkiller)

Goratrix:
Change description to better indicate that only monsters hit
by the initial wave will be affected by silence, not monsters that
walk into the water later.


Goratrix:
Noticed another error in items.txt, will do a new pull for both.

PR #1383, opened 2020-05-02
Change show_more default to false

alexjurkiewicz:
For new players, it's confusing and inconsistent to sometimes be
prompted to press space while waiting to play your next turn.
Additionally, the prompt to press space is small and easily overlooked
(and it's not clear what the --MORE-- prompt means unless you are
familiar with Unix command line environments).

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -1425,7 +1425,7 @@ clear_messages = false
         between player actions (default is false which will delay the
         clearing of messages until the message space is full).
 
-show_more = true
+show_more = false
         Setting this option to false will cause the game not to prompt
         if more than a window-full of messages are output at once. This
         option has no effect if clear_messages is set. Additionally, it
--- a/crawl-ref/source/initfile.cc
+++ b/crawl-ref/source/initfile.cc
@@ -196,11 +196,7 @@ const vector<GameOption*> game_options::build_options_list()
         new BoolGameOption(SIMPLE_NAME(note_chat_messages), false),
         new BoolGameOption(SIMPLE_NAME(note_dgl_messages), true),
         new BoolGameOption(SIMPLE_NAME(clear_messages), false),
-#ifdef DEBUG
         new BoolGameOption(SIMPLE_NAME(show_more), false),
-#else
-        new BoolGameOption(SIMPLE_NAME(show_more), !USING_TOUCH),
-#endif
         new BoolGameOption(SIMPLE_NAME(small_more), false),
         new BoolGameOption(SIMPLE_NAME(pickup_thrown), true),
         new BoolGameOption(SIMPLE_NAME(show_travel_trail), USING_DGL),


gammafunk:
I think the word "More" is a pretty universal way to convey that additional information is forthcoming; I see that text in many games. Not saying that the precise way we render that indicator is absolutely ideal, but it definitely doesn't require any esoteric knowledge to understand.

We do have to train new players to pay attention to the message log as it conveys important information, and the MORE prompt is one reasonable way we manage that presentation. Just immediately scrolling past a large bit of potentially important text by default isn't going serve them well. I'm not sure if this is a larger UI issue that can be addressed in other ways, but there are benefits to getting players to slow down a bit and read why they're dying.

alexjurkiewicz:
@ebering also had similar reservations in #1378.

How about this proposal as an alternative?

Instead of showing `--MORE--` for every "screen" of text, print all text (like `show_more = False` now). However, then print a new transient line `7 other messages hidden (press ctrl-p to view)` at the bottom of the message window.

This would be the new default behaviour of `show_more = normal`, which would become the default. The existing behaviours of `true` and `false` would remain unchanged to cater for existing players with existing preferences.

alexjurkiewicz:
```
03:09:16 +gammafunk | that might work, but I do question the idea that we want to
                    | by default hide or at least obfuscate the fact that the
                    | player did something to generate a lot of text
03:09:31 +gammafunk | the issue being that probably something happened they should
                    | be aware of
03:10:35 +gammafunk | I'm not really sure if there's a good general approach to an
                    | issue like this; if the aim is getting players to learn,
                    | either messages are helpful to that end, and slowing them
                    | down to read it, or messages are bad to that end and we
                    | shouldn't be generating them, it feels like
03:10:54 +gammafunk | s/down to read it/down to read it is good/
03:11:34    +aidanh | i think the right way is just to flash the more prompt if the
                    | player hits keys while it's active
03:11:35 +gammafunk | it is possible that something like message coloration and
                    | priority (sufficiently important ones get a color and either
                    | do trigger more or would if they would be missed)
03:12:06    +aidanh | and possibly switch --more-- to --more (press space to
                    | continue)--
03:12:27 +gammafunk | hrm, how is that different from the current default?
03:12:47 +gammafunk | what would you see if you generted a more but hit nothing?
03:12:52 +gammafunk | *generated
03:15:39    +aidanh | currently there's no visual feedback whatsoever
03:15:57    +aidanh | hit nothing?
03:18:45 +gammafunk | I guess I don't know what you mean by flash the more prompt;
                    | flash it onscreen, out of the message area?
03:18:58 +gammafunk | so, an additional indicator essentially?
03:21:37    +aidanh | i mean literally flash the piece of text on the screen saying
                    | --more-- :)
03:21:51    +aidanh | e.g. flip the background and foreground colours for 0.5
                    | seconds
03:24:05 +gammafunk | ok, I guess that's the same effect as what I said, in
                    | essense; you're just saying we'd make the prompt more
                    | visually distinct, so the player even gets that it's there
03:24:23 +gammafunk | to address the "I don't know why this is pausing, but I
                    | figured out I can hit space to carry on" issue
03:24:32    +aidanh | yeah, exactly
03:25:46 +gammafunk | seems reasonable to me
03:50:21     +advil | re this more stuff, I think it's definitely a bad idea to
                    | make any changes to it this close to a release
03:50:30     +advil | all of the alternatives I'm hearing would need testing IMO
03:51:05     +advil | and removing default more seems like not the way to go
03:51:15     +advil | if this is all such a concern, why not add something to the
                    | tutorial about it?
03:51:35     +advil | *concern for new players
03:59:26 +gammafunk | yeah, I'm not in any particular rush to change the --more--
                    | prompt either
04:03:22    +aidanh | i think a more prompt change would be relatively low-risk;
                    | the relevant code in messages seems reasonably
                    | straight-forward
04:04:12     +advil | I'm talking about testing by users :-)
04:04:27     +advil | the risk is that an alternative wouldn't work very well
04:04:59     +advil | and/or be even worse for newer players
04:05:25    +aidanh | hm, it's possible flashing the prompt wouldn't work, but I
                    | think some form of attention-getting visual feedback is
                    | required, and I don't see how to do it any other way
04:06:25    +aidanh | flashing focus-grabbing ui elements is also a common pattern
                    | at the OS level, e.g. for dialog boxes
```

petercordia:
I like the alternative you propose. No information gets lost if it's telling you you can see the other messages with control-p, and it's really annoying when you have to press spacebar a lot. On characters that have the more prompt show up every other turn I wouldn't read any of the 'more' messages anyway, I only do that if it is a rare occurrence, and in such cases I would notice the control-p prompt.
I use a lot of default options.

RojjaCebolla:
yeah, i feel like turning off default showmore is one of the most common things i see in RC files.

https://github.com/HilariousDeathArtist/DCSSConfigFile/blob/master/HilariousDeathArtist.txt#L480
https://crawl.project357.org/rc-files/trunk/Ultraviolent4.rc
https://github.com/gammafunk/dcss-rc/blob/master/gammafunk.rc#L358 (this appears to do the same thing)

ebering:
I think the conclusion here was that this particular change: defaulting show_more to false is not one we're going to pick up. The discussion has some good alternatives for making the `-- more --` prompt more obvious to a new player, if they get implemented a PR for them would be great but I'm going to close this for now.

PR #1384, opened 2020-05-03
Fix resource leak if allocating PNG structs fails.

Quipyowert2:
Cppcheck complains "Resource leak: fp" at lines 254 and 260 where the
function write_png returns on error without closing the file opened
at the beginning of the function. After write_png returns, the file
would still be open, but the pointer to it has been lost.

Cppcheck errors:
```
[crawl-ref/source/rltiles/tool/tile_colour.cc:254] (error) Resource leak: fp [resourceLeak]
[crawl-ref/source/rltiles/tool/tile_colour.cc:260] (error) Resource leak: fp [resourceLeak]
```


aidanholm:
Thanks!

PR #1390, opened 2020-05-09
Add IRC build failure notification

alexjurkiewicz:
You'll need to edit ci.yml after merging this. The value for `channel` should be put into a repo secret.

You can access the secret page here: https://github.com/crawlcrawl/settings/secrets

Add a secret called (eg) `IRC_CHANNEL` with value `##crawl-dev`. Then change:

```yaml
channel: "##crawl-dev"
```

to:

```yaml
channel: ${{ secrets.IRC_CHANNEL }}
```


alexjurkiewicz:
merged

PR #1391, opened 2020-05-09
Notify IRC for every build result

alexjurkiewicz:


alexjurkiewicz:
merged (and reverted)

PR #1394, opened 2020-05-10
Remove the bonus GDR of Gargoyles

alexjurkiewicz:
If this is too much of a nerf, they could get more AC (either base AC, or per-XL).

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -6051,7 +6051,7 @@ int player::gdr_perc() const
 
     const item_def *body_armour = slot_item(EQ_BODY_ARMOUR, false);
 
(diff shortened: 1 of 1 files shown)


ByrelMitchell:
Personally I would be inclined to give small amounts of GDR to other flat AC sources (like draconian scales) instead. The hidden mechanic of GDR contributes a lot to the general feel that AC has, and generally scales up with your AC. So long as it's a hidden mechanic from the player, I think less spoilery if it's more universally correlated with AC, rather than less.

ebering:
Gargoyles are in a fine place balance-wise imo.

I strongly prefer the approach begun in #1400 to reveal and explain the mechanic to removing it simply for lack of transparency.

PR #1397, opened 2020-05-11
Shuffle amulet/scarf/cloak egos

ebering:
The last amulet reform introduced some cool amulets, and centered the theme of the slot on "no tactical swapping, mostly passive effects that shift the emphasis of a character". Unfortunately, it didn't remove enough of the old ones, leaving the amulet slot with a lot of clutter.

Scarves were introduced to offer a trade-off between AC and more powerful effects in the cloak slot, and also to soak up the old RMsl status and turn it into an item. However, the most powerful effect in the cloak slot remained +Invis which stayed on cloaks!

This batch of commits shuffles the egos around to try and line things up better. Cloaks gain stealth and rCorr (looking thin with just rPois and MR, but boots have almost no egos so maybe it's fine without these two). Scarves gain Invis and Harm and drop spirit (dups the amulet) and cloud immunity. Amulets get pruned, losing harm (not good in the amulet spot because the swap penalty doesn't actually prevent swapping), reflection and gourmand (the latter two for not really fitting with the stronger amulets).

Outstanding todos:
- [ ] Finalize decision on cloak egos (gammafunk expressed some reservation wrt duplicating rings).
- [x] Consider: merging the effects of the two regen amulets into one. _Leaving them separate_ The merged item isn't much more compelling than each on their own, and still doesn't stack up against the faith/spirit/acrobat trio. A better revision can be looked at in 0.26
- [x] Consider: moving rage to a scarf (so no more active amulets other than unrands) or just removing rage (there's zerk pots and trog!). _Went with removal_
- [x] Decide if this should go into 0.25 or wait for 0.26 _No big objections to landing this in 0.25 have been voiced_


gammafunk:
I guess you've already covered pushing the effect reworks of regen/mpregen to 0.26, so that seems fine. I'm also of the opinion that those two effects aren't compelling, at least not as-is. There are quite a lot of mechanics in the game that let you sustain combat in very similar says. 

I don't think scarf of harm is good idea, since harm is quite bad if you have no way to swap it. The set of characters who could by some thin thread of reasoning convince themselves to wear it is just too small. Harm sort of has a problem where we already have effects like archmagi/+int and slaying to offer functionally similar advantages; the difference is that those usually come with a strategic cost. So it would probably be better for harm to be distinctive and have a tactical downside like it does now, but if we don't like the swapping aspect, I'm not sure what downside to give it.

ebering:
I disagree with the assessment of harm vs. slaying or archmagi. The increased incoming damage is a tradeoff but harm applies uniformly to _all_ sources of damage, including invocations and evocations. Offering such a uniform damage boost with a tactical tradeoff instead of a strategic tradeoff does provide some character niches with an interesting choice: heavy invo/evo users, early characters on more robust species that haven't yet made a strategic commitment to a damage source, characters that have made the strategic commitment to lower skill-cap lower damage weapon classes.

The decision "can I kill things faster enough to mitigate the extra risk of incoming damage" does more often become "no, since I can't rip it off at a moments notice if things go extra bad", but IMO there are still enough characters where the answer is "yes" or "yes, but" to make the ego worthwhile.

ByrelMitchell:
If it's marginal, and you want to increase the number of characters than can justify wearing harm without swapping, perhaps it should be a cloak ego. Losing 3-4 AC AND taking harm damage is harder to justify than just taking harm damage.

gammafunk:
Harm being a damage increase that applies sources that otherwise can't be enhanced won't push it over the edge for enough characters to justify it over what's effectively frail 2. Characters with multiple sources of damage are effective at doing damage as it is, since those different forms have different costs. Obviously you could find the ego early when you simply don't have much damage, but I don't think harm as an early ego is too compelling. Perhaps it is a bit more so with the ID game in play, which is partly why harm was conceived in the first place. Note that archmagi and slaying aren't so bad that they're easily overtaken by other options later like harm with no abilitiy to swap out of it would be.

It may be true that if you're going to make an effect that is so similar to archmagi/slaying, that it should be extreme in the sense that harm is and have a strong, unavoidable downside, but this type of mechanic does tend to have a quite bad "playfeel".

ebering:
I'm not sure if your indictment is a general indictment of any sort of "increased damage" downside (beyond the loss of cloak AC and alternative choices in the slot) or an indictment of the increased damage number in particular.

If it's the latter, we can tune the number. If it's the former, well, that's a bit trickier. I'm not convinced its the former, however; there are players who enjoy a play style that manages risk by not taking hits as opposed to being able to take more hits.

I think I'll merge this and we can see how it goes, if scarves of harm are no good we can tune the numbers in the coming weeks, or just remove them before release.

ebering:
I didn't squash this as much as I would typically to make reverts a bit easier if there's strong disagreement. The only tricky one to revert is scarves of harm, but I'd prefer to try harm with say a 10% received damage increase before axing it.

Ash-R:
Oh, you merged in the few minutes it took to log into my github account :(.
IMHO invis is overpowered on cloaks and will still be overpowered on scarves but at least the fact they appear less and have more possible ego outcomes makes getting one less likely. Sealth is a wonderful ego for cloak. rCorr I really have no thougts on other than "sure padding".

Harm should just completely removed from the game (unless it remains simply as an wear ID trap in early dungeon)
Gourmand should stay until there's no such thing as spell hunger, I still use them.
Completely disagree with the removal of reflection amulet and disagree with your estimation of newbie trap.
Never liked Rage+ ammy for the reason you described.

At the end this feels like a disappointing "mass-removal" than an interesting update, much like the most recent of evocable gutting episodes.

PR #1400, opened 2020-05-12
Show GDR in body armour descriptions

alexjurkiewicz:


ByrelMitchell:
Minor correction: The description here states that GDR only applies to melee attacks. That's not true; it applies to ranged attacks as well.

ByrelMitchell:
Ok, now you added 'from weapons'; it also applies to UC.

alexjurkiewicz:
Oh my god. This gets harder and harder to phrase. Any suggestions?

semitonal:
FWIW I strongly disagree that listing a lot of arbitrary percentages and confusing conditions on them is a better approach than removing GDR outright (and adjusting the AC provided by things that provided it, if necessary). I think it'd need to be substantially reworked and simplified if it were to be something worth keeping.

alexjurkiewicz:
I agree, I would prefer a DCSS with a much simplified GDR mechanic. Until then, it would be nice to make this mechanic a little more transparent to the player.

Related conversation from ##crawl-dev today

```
05:19:02 alexjurkie+| when testing the body armour description GDR stuff, I
                    | discovered that GDR is capped to [half your AC]% so it's
                    | nearly impossible to reach the true GDR of heavier
                    | armours
05:19:19 alexjurkie+| and also that gargoyle bonus GDR is huge
05:20:09 +gammafunk | there are further issues
05:20:26 +gammafunk | GDR's stated reduction is also false advertising; some
                    | aspects of damage it simply doesn't reduce
05:20:45 +gammafunk | I need to go back into attack.cc to see the details, but
                    | it makes its calculations without considering all parts
                    | of damage
05:21:04 +gammafunk | and for some attacks this is a big difference between
                    | the stated reduction percentage and the final damage
                    | actually done
05:21:30 +gammafunk | which, the real solution is to remove GDR and just
                    | rework it somehow into AC/armour/forms
05:27:22 alexjurkie+| it doesn't apply to the brand damage, right?
05:27:40 +gammafunk | I've actually forgotten the detail
05:27:45 +gammafunk | I want to say stuff like berserk as well?
05:27:58 alexjurkie+| and because the reduction percentage is applying to each
                    | roll individually, I'm guessing the quantisation makes
                    | small and even moderate changes in GDR irrelevant
05:28:39 +gammafunk | yes in practice you don't reason independently about
                    | changes in GDR since there's an accompanying change in
                    | AC that's far more meaningful
05:29:15 +gammafunk | if you didn't have GDR, the relevant attacks would do
                    | significantly more damage, of course
05:29:33 +gammafunk | but comparing 35% gdr to 45% gdr is just not meaningful
05:33:01 alexjurkie+| perhaps set GDR to a fraction of AC
05:33:21 alexjurkie+| i guess at that point it's just changing the min AC roll
                    | to a percentage of max
05:33:41 alexjurkie+| "just" = "i hope nobody expects me to edit combat code"
05:34:04 +gammafunk | yeah, there's a concern about balance changes for e.g.
                    | forms that provide no GDR, but something more reasonably
                    | tied to AC might work
05:34:40 +gammafunk | maybe something like the draconian mutation becomes a
                    | bit too strong, but a lot of that seems like it might be
                    | adjustable
05:34:49 +gammafunk | just give less AC for those things
```

ByrelMitchell:
I agree with the idea of making GDR simply a feature of AC (rather than particular sources of AC) and nerfing AC amount from existing non-GDR-granting AC sources. However, that has balance implications; it's possible that such a change would be put off to 0.26.

Just in case this display PR ends up being accepted for 0.25, I went through the current GDR code. You're not quite understanding the AC/2 limit properly; that's not a limit on the percentage minimum damage, but on the absolute minimum damage. For instance if you have 50% GDR and 30 AC, your minimum damage reduction is either 50% of max damage or 15 damage, whichever is lower.

I would suggest a phrasing like 'This reduces incoming melee and ranged damage by at least GDR% or 1/2AC damage, whichever is lower.'

crawl-ref/source/describe.cc
```diff
@@ -1699,6 +1700,39 @@ static string _describe_armour(const item_def &item, bool verbose)
...
             }
+
+            if (is_body_armour)
+            {
+                const armour_type body_type =
+                    static_cast<armour_type>(item.sub_type);
+                const int item_gdr_base = armour_gdr(body_type, false);
+                const int item_gdr_actual = armour_gdr(body_type, true);
+                const bool worn = _you_are_wearing_item(item);
+                if (item_gdr_actual) {
```
randomMesh, line 1711: The curly brace here is on the same line but they are on the next line elsewhere.

semitonal:
Closing this now since it's obsoleted by ff078d554e3f.

PR #1401, opened 2020-05-13
Reflect Bounces Back

PleasingFungus:
- Bring back amulets of reflection, but always +5 (no variable plusses), to simplify our much-maligned ID game.
- Give piety for reflected missiles, poison, and omnireflected Banishments.
- Add a delay before reflect amulets turn on, to avoid the dreaded Swap Micromanagement. It's always exactly 100 turns right now - doesn't seem like there's much point randomizing it, since there's no real hidden information or tactical gameplay around managing it?

I'll squash up some of these commits before merging if they're approved.

### DIFF
--- a/crawl-ref/source/actor.cc
+++ b/crawl-ref/source/actor.cc
@@ -272,6 +272,11 @@ bool actor::no_cast(bool calc_unid, bool items) const
     return items && scan_artefacts(ARTP_PREVENT_SPELLCASTING, calc_unid);
 }
 
+bool actor::reflection(bool calc_unid, bool items) const
+{
+    return items && wearing(EQ_AMULET, AMU_REFLECTION, calc_unid);
+}
+
 bool actor::extra_harm(bool calc_unid, bool items) const
 {
     return items &&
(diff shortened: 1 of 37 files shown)


alexjurkiewicz:
I think you may need to tweak autopickup logic so that after finding one you stop picking up more, like amulets of regeneraton and unlike rings of strength.

PleasingFungus:
Good point - done!

RojjaCebolla:
thanks very much for working to preserve one of my favorite mechanics.  it's so fun to see some nasty casting creature blow itself up especially when i lack good ranged attacks and resists.  actually, not only fun, a big relief!

if egos are becoming inherently +5, will that apply to shield egos too, or is that for amulets specifically?

PleasingFungus:
> if egos are becoming inherently +5, will that apply to shield egos too, or is that for amulets specifically?

@RojjaCebolla good question! Shield reflection egos don't actually affect SH at all - they're totally independent. Amulets of reflection are the only things being set to +5 SH.

PR #1402, opened 2020-05-13
Simplify ring plusses

RojjaCebolla:
Put +6 on stat rings, +4 on the stronger rings, and curses are only -3 or -4, instead of somewhere between 2 to 6.

### DIFF
--- a/crawl-ref/source/makeitem.cc
+++ b/crawl-ref/source/makeitem.cc
@@ -1610,26 +1610,11 @@ static bool _try_make_jewellery_unrandart(item_def& item, int force_type,
 }
 
 /**
- * Generate a random 'bad' plus for a ring type that cares about plusses.
- *
- * @return a bad 'plus', between -2 and -6 (inclusive).
- */
-static int _bad_ring_plus()
-{
-    int plus = -2;
-    if (coinflip())
-        --plus;
-    if (one_chance_in(3))
-        plus -= random2(4);
-    return plus;
-}
-
-/**
- * Generate a random 'good' plus for a ring type that cares about plusses.
+ * A 'good' plus for stat rings is +6, for other rings it's +4.
  *
  * @param subtype       The type of ring in question.
- * @return              Between 1 and 6 (inclusive); 2-6 for statrings.
- *                      (+1 stat rings are extremely boring.)
+ * @return              4 or 6.
+ *                      (minor numerical variations are boring.)
  */
 static int _good_jewellery_plus(int subtype)
 {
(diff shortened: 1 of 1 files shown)


crawl-ref/source/makeitem.cc
```diff
@@ -1612,24 +1612,22 @@ static bool _try_make_jewellery_unrandart(item_def& item, int force_type,
 /**
  * Generate a random 'bad' plus for a ring type that cares about plusses.
  *
- * @return a bad 'plus', between -2 and -6 (inclusive).
+ * @return a bad 'plus', between -3 and -4 (inclusive).
```
ebering, line 1615: Just make bad rings always `-4` and delete this whole function.

crawl-ref/source/makeitem.cc
```diff
@@ -1612,24 +1612,22 @@ static bool _try_make_jewellery_unrandart(item_def& item, int force_type,
...
 
 /**
- * Generate a random 'good' plus for a ring type that cares about plusses.
+ * A 'good' plus for stat rings is +6, for other rings it's +4
  *
  * @param subtype       The type of ring in question.
- * @return              Between 1 and 6 (inclusive); 2-6 for statrings.
- *                      (+1 stat rings are extremely boring.)
+ * @return              4 or 6.
+ *                      (minor numerical variations rings are boring.)
```
ebering, line 1630:
```suggestion
 *                      (minor numerical variations are boring.)
```

Grammar

crawl-ref/source/makeitem.cc
RojjaCebolla: Currently, 50% of rings are -2, 33% of rings are -3, and 17% of rings are -4 or more.  Because t average cursed ring is -2.83, i'm inclined to pick -3 if i'm going to pick a static number.  But maybe you're suggesting bumping it up to -4 because cursed items aren't meaningfully punishing at present?

ebering: Cursed items are pretty vestigial in crawl at this point, but yes, as long as they still exist they should be more impactful.

RojjaCebolla:
The next thing i wish to fix is, probably autopickup is still going to want you to id the plusses on every str/int/dex ring that you find.  But that's no longer necessary, since they're always +6.  Can you point me in the direction of the relevant code? @ebering

ebering:
I'm on my way out the door to do some shopping; you'll get a quicker pointer in ##crawl-dev on free node irc. Otherwise I'll try to point you in more detail when I get back.

The quickest I can offer is you could look at the auto pickup commit for reflection amulets.

RojjaCebolla:
> I'm on my way out the door to do some shopping; you'll get a quicker pointer in ##crawl-dev on free node irc. Otherwise I'll try to point you in more detail when I get back.
> 
> The quickest I can offer is you could look at the auto pickup commit for reflection amulets.

still wondering if we went a little high on this. 5 and 3 are more "pleasing" numbers than 6 and 4.

PR #1403, opened 2020-05-13
Sometimes generate Crazy Yiuf neutral

alexjurkiewicz:


ebering:
Yuif's hut isn't tagged no_monster_gen; neutral monsters spawned by vaults tend to behave in strange ways, robbing the player of XP by killing nearby monsters.

A neutral Yuif with his chaos quarterstaff might also be worth the XP tradeoff to let him hit monsters; if he gets a good chaos effect, sneak in and kill-steal; if he gets a bad one let the monster kill him safely and receive the chaos rewards.

So I'm closing this.

PR #1405, opened 2020-05-14
Formicid flavour: no bad shafts, no swiftness

RojjaCebolla:


ebering:
If you could rebase this to drop the merge commit and squash the fakemut into the first commit I think we can sneak this in even though its technically the start of feature freeze.

PR #1411, opened 2020-05-16
Display Zz and ? status icons for allies

nlavsky:
Having this information on the main view is useful because confused
allies can attack the player, and paralysed allies block movement.
Also, since the player gets the top-tier stabbing bonus when
attacking sleeping/paralysed/petrified allies, that should be
indicated by the Zz's, TILE_FLAG_STAB, too.


ebering:
Thanks for the patch!

PR #1413, opened 2020-05-17
Make faded Xom altars less disappointing

RojjaCebolla:
The player's still stuck with Xom, but at least they get two gifties 
and a random Xom action.

This is particularly intended to assist unlucky tournament players 
who, in aiming for the Inheritor (Hepliaklqana) banner, wind up 
worshipping Xom instead of say, Makh or the like.

In a not-very-exhaustive wizmode test, it seemed like the items
tended to be not very great, but I wasn't able to make heads or
tails of the variables "sever" and "tension".  Those could be set
to any ol' value if someone better understands what they do!

### DIFF
--- a/crawl-ref/source/god-prayer.cc
+++ b/crawl-ref/source/god-prayer.cc
@@ -26,6 +26,7 @@
 #include "stringutil.h"
 #include "terrain.h"
 #include "unwind.h"
+#include "xom.h"
 
 string god_prayer_reaction()
 {
@@ -89,6 +90,12 @@ static bool _pray_ecumenical_altar()
 
         if (you_worship(GOD_RU))
             you.props[RU_SACRIFICE_PROGRESS_KEY] = 9999;
+        else if (you_worship(GOD_XOM))
+        {
+            xom_take_action(XOM_GOOD_ACQUIREMENT, 1);
+            xom_take_action(XOM_GOOD_RANDOM_ITEM, 1);
+            xom_acts(1, MB_TRUE, 0);
+        }
         else
             gain_piety(20, 1, false);
 


rawlins:
Not sure if you've gotten any feedback on this elsewhere but my first reaction (not speaking for the rest of the devteam) is that the idea of making this roll "less disappointing" isn't really in the spirit of either Xom or the faded altar. My perspective is that both are effectively gambling mechanisms that you take if you want a challenge that requires you to adapt to something potentially unexpected, i.e. a sort of difficulty randomizer, and the fact that xom is possible on a faded altar is already part of what you are accepting by trying the altar. The point of tournament banners like this is to provide encouragement to players to try these randomizers. I do think that adding a bit of flavor to this case might be good, as presumably xom should be fairly entertained by being rolled on a faded altar and piety is meaningless. So I'm not opposed to making Xom do something analogous to piety gain if you roll the god (a la ru giving you a sac immediately), but the question is what that analogous thing should be. If there is a custom event list, I wouldn't necessarily think it should automatically aim for making things less disappointing (though probably shouldn't be automatically terrible). One alternative is to call `xom_acts` with maybe_good set to true. An even simpler alternative would be to do a `xom_is_stimulated` call with a relatively high value, which IIRC just does some flavor stuff and resets the gift timer. I do think two item drops is too much.

Re the custom event option: on the other hand I think what's happening in practice here may actually be more in line with my preferences than one might think. Xom acquirement has a large set of custom exceptions that are basically trying to troll the player -- so while it can generate something good, it also often generates something that can range from pointless to intentionally unusable. So a xom acquirement drop + perhaps something cosmetic (cloud trail, pseudo-miscast, and the like) isn't so bad. I'm not sure that the substantive thing should be limited to acquirement though, possibly good mutations, chaos branding, and a few other things should be possible.

(I'll also suggest that xom is far from the most disappointing thing that can happen with a faded altar, something like MiFi of Sif definitely more annoying. In fact Xom chars with good luck can be extremely strong; if I'm doing faded altar in the first place I'm not sure I'd actually be all that disappointed by xom...)

RojjaCebolla:
Thanks, well it still bugs me that a minotaur of sif gets some piety and a minotaur of xom just gets a challenge god.  Xom can make a tough situation deadly, sif at least doesn't do that!

I'll resubmit this without the forced gifties, thanks for your recommendations and thoughts!

alexjurkiewicz:
Generating a single Xom acquirement on worship seems like a fun little thing to me. Yes, it will often be worthless and occasionally even completely useless, but I think that's ok. Xom is Xom, after all, and it would be a nice condolence prize.

RojjaCebolla:
alternate version of this: we could just run `xom_acts(200, MB_TRUE, 200);` a few times

another alternate version of this: https://paste.ubuntu.com/p/9Qwh4vv3Km/

semitonal:
I'm not sure there's a good justification to have Xom immediately act, and especially not to be biased towards acting positively (and tournament banner details aren't really a reason to make design changes). It might be reasonable to just do a xom_is_stimulated(), since that's what happens usually when something amuses Xom and it's also probably the closest analogue to other gods' bonus piety.

As mentioned already though, faded altars by design are going to be disappointing a lot of the time and there's certainly no guarantee of any additional bonus (Uskayaw and Gozag also don't benefit from the extra piety, for example).

ebering:
I went with Kate's suggested approach of calling `xom_is_stimulated`. This hasn't been back ported to 0.25 since it's past feature freeze. I'm in favor of including the change but I'll let the rest of the team weigh in.

PR #1414, opened 2020-05-17
Flesh out randart grammars

RojjaCebolla:
This PR adds a section for verbs and a section for adjectives to the randart name generator.  Some categories are incompatible with the new additions because the category includes some things that start with "the" and others that don't: see _flower_name_ for an example.

Some categories that can be combined with a verb or adjective have been updated to do so, but not exhaustively.

Here's some names generated in wizmode: https://imgur.com/e12Bgzo.png  "Tired" and "Beastly" are from the new set. And here's some verbed randarts from a wizmode test with extremely high weight placed on the new words: https://i.imgur.com/MWk6Ixy.png


RojjaCebolla:
I'd love to get this merged in time for the tourney, because it's been pretty gratifying to see my other randart names occasionally showing up in Discord chat.  Let me know if you have any concerns or suggestions that i can address!

Also, there's an easy wizmode button to generate a ton of **un**randarts.  Is there an easy way to make a ton of **rand**arts in wizmode?  Doing it one at a time is somewhat tedious.

ebering:
You can use the command repeat function `0` or a macro, maybe?

Anyway, I've merged this after tweaking the commit history. Please try to keep your commits atomic (so don't go editing nouns in the verb commit &c.). For database changes like this maintaining an easy to follow `git blame` history isn't so important, but its a good habit to be in for future contributions!

PR #1416, opened 2020-05-19
Pass some parameters by reference

Quipyowert2:
According to Cppcheck, this is "usually faster" than passing by value
and "recommended in C++."

### DIFF
--- a/crawl-ref/source/database.cc
+++ b/crawl-ref/source/database.cc
@@ -545,7 +545,7 @@ static void _store_text_db(const string &in, DBM *db)
     _parse_text_db(inf, db);
 }
 
-static string _chooseStrByWeight(string entry, int fixed_weight = -1)
+static string _chooseStrByWeight(const string &entry, int fixed_weight = -1)
 {
     vector<string> parts;
     vector<int>    weights;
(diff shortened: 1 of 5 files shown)


Quipyowert2:
from ##crawl-dev (2020-05-18):
(8:36:22 PM) Cheibriados: New branch created: pull/1416 (1 commit) https://github.com/crawl/crawl/pull/1416
(8:36:22 PM) Cheibriados: Quipyowert2 https://github.com/crawl/crawl/pull/1416 * 0.25-a0-1129-g0a8e893: Pass some parameters by reference (15 minutes ago, 11 files, 46+ 46-) https://github.com/crawl/crawl/commit/0a8e893a68a0
(8:47:50 PM) |amethyst: hm, with a C++11 compiler I expect several of those changes result in copying when call-by-value would be able to use a move.  Specifically for the cases where the function or constructor is called with an rvalue
(8:49:07 PM) |amethyst: (and where it uses the parameter to initialize something else, e.g. the constructors)

Quipyowert2:
I undid the changes to constructors and changed the `get_desc_quantity` declaration to match the definition

wheals:
Thanks for the patch!

PR #1417, opened 2020-05-19
Rework the Sword of Power

alexjurkiewicz:
Increase the enchantment to hp / 5, remove the cap of 27, and add the
properties Harm, *Contam, *Curse.

Sword of Power is fun and evocative, but a little underwhelming in terms
of power level. Doubling the bonus will make the weapon much more
competitive.

Without any swap penalty, players are incentivised to carry their next
best long blade as a swap when the damage falls too much. The Sword of
Power requires commitment from the player, and so gains the
quasi-strategic costs of *Contam and *Curse.

Finally, adding Harm is an obvious choice. It's newly available for
artifacts (so no other unrand uses it), and fits perfectly with this
sword's theme.

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -80,6 +80,7 @@
 # * fragile:  Is destroyed when unequipped.
 # * fog:      Lets wearer use the evoke fog ability.
 # * evil:     Is treated as an evil item (hated by good gods).
+# * harm:     Provides harm effect.
 # * holy:     Is treated as a holy item.
 # * inv:      Lets wearer evoke invisibility ability.
 # * fly:      Lets wearer evoke flight ability.
@@ -220,7 +221,7 @@ COLOUR:  RED
 TILE:    spwpn_sword_of_power
 TILE_EQ: sword_of_power
 VALUE:   800
-BOOL:    special, no_upgrade
+BOOL:    special, no_upgrade, mutate
 
 NAME:    staff of Olgreb
 INSCRIP: rPois∞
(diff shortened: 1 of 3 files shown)


PleasingFungus:
I like the basic idea, but it seems like an awful lot is going on here. Do we need both `*Contam` *and* `*Curse`? How, exactly, does Harm relate to the sword's theme? You say it fits perfectly, but I'm really missing it, sorry.

alexjurkiewicz:
You're right. Contam is enough, no need for Curse.

I thought Harm was good because it is also a push your luck mechanic. But I'm not particularly wedded to it.

On IRC, other devs had some reservations about the maximum plus. +54 was seen as too high. I still think it's OK, especially with a swap penalty.

gammafunk:
We discussed an entirely different rework in irc that still needs some design, but I think would be a more interesting design. I can write a summary of Kate’s Zelda sword idea (!) later before working on a branch for it.

PleasingFungus:
Sure, I'd love to hear it!

semitonal:
I don't have a solid design in particular, but it's an old idea that I've always thought was kinda cute and fits the sword of Power's theme, of firing an energy bolt in the direction of your attack when attacking while at full health. To make it work in Crawl it'd probably need to be something like a chance to activate based on your HP%, and to only activate when actually swinging at a monster, not at empty space. Not sure how well it would work in practice but I'm fond of the concept, the details of range/chance to activate/damage etc would still need figuring out.

ebering:
I'm strongly in favor of Kate's power beam mechanic over a variable plus. Variable plusses just don't do it for me. Let me know if you're interested in taking that route—if you do I'd ask that the codepath be sufficiently general to allow the staff of Olgreb's venom bolt to be moved to the same place.

alexjurkiewicz:
I much prefer Kate's idea. But I won't put my hand up to implement it. I'll
close this PR, but I think it would be good for the game if the Sword of
Power gets a simple numeric buff as an interim measure since the Zelda
Sword remake might be months off.

PR #1421, opened 2020-05-22
force autopickup with ;;

petercordia:
; usually displays the items in the stack the player is standing
of if the player is 'unsafe', and forces autopickup otherwise.
As of this commit autopickup can be forced even if the player
feels unsafe by pressing ; twice.

### DIFF
--- a/crawl-ref/source/items.cc
+++ b/crawl-ref/source/items.cc
@@ -129,8 +129,6 @@ static bool _merge_items_into_inv(item_def &it, int quant_got,
 static bool will_autopickup   = false;
 static bool will_autoinscribe = false;
 
-static int last_auto_pickup_try = 1;
-
 static inline string _autopickup_item_name(const item_def &item)
 {
     return userdef_annotate_item(STASH_LUA_SEARCH_ANNOTATE, &item)
@@ -3313,20 +3311,12 @@ static void _do_autopickup()
 
 void autopickup(bool forced)
 {
-    // register when ; is pressed twice in a row, and force autopickup.
-    if (forced)
-    {
-        if (last_auto_pickup_try == you.num_turns)
-            _do_autopickup();
-        last_auto_pickup_try = you.num_turns;
-        return;
-    }
-
     _autoinscribe_floor_items();
-    if (!can_autopickup())
-        item_check();
-    else
+    // pick up things when forced (by input ;;), or when you feel save
+    if (forced || can_autopickup())
         _do_autopickup();
+    else
+        item_check();
 }
 
 int inv_count()
--- a/crawl-ref/source/items.cc
+++ b/crawl-ref/source/items.cc
@@ -129,6 +129,8 @@ static bool _merge_items_into_inv(item_def &it, int quant_got,
 static bool will_autopickup   = false;
 static bool will_autoinscribe = false;
 
+static int last_auto_pickup_try = 1;
+
 static inline string _autopickup_item_name(const item_def &item)
 {
     return userdef_annotate_item(STASH_LUA_SEARCH_ANNOTATE, &item)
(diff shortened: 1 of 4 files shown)


petercordia:
this is fulfilling a UI request from the wiki.

I think it's all working as intended.

Looking at the code of items.cc, I actually think that request_autopickup() is a kindof degenerate function. It is implementing a global variable in a really obtuse manner. The scope of this project could be increased to remove it everywhere, but that would be a much bigger and more difficult project.

petercordia:
In case you were wondering from the code, I moved the can_autopickup() check from _do_autopickup() to autopickup(), and this shouldn't cause problems because _do_autopickup is only called from inside autopickup.

crawl-ref/source/items.cc
```diff
@@ -129,6 +129,8 @@ static bool _merge_items_into_inv(item_def &it, int quant_got,
 static bool will_autopickup   = false;
 static bool will_autoinscribe = false;
 
+static int last_auto_pickup_try = 1;
+
```
ebering, line 133: I'm not happy about this adding ui-influencing global state, but I don't know where something like this would go otherwise. @aidanholm can you comment here?

crawl-ref/source/main.cc
```diff
@@ -1797,9 +1797,9 @@ void process_command(command_type cmd)
         break;
 
     case CMD_INSPECT_FLOOR:
-        request_autopickup();
         if (player_on_single_stack() && !you.running)
             pickup(true);
+        else autopickup(true);
```
ebering, line 1802:
```suggestion
        else
            autopickup(true);
```

Crawl house style tweak.

petercordia:
I just updated the manual and compressed the 3 commits into 1.

I had a bit of trouble with git; hope it didn't mess anything up.

crawl-ref/source/items.cc
aidanholm: Yeah, this is not a good idea; the better approach is to do this logic in `main.cc` and check `prev_cmd` there.

petercordia: I was wondering whether there was something like that in the code base, but couldn't find it.
Do you also have a suggestion oh how to avoid triggering pickup when one has just closed a different window with ; ?

aidanholm: Sorry, not sure I understand the question; `crawl_state.prev_cmd` should only be set to `CMD_INSPECT_FLOOR` if that was the previous command run, so you shouldn't be able to get to a state where you closed some other window with `;` without `crawl_state.prev_cmd` being set to something else.

petercordia: I misunderstood the way prev_cmd worked. You're right, the bug I described can't occur. I have now updated the logic.

petercordia:
Just found a bug. I should have checked before pushing. The problem is probably that prev_cmd gets assigned to CMD_INSPECT_FLOOR before we check prev_cmd == CMD_INSPECT_FLOOR

petercordia:
some minor things of which I wonder whether they should be renamed:
`autopickup()` -> `try_autopickup()`
'CMD_INSPECT_FLOOR` -> 'CMD_TRY_AUTOPICKUP` (?)

petercordia:
@ebering, I have implemented the changed you requested, haven't I?

moving 'prev_cmd = cmd` until after 'process_command(cmd)` has a lot of potential to create bugs, but it is necessary to implement @aidanholm's suggestion, and based on the fact he thinks prev_cmd aught to work that way it seems proper.

the previous implementation of CMD_INSPECT_FLOOR would have been broken by this change, though it was a bit hacky to begin with.

aidanholm:
@petercordia main.cc now has a `prev_cmd` parameter you can use; feel free to jump on IRC to find me if you run into any issues :)

ebering:
@petercordia Yes, you had implemented the changes I'd requested, but Aidan's comment was also a request for changes (use the new prev_cmd parameter in process_command). I've gone ahead and done that tweak and merged this.

Thanks for the patch!

petercordia:
Thanks for polishing and committing!
It would have taken me a while to get back to this and finish it.

PR #1423, opened 2020-05-25
Plug an information leak in summoner highlighting

ufshaikh:
During the do_look_around loop, placing the cursor over an invisible or out of
los monster highlighted the summoner. This commit makes the highlighting code
call can_see to make sure the character should know about the summon.

(A leak with respect to invisible monsters in los remains if you scan with the
cursor while scrying without sinv. This is leaked anyway in that case by the
monster description in the message log area, and there are other leaks with
respect to invisibility while scrying. I've left this alone here on the grounds
that it is part of a bigger, separate problem with scrying and invisibility, and
deserves a more general fix.)

### DIFF
--- a/crawl-ref/source/directn.cc
+++ b/crawl-ref/source/directn.cc
@@ -1767,10 +1767,15 @@ coord_def direction_chooser::find_summoner()
 {
     const monster* mon = monster_at(target());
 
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Nice fix, thanks!

PR #1425, opened 2020-05-27
Add autofight_danger_stop option.

aaronegolden:
Adds an autofight_danger_stop option, disabled by default. When
enabled, the option prevents autofighting when any monster of
the given threat level is visible. For example, with

autofight_danger_stop = 4

autofighting will refuse to act if any "extremely dangerous" enemy
is visible.


rawlins:
Is it possible this was closed accidentally? It doesn't seem to have been merged (though I can't even see the commit right now for some reason) and is probably something we would merge; I think you just didn't get an immediate response because of the release/tourney.

PR #1427, opened 2020-05-27
Added habitat description (11224)

elliotd123:
Idea for adding a description of a monster's habitat (If it's amphibious).
Mantis ticket: https://crawl.develz.org/mantis/view.php?id=11224


neilmoore:
You might want to test with zombies of amphibious monsters, and salamander simulacra.  See mons_habitat() in mon-util.cc: you can't use that function directly (it takes a monster rather than a monster_info), but you can see what it does with the type and base_type.

elliotd123:
Just tested and you're right, doesn't work correctly with Zombies.

crawl-ref/source/mon-util.cc
```diff
@@ -3326,13 +3326,20 @@ static habitat_type _mons_class_habitat(monster_type mc,
     return ht;
 }
 
+habitat_type mons_habitat_types(monster_type t, monster_type base_t,
```
aidanholm, line 3329: Nit: this should probably be called `mons_habitat_type`

crawl-ref/source/mon-util.cc
```diff
@@ -3326,13 +3326,20 @@ static habitat_type _mons_class_habitat(monster_type mc,
...
+                                bool real_amphibious) {
+    return _mons_class_habitat(fixup_zombie_type(t, base_t),
+                               real_amphibious);
```
aidanholm, line 3332: Can we add some test cases for this? In `catch2-tests/test_mon-util.cc`. It should be enough to test a few different monsters of different types including zombie types, simulacrum types, etc.

It would be best to also test `_describe_monster_hab()` separately (in a new `test_describe.cc` file): just testing one monster of each type (water, lava, none) is sufficient there.

crawl-ref/source/describe.cc
```diff
@@ -4011,6 +4011,28 @@ static void _describe_monster_mr(const monster_info& mi, ostringstream &result)
...
+        switch (h) {
+            case HT_AMPHIBIOUS:
+                result << "It can travel through water." << "\n";
+                break;
+            case HT_AMPHIBIOUS_LAVA:
+                result << "It can travel through lava." << "\n";
+                break;
+            default:
+                break;
+        }
```
aidanholm, line 4033: This block is over-indented

crawl-ref/source/describe.cc
```diff
@@ -4011,6 +4011,28 @@ static void _describe_monster_mr(const monster_info& mi, ostringstream &result)
...
+ * @param mi[in]            Player-visible info about the monster in question.
+ * @param result[in,out]    The stringstream to append to.
+ */
+static void _describe_monster_hab(const monster_info& mi, ostringstream &result)
+{
+    habitat_type h = mons_habitat_types(mi.type, mi.base_type);
+
+        switch (h) {
+            case HT_AMPHIBIOUS:
+                result << "It can travel through water." << "\n";
```
aidanholm, line 4026: Just `<< "It can travel through water.\n";` is better.

crawl-ref/source/describe.cc
```diff
@@ -4011,6 +4011,28 @@ static void _describe_monster_mr(const monster_info& mi, ostringstream &result)
...
 
+/**
+ * Append information about a given monster's habitat to the provided stream.
+ *
+ * @param mi[in]            Player-visible info about the monster in question.
+ * @param result[in,out]    The stringstream to append to.
+ */
+static void _describe_monster_hab(const monster_info& mi, ostringstream &result)
+{
+    habitat_type h = mons_habitat_types(mi.type, mi.base_type);
```
aidanholm, line 4022: super nit: should be `const auto habitat` or `const habitat_type habitat`
 - use const; the lack of const indicates that it's mutated later (but it isn't in this case)
 - `h` is an unclear name; prefer avoiding abbreviations if possible

crawl-ref/source/describe.cc
```diff
@@ -4011,6 +4011,28 @@ static void _describe_monster_mr(const monster_info& mi, ostringstream &result)
     result << "\n";
 }
 
+/**
+ * Append information about a given monster's habitat to the provided stream.
+ *
+ * @param mi[in]            Player-visible info about the monster in question.
+ * @param result[in,out]    The stringstream to append to.
+ */
+static void _describe_monster_hab(const monster_info& mi, ostringstream &result)
```
aidanholm, line 4020: Can we rename this to `_describe_monster_habitat`?

crawl-ref/source/describe.cc
```diff
@@ -4347,6 +4368,7 @@ static string _monster_stat_description(const monster_info& mi)
     result << _monster_attacks_description(mi);
     result << _monster_missiles_description(mi);
     result << _monster_spells_description(mi);
+    _describe_monster_habitat(mi, result);
```
aidanholm, line 4371: Maybe it would be easier to make this function return a string; that way you can use the same `result << ` pattern. Might make your tests slightly cleaner too. What do you think?

crawl-ref/source/catch2-tests/test_mon-util.cc
```diff
@@ -26,3 +30,51 @@ TEST_CASE("can get names for removed monster types", "[single-file]" ) {
...
+        REQUIRE(habitat == HT_AMPHIBIOUS);
+    }
+
+    SECTION("Giants should show up as HT_AMPHIBIOUS unless explicitly told not to") {
+        habitat_type habitat = mons_habitat_type(MONS_CYCLOPS,MONS_CYCLOPS);
+        REQUIRE(habitat == HT_AMPHIBIOUS);
+
+        habitat = mons_habitat_type(MONS_CYCLOPS,MONS_CYCLOPS,true);
+        REQUIRE(habitat != HT_AMPHIBIOUS);
+    }
```
aidanholm, line 74: nit: I'd recommend splitting this into two separate sections. It's best to have one assertion per section wherever possible.

Also: amphibious giants should still show up as amphibious even if `real_amphibious` is true, so that would be another good test case.

crawl-ref/source/catch2-tests/test_mon-util.cc
```diff
@@ -26,3 +30,51 @@ TEST_CASE("can get names for removed monster types", "[single-file]" ) {
...
+
+    SECTION("Simulacrum Salamanders should NOT still be HT_AMPHIBIOUS_LAVA") {
+        const habitat_type habitat = mons_habitat_type(MONS_SIMULACRUM,MONS_SALAMANDER);
+        REQUIRE(habitat != HT_AMPHIBIOUS_LAVA);
```
aidanholm, line 78: what should they be? probably better to explicitly expect a given habitat

crawl-ref/source/catch2-tests/test_mon-util.cc
```diff
@@ -2,8 +2,12 @@
 
 #include "AppHdr.h"
 
+#include "mon-enum.h"
+
 #include "mon-util.h"
 
+#include "monster-type.h"
```
aidanholm, line 9: again, no need for the extra line breaks in the last group of headers

crawl-ref/source/describe.cc
elliotd123: I was following the same patter as the functions:
`    _describe_monster_hp(mi, result);
    _describe_monster_ac(mi, result);
    _describe_monster_ev(mi, result);
    _describe_monster_mr(mi, result);`

I ended up moving that function call to a later part of describe because I thought it belonged somewhere else.

Which pattern works better? I'd be happy to switch it to a function that returns a string instead.

elliotd123: I switched it to use a string, and renamed the function to be consistent with those other description functions. I think it's cleaner to do it this way. Let me know what you think!

elliotd123:
I'm guessing this will need to be reconciled with these changes at some point:

[https://github.com/crawl/crawl/commit/2a36f74b56e0](url)

aidanholm:
No worries, I'll merge this one first.

Looks good; two more minor bugs: this doesn't use the correct article for some monsters, e.g. Cerebov. The habitat description also needs to be moved above the `_monster_spells_description`.

crawl-ref/source/catch2-tests/test_mon-util.cc
```diff
@@ -2,27 +2,103 @@
 
 #include "AppHdr.h"
 
+#include "mon-enum.h"
 #include "mon-util.h"
+#include "monster-type.h"
```
aidanholm, line 7: nit: keep these sorted

crawl-ref/source/describe.cc
```diff
@@ -4347,6 +4367,7 @@ static string _monster_stat_description(const monster_info& mi)
     result << _monster_attacks_description(mi);
     result << _monster_missiles_description(mi);
     result << _monster_spells_description(mi);
+    result << _monster_habitat_description(mi);
```
aidanholm, line 4370: Flip these; otherwise it looks weird when looking at some monsters' descriptions, e.g. Cerebov.

crawl-ref/source/mon-util.cc
```diff
@@ -3326,13 +3326,20 @@ static habitat_type _mons_class_habitat(monster_type mc,
...
 {
-    return _mons_class_habitat(fixup_zombie_type(mon.type,
-                                                 mons_base_type(mon)),
-                               real_amphibious);
+    return mons_habitat_type(mon.type, mons_base_type(mon),
+                              real_amphibious);
 }
 
+
+
```
aidanholm, line 3342: Please remove these extra lines.

aidanholm:
Merged manually at 0.26-a0-88-g5ee4766; thanks for the PR!

PR #1428, opened 2020-05-27
Prevent draconians from getting armour troves

RojjaCebolla:
These are mostly useless unless there's gauntlets of war, boots of running, etc in the trove.


ebering:
From the mantis ticket discussion there is some mention that an important part of these armour tolls is the investment of ?EA scrolls. However, I think it better (if we want ?EA to be a toll) to make that a direct toll. Merging this, thanks for the patch!

PR #1429, opened 2020-05-27
Simplify Phial of Floods

ByrelMitchell:
Previously, phial of floods did three things:
 a) It fired a Primal Wave at the target
 b) It created a pool around the target
 c) It waterlogged monsters in the pool

Primal Wave deals damage, potentially knocks back a tile, and creates its own pool. This led to phial of floods creating two different overlapping temporary pools, one waterlogged and the other not.

This commit consolidates everything into Primal Wave. Phial no longer manually creates its own pool; instead player-created Primal Waves now scale with power like the Phial pool did, and apply waterlogged appropriately.

This commit also reworks Primal Wave's pool generation. Previously, it created the center tile manually, and then randomly added tiles on a 2-radius circeLOS ring to it. I switched it to use the Phial pool generation method which uses a fair distance_iterator to fill the pool from the center. The results are fairly similar without the weird spoilery effects of circle-LoS math.

This commit also incidentally solves the issue where the actual target of the phial could be pushed out of the pool by the primal wave and miss getting waterlogged. Now that we silence based on the Primal Wave pool (which was always centered on the pushed-back monster) you're guaranteed to at least waterlog your target (if it's vulnerable.)

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -2067,42 +2067,28 @@ void fire_tracer(const monster* mons, bolt &pbolt, bool explode_only,
     pbolt.is_tracer = false;
 }
 
-static coord_def _random_point_hittable_from(const coord_def &c,
-                                            int radius,
-                                            int margin = 1,
-                                            int tries = 5)
-{
-    while (tries-- > 0)
-    {
-        const coord_def point = dgn_random_point_from(c, radius, margin);
-        if (point.origin())
-            continue;
-        if (!cell_see_cell(c, point, LOS_SOLID))
-            continue;
-        return point;
-    }
-    return coord_def();
-}
-
-void create_feat_splash(coord_def center,
+vector<coord_def> create_feat_splash(coord_def center,
                                 int radius,
-                                int nattempts)
+                                int number,
+                                int duration)
 {
-    // Always affect center, if compatible
-    if ((grd(center) == DNGN_FLOOR || grd(center) == DNGN_SHALLOW_WATER))
+    vector<coord_def> splash_coords;
+
+    for (distance_iterator di(center, true, false, radius); di && number > 0; ++di)
     {
-        temp_change_terrain(center, DNGN_SHALLOW_WATER, 100 + random2(100),
-                            TERRAIN_CHANGE_FLOOD);
+        const dungeon_feature_type feat = grd(*di);
+        if ((feat == DNGN_FLOOR || feat == DNGN_SHALLOW_WATER)
+            && cell_see_cell(center, *di, LOS_NO_TRANS))
+        {
+            number--;
+            int time = random_range(duration, duration * 3 / 2) - (di.radius() * 20);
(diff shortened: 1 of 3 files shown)


alexjurkiewicz:
Does this affect monsters too? I forget if elemental wellsprings are still in the game...

ByrelMitchell:
Yes and no. The most common monster with Primal Wave is a merfolk aquamancer. The change in how the puddle is made does affect them. I tuned it to be similar to their previous puddle-size, but the generation method is now based on a distance-iterator instead of randomly selecting a degree out of 360 and projecting 2 tiles out onto the dcss grid, so there's liable to be some change in shape distribution.

Other than that, everything should be the same for the monster use-cases.

crawl-ref/source/beam.cc
```diff
@@ -2369,9 +2356,26 @@ void bolt::affect_endpoint()
             noisy(spell_effect_noise(SPELL_PRIMAL_WAVE),
                   pos(), "You hear a splash.");
         }
-        create_feat_splash(pos(), 2, random_range(3, 12, 2));
+        const int num = agent()->is_player() ? div_rand_round(ench_power * 3, 20) + 3 + random2(7)
```
ebering, line 2359: `agent()` can be null, since Fedhas uses primal wave for wrath.

```suggestion
        const int num = agent() && agent()->is_player() ? div_rand_round(ench_power * 3, 20) + 3 + random2(7)
```

crawl-ref/source/beam.cc
```diff
@@ -2369,9 +2356,26 @@ void bolt::affect_endpoint()
...
+                                             : random_range(3, 12, 2);
+        const int dur = div_rand_round(ench_power * 4, 3) + 66;
+        vector<coord_def> splash_coords = create_feat_splash(pos(), agent()->is_player() ? 2 : 1, num, dur);
+        dprf(DIAG_BEAM, "Creating pool at %d,%d with %d tiles of water for %d auts.", pos().x, pos().y, num, dur);
+        if (agent()->is_player())
```
ebering, line 2364:
```suggestion
        if (agent() && agent()->is_player())
```

crawl-ref/source/evoke.cc
```diff
@@ -939,38 +939,7 @@ static bool _phial_of_floods()
...
-                monster* mons = monster_at(*di);
-                if (mons && !mons->res_water_drowning())
-                {
-                    simple_monster_message(*mons, " is engulfed in water.");
-                    mons->add_ench(mon_enchant(ENCH_WATERLOGGED, 0, &you,
-                                               time));
-                }
-            }
-        }
+        beam.fire(); // Delay firing the Primal Wave till after waterlogged has been applied.
```
ebering, line 942: This comment looks out of place now? Is it from an earlier draft.

ByrelMitchell:
Alright, I fixed those issues. I think applying waterlogged to players is a pretty sizable task (given my skill level in crawlcode wrangling); I'd rather just spin up a new PR for that.

ebering:
A separate PR to add waterlogged to players would be great!

PR #1430, opened 2020-05-28
Remove blood and web sense

ufshaikh:
This PR removes web sense and blood scent, and has a couple minor code clean-ups along the way. There are some justificatory remarks in the commit messages.

Sorry about the bad timing. I realize it may be a while until anyone looks at this.

I'm happy to make adjustments, of course.


PleasingFungus:
I like these changes overall, but I do think it'd be nice to have something in spider to help encourage swarms of enemies to descend upon hapless players. Screaming (camel?) spiders? Extremely noisy orb weavers? Something else? My inbox is open.

semitonal:
Ambient branch noise is an existing mechanic that tries to make that kind of swarming more likely, so changing Spider to be ambiently quiet seems like a good way to replace web sense. If nobody else gets to it first I'll try and take a look and merge this and make that change.

rawlins:
Ambient branch noise sounds reasonable to me, and it has the plus side that it might work (I'm skeptical that web sense ever did much).

PleasingFungus:
Using ambient branch noise seems like a great idea.

ufshaikh:
OK, I've tweaked ambient noise to quiet.

ebering:
This looks great, I'm going to hit merge. If after playing spider doesn't always have the swarm feel it used to we can further get the web-sense effect by making struggling against a web noisy.

PR #1431, opened 2020-05-28
Fix description of amulets of reflection

nlavsky:
After 324b7566, amulets of reflection lost their plusses, and as
a result no longer have a part of their description that mentions
the shielding bonus.


PleasingFungus:
Great fix, thanks!

PR #1434, opened 2020-05-29
Fix up some typos in the docs folder.

aaronegolden:
Why yes I **am** fun at parties.


semitonal:
Good catches, thanks!

PR #1435, opened 2020-05-29
Fix segfault for Summon Mushrooms spell

elliotd123:
https://crawl.develz.org/mantis/view.php?id=11889

In wizmode, attempting to cast summon mushrooms (A monster-only spell) causes a segfault.
This is because it's targeted on a foe, which is a nullptr for wizmode monster casts.


elliotd123:
Also maybe it needs to output some failed message of some kind?

aidanholm:
Thanks for the fix! Merged manually at b94faee717

PR #1436, opened 2020-05-30
Allow player know the full name of dead orc during Resurrect

Gittourarmy:
Now, the players don't know the role of the dead orc if the one is  blessed, when they use Resurrection ability onto the corpse of dead orc.
For example, if you tried to resurrect Borug the orc warlord, then you will see only Borug.
The point of this commit is, if there is another orc corpse that you don't know the role(possibly, just orc) and has its own name, then player can make mistakes.

This fix may help Beogh players to use their ability in intened way.


aidanholm:
Thanks for the fix! Merged manually at 0.26-a0-120-g33945c5

PR #1437, opened 2020-05-30
Reword hailstorm description to be less hand-wavy

anderspapitto:


PleasingFungus:
Clean change. Thanks!

PR #1439, opened 2020-05-31
Let vampires quaff from the 'i' menu again (blerg)

ufshaikh:
It looks like vampires accidentally lost this when they were reworked to
be foodless.


PleasingFungus:
Pushed with a bunch of other fixes as a486cdfe0b6ce2322cc10292abe60b7c83717423, thanks!

PR #1443, opened 2020-06-03
Rename Spiny Frogs to Cane Toads

ByrelMitchell:
There's a handful of reasons why these amphibians make more sense as cane toads than 'spiny frogs'.

1.) Spiny is an actual mechanic in the game, and spiny frogs don't have it.

2.) Spiny frogs have way more AC and less EV than other frog-genus monsters, which is one major difference between toads and frogs.

3.) The spiny frog tile is the typical color for toad camouflage rather than frog, and has what appears to be warts on its side.

4.) This is an Australian game, and it makes sense for their 'bullfrog' equivalent to be the ultimate mundane frog-genus monster.

These are in order of descending importance; 1 has the potential to mislead players, and 2 is more accessibly indicating AC/EV. 3 and 4 are more aesthetic and so not very important (or objective.)


rawlins:
Seems like good arguments against "spiny frog" at least. Though re 4, TIL while following up on this that cane toads are invasive to Australia (introduced in 1935).

alexjurkiewicz:
It's alright, we claim them as Australian by now.

PleasingFungus:
Great suggestion. Merged with small technical tweaks as f02c06f1080ff5, thanks!

PR #1444, opened 2020-06-03
Update `restart_after_save` default in the options guide

nlavsky:
The default value was changed to `true` in 750a858e.


PleasingFungus:
Good catch, thanks!

PR #1445, opened 2020-06-04
Fix foxfire attacking out of LoS

ByrelMitchell:
Currently a foxfire will attack a monster it's aggroed to that is just outside LoS (while the foxfire is in LoS.) Since a foxfire attacks 5 times per turn for 1 damage, this is actually somewhat usefully abusable in the early game vs 0 AC monsters (you can get 3-4 extra damage out of a foxfire by manipulating this.)

This change makes it follow the rules of normal summons and refuse to attack monsters out of player LoS.


PleasingFungus:
Nice fix, thanks!

PR #1446, opened 2020-06-04
Make the Crypt monster list tougher

alexjurkiewicz:
Remove shadows and wights entirely, they are complete popcorn by this
point.

Push the placement weights earlier for flying skulls, necromancers and
wraiths. These are all very on-brand but still quite easy.


PleasingFungus:
Players weren't dying enough anyway, imo.

Issue #1363, opened 2020-04-18
fatal error: SDL_image.h


lembark:
Compiling crawl on linux (gentoo) using "gcc (Gentoo 9.2.0-r2 p3) 9.2.0".

This happens using HEAD:

commit 5b70d41e488af14cffbb0d0b6c7121715eae9c6a (HEAD -> master, origin/master, origin/HEAD)
Author: Goratrix <goratrix@chello.sk>
Date:   Fri Apr 17 17:12:18 2020 +0200

or also with tag:

HEAD is now at 33895e9a04 Changelog for 0.9.2

git clone https://github.com/crawl/crawl.git;
git submodules update --init;
make -wk TILES=y;
<drink a beer... it errors out.>
make -wk TILES=y;  # re-run make to get more specific error
https://pastebin.com/1RzgnuPs

windowmanager-sdl.cc:23:11: fatal error: SDL_image.h: No such file or directory

Q: Is SDL_image.h part of a sub-module or provided by the O/S?
I cannot find it in any gentoo package.

lembark:
Checking libsdl & libsdl2:
  $ equery files  media-libs/libsdl | grep 'SDL_image.h';
  $ equery files  media-libs/libsdl2 | grep 'SDL_image.h';
i.e., nada.

alexjurkiewicz:
If I understand Gentoo correctly it's part of this package:
https://packages.gentoo.org/packages/media-libs/sdl2-image

(I found it by googling "Gentoo SDL image".)

alexjurkiewicz:
Also, if you haven't seen it already you may want to read the install instructions. There are other packages you need to install:
https://github.com/crawl/crawl/blob/master/crawl-ref/INSTALL.md#other-linux--unix
Although it's easier to use the packaged submodules, if you're compiling for personal use.

lembark:
I went through the instructions, they didn't include Gentoo, however. 
The missing library was:

*  media-libs/sdl-image
      Latest version available: 1.2.12-r2
      Latest version installed: 1.2.12-r2
      Size of files: 2,179 KiB
      Homepage:      http://www.libsdl.org/projects/SDL_image/
      Description:   image file loading library
      License:       ZLIB


Might want to add "sdl-image" to the "Prerequisites" along with ncurses.

alexjurkiewicz:
Both SDL2 and ncurses are mentioned as required in the section of guide I
linked. I rewrote it recently so happy to take feedback on how to improve
it!

[...]

PR #1447, opened 2020-06-08
Added SDL_image to dependency requirements in INSTALL.md

elliotd123:
Most distros package SDL2 separately from its child libraries (SDL2_image, SDL2_ttf, etc).
Adding this to the dependencies for other OS distributions to reduce confusion.
Should clear up #1363


aidanholm:
Thanks, and sorry for the delay in merging!

PR #1448, opened 2020-06-08
Change = to == in checking crawl_state.type

shummie:
Every game is displaying the seed, whether or not it's a custom game since the statement evaluates as true.


rawlins:
Thanks for catching this so quickly, could have been fairly epic...just affects a few cko games I think.

PR #1450, opened 2020-06-10
Get rid of MSG_NO_ENERGY canned message.

aaronegolden:
This canned message was used both for when the player is too hungry
to cast a specific spell and for when the player is too hungry to
cast *any* spells. It was confusing in the "any spell" case, because
"You don't have enough energy to cast that spell." suggests that some
specific spell was attempted. When it happened to me I was convinced
that I had some macro attached to 'z' by mistake.

The patch removes MSG_NO_ENERGY and replaces both uses of it with
more specific messages: "You don't have enough energy to cast that
spell." for the case where a specific spell was tried (same as
before) and "You don't have enough energy to cast any spells." for
the case where no specific spell was selected.


aidanholm:
Thanks for the PR! Looks good, but you need to add your replacement messages to `dat/defaults/messages.txt` and remove the original one.

aaronegolden:
> Thanks for the PR! Looks good, but you need to add your replacement messages to `dat/defaults/messages.txt` and remove the original one.

Thanks for the tip! I added the new message to dat/defaults/messages.txt. I didn't remove the old one, "You don't have the energy to cast that spell", because it's still used in the case where the user has attempted a specific spell.

PleasingFungus:
Great change, thanks!

PR #1458, opened 2020-06-22
Fix scarf of harm not working

Goratrix:
Commit 19c8d55 added harm to scarves, but used the 'wearing' function,
which is only valid for getting jewellery subtypes. Need to use
'wearing_ego' to get the scarf ego instead.


aidanholm:
Thanks for the fix!

PR #1465, opened 2020-06-28
Rename Assassin -> Bandit (Notta)

PleasingFungus:
If you see that you're playing an 'assassin' with a dagger, your
playstyle is very obvious: you sneak around and stab unsuspecting
enemies. But that's not what Crawl's Assassin is!

If you want to stab monsters after the first few floors, you want
to be playing an Enchanter, Earth Elementalist, or someone else with
the ability to incapacitate foes. Crawl's stealth mechanic isn't
meant to support 'purely mundane' stabbing, but the Assassin's very
powerful theming is extremely effective at convincing new players
otherwise, and then frustrating them when it doesn't work.

A Bandit ia more brutish figure, an outlaw who leaps from the woods
with an axe and chops you up. They're not sneaky in the same way that
an Assassin is, and so, I hope, won't encourage the same confusion.

Other considered names: 'ruffian', 'thug', 'brigand', 'skirmisher'.


ByrelMitchell:
+1 for 'brigand'. Bandit doesn't have the right flavor for DCSS IMO. Too italian/american wild west instead of english/australian.

rawlins:
I would prefer that this not be merged without a concrete proposal (or even better, PRs) for how to handle the rename in at least Sequell/scoring...

ByrelMitchell:
dASperado? :P

PleasingFungus:
@rawlins I have very little familiarity with the Sequell codebase. Can you talk a bit about what you'd want here? Some kind of aliasing between`As` and `Ba`, conversion of older data...?

alexjurkiewicz:
Most species and background renames have preserved the short code, eg Grotesque to Gargoyle. But I don't think it's necessary here.

Since nothing is going to use the old Assassin name, scoreboards can map any Assassin game to the new name. Sequell can alias `as` to `as|bn` (and vice versa).

BTW, it would be nice to have a short code which doesn't overlap with Barachian. It makes sequell queries harder, as seen for Felid Elementalists.

PleasingFungus:
I have no opinion on brigand vs bandit; I opted for bandit here because it's slightly shorter, but the words seem extremely similar to me. `Br` would presumably solve the short code issue, though at the terrible cost of creating fewer words.

PR #1466, opened 2020-06-28
Rename Wizards to Hedge Wizards

PleasingFungus:
Wizards are a weird background. Their name and starting stats both
point towards a 'pure magic' playstyle, but their starting book
is quite poorly suited to such an approach - they aren't
well-equipped to 'blast' foes, and are much better advised taking
advantage of their imps and mephitic clouds to do a little
backstabbing.

This playstyle seems fine and distinctive, so let's make it
clearer to players. Renaming them to 'hedge wizards' shows that
they aren't expected to be masters of 'high magic' (pure blasting),
as does tweaking their stat spread closer to skalds'. (If we want
the extremely high-int spread to be on some background, Cj would
be a reasonable choice.)

### DIFF
--- a/crawl-ref/source/dat/descript/backgrounds.txt
+++ b/crawl-ref/source/dat/descript/backgrounds.txt
@@ -62,6 +62,11 @@ Gladiator
 Gladiators are ready for the arena with light armour, nets, and a weapon of
 their choice.
 %%%%
+Hedge Wizard
+
+Hedge Wizards learn many minor magics, letting them backstab or escape their
+bewildered foes.
+%%%%
 Hunter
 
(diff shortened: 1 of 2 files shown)


petercordia:
this would pretty much ruin the way wizards are currently played: as powerful magic users with a strong early&end game, and a weak mid game.

alexjurkiewicz:
Removing 4 int does not "ruin" the background. Their early game is essentially unchanged, and 4 int in late game is not a major thing.

However, it is nice to have a background with the opposite stats to berserker. Perhaps Conjurer could get them instead?

PleasingFungus:
I suggested giving the int to cj at the end of the PR description :)

PR #1467, opened 2020-06-28
New Centaur

PleasingFungus:
Centaurs are a classic, flavorful race, but their current implementation is fairly one-dimensional and doesn't encourage the fun tactical gameplay that Crawl specializes in. If players really want to kite monsters using innate movespeed, there are other species that offer more interesting tradeoffs.

Let's swap out that innate movespeed bonus for a new 'Charge' ability. This allows the player to move up to 3 spaces and attack a monster, dealing double damage on a hit. If the player is attacking a monster, their warlike blood rises, and the only cost of the Charge is 10 noise. Otherwise, they get tired enough to have `-Charge` until they rest in place for a while, exactly like Barachian's `-Hop`. In my testing, this was very fun - possibly a bit too strong, but we can nerf it in a few different ways if so.

As an additional perk for the swap, new Centaurs have their apts adjusted a bit, though that's still tentative. See notes at bottom.

## Design notes
Charge differs from Barachian hop in being beam-targeted (not fuzzy smite-targeted), getting a free attack, and being re-usable on attack. It felt very different in play, despite superficial similarities.

Since the ability to escape is generally more important than the ability to fight, we allow re-use of any Charge that attacks (not on hit) so that players can use it without losing an important escape tool. See Power Leap for an example of the other approach.

As with Barachians, `-Charge` only recharging while the player stays in place makes the ability not tediously optimal to use to move around the dungeon.

I considered a few alternate versions of the Charge ability.
- A targeterless version (always charging toward the furthest enemy?) would be interesting in some ways, but could be confusing for players and would only really be useful for melee characters. (The current ability can be used by any character in the 'escape' mode.) 
- Giving a `Charging` status and bonus move speed for a couple of player turns, rather than going off as one action. However, it'd be easy for this ability to 'miss' (if the target steps away while you're en route, say), which makes it much less safe to use offensively. It might also encourage riding around in tight loops to attack an adjacent enemy for bonus damage.

It seemed fun to tweak the Black Knight's Barding to do something that only a centaur item could, so I made it increase charge damage by 50%, removing the old 'partially undead' theming. If we liked that theme, we could always put it back on another unrand.

## Standing questions

I haven't put a lot of thought into the apts, stats, or other innate muts yet. It's possible that the apts should be flattened further. The +1 polearms is critical for theme, however, and must stay.

These centaurs are now quite different from the monsters. One of them probably needs to be rethemed - player centaurs could perhaps become 'springtaurs' (half springbok, half human!) or something even further away. We could alternately do something sillier - renaming monster centaurs to 'cave centaurs', or player centaurs to 'dale centaurs', 'free centaurs', etc.

It's possible that some save compatibility logic is needed - I did some testing early on, but should double check.


ByrelMitchell:
Bikeshedding:

Possibly Rhinotaurs; rhinos are noted for charge-based tactics.

PleasingFungus:
Closing this while I work on a new version.

PR #1468, opened 2020-06-28
New background: Delver

PleasingFungus:
Shaft traps are fun. What if you had a background that started shafted all the way to D:5, and needed to find their way back up to survive?

To make it fair, you start with a nice set of consumables: a wand of digging, ?mapping, ?fog, !haste, and !invis. In practice, I found that this was *usually* enough to avoid sudden centaur- or bee-based death and to make it back to the top, and that it was a pretty fun and distinctive start. It's quite possible that the skills or starting equipment should be buffed further, to make the background closer in strength to others.

The theme is tentative. I initially had the background start in Temple as a bigger perk, but the variance in viability due to starting location (D:4-7) was quite high, and temple entrance vaults weren't designed as entrances - something like the giant transparent spiral would be a nightmare to start in. It also strongly encourages startscumming for players looking to start with a specific god. If we were OK with making the background's temple entrance always generate at D:5 exactly and maybe make an initiate-specific entrance vault, we could revisit that idea, but I'm not extremely positive about it.

TODO: ban this background in sprint.


ebering:
Some initial observations based on playing a few HuDe, NaDe, BaDe, and SpDe from start to lair. I don't have strong suggestions from these at the moment.

1. This feels super cool; radically different from Wn start. Very very exciting opening.
1. The starting package feels too good. Maybe remove ?mapping or !invis. Especially leaning towards the latter, I found that starting short blades and using !invis to get a few opportune high XP stabs quickly got my characters to "progress parity" between XL and dungeon floor. I rarely felt the need to ascend above D:3 before "starting in earnest". Even without !invis I never used the whole kit to get online and to a place where the character felt stable. The whole kit together was also enough to tackle an ossuary and a sewer, after which I was definitely at parity.
1. I get a strong (possibly tedious optimal) incentive to take an exploration-piety god if I luck into early temple: the process of climbing grants piety and someone like Nemelex (with the early card gift) really shines absurdly here. That's cool (piety conducts really mattering!) but it drifts in the no-brainer direction and leaves the delver who's ready to progress with an incentive to go all the way up to D:1 to get the piety for exploration even after feeling stabilized as a character. By that point, D:1 isn't threatening. This incentive still exists for normal starts that get shafted to near an exploration piety altar, but the incentive is much greater for the Delver.

PleasingFungus:
1. Thanks for the feedback!

2. Good point re invis - I was thinking of it as purely an escape tool, but early invisibility is definitely very strong offensively. Do you think it'd be too strong to have a ?fear instead?

3. I don't feel strongly about the piety thing - I generally left d:5 and d:4 too quickly to find temple, but certainly that early piety would be quite strong. We could try doing something like forcing delver temple to D:5+, though seed stability would probably be a problem for that.

I'm not sure if the problem here is with Delver or with Nem, though...

RojjaCebolla:
>     3\. leaves the delver who's ready to progress with an incentive to go all the way up to D:1 to get the piety for exploration even after feeling stabilized as a character

But you have to clear to D1 anyways to prep for orbrun.  And you may as well do this at the start of the game, when those monsters are worth at least a little bit of exp, rather than go seek higher level challenges and clear to D1 later on.

crawl-ref/source/ng-init.cc
```diff
@@ -2,7 +2,7 @@
  * @file
  * @brief Initializing non-player-related parts of a new game.
 **/
-/* TODO: 'you' shouldn't occur here.
+/* TODO: 'you' shouldn't occur here. (Except for Initiate?)
```
RojjaCebolla, line 5: update role name

PleasingFungus: Good catch! Actually this line should just be deleted - it referred to the Temple location setting that Delver doesn't do.

Diojeanne:
Guaranteed temple on D:5 doesn't seem so bad to me on its own, but paired with the starting kit (just having these things IDed is powerful) it seems like the background has two powerful core ideas that don't really need to be in the same background: A start with a lot of non-wand consumables in inventory/IDed, and a start for devoted temple-divers. 

Would you consider breaking into two new backgrounds? A scroll/potion-heavy start that begins on D:1 and another background (archaeologist?) which starts on D:5 with a guaranteed D:5 temple. Could give an ID scroll, a whip, and a full set of clothes to start. A full set of clothes is a pretty desirable start but it's still many levels (usually) before it reaches full benefit. Gloves at least seem thematically essential.

PleasingFungus:
It's a reasonable suggestion, but I'm not interested in having any 'temple-scumming start' at all. If we wanted to let you start with a god of your choice, we could just... let you choose a god, rather than repeatedly quitting and restarting?

PR #1470, opened 2020-06-29
Give Artificers a choice of starting weapon

mainiacjoe:
Artificers have neither an in-game reason to start with a particular weapon (UC for Tm, dagger for En) nor a thematic reason (dagger for Assassin).  The short sword they start with in 0.25 is an example of WCHOICE_PLAIN, so give that to Ar and take the explicit short sword from their starting inventory.

As far as I can tell this is the only change needed in the code.  The Artificer in-game description does not refer to the starting weapon.

This is my first pull request so please let me know how I ought to do things differently as needed.


PleasingFungus:
Short sword is probably the weakest of the starting weapons - giving weapon choice would be a significant buff, though probably not overwhelming.

Cons:
- Adds another choice on startup, which clutters the game loop.

Pros:
- Pushes players towards understanding this background as "a melee start with some wands", rather than "a wand start".

I'm undecided but open to others' opinions.

mainiacjoe:
I looked at quantifying the skill points an Ar would gain by starting with their species' best weapon skill (excl. Stv) instead of SBl, if they choose in the early game to switch to a found weapon.  Ar only gives one level of weapon skill.  The values below are the skill points started with in SBl plus the skill points needed to reach 1.0 in the best of the remaining weapons skills (even if it is equal to or worse than SBl, on the lines that short sword is the worst plain weapon).

- For +3 Evo, Sp has SBl best already and would lose 130, whereas DD loses 84 SP
- For +2 Evo, Ko has SBl best already and would lose 134.
- For +1 Evo, DE has SBl best already and would lose 109, Fo and Op lose 100, Fe gets UC regardless, and Ba loses 94.
- For +0 Evo, HO loses 80, Mf loses 96, Dr, Hu, and Na lose 100, Ha has SBl best already and would lose 121, Ds and Te lose 101.
- For -1 Evo, Mi loses 94, Gr, Ce, and VS lose 92, Gh loses 84, and Vp and Dg have SBl best already and would lose 101.
- For -2 Evo, Og loses 101, and Mu loses 106.
- For -3 Evo, Tr loses 85.
- Gn loses 200 by abandoning SBl.

Ignoring species such as Sp, Ko, Ha, etc. which might actively choose SBl anyway, it looks like characters spend an average of 95 skill points to switch to a better weapon.  If short blade is indeed the worst plain weapon, then this is the right decision for most species, meaning that the background has built into it the need to make the strategic choice to stall character growth in the early game for a better mid-to-late game.  If Ar is meant to be a late blooming background, fine, but starting with a short sword doesn't really signal this, especially since Ar starts with skill training in SBl.

It seems that we can signal two truths about Ar to the new player with the starting inventory and skills.
1.  As you suggested, communicate that it is a melee background that starts with wands.  Giving them the same choice of weapons as  Mo, Be, AK, CK, Sk, and Wr (this pull request) does this.
2.  Communicate that you need to pick and train a weapon ASAP.  Giving Ar a uniquely crappy weapon (a club) and no weapon skill does this.  Even the new player will quickly realize that there aren't enough charges and will naturally pick up a weapon and start training it, but without losing the skill points they started with in SBl and without being forced to waste charges on rats before finding a weapon.  Since they start with leather armor, give them 50 skill points of Armor skill, or even 1 or 2 in Armor and Dodging based on aptitudes like AK and CK get.

Starting with a short sword and 50 SP in SBl does a poor job of communicating either of these ideas.  I know how to code option 2 if you decide it's better--I picked a simple project for my first pull request.

RojjaCebolla:
Alternate solution: just start with 1 skill point in **every** weapon skill.

This could be generalized to all backgrounds, imo: getting 1 or 2 points from { SK_WEAPON, 1 } along with WCHOICE_PLAIN is not terribly interesting.  Just start the character with some weapon or another, and some skill points in the other skills to reduce fretting about wasted experience.

Diojeanne:
1 skill point in every weapon would make a new player's skill screen a mess from the get-go.

I don't think there's any effective way to implicitly communicate to a player that they ought to consider their weapon choice (give them a club, they'll pick up a whip or flail every time), so an explicit choice is warranted.

mainiacjoe:
Closed in favor of PR#1484.  I realize now that I could have edited this one instead of making a new one.

PR #1471, opened 2020-06-30
Fixed TSO stabbing comment

ByrelMitchell:
TSO blocks all stabs, not just on intelligent monsters.


PleasingFungus:
Good catch, thanks!

PR #1473, opened 2020-07-04
Replace Centaurs with Palentongas

PleasingFungus:
Centaurs are a classic, flavorful race, but their implementation was
one-dimensional and didn't encourage the fun tactical gameplay that
Crawl specializes in. If players really want to kite monsters using
innate movespeed, there are other species that offer more interesting
tradeoffs, and palentongas offer a more exciting take on fast movement.

Palentongas are a two-armed, four-legged, scaled mammal species the
size of a horse. Their special power is the ability to curl into a
ball and launch a rolling charge at enemies, dealing extra damage
proportionate to the distance traveled. This is a dynamic way to
start fights (as long as you don't mind some extra noise!), and in
larger engagements can let the player ping-pong between enemies for
huge damage. It's also extremely fun against phantoms.

Palentongas have flat melee apts, slightly weak spell apts (with
some preference given to schools that seem more interesting), and
fairly weak ranged apts. Players with good throwing or slings skill
will be forced to ask themselves "do I do more damage with 2-4
ranged attacks than I do by charging?", which doesn't sound very fun.

Like centaurs, they have +3 base AC and extra HP, along with an aux
attack - an upgraded tail slap instead of hooves. The tail slap does
the same damage as hooves 3 (10), but has a coinflip chance instead of
being guaranteed, since I didn't want to emphasize melee *too* much.
That's tentative, though.

Stats skew towards strength and away from dex, to give the 'large
race' feeling. As with centaurs, Palentonga have deformed body and
bad armour/dodging apts.

In the Lore, Palentonga are an up-and-coming power, displacing the
orcs, trolls, and ogres of the eastern wastes. Their generals ride
elephants into battle, goading them forward with spiked ankuses.

Art is pending.


ebering:
Some initial feedback based on some PaNe, PaBe, PaFi, and PaWr played to lair:

Gameplay feedback:

- The noise from the charge attack is really noticeable early on. With the PaNe I was trying to use it when caught after a fight without mp to quickly dispatch and got into trouble with it. Good.
- Charge + blink is really great playfeel, as is Charge + Borgs vile clutch. Haven't yet tried charge w/confusing abilities but I expect it to be similarly fun.
- The targeting seems mostly unnecessary imo, I am almost always targeting nearest. It might also be nice to incorporate charge into autofight somehow; currently I've macro'd "charge nearest" to the newly free c key.
- This could maybe go in the flavor section, but my PaFi and PaBe both greatly benefitted from the availability of boots over bardings in getting strong defenses. (Unified) bardings might be a better fit.

Trivial optimal feedback:

- Charge can target firewood, which makes fighting in a place where blinking to a plant is a nice escape option incentivized; since charge is supposed to be a combat ability, it should probably not target firewood.

UI/Bug Feedback:

- The targeter displays funny in console: no range shading, the endpoint highlighting doesn't seem quite correct, and there is no "out of range" message when I select something too far away.

Flavour/Aptitude Feedback:

- Flat apts are generally preferred but if Palentonga generals wielded ankuses is the lore then maybe they should get +1 aptitudes. (I'm not super serious here.)
- On the varied magic apts, while I dig that scaly Pangolin type things should have some affinity for earth, a shakeout from the magic overhaul was a desire to make necromancy more "evil melee support"; I'm not sure how much more elaboration that idea will get in 0.26, I hope to give it some, but with the remaining necro damage spells other than Pain made melee range it fits better with a species that closes to melee quickly than does Earth magic, so a 0 necro -1 earth is preferred imo.
- Elaborating on the previous, the aptitudes are meant to tilt towards melee, and I think that to the extent that Pa is good at magic that tilt should match the apts given to ranged: give Hex, Chrm, Necro, Tloc, and Tmut apt 0 and the rest -2. (Summons, while better when you fight along side them, get in the way of charging and the school doesn't also give you cool melee range spells.) Let me know what you think about this!

PleasingFungus:
Hey, thanks for the feedback!

> The targeting seems mostly unnecessary imo, I am almost always targeting nearest. It might also be nice to incorporate charge into autofight somehow; currently I've macro'd "charge nearest" to the newly free c key.

I'd considered a version that always charged at the nearest enemy (no targeter), but it seemed hard to communicate to the player and also quite tactically limiting (wrt charging back and forth.) Do you think it'd be better to set up a preview UI for that instead?

> This could maybe go in the flavor section, but my PaFi and PaBe both greatly benefitted from the availability of boots over bardings in getting strong defenses. (Unified) bardings might be a better fit.

If their defenses are too strong, there are various ways to tweak that. I don't think bardings are particularly fun or something that we want to have on more than one species - do you?

> The delay is equal to attack speed, which seems weird. It should be a fixed delay or scale with distance. Quickblade charges covering 3 cells in 3 auts (plus attacking) is overpowered (especially in conjunction with dispersal or disjunction or a dispersal trap).

That is a bug. Oops. It should definitely be a fixed 10 aut delay.

> Charge can target firewood, which makes fighting in a place where blinking to a plant is a nice escape option incentivized; since charge is supposed to be a combat ability, it should probably not target firewood.

Sure, that seems fine. Any opinion on charging at friendlies?

> Charge works while held in a web or net.

It's supposed to give you an escape attempt and fail if you don't wiggle out, same as blink, hop, etc. Is that a problem?

> The targeter displays funny in console: no range shading, the endpoint highlighting doesn't seem quite correct, and there is no "out of range" message when I select something too far away.

Oops, I didn't really test console. I can take a look at that.

> Flavour/Aptitude Feedback...

These are great points. I'm fairly torn on how much I want to make the apts push the species towards the melee playstyle, beyond the advantages already provided by the charge and the aux attack. I'll think about it and hopefully push some of this in the next changeset - thanks!

ebering:
> I'd considered a version that always charged at the nearest enemy (no targeter), but it seemed hard to communicate to the player and also quite tactically limiting (wrt charging back and forth.) Do you think it'd be better to set up a preview UI for that instead?

I hadn't heard the "preview UI" term before but that's something that the positional spells with interesting stencils need! Independent of Charge being targeted or not this tech would see good use throughout the crawl codebase.

I guess I wasn't tactically sharp enough to set up good multi-charges without (ab)using mindelay quick blade charges. I agree that it's limiting in theory, but in practice monsters swarming and generally only coming at me from one side meant the opportunity to do so was rare.

> If their defenses are too strong, there are various ways to tweak that. I don't think bardings are particularly fun or something that we want to have on more than one species - do you?

I find bardings fun; they have a special loot feel different from other aux armour which can be frustrating at times but very exciting others. They give an extra incentive to fight the uniques who bear bardings.

Loot feel alone isn't enough to justify them though, so I'll make the rest of my case.

**Mechanics:** Deformed body is a limitation to the benefit from body slot armour, this is offset for Ce and Na by the existence of bardings; other body slot restricted species are either good at dodging (Sp, Op, Fe),  can get full benefit from their available body slot types and have big HP to carry them to that point (Tr, Og), or get innate AC (Dr). Longer term (after playing a few more longer Pa games) the lack of defense begins to show as the innate +3 AC that Pa get becomes less good since it doesn't scale; shorter term the availability of early boots allows this to be offset faster than for Na or Ce at a place in the power curve where Pa doesn't feel like it needs it.

**Flavour:** Pa have four legs, why don't they get two boots slots! Endless "If a Palentonga wore boots would it wear them like this or like this memes" (unsure if pro or con on this one). Palentonga have scales so unified bardings make some sense for the lightning scales (if vague sense). Allowing them to wear boots would put Pa in a weird flavour place as a species with a weird lower part but normal boots. Of course all of this could be changed by making them have two legs instead of four, but see the mechanics paragraph for more important reasons.

> That is a bug. Oops. It should definitely be a fixed 10 aut delay.

I think it should be a current move speed delay; no speedy charges while slow!

> > Charge can target firewood, which makes fighting in a place where blinking to a plant is a nice escape option incentivized; since charge is supposed to be a combat ability, it should probably not target firewood.
> 
> Sure, that seems fine. Any opinion on charging at friendlies?

This should also be a no go; see e.g. charging your lightning spire to retreat, or charging butterflies to zoom around in some kind of passage of Golubria. 
 
> It's supposed to give you an escape attempt and fail if you don't wiggle out, same as blink, hop, etc. Is that a problem?

It's possible I wasn't paying close attention. I was fighting a jumping spider so I may have escaped, charged, and then been hit and webbed by the spider. I'll keep an eye out on this more carefully.
 
> > Flavour/Aptitude Feedback...
> 
> These are great points. I'm fairly torn on how much I want to make the apts push the species towards the melee playstyle, beyond the advantages already provided by the charge and the aux attack. I'll think about it and hopefully push some of this in the next changeset - thanks!

Yeah, the overall depth provided by flatter apts is generally richer. That said, a bit of apt tilting to encourage/discourage playstyle or enhance flavor is good. Maybe -2 vs. 0 is too far, but if there's to be a tilt I'm strongly in favor of the schools I identified receiving the better of the tilt.

PleasingFungus:
> I guess I wasn't tactically sharp enough to set up good multi-charges without (ab)using mindelay quick blade charges. I agree that it's limiting in theory, but in practice monsters swarming and generally only coming at me from one side meant the opportunity to do so was rare.

Yeah, I think that'll be fairly uncommon. It's so cool when it happens, though... well, I'll keep the targeter version for now but think about it.

> I think it should be a current move speed delay; no speedy charges while slow!

Sure, 10 aut times (fast or slow). Then we get into the issue of extremely slow attacks (e.g. low-skill dark maul attacks), so we might need to do max(10, weapon delay)... I remember something else (jump attack?) had this exact problem, but don't recall how it solved it.

> I find bardings fun; they have a special loot feel different from other aux armour which can be frustrating at times but very exciting at others. They give an extra incentive to fight the uniques who bear bardings.

Sure, but less incentive to care about other boots, of which we have many more!

I agree that variety is good, which is why it's fun for nagas to have the barding game. But do two species need to have it?

>  Longer term the lack of defense begins to show as the innate +3 AC that Pa get becomes less good since it doesn't scale; shorter term the availability of early boots allows this to be offset faster than for Na or Ce at a place in the power curve where Pa doesn't feel like it needs it.

Lightli was saying the same thing. Seems easy to fix without needing bardings. We can remove deformed body entirely (and nerf something else if we feel like it) or make scales grow in over time, like nagas (and buff something else, e.g. armour skill).

I'm inclined to just remove deformed body, since I don't think it's core to the concept or the flavor of the race. It's mostly just a leftover from centaur.

> Flavour: Pa have four legs, why don't they get two boots slots! Endless "If a Palentonga wore boots would it wear them like this or like this memes" (unsure if pro or con on this one)

I do love memes (<- quote this), but right now there's no particular reason that palentongas need to be four-legged. We could just make em ogre sized pangolins and call it a day.

PleasingFungus:
I've pushed everything to the delpal branch except for the barding thing, since we're still discussing that.

ebering:
> > I guess I wasn't tactically sharp enough to set up good multi-charges without (ab)using mindelay quick blade charges. I agree that it's limiting in theory, but in practice monsters swarming and generally only coming at me from one side meant the opportunity to do so was rare.
> 
> Yeah, I think that'll be fairly uncommon. It's so cool when it happens, though... well, I'll keep the targeter version for now but think about it.
> 
> > I think it should be a current move speed delay; no speedy charges while slow!
> 
> Sure, 10 aut times (fast or slow). Then we get into the issue of extremely slow attacks (e.g. low-skill dark maul attacks), so we might need to do max(10, weapon delay)... I remember something else (jump attack?) had this exact problem, but don't recall how it solved it.

Wu Jian had this issue as well, the solution is a move speed in attack speed randomized chance at getting the attack. That could work here, but if you go with the "target nearest" approach then the "charge for faster attacks with slow weapons" problem goes away and charge gets a bit of an extra incentive with a slow weapon: you get to make this one attack a little faster!
 
> > I find bardings fun; they have a special loot feel different from other aux armour which can be frustrating at times but very exciting at others. They give an extra incentive to fight the uniques who bear bardings.
> 
> Sure, but less incentive to care about other boots, of which we have many more!
> 
> I agree that variety is good, which is why it's fun for nagas to have the barding game. But do two species need to have it?
> 
> > Longer term the lack of defense begins to show as the innate +3 AC that Pa get becomes less good since it doesn't scale; shorter term the availability of early boots allows this to be offset faster than for Na or Ce at a place in the power curve where Pa doesn't feel like it needs it.
> 
> Lightli was saying the same thing. Seems easy to fix without needing bardings. We can remove deformed body entirely (and nerf something else if we feel like it) or make scales grow in over time, like nagas (and buff something else, e.g. armour skill).
> 
> I'm inclined to just remove deformed body, since I don't think it's core to the concept or the flavor of the race. It's mostly just a leftover from centaur.

The way I see it is there are currently 4-ish defensive item-use paradigms for crawl species:

1. Normal body armour (with some aux limitations possibly, most species are here)
2. Only skins and robes (Sp, Tr, Og) with some form of alternate defense (dodge + movespeed or HP)
3. (almost) no armour items with some innate tradeoff (Dr, Op, Fe)
4. Deformed body plus barding (Na, Ce)

Currently Pa is in none of the above. Making them Ogre sized Pangolins with the attendant size restrictions (and abilities, would they be big enough for giant clubs?!) gives four species in category 2 and one in category 4. I'm not sure whether having a few different species option in each category or having Na be the only category 4 species is better for variety. Definitely the species within each category are strongly distinct, and Pa is very different from Na in playfeel.

I've already laid out my subjective case for the loot feel of category 4; I think that category 4 is also a better fit for charge attack mechanically, when compared to category 2.

With a -2 armour apt, finding sufficient AC to feel good charging into battle before good scales armour has dropped in the midgame seems like it will be scarce. Since Pa don't have Tr or Og HP and poor other defensive apts, I worry that it will be rare to find sufficient defenses to want to charge in that stretch of midgame where there aren't yet good scales but Pa's innate AC isn't that relevant. This can be adjusted in other ways of course to get Pa to fit in category 2.

But! I think this gap is bridged better by giving them the deformed body + barding combo to keep Pa charging through the game. The previously mentioned falloff point is around the time when acquirement scrolls and barding dropping enemies appear, to better keep Pa charging through the game.

Both of these are based on speculation; since this is an experimental we could try it both ways! To conclude, for me what's more important than making the case for bardings is making the case that Pa currently create a fifth category of armour item interaction which doesn't feel as good as categories 1-4.

> I do love memes (<- quote this)

Done

ByrelMitchell:
> but if you go with the "target nearest" approach then the "charge for faster attacks with slow weapons" problem goes away and charge gets a bit of an extra incentive with a slow weapon: you get to make this one attack a little faster!

Note the flip side; if you have a weapon faster than your move delay (the usual case in the midgame) then you're facing a sizable penalty for charging. Wu Jian attacks solve this by executing 'div_rand_round(action_delay, attack_delay)' attacks. So for the usual case of 0.7 mindelay, 1.0 move delay, you have a 42% chance of doing 2 attacks, and a 58% chance of 1.

sdynet:
I like that. But I feel sorry that the archer-like species are disappearing. After you add the Palentongas separately, don't you want to make the Ce a more archer-like species? I'm looking forward to your archer-like species. :)

PleasingFungus:
Merged.

PleasingFungus:
(@sdynet , your wish is granted! )

RojjaCebolla:
NB: a rolling charge is no longer extremely fun against phantoms.

PleasingFungus:
Life is cruel.

PR #1480, opened 2020-07-06
Adjust Sprint III cloud generators

EugeneAbramchuk:
Fixes issue 12302. Existing cloud generators kill sleeping monsters in several places across the map within the first few hundred turns. This tones down aggressive cloud spawn, while keeping the danger in the affected rooms.

### DIFF
--- a/crawl-ref/source/dat/des/sprint/sprint_mu.des
+++ b/crawl-ref/source/dat/des/sprint/sprint_mu.des
@@ -516,7 +516,7 @@ KMONS: i = hydra
 KMONS: j = patrolling spriggan ; q:5 dart ego:atropa | q:5 dart ego:datura
 KMONS: k = patrolling spriggan druid
 KMONS: q = electric eel
-KFEAT: q = deep_water
+KFEAT: q = shallow_water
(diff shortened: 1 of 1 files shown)


EugeneAbramchuk:
https://crawl.develz.org/mantis/view.php?id=12302

There is one remaining affected vault (fire and ice Swamp), where spawned dragons could be endangered by the clouds, but the random nature of monster placement calls for minor vault redesign, rather than cloud generator change.

rawlins:
Thanks for the PR, this looks good to me! For future reference we prefer commit messages to be wrapped at 72 columns.

rawlins:
also, how would you like to be listed in `CREDITS.txt`? (at least, I'm not spotting any names that match you in there, sorry if you are in there and I'm missing it.)

EugeneAbramchuk:
Oh yeah, I realised the wrapping was amiss shortly after I submitted the PR (+ it was in that git commit conventions you shared with me). Will configure vim to help with that.

"Eugene Abramchuk" would be super nice, thank you. This is my first contribution to the project.

PR #1485, opened 2020-07-10
Fixes several issues with the xv monster evasion chance calculation.

ByrelMitchell:
Note that this modifies PR #1464 by Pleasing Fungus, not master.

* Generalizes '_apply_defender_effects' to all post-roll modifiers (including things like a confused player or inaccuracy.)
* Eliminated randomizing EV, since that's only done for monster attacks
* Added iterating over the mhit roll, and enumerating hits there.
* Moved applying post-roll modifiers to post-roll to properly handle repel missiles (whose post-roll modifier is a function of the rolled mhit.
* Switched auto-hit calculation to use Bayes theorem, because I think it's a lot more readable.
* Factored out post-roll modifiers in the attack classes into their own functions.
* Made the lighting modifiers in attack.cc use the relevant consts.

### DIFF
--- a/crawl-ref/source/attack.h
+++ b/crawl-ref/source/attack.h
@@ -13,8 +13,8 @@ const int HIT_WEAK   = 7;
 const int HIT_MED    = 18;
 const int HIT_STRONG = 36;
 
-const int BACKLIGHT_TO_HIT_BONUS = 12;
-const int UMBRA_TO_HIT_MALUS = -8;
+const int BACKLIGHT_TO_HIT_BONUS = 5;
+const int UMBRA_TO_HIT_MALUS = -3;
 
 class attack
 {
@@ -98,7 +98,7 @@ public:
     // To-hit is a function of attacker/defender, defined in sub-classes
     virtual int calc_to_hit(bool random);
     int calc_pre_roll_to_hit(bool random);
-    int post_roll_to_hit_modifiers(int mhit);
+    virtual int post_roll_to_hit_modifiers(int mhit, bool random);
 
     // Exact copies of their melee_attack predecessors
     string actor_name(const actor *a, description_level_type desc,
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -3884,49 +3884,49 @@ static void _add_energy_to_string(int speed, int energy, string what,
 }
 
 /**
- * Calculate effects on the player's to-hit from a defending monster.
+ * Calculate post-randomization effects on the player's to-hit.
  *
- * @param base_to_hit   The to-hit with only attacker-side effects applied.
+ * @param to_hit        The actual randomized to-hit.
  * @param mi[in]        Player-visible info about the monster in question.
  * @param melee         Whether the attack is melee or ranged.
  */
-static int _apply_defender_effects(const monster_info& mi, int to_hit, bool melee)
+static int _post_roll_to_hit_modifiers(const monster_info& mi, int to_hit, bool melee)
 {
     int effect = 0;
-    // Lighting effects. Subtracting these from EV is weirdly equivalent to adding
-    // them to the post-roll to-hit (the to_hit we have here is pre-roll).
+
+    effect -= 5 * you.inaccuracy();
+
+    if (you.confused()) effect -= 5;
+
+    // Lighting effects.
     if (mi.is(MB_GLOWING)       // corona, silver corona (!)
         || mi.is(MB_BURNING)    // sticky flame
         || mi.is(MB_HALOED))
     {
         effect += BACKLIGHT_TO_HIT_BONUS;
-    } else if (mi.is(MB_UMBRAED) && !you.nightvision())
+    }
+    else if (mi.is(MB_UMBRAED) && !you.nightvision())
         effect += UMBRA_TO_HIT_MALUS;
 
     if (melee) {
-        // Duplicates melee_attack::calc_to_hit().
-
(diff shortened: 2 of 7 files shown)


PleasingFungus:
Were you planning on rewriting this to avoid leaking information with `monster_at`? Either way, could you please summarize the changes in this PR? There's a lot of stuff going on in one commit.

Zireael07:
Working implementation of _what_? PR title is super unclear...

PleasingFungus:
This is a variant of https://github.com/crawl/crawl/pull/1464 based on conversations had in Discord. I agree it would be nice to have a clearer description.

ByrelMitchell:
Yeesh. I didn't realize this would show up in the normal PR list.

I won't have time to work on this till Monday; I can fix and clarify then.

PleasingFungus:
Sounds good!

crawl-ref/source/attack.h
```diff
@@ -13,8 +13,8 @@ const int HIT_WEAK   = 7;
 const int HIT_MED    = 18;
 const int HIT_STRONG = 36;
 
-const int BACKLIGHT_TO_HIT_BONUS = 12;
-const int UMBRA_TO_HIT_MALUS = -8;
+const int BACKLIGHT_TO_HIT_BONUS = 5;
+const int UMBRA_TO_HIT_MALUS = -3;
```
PleasingFungus, line 17: Hm, wasn't this a lot smaller than the old values effectively were? Or am I misremembering?

crawl-ref/source/describe.cc
```diff
@@ -3884,50 +3884,50 @@ static void _add_energy_to_string(int speed, int energy, string what,
...
+    else if (mi.is(MB_UMBRAED) && !you.nightvision())
         effect += UMBRA_TO_HIT_MALUS;
 
     if (melee) {
-        // Duplicates melee_attack::calc_to_hit().
-
+        // Duplicates melee_attack::post_roll_to_hit_modifiers().
         if (you.duration[DUR_CONFUSING_TOUCH])
-            effect += you.dex() / 2;
+            effect += maybe_random_div(you.dex(), 2, false);
```
PleasingFungus, line 3914: This is literally `you.dex() / 2`, right?

crawl-ref/source/describe.cc
```diff
@@ -3884,50 +3884,50 @@ static void _add_energy_to_string(int speed, int energy, string what,
...
-static int _apply_defender_effects(const monster_info& mi, int to_hit, bool melee)
+static int _post_roll_to_hit_modifiers(const monster_info& mi, int to_hit, bool melee)
 {
     int effect = 0;
-    // Lighting effects. Subtracting these from EV is weirdly equivalent to adding
-    // them to the post-roll to-hit (the to_hit we have here is pre-roll).
+
+    effect -= 5 * you.inaccuracy();
+
+    if (you.confused()) effect -= 5;
```
PleasingFungus, line 3899: It seems like we can factor out all of the post-roll effects that don't depend on the defender to methods on attack, melee_attack, and ranged_attack, right? (I can do this myself after merging your PR into mine, if you prefer.)

Then we'd only have lighting effects and rmsl here.

crawl-ref/source/describe.cc
```diff
@@ -3936,46 +3936,37 @@ static int _apply_defender_effects(const monster_info& mi, int to_hit, bool mele
...
-    // final hits are 97.5% of rolled hits + the 2.5% of misses that convert into hits
-    const int non_automissed_hits = (100 - MIN_HIT_MISS_PERCENTAGE/2) * base_hit_perc / 100;
-    const int autohit_misses = (MIN_HIT_MISS_PERCENTAGE/2) * base_miss_perc / 100;
-    dprf("to_land: %d, ev: %d, hits: %d, total rolls: %d, base_hit: %d, base_miss: %d, non_automiss: %d, autohit: %d",
-        to_land, ev, hits, total_rolls, base_hit_perc, base_miss_perc, non_automissed_hits, autohit_misses);
-    return non_automissed_hits + autohit_misses;
-}
 
+    double hit_chance = ((double)hits) / to_land;
+    // Apply Bayes Theorem to account for auto hit and miss.
```
PleasingFungus, line 3963: Nice simplification!

crawl-ref/source/attack.h
ByrelMitchell: Original code was:
`        if (defender->backlit(false))
            mhit += 2 + random2(8);
        else if (!attacker->nightvision()
                 && defender->umbra())
            mhit -= 2 + random2(4);`
Means of 5.5 and 3.5, so we can't precisely model them with a constant. I just used the same constant values that you'd used in the attack.cc code, which was rounding them both down.

crawl-ref/source/describe.cc
ByrelMitchell: No, it's you.dex()/4. maybe_random_div in random mode puts out a random number that's between 0 and ceil(nom/denom), so half of that is nom/2/denom.

But I figured using the maybe_random_div's built-in non-random mode was good because it allows future devs to easily change what we use as the int approximation of the mean of a distribution. (Notably, the existing code is not actually the nearest int to the actual mean in all cases; sometimes rounding up would be more accurate. But this keeps it standardized, and a change can happen in one place if anyone decides to improve it.)

crawl-ref/source/describe.cc
ByrelMitchell: Sure, that works. I'll let you handle that. :)

PR #1487, opened 2020-07-13
Charming items (aka Charms overhaul)

ebering:
This PR implements [the planned charms overhaul](https://github.com/crawl/crawl/wiki/Charms-Overhaul). It is currently reasonably close to the design doc, with some of the outstanding questions there answered.

### Specifics (where different from the design doc or where the design doc wasn't concrete)

- Infusion was removed outright, didn't seem like it would be a good amulet
- Song of Slaying reflavoured "Wereblood" and moved to Tmut (replacing S2S in the starter book), gives noise on kill instead of continuously.
- Spectral Weapon becomes the "spectral" ego found on staves and clubs and (rarely) 2-handed randarts; flavored as bonding with the tree spirit still contained in the weapon's wood.
- Ring of Flames becomes an amulet of flames, giving the spell's effect with the cloud ring scaling with evo.
- Darkness becomes a scarf of shadows, reducing LOS and giving a spell de-enhancer
- Ozocubu's armour works with all armour, the effect scales down with ER
- Excruciating Wounds gives -Wiz and prevents unwielding for the duration.
- Death's Door is single-school Level 9

### Outstanding Issues/Tasks

- [x] The amulet of flames might be too flashy for a mundane item and be better as an unrand. Discuss.
- [x] Should artificers get a spectral club to start? (Pro: uses evo, cool, good flavour, unique, clubs aren't that good; Con: background is already strong with wands)
- [x] Is Tm too weak without S2S as a crutch, and if so what should be done (if anything)?
- [x] Beastly Appendage is "good to always cast if you fight in melee", fix it. (This may address the above).
- [x] Finish the WIP battlesphere adjustments.
- [x] Double check save compat with skill removal.
- [x] Do the tiles (see next checklist):

### Needed tiles

- [x] Shadows ego icon
- [x] Amulet of flames icon
- [x] Spectral ego icon
- [x] Wereblood spell icon


sdynet:
IMHO the shadow scarf doesn't have to have a de-enchanter. Even without such shortcomings, shadow scarf will never be overwhelmingly superior because other scarves are attractive and powerful enough.

ebering:
@sdynet Interesting to hear that; from other places I've heard that shadows will be a highly valued scarf, even with the enhancer. I think it's something that will need to be watched in play testing.

PleasingFungus:
i'd think that shadow scarf would be weakest for '''' casters '''' and strongest for people who don't care about spell enhancers either way (since the former are more likely to have ranged attacks). haven't thought super hard about it though

PleasingFungus:
could excruc go down a level with the changes?

ebering:
> could excruc go down a level with the changes?

I don't think so; I used EW on both of my recent trunk games with a lot of stairs, digging some scream-closets &c. In local tests it felt similarly strong even with -Wiz; since necro skill improves the pain damage the player is incentivized to keep training necro if they use the spell, overcoming the wizardry penalty for their necro spells. Since it's now easier to bring on line single-school and pain is very strong my main worry is that it should maybe go up a level (even with -Wiz)...

ebering:
@PleasingFungus I saw on discord you mentioned you were planning on making Beastly Appendage into a form; I was thinking similarly. Do you have details/implementation in progress?

My sketch is: meld aux slots and weapon, give lvl 1 aux attacks in each slot. A bunch of auxes instead of one big one has a better interaction with the slay bonus from Wereblood, and so more incentives to use it early despite the noise.

PleasingFungus:
All I have is a suggestion from Hellmonk, who posed:

>  I would suggest making it a unique form that melds body slot, gives 2 ac and equivalent damage to talons 3

I like your idea of an omniaux form, but I wonder if it'd still encourage og or other severely restricted aux slot races to cast it throughout the game.

ebering:
> I like your idea of an omniaux form, but I wonder if it'd still encourage og or other severely restricted aux slot races to cast it throughout the game.

Hm. Thats why I suggested melding weapon (and shield) too. Perhaps a synthesis of these ideas: "Monstrous Form" (name pending), that melds all non-jewellery slots, gives lvl1 omniaux and 2 ac.

RojjaCebolla:
correction: hedge wizard has a "magic _stab_ bag"

crawl-ref/source/mon-death.cc
```diff
@@ -2028,13 +2029,15 @@ item_def* monster_die(monster& mons, killer_type killer,
 
     // Adjust song of slaying bonus. Kills by relevant avatars are adjusted by
     // now to KILL_YOU and are counted.
-    if (you.duration[DUR_SONG_OF_SLAYING]
+    if (you.duration[DUR_WEREBLOOD]
```
rawlins, line 2032: weird place for an effect like this, it would be better if this were somehow done not as a side effect. (E.g. just keep track of deaths in a prop and do the howling in player_reacts.)

rawlins: (Or, I get the flavor of this, but it might be simpler for the player if it just had old SoS behavior and always made noise.)

crawl-ref/source/spl-data.h
```diff
@@ -734,7 +723,7 @@ static const struct spell_desc spelldata[] =
 
 {
     SPELL_BERSERKER_RAGE, "Berserker Rage",
-    spschool::charms,
+    spschool::hexes,
```
rawlins, line 726: doesn't really matter but hexes seems odd here, maybe earth or something?

crawl-ref/source/spl-data.h
```diff
@@ -2201,7 +2157,7 @@ static const struct spell_desc spelldata[] =
 
 {
     SPELL_INJURY_BOND, "Injury Bond",
-    spschool::charms,
+    spschool::hexes,
```
rawlins, line 2160: or maybe necro?

crawl-ref/source/spl-data.h
```diff
@@ -3071,7 +3005,7 @@ static const struct spell_desc spelldata[] =
 
 {
     SPELL_AURA_OF_BRILLIANCE, "Aura of Brilliance",
-    spschool::charms,
+    spschool::hexes,
```
rawlins, line 3008: also odd

crawl-ref/source/spl-summoning.cc
```diff
@@ -2543,61 +2439,22 @@ bool aim_battlesphere(actor* agent, spell_type spell, int powc, bolt& beam)
...
-        zappy(ztype, powc, false, testbeam);
-
-        battlesphere->props["firing_target"] = beam.target;
         battlesphere->props.erase("foe");
-        if (!actor_at(beam.target))
-        {
-            testbeam.fire();
 
-            for (const coord_def c : testbeam.path_taken)
+        // Pick a random baddie in LOS
```
rawlins, line 2444: I'm not sure about this change at all, why is it bad to target whatever the main spell is aiming at, regardless of how that aim is achieved? Randomness just seems to add player lack-of-control here, and I don't see the gain

crawl-ref/source/mon-death.cc
ebering: An earlier drat had the constant noise retained. The constant noise made the spell weaker, and imo too weak. 

In terms of the howl being in `monster_die` as a side effect, that function is where we give out XP, heal on kills, &c. so it's the correct place for an on-death side effect.

crawl-ref/source/spl-data.h
ebering: Picked this because Summon Guardian Golem is Summon/Hex

crawl-ref/source/mon-death.cc
rawlins: I mean, all side effects are bad (and death side effects are a constant source of bugs) so I'm not in favor of adding more, and both of the ones you mention lead to issues too. In fact the xp one combine with skill targets is in major need of (difficult) refactoring because of the crash on disconnect bug.

crawl-ref/source/spl-summoning.cc
ebering: I went with a consistent targeting scheme across all spells instead of the special-case different targeting rules for different spells for clarity of effect. 

The old targeting scheme was (roughly):

- If aimed at a target hit that,
- Else if there's a target in the beam path of the aim hit that,
- Else if the mirrored spell is an explosion hit something it explodes on

And I don't see the gain of adding an "else if the mirrored spell is one of the positional spells hit something that the spell hit" compared to simplifying the targeting scheme to always shoot something.

crawl-ref/source/mon-death.cc
ebering: Hm. I think I'll go back to the old version but make the noise happen on a slay-bonus-in-max chance in `player_reacts`.

crawl-ref/source/spl-data.h
ebering: swapped to conjurations; monster buffs sadly don't have a good thing to put here, but it's not used.

crawl-ref/source/spl-data.h
ebering: Did earth

crawl-ref/source/spl-summoning.cc
ebering: We had some back-and-forth in IRC about improved battlesphere targeting; I'm not convinced its worth the effort and complexity, but I made the simplification commit separate for an easy revert.

PR #1488, opened 2020-07-15
Fix fleeing Harold speech

Goratrix:
Bug reported by Mintice in Discord, the speech shows yells with
at signs around it. I suppose the intention here was to have the
verb randomized as per the _yells_ list.


ebering:
Thanks for the patch!

PR #1489, opened 2020-07-15
Add Pandemonium Lord cloud rings

alexjurkiewicz:
Opening as a draft for feedback.

To do:

- [x] Fix Antaeus
- [x] Check Bai Suzhen is still alright
- [x] Fix chance of cloud ring
- [x] Add ice cloud ring
- [x] Add poison cloud ring?
- [x] Add miasma cloud ring?
- [x] Do any of these rings affect the pan lord / their movement?


ebering:
This is in the draft phase, so I'll hold off on detailed feedback, but some initial reactions

- Cloud rings on melee pan lords sounds cool
- Cloud rings have the problem that they kill allies which is less cool; that limits your cloud types
- Possibly controversial suggestion: make all demons immune to chaos clouds (such flavour)

alexjurkiewicz:
I think this is in a good place, mechanics-wise.

On the code side, I have a question. Should the existence of a monster's cloud ring be exposed to the player via a) spell and/or b) status text? Currently, Bai Suzhen & Asmodeus have a spell, but no visible status. Random panlords get neither. This should be made consistent between the two sorts of monster, but I'm unsure what end state is best. Thoughts?

Otherwise, this is ready for review.

crawl-ref/source/cloud.cc
```diff
@@ -1899,3 +1899,24 @@ void end_still_winds()
...
+{
+    const coord_def pos = mons->pos();
+    const cloud_struct* c = cloud_at(pos);
+    if (c && c->type == cloud)
+        delete_cloud(pos);
+    for (adjacent_iterator ai(pos); ai; ++ai)
+    {
+        if (!cell_is_solid(*ai)
+            && (!cloud_at(*ai)
+                || cloud_at(*ai)->type == cloud))
```
ebering, line 1917: Skip this check, let cloud rings overwrite their own clouds (so they don't appear to sputter).

crawl-ref/source/ghost.cc
```diff
@@ -95,7 +96,6 @@ static spell_type search_order_misc[] =
     SPELL_PETRIFY,
     SPELL_POLYMORPH,
     SPELL_FORCE_LANCE,
-    SPELL_SLOW,
```
ebering, line 98: Don't include a sneaky change like this. Slow checks MR and should stay in the misc spells.

crawl-ref/source/ghost.cc
```diff
@@ -251,6 +253,37 @@ void ghost_demon::set_pan_lord_special_attack()
...
+void ghost_demon::set_pan_lord_cloud_ring()
+{
+    if (brand == SPWPN_ELECTROCUTION)
+        cloud_ring_ench = ENCH_RING_OF_THUNDER;
+    else if (brand == SPWPN_FLAMING)
+        cloud_ring_ench = ENCH_RING_OF_FLAMES;
+    else if (brand == SPWPN_CHAOS)
+        cloud_ring_ench = ENCH_RING_OF_CHAOS;
+    else if (brand == SPWPN_DISTORTION)
+        cloud_ring_ench = ENCH_RING_OF_MUTATION;
```
ebering, line 265: This flavour doesn't fit imo.

crawl-ref/source/cloud.cc
```diff
@@ -1899,3 +1899,24 @@ void end_still_winds()
     env.level_state &= ~LSTATE_STILL_WINDS;
     mpr("The air resumes its normal movements.");
 }
+
+/**
+ * Surround a monster with clouds of a certain type. This also deletes the same
+ * cloud if it's on top of the monster (which will happen as they walk around).
+ */
+void surround_mons_with_cloud(const monster* mons, cloud_type cloud)
```
ebering, line 1907: Could you make this work with an `actor *` and refactor `_manage_fire_shield` in `art-func.h` to use it, so that our cloud rings use a unified codepath?

crawl-ref/source/mon-act.cc
```diff
@@ -2292,18 +2279,28 @@ static void _post_monster_move(monster* mons)
...
+        surround_mons_with_cloud(mons, CLOUD_GREY_SMOKE);
+
+    if (mons->has_ench(ENCH_RING_OF_ICE))
+        surround_mons_with_cloud(mons, CLOUD_COLD);
+
+    if (mons->has_ench(ENCH_RING_OF_DRAINING))
+        surround_mons_with_cloud(mons, CLOUD_NEGATIVE_ENERGY);
+
+    if (mons->has_ench(ENCH_RING_OF_ACID))
+        surround_mons_with_cloud(mons, CLOUD_ACID);
```
ebering, line 2303: This should be replaced with a `static map<mon_ench_type, cloud_type>` that you iterate over to pick the cloud type and update. That iteration logic should be in its own function in `mon-ench.cc` called something like `update_mons_cloud_ring` and used below as well.

crawl-ref/source/mon-ench.cc
```diff
@@ -317,14 +304,42 @@ void monster::add_enchantment_effect(const mon_enchant &ench, bool quiet)
...
+        break;
+
+    case ENCH_RING_OF_DRAINING:
+        surround_mons_with_cloud(this, CLOUD_NEGATIVE_ENERGY);
+        break;
+
+    case ENCH_RING_OF_ACID:
+        surround_mons_with_cloud(this, CLOUD_ACID);
+        break;
+
```
ebering, line 342: These should be deduplicated, have a block of fallthroughs for each of these, then

```
// .... cases
case ENCH_RING_OF_ACID:
    update_mons_cloud_ring( /* args */);
    break;
```

crawl-ref/source/timed-effects.cc
```diff
@@ -744,6 +744,7 @@ void monster::timeout_enchantments(int levels)
         case ENCH_FRIENDLY_BRIBED: case ENCH_CORROSION: case ENCH_GOLD_LUST:
         case ENCH_RESISTANCE: case ENCH_HEXED: case ENCH_IDEALISED:
         case ENCH_BOUND_SOUL: case ENCH_STILL_WINDS: case ENCH_RING_OF_THUNDER:
+        case ENCH_RING_OF_FLAMES:
```
ebering, line 747: Why are Ring of Thunder and Ring of Flames (both permanent enchantments anyhow) in this function? Neither should be here as far as I can tell...

alexjurkiewicz:
> I'm glad to see that the panlords who get this don't have summon spells. I still worry that they will harm their allies/band members with their clouds but if it's rare enough it's ok? Have you tried this on an orb run or in a Ziggurat?

They absolutely harm allies. Since there's no AI for dealing with their cloud ring, they will quite happily roast monsters. I felt it was a cool flavour thing, and not too major an issue. However I didn't think about packed Zig floors...

crawl-ref/source/cloud.cc
alexjurkiewicz: Doesn't the logic do this? "If (cell can have a cloud and (no cloud OR cloud type matches what we want to place)) { place cloud }"

crawl-ref/source/ghost.cc
alexjurkiewicz: It wasn't to be sneaky! I put it into its own commit. I'll drop from this PR.

crawl-ref/source/timed-effects.cc
alexjurkiewicz: Right. They are both getting removed entirely

alexjurkiewicz:
Rebased against master and addressed review in 8a6071d  onwards 👍 

TODO:

- [ ] Address pan lord ally harm (esp Zig floors)
- [ ] Add status text for cloud rings
- [ ] Fix cloud rings for Asmodeus & Bai Suzhen
- [ ] Add a "ring of miasma" cloud ring

crawl-ref/source/cloud.cc
ebering: Oh, yes it does.

ebering:
I think a sufficient solution would be to not place cloud ring clouds on allies of the monster generating the cloud ring. Monsters might still wander into a cloud ring that already exists, but at least "sit an wait on a pan lord floor" won't be a helpful strategy.

alexjurkiewicz:
OK, implemented that. And everything else, I think...

gammafunk:
I recommend you always make a short description of the motivation for why you want to make a gameplay change like this, even if it's only a few lines, and even if the feature is relatively small. It'll help others understand your motivation for the change in the first place. I've also found it useful as a way to make myself think a little more about why I'm making a change. I may have an initial set of reasons, but upon writing the summary, some of those fall out, and I'm left with the ones that are important. This helps narrow the scope of the change and sometimes leads to better alternatives for getting at the same idea.

Monster cloud rings are trying to be an anti-melee effect, so giving pan lords chaos clouds rings gives me pause, since it's basically never sane to stay be in a cloud that has a chance to paralyze you when a pan lord is adjacent. Apocalypse crabs are relatively weak monsters where you might justify fighting them in chaos clouds, since paralysis is just not that dangerous versus that monster. However, pan lords tend to be fast with either dangerous spells or nasty melee. On the other hand, one primary role of pan lords is to chase you on the orb run, so it's kind of neat as a penalty for letting them get adjacent, I guess.

The problem is that in extended where you'd more realistically want to kill a pan lord, chaos cloud rings make the pan lord much more a target for ranged and allies. These two ways of killing that already get a lot of preferential treatment in extended as it is. We have already veered in this direction by given pan lords af_chaos, but if something like chaos cloud ring is added, it probably needs to be a rare one to prevent "never melee" pan lords from being too common a thing. 

Another issue is that the more mundane elemental clouds are going to be a mostly harmless annoyance, which is unfortunate. It's possible that there needs to be some mechanism for these clouds to be enhance so that the damage is actually relevant. I don't know if new cloud types with higher damage numbers (e.g. "deep freeze cloud") would work here, or perhaps a synergistic monster/spell. Something else to consider is that said enhancement could happen only in Pan to prevent Orb Run pan lords form being insane. The pan Lord being in its native realm of Pandemonium could make its cloud magic more powerful, etc. The flavor of the cloud ring is fun, but I think any effect we do add shouldn't be something that hasn't been scary since you left a Volcano on Lair:2.

alexjurkiewicz:
Here's what I'm planning to add as part of the commit message to explain / justify this change:

```
Add cloud rings to (some) random Pandemonium Lords

This effect already exists in the game, but only for Asmodeus and Bai
Suzhen (in dragon form). Let random pan lords have a chance at getting
one too.

The chance of a pan lord generating with a cloud ring is about 1 in 12,
which might have to be adjusted upwards or downwards in future. The
chance for each cloud type is a little more complex to calculate given
the dependency on pan lord brand / attack flavour, but the rarest brands
will also appear about 1 in 12 times.

Depending on the cloud type, these cloud rings can be terrifying or
ignorable. For example, chaotic clouds can instantly paralyse the player
and negative energy clouds can heavily drain them within a few turns. On
the other hand, clouds of fog and mutation do no direct damage.

Pan lords with cloud rings should feel qualitatively different to fight
than those without. Mostly this comes from increased danger being in
melee range, but also from the impact on battlefield terrain that
dropping clouds everywhere provides. In some cases, this can even work
to the benefit of the player.

One aspect of pandemonium lords I hope this change emphasises is that
they should be considered the "apex predator" of DCSS. They should
always be a scary fight, no matter how prepared you are. This is
especially true since they spawn only in either the optional branch of
Pandemonium (where it's even possible to collect the demonic rune
without fighting one) and on the orb run (where running is already
encouraged). This change should increase the number of cases where
players run (in careful ways!) from pan lords.
```

alexjurkiewicz:
I've also reduced the weight of chaos cloud ring. It should occur on around ~0.8% of pan lords (1 in 144-ish).

crawl-ref/source/mon-info-flag-name.h
```diff
@@ -55,9 +54,18 @@ static const vector<monster_info_flag_name> monster_info_flag_names = {
...
     { MB_GOZAG_INCITED, "incited", "incited by Gozag", "incited"},
+    { MB_CLOUD_RING_THUNDER, "", "surrounded by a ring of thunder", "" },
+    { MB_CLOUD_RING_FLAMES, "", "surrounded by a ring of flames", "" },
+    { MB_CLOUD_RING_CHAOS, "", "surrounded by a ring of chaotic energy", "" },
+    { MB_CLOUD_RING_MUTATION, "", "surrounded by a ring of mutagenic energy", "" },
+    { MB_CLOUD_RING_FOG, "", "surrounded by a ring of fog", "" },
+    { MB_CLOUD_RING_ICE, "", "surrounded by a ring of freezing clouds", "" },
+    { MB_CLOUD_RING_DRAINING, "", "surrounded by a ring of negative energy", "" },
+    { MB_CLOUD_RING_ACID, "", "surrounded by a ring of acid", "" },
+    { MB_CLOUD_RING_MIASMA, "", "surrounded by a ring of miasma", "" },
```
ebering, line 68: I think that if I type this suggestion I can then use the web UI to commit it? Just some display tweaks so that this appears in the HUD and doesn't take too much space in the targeter/^x.

```suggestion
    { MB_CLOUD_RING_THUNDER, "clouds", "surrounded by thunder", "clouds" },
    { MB_CLOUD_RING_FLAMES, "clouds", "surrounded by flames", "clouds" },
    { MB_CLOUD_RING_CHAOS, "clouds", "surrounded by chaotic energy", "clouds" },
    { MB_CLOUD_RING_MUTATION, "clouds", "surrounded by mutagenic energy", "clouds" },
    { MB_CLOUD_RING_FOG, "clouds", "surrounded by fog", "clouds" },
    { MB_CLOUD_RING_ICE, "clouds", "surrounded by freezing clouds", "clouds" },
    { MB_CLOUD_RING_DRAINING, "clouds", "surrounded by negative energy", "clouds" },
    { MB_CLOUD_RING_ACID, "clouds", "surrounded by acidic fog", "clouds" },
    { MB_CLOUD_RING_MIASMA, "clouds", "surrounded by foul miasma", "clouds" },
```

crawl-ref/source/tags.cc
```diff
@@ -2333,6 +2333,7 @@ static spell_type _fixup_removed_spells(spell_type s)
         case SPELL_SPECTRAL_WEAPON:
         case SPELL_RING_OF_FLAMES:
         case SPELL_DARKNESS:
+        case SPELL_RING_OF_THUNDER:
```
ebering, line 2336:
```suggestion
```

This is for player spells.

ebering:
Thanks for the patch!

gammafunk:
Quick comment: you don't want the cloud effects to be ignorable (which the commit mentions), otherwise it's not making the pan lord more dangerous. Specifically I think the fire, ice, and obviously fog clouds are ignorable. The cited example of mutagenic fog isn't quite ignorable, since the contam build-up results both in arbitrarily damaging explosions as well as negative mutations or stat drain. 

Of the added effects, the ones I'd recommend not keeping would be fog, freezing cloud, and flames, particularly the harmless fog clouds. Asmodeus likewise has a problem with his current flame clouds not being dangerous.

alexjurkiewicz:
I think fog is the most interesting of all cloud effects! It will affect
the visible squares "behind" the pan lord.

So I don't mean it is always ignorable. I mean that sometimes you'll decide
to kill the pan lord and nothing will walk out of fog while you do so. But
especially on the orb run this will be an interesting risk.

Similarly, mutagenic energy says "you can fight me in melee, but you will
pick up a bad mutation for your trouble". Also an interesting trade off, I
think.

I do agree it would be good to add higher tier versions of the "base"
clouds. But it's not a new problem introduced by this PR.

PR #1490, opened 2020-07-17
Invoke the malevolent chronology of Zot

PleasingFungus:
The natural tendency for roguelikes is to encourage cautious behavior.
It's over once you die once, after all, so you should take your time.
That's not a bad thing, but to balance things out and discourage
extremely time-consuming tactics (which tend not to be very fun!),
it's good to have a counterpressure encouraging players to move quickly.

Food was intended to be such a clock, but for a variety of reasons, it
never worked as a clock for anything but the most egregious abuses.
(And not even for that for foodless races, of course.) Scores is a
compelling motivation for some players, but not many - most are just
playing to win. How can we make the most viable way to win also the
most fun, while also making a system that doesn't get in players' way?
(That is, can we build something that no one playing normally, even
if they're fairly slow, will ever encounter?)

This commit adds a new 'Zot Clock'. It ticks up over time, and jumps
down whenever you enter a new level - roughly 5000 turns of clock per
level. Once you hit turn 14500 or so on the clock, the malevolent and
unexplained entity Zot strikes, draining you immediately and then again
once every 100 turns or so. At turn 15000, you die.

5000 turns per level allows for a 250,000 turn minimal 3-rune game,
which is very generous. The effective cap of 15000 turns 'stored' at any
given time means that there's somewhat less slack than that implies,
but a slow character can usually do levels in under 2500 turns and
inter-level travel in under 350, so I'm hopeful that normal characters
will rarely if ever see any of this system.

We use 'levels seen' here as a proxy for progress because that's
simple to conceptualize and relatively difficult to abuse - it's hard
to "stash" entire levels as a clock reserve in the same way you could
stash "unexplored map tiles" or "weak monsters". (Temple is easy to
'stash', so it gives no clock.)

The clock is paused while you're in Abyss, since that doesn't fit the
exploration model well and isn't usually that wise to linger in, and
turned off entirely when you get the ORB, since (a) there's a new
clock then and (b) it'd feel awful to die of time-out on the orb run.
It's possible to scum Pan for time, but I mean, if you're scumming
Pan, just win already!

The clock is implemented as 10% chance of increase per 1 aut,
which averages out to about 1 'clock tick' per turn. Over 100 turns,
the clock will usually go up by somewhere between 88 to 112 or so.
This discourages trying to micromanage the clock and count turns
exactly, as with how we fuzz contamination and other similar things.

Chei gives you some extra time, in keeping with the theme of the
god. It's possible that races need different clocks based on their
movement speeds, but I'm hopeful that the difference there is fairly
small relative to the slack in the clock.


ebering:
Is there a reason I'm not seeing as to why this is stored as a "roughly turn counter with probability" and not just straight counting auts taken and scaling the thresholds by Chei piety?

PleasingFungus:
Fuzzing. I don't want players to be counting the number of auts they've taken to ensure that they're perfectly safe!

ebering:
Oh, excellent point!

PleasingFungus:
Summarizing discussion from this morning:
- There are concerns about very long levels (tomb, v:$, z:$) and about players who want to start one branch, bail, start another branch... (I think that the expanded clock bank should handle this, but it's hard to tell until we test.)
- Advil is going to turn on `turns_by_place` in trunk for a few days to see how long people are empirically spending in different places in normal games before we merge this.

PleasingFungus:
Closing in favor of #1493.

PR #1493, opened 2020-07-21
Zot branch bank

PleasingFungus:
The natural tendency for roguelikes is to encourage cautious behavior.
It's over once you die once, after all, so you should take your time.
That's not a bad thing, but to balance things out and discourage
extremely time-consuming tactics (which tend not to be very fun!),
it's good to have a counterpressure encouraging players to move quickly.

Food was intended to be such a clock, but for a variety of reasons, it
never worked as a clock for anything but the most egregious abuses.
(And not even for that for foodless races, of course.) Scores is a
compelling motivation for some players, but not many - most are just
playing to win. How can we make the most viable way to win also the
most fun?

This commit adds a new 'Zot Clock'. It ticks up over time, and jumps
down whenever you enter a new level - roughly 5000 turns of clock per
level. Once you hit turn 14,500 or so on the clock, the malevolent and
unexplained entity Zot strikes, draining you immediately and then again
once every 100 turns or so. At turn 15,000, you die. Each branch tracks
the 'clock' separately.

5000 turns per level allows for a 250,000 turn 3-rune game, which
seems very generous. The effective cap of 15,000 turns 'stored' at any
given time means that there's somewhat less slack than that implies,
but a slow character can usually do levels in under 2500 turns and
inter-level travel in under 500, so I'm hopeful that normal characters
will rarely if ever see any of this system.

We use 'levels seen' here as a proxy for progress because that's
simple to conceptualize and relatively difficult to abuse - it's hard
to "stash" entire levels as a clock reserve in the same way you could
stash "unexplored map tiles" or "weak monsters". The separate clock
per branch allows players to dip into branches and then bail without
being penalized - the per-branch clock is paused while they're away.

The clock is paused while you're in Abyss, since that doesn't fit the
exploration model well and isn't usually that wise to linger in, and
turned off entirely when you get the ORB, since (a) there's a new
clock then and (b) it'd feel awful to die of time-out on the orb run.
It's possible to scum Pan for time, but I mean, if you're scumming
Pan, just win already!

Chei gives you some extra time, in keeping with the theme of the
god. It's possible that races need different clocks based on their
movement speeds, but I'm hopeful that the difference there is fairly
small relative to the slack in the clock.

(This is an alternative version of #1490.)


alexjurkiewicz:
Is it worth documenting why the clock is tracked per-branch? From what I understand, it's to prevent players from getting more time to clear Tomb:3 by diving Elf, and other similar situations.

The messaging to players doesn't make it clear the timer is per-branch. I think it's a problem, but if the zot clock fires rarely enough it is probably a minor issue that doesn't have to be addressed. Communicating this to the player seems like it makes the concept much more complicated.

PleasingFungus:
> Is it worth documenting why the clock is tracked per-branch? From what I understand, it's to prevent players from getting more time to clear Tomb:3 by diving Elf, and other similar situations.

Sorry, I know there's a lot of words here, so it's easy to miss. End of paragraph 5:

```
The separate clock
per branch allows players to dip into branches and then bail without
being penalized - the per-branch clock is paused while they're away.
```

This came out of a discussion of a player who dove through all of Zot, realized they were extremely outclassed, left to go do Elf, and then came back to try Zot again. We do want to support that kind of backtracking for now (or failing at Snake and doing Swamp, or w/e), but it'd be badly penalized under the previous PR's . (They'd dive Crypt or die, basically.)

You're quite right that this also prevents players from using one branch as a time pool for another (except for resting, briefly), which is probably a plus.

> The messaging to players doesn't make it clear the timer is per-branch. I think it's a problem, but if the zot clock fires rarely enough it is probably a minor issue that doesn't have to be addressed. Communicating this to the player seems like it makes the concept much more complicated.

I'm open to suggestion for better messaging :)

crawl-ref/source/timed-effects.cc
```diff
@@ -1258,3 +1259,104 @@ int speed_to_duration(int speed)
...
+// Decrease the zot clock when the player enters a new level.
+void decr_zot_clock() {
+    if (!zot_clock_active()) {
+        return;
+    }
+    if (bezotted()) {
+        mpr("As you enter the new level, Zot loses track of you.");
+    }
+    int &zot = _zot_clock();
+    zot = max(0, zot - ZOT_CLOCK_PER_FLOOR);
```
rawlins, line 1319: just to double check, is it intentional that whatever remains of the delver's secret extra bonus gets wiped out when they change a level?

crawl-ref/source/timed-effects.cc
```diff
@@ -1258,3 +1259,104 @@ int speed_to_duration(int speed)
...
+        mpr("You have lingered too long in familiar places. Zot approaches. Travel to new levels before it's too late!");
+        drain_player(150, true, true);
+    } else if (bezotting_level() > old_lvl) {
+        mpr("Zot draws near...");
+        drain_player(75, true, true);
+    }
+}
+
+void set_initial_zot_clock(bool start_deep) {
+    // Give Delver a pile of extra time to compensate for 'wasted' levels
```
rawlins, line 1357: they still get the clock decr when the enter these levels, right (setting the `max(0,...)` aside)? Wondering if 3 levels worth extra is too much.

crawl-ref/source/timed-effects.cc
```diff
@@ -1258,3 +1259,104 @@ int speed_to_duration(int speed)
...
+
+    if (_zot_clock() >= MAX_ZOT_CLOCK)
+    {
+        mpr("Zot has found you!");
+        ouch(INSTANT_DEATH, KILLED_BY_ZOT);
+        return;
+    }
+
+    if (!old_lvl) {
+        mpr("You have lingered too long in familiar places. Zot approaches. Travel to new levels before it's too late!");
```
rawlins, line 1348: perhaps it would be fair to give at least one warning without consequences before the first drain effect? Maybe at threshold - 100 or something? It might even be worth considering having the status light start early with a different color.

crawl-ref/source/timed-effects.cc
```diff
@@ -1258,3 +1259,104 @@ int speed_to_duration(int speed)
 
     return div_rand_round(100, speed);
 }
+
+static int& _zot_clock_for(branch_type br)
+{
+    static const string CLOCK_KEY_ROOT = "ZOT_CLOCK_FOR_";
+    const string branch_name = branches[br].abbrevname;
+    return you.props[CLOCK_KEY_ROOT + branch_name].get_int();
```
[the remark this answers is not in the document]
PleasingFungus: I considered that, but it seemed more complex to me. Why do you think it's more robust?

crawl-ref/source/timed-effects.cc
```diff
@@ -1258,3 +1259,104 @@ int speed_to_duration(int speed)
...
+}
+
+static int& _zot_clock() {
```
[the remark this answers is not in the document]
PleasingFungus: Hah. I think I've fixed most of em - let me know if you see any more.

crawl-ref/source/god-passive.cc
```diff
@@ -324,6 +324,9 @@ static const vector<god_passive> god_passives[] =
         {  0, passive_t::slow_abyss,
               "GOD will NOW slow the Abyss"
         },
+        {  0, passive_t::slow_zot,
+              "GOD will NOW slow Zot's hunt for you"
+        },
         // TODO: this one should work regardless of penance
```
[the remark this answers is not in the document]
PleasingFungus: Definitely doesn't apply to the old one! Good catch.

crawl-ref/source/timed-effects.cc
```diff
@@ -1258,3 +1259,104 @@ int speed_to_duration(int speed)
...
+        return base_odds - div_rand_round(you.piety, 6);
+    }
+    return base_odds;
+}
+
+void incr_zot_clock() {
+    const int clock_incr = binomial(you.time_taken, _zot_clock_odds(), 1000);
+    const int old_lvl = bezotting_level();
+    _zot_clock() += clock_incr;
+    if (!bezotted()) return;
```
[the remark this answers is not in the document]
PleasingFungus: If I must.

crawl-ref/source/timed-effects.cc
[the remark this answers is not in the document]
rawlins: because it would make all zot clock data easily accessible and manipulable in a single data structure? For example easy to iterate through, etc. Also easier to rename things if it ever came to that, keeps the player prop namespace less polluted. There's no cost to doing it as its own hash table that I'm aware of.

crawl-ref/source/timed-effects.cc
PleasingFungus: It was not. I'll drop the secret extra clock for now - I think delver should be fine anyway. Fingers crossed?

crawl-ref/source/timed-effects.cc
PleasingFungus: I considered that, but it just means that you can scum completely consequence free until you hit the timer. Seems counterproductive.

crawl-ref/source/timed-effects.cc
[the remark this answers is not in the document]
PleasingFungus: Iterating through it seems nice. I can do that.

rawlins:
From looking at some delver early game data it doesn't seem to me like they use much if any extra time, so I think that change should be fine. Not a lot of data for full games yet but very few that would come even close to running up against this version of the clock so far.

PleasingFungus:
It's not so much that they use extra time as that they end up 'losing' about two floors worth of time to bank limits - they get 15k turns for the first 5 floors, not 25k like a normal character would. I think that'll still be more than plenty in practice, though.

rawlins:
Maybe I'm misunderstanding, but why would they be any different? They may be rushing to ascend so wouldn't use as much of the time per level as a character playing slowly from D:1 would, but they still get their clock decr for entering D:4, D:3 etc and still probably spend some time on those levels as they proceed.

PleasingFungus:
I'm talking about on the way back down, where they're actually trying to clear D:4 and D:5 (presumably).

It's probably not worth talking about - I doubt it'll matter.

crystalsolenoid:
So I got a confusing message about Zot watching me and then immediately losing interest right when entering the temple for the first time. I don't know how to use the ttyrec and get the exact timing of the messages, but I made a note so I have the turn number. It's the only game I've played so far on the testing server http://13.59.137.204/crawl/morgue/purplemustard/ (number in the filenames is 20200721-201352) and the turn it was weird is shortly after 4981 and right before 5001.

rawlins:
> So I got a confusing message about Zot watching me and then immediately losing interest right when entering the temple for the first time. I don't know how to use the ttyrec and get the exact timing of the messages, but I made a note so I have the turn number. It's the only game I've played so far on the testing server http://13.59.137.204/crawl/morgue/purplemustard/ (number in the filenames is 20200721-201352) and the turn it was weird is shortly after 4981 and right before 5001.

Thanks for the report @crystalsolenoid ! I think the commit I just pushed fixing empty clocks should fix this bug once it is up on the server (also reported in another variant by @semitonal).

crawl-ref/source/timed-effects.cc
Diojeanne: Perhaps it would be even more fair to just display it. Having it a mystery just produces constant anxiety: even if you're still 10,000 turns from being flogged by Son of Food Clock, how would you know? The reasons for even implementing this clock are mysterious to me, even after reading your explanations. Is it about trapfinding? If so, how does it even discourage that playstyle? Trapfinding is more far more wasteful of real time than turns.

In my view, time limits are not intrinsically more fun, and Son of Food Clock is worse than Food Clock on two counts: one, it's entirely implicit, the Food Clock could be monitored simply by keeping track of how many rations you have. two, it even affects undead, for whom previously freedom from the food clock was one of very few selling points. If you merge this, and really I think you shouldn't, it increases the urgency for some kind of buff for non-eating species.

EugeneAbramchuk:
I might be wrong (please correct), but is not this aimed at Xomscumming? If it is, I'd rather have a solution which remedies that particular behaviour (Xom to penalize idling. Say, bad mutations/hostile summons/deterioration only).

gammafunk:
This branch isn't focused on Xom, although it will certainly put some limits to it. Xom can be addressed directly by changing how that god operates. 

This branch is focused on preventing unlimited use of the most tedious tactics, e.g. fighting nearly every monster safely from stairs on every level over 10k+ turns. With these changes, the player can use those tactics, but only to a limited degree, and it applies to all species. Other tedious tactics that might be beneficial to perform over large amounts of time are also covered, even if they're specific to one species or god and even if they're not available or simply haven't been thought of yet.

EugeneAbramchuk:
I started penning out my concerns, but @Diojeanne did it more eloquently above. 

Still, the nature of the limit seems superficial in my eyes, and I am not saying that because it affects me (my longest farming all rune run was 212k moves); but different players have different play style (especially the new ones) and it might be punishing to them. The extended endgame is already way more challenging than it was years ago (haste removal, Tomb rework spring to my mind).

Without extended, this will effectively force the player to win under 35k turns for a bare minimal 3-rune run.

rawlins:
> Without extended, this will effectively force the player to win under 35k turns for a bare minimal 3-rune run.

This is off by nearly an order of magnitude...

gammafunk:
Tedious play styles that involve spending vast amounts of turns each level aren't a category of play styles we're trying preserve here; directly the opposite, in fact. Normal play and even relatively safe play won't see this clock come into effect. As the PR writeup notes:

> 5000 turns per level allows for a 250,000 turn 3-rune game, which seems very generous.

I don't know what "bare minimal 3 rune" is referring to, but players absolutely will not need to win in 35k turns with 3 runes, even if they only visit D, Depths, Zot, and exactly three non-extended rune branches. The clock grants a very generous 5k turns per level and the clock is tracked per-branch. 

Extended won't be meaningfully affected by this clock, but more importantly we're not interested in allowing more scumming in extended branches either.

EugeneAbramchuk:
> The clock grants a very generous 5k turns per level and the clock is tracked per-branch.
Oh, I mistook that for 5k _per branch_. But even then, if someone - for whatever reason they might possibly have - is enjoying tedious play, why punish it?

Checking through my recent game, I see I spent 17k turns in Pan and I was not even scumming, I just was looking for an exit.
https://crawl.develz.org/morgues/git/hayenne/morgue-hayenne-20200717-211411.txt
(here I have to take back my words of the change not affecting me)

crawl-ref/source/timed-effects.cc
rawlins: Hi! Thanks for the feedback. However, you're posting this as a followup to a specific PR review comment, which is not a good place to post general feedback and background questions. If you want to have this discussion in this PR (I'll let PF decide whether that is appropriate) please repost or followup as a general comment, not in this review thread.

I'll add that clocks are something where there's been an awful lot of discussion both by the community and the devteam, so as usual this PR is not really in a vacuum, and it may be worth looking through some of that before commenting further, as I do think this comment contains some incorrect assumptions about both what this PR is about and what it would do, as well as what the food clock (rip) actually did. The main targets of this are behaviors in the tedious but optimal category that won't resemble a "typical" dcss playstyle at all: extremely excessive luring (i.e. 10s of thousands of turns), certain other kinds of waitscumming (i.e. placing a very outer limit on what kind of xomscumming is possible). One recent community discussion that has a useful summary is this tavern thread from hellmonk: https://crawl.develz.org/tavern/viewtopic.php?f=8&t=26941

rawlins:
> if someone - for whatever reason they might possibly have - is enjoying tedious play, why punish it?

This is not a game design philosophy that the devteam ascribes to (though it's certainly one that [has existed](https://nethackwiki.com/wiki/Pudding_farming)). One nice way of putting it, I believe originally due to @PleasingFungus himself ([https://www.youtube.com/watch?v=-MGNS39D0B0](see his 2016 roguelike celebration talk here)), is that if you lose a game and think back and realize that you died because you didn't engage in extremely tedious behavior, this is not a good feeling. I do agree that this general philosophy issue is (unusually) on topic for this particular PR, but do be aware that this PR is presupposing this particular philosophical viewpoint rather than proposing it as something new.

> Checking through my recent game, I see I spent 17k turns in Pan and I was not even scumming

You visited 12 levels during those 17k turns, where each new level in pan gets you 5k turns on the clock, the clock would generally be hovering near 0 for this kind of play -- there was no danger of this clock coming into play for this case.

crawl-ref/source/timed-effects.cc
damerell: > I considered that, but it just means that you can scum completely consequence free until you hit the timer. Seems counterproductive.

Unless I'm missing something, can't Hypothetically Optimal Man just track the state of the bank by doing a little arithmetic? In which case, this seems like a thing a spoiled player can know by doing some tedious tracking, and my expectation would be that DCSS just tells me such a thing. I can't see any random element for non-Chei which would prevent it being tracked exactly by HOM.

rawlins: > I can't see any random element for non-Chei which would prevent it being tracked exactly by HOM.

The clock increment currently has some fuzzing to avoid this, see `incr_zot_clock()`. (For clarity's sake, my review comment here was only concerned with how to warn players that they are getting close to a limit.)

petercordia:
It's not clear to me that this system is better than a completely goldfield ration system. (A completely goldified ration system is one where rations are goldified, eating does not take time, and eating is done automatically (the same way chunks were handled).)
Some minor differences are:
* this system allows a specific clock per branch
* this system allows the amount of time accrued to be capped
* rations allow the player can 'buy' extra time via rations shops and amulets of gourmand. This could allow a tighter clock, and since the time is bought at a strategic cost it still discourages tedious behaviour. Another way of saying the same thing: the ration system can be more flexible
* this system does not work well in Pan (apparently) and I expect it'll make Hell harder.
* rations were distributed randomly, but this time is distributed deterministically.

I'm sure it will be fine, but it's not clear to me why you would make this new system from scratch without first seeing whether a completely goldfield food clock has any issues.

One small issue with this system:
If I see I'm almost at the turn cap I'm incentivised to play as Hypothetical Optimal Man, with tedious behaviour. With the rations I'd want to do D quickly so that I'd have more time for Hell if I decided to go that way. I know this is minor, and in the reverse this system doesn't let players feel they have all the time in the world in Zot.

rawlins:
> It's not clear to me that this system is better than a completely goldfield ration system. (A completely goldified ration system is one where rations are goldified, eating does not take time, and eating is done automatically (the same way chunks were handled).)

By far the biggest reason to drop food as a clock mechanically is that as long as rations are generated on the floor there's a huge amount of variability in the clock size and levelgen doesn't provide many tools for addressing that. Once you have something where food-like ticks are being added on kill, or on level entry, you end up much closer to the clock in this branch or one of the alternatives that has been discussed, and it's better IMO to drop the food metaphor.

I don't really think allowing buying time on the clock is a good mechanism. In the case of ration shops, it served mainly to paper over shortcomings of the food clock IMO.

> this system does not work well in Pan (apparently) and I expect it'll make Hell harder.

Not sure what you have in mind by these -- I expect no impact on either hell or pan.

PleasingFungus:
Thanks for all the thoughtful comments! Rawlins covers my own thoughts perfectly, so I mostly have one question:

> If I see I'm almost at the turn cap I'm incentivised to play as Hypothetical Optimal Man, with tedious behaviour. 

I'm a bit confused by this, sorry. Can you expand on how you see that you're almost at the turn cap (when you've already been bezotted?), and on what tedious behaviour you have in mind?

jfcaron3:
Is there a way to disable the Zot-clock-stuff for floors that have been fully explored (or fully cleared of monsters)?  This system would make me question the wisdom of going back to early Dungeon levels from e.g. Zot to grab something off the floor or from a shop.  Or maybe a smaller clock allotment per branch when you enter not-a-new floor (like 200 turns, enough to grab items, go to shops, swap armour, etc)?

PleasingFungus:
That sort of backtracking doesn't take very long, and it'd be better not to add more special cases.

rawlins:
> This system would make me question the wisdom of going back to early Dungeon levels from e.g. Zot to grab something off the floor or from a shop.

One of the challenges that this clock has is that a lot of people have this initial reaction, but it's just not grounded in how the clock works -- there's a huge margin in normal play (or even fairly abnormal play) for backtracking built in. In most of the data I've looked at people will typically have thousands (or more) of spare turns in any branch, way more than enough for multiple backtracking all over the place.

I don't think we'll be adding any of the features you suggest because they make it possible, just more tedious or scripted, to scum (for example someone could write a script that alternates between cleared levels 200 turns at a time). But, I would like to figure out a way to present this clock that mitigates this reaction.

jfcaron3:
Fair enough.  

This gets a bit conceptual, but I think these clocks (including the removed food clock) is trying to patch a problem that originates elsewhere: that long periods of waiting can be beneficial in the first place.  

If people are Xomscumming, why not make Xom's gifts timed with XP gain (or tiles explored) instead of turns?  At least then they'd have to Xomscum in Abyss or Pan which has inherent risks.  If people are shouting & waiting at the stairs, change the noise & wandering mechanics so that this only works the first time and repeated shouts do nothing.  

I don't mean to dismiss all the work that's gone into this, and probably this discussion has already happened among the devs, but I guess I wanted to record my thoughts on it too.

PleasingFungus:
> This gets a bit conceptual, but I think these clocks (including the removed food clock) is trying to patch a problem that originates elsewhere: that long periods of waiting can be beneficial in the first place.

Interesting!

Think about it this way: without any kind of clock, it's natural for it sometimes to be optimal to act very quickly (e.g. in a fight), and sometimes to act slowly. There are lots of ways the game can end up incentivizing slow play, and I don't think you're ever going to eliminate all of them. It's like trying to plug holes in a dike with with your finger.

A clock provides a systemic response to the problem. Rather than trying to patch out every incentive to play slowly as you discover them, you add an across the board incentive to play quickly (or at least less slowly). This isn't just efficient, it's also a good fit for the underlying problem. If we want 'optimal play' to be exciting and challenging and fun, adding a mechanism that says 'extremely slow play is not optimal play' just makes sense.

>  If people are shouting & waiting at the stairs, change the noise & wandering mechanics so that this only works the first time and repeated shouts do nothing.

How?

> I don't mean to dismiss all the work that's gone into this, and probably this discussion has already happened among the devs, but I guess I wanted to record my thoughts on it too.

No need to worry about dismissing anything! I appreciate your thoughts - it's valuable to have outside opinions on something like this.

jfcaron3:
I'll be honest I don't understand exactly why shouting at stairs and waiting works beyond the first shout.  What follows is just wild speculation & ideas.

My guess without code-diving is that a shout has a certain range but monsters at the edge of that range might have only a small chance of waking up.  So if it's only a 1% chance, then shouting many many times will eventually grab more and more monsters.  If this is how it works, then the chance to wake up needs to be made more deterministic, maybe add a fixed ~30% chance (on top of the variable chance due to distance) so that a few shouts pretty much wakes up everyone in range.  Another way would be to make shouts "stack" over multiple turns for the purpose of waking up enemies, so that after a few shouts reliably all monsters are woken up.

Then there's the issue of wandering-awake-but-unaware monsters.  Brutally I would add a small bias to the monster-wandering algorithm to either lead them towards or away the player on average (maybe based on monster type?  or randomly generated per-monster?).  Thus a monster that's biased towards the player will pretty much always end up there after N turns, and those biased away will never come into view.  Thus the player never benefits from waiting more than some number of turns after shouting.

Another idea is to remove shouting from the game.  This seems bold but it would nicely make scrolls of noise valuable (some people already carry them for this purpose).  I guess a dedicated shout-and-wait player could just get a loud spell castable and  spam lightning bolts at the wall...without spell hunger there is no cost to this, but at least they'd presumably spend spell levels and maybe even skill XP. ^_^

Diojeanne:
> I'll be honest I don't understand exactly why shouting at stairs and waiting works beyond the first shout. What follows is just wild speculation & ideas.
> 
> My guess without code-diving is that a shout has a certain range but monsters at the edge of that range might have only a small chance of waking up. So if it's only a 1% chance, then shouting many many times will eventually grab more and more monsters. If this is how it works, then the chance to wake up needs to be made more deterministic, maybe add a fixed ~30% chance (on top of the variable chance due to distance) so that a few shouts pretty much wakes up everyone in range. Another way would be to make shouts "stack" over multiple turns for the purpose of waking up enemies, so that after a few shouts reliably all monsters are woken up.

Standing in one place and shouting is indeed not a workable shoutscumming strategy. Once you've drawn in everyone who can hear you shouting from the stairs, you take one step away from the stairs and shout there, then step back to the stairs and shout there for a while waiting for anyone to show up again.  Then you take two steps away from the stairs and shout  there, step back to the stairs, shouting at every step, and wait for anyone to show up again.  And so on. It's not a fun way to play.  I see this whole thing as a fight between two groups of very sick people: people like myself who will have their play rushed simply by knowing the zot clock is out there, even if by any reasonable measure it's sure to be 30,000 turns away from flogging them, and people who Must. Play. Optimally, even if it takes 400 hours and sucks the whole time.
 
> Another idea is to remove shouting from the game. This seems bold but it would nicely make scrolls of noise valuable (some people already carry them for this purpose). I guess a dedicated shout-and-wait player could just get a loud spell castable and spam lightning bolts at the wall...without spell hunger there is no cost to this, but at least they'd presumably spend spell levels and maybe even skill XP. ^_^

Removing shouting is a bold suggestion, but I don't think it'll help much: there are other noisy things you can do apart from the scroll of noise, such as most spells.

PR #1499, opened 2020-07-26
New demonspawn scales facets + adjust thin skeletal structure

hellmonk2:
This pr adds two scales tier facets for demonspawn: Big Brain gives 2/4/6 int and provides wizardry at rank 3, and Sharp Scales provides 1/2/3 AC and Slaying. Additionally, it doubles the stealth bonus provided by thin skeletal structure to a full pip per mutation level. The goal is to create more scales facets that aren't simply 2/3/4 AC plus resistance. Ideally it'll put a little bit more pressure to "adapt" to the scales as well, but I don't really expect it to do that.

Both new mutations are currently demonspawn exclusive, but obviously don't have to be.


RojjaCebolla:
block helmets for big brain?

ebering:
Ok, no one else on the team stopped by to bikeshed Big Brain so I'm going to fix this up and merge it later this afternoon. Again will be sped along if you fix push but if not I'll get to it.

PR #1500, opened 2020-07-27
Solidify hangedman_spin_cycle

alexjurkiewicz:


alexjurkiewicz:
```05:46:56 +gammafunk | alexjurkiewicz: I was going to make that area a little transporter area, which seems like a nice way to keep it
05:47:22 +gammafunk | randomly place a transpoter to the area in one of the outer cells, and have a transporter out
05:47:32 +gammafunk | then place glass so the player can see inside to the center
05:48:01 +gammafunk | only issue with that is the center is slightly too large to see in fully with LOS 7```

gammafunk:
Ended up doing this in acbe4010. Thanks for the reminder to make this change. 

I leave commit comments as to how it might be improved, and the original vault author might want to do something different.

PR #1503, opened 2020-07-29
Fix crawlrc-dependent blank screen/crash on turn 0

joshgelbard:
After b274e4624c7c, we print "You add the spells ... to your library"
even when memorizing from a starting spellbook on turn 0.  This can
cause issues for players who have a force_more_message that matches that
text: for those players, any spellbook background starts with a blank
screen (a promptless force_more), and Earth Elementalists just crash.

Two changes:
* Fix the crash by moving add_held_books_to_library down a few lines, so
it happens after items are linked.
* Change the printed message so that on turn 0 it says "You begin with
the spells" instead of "You add the spells"; that way, it's possible to
target (only) the mid-game version with a force_more_message.

### DIFF
--- a/crawl-ref/source/spl-book.cc
+++ b/crawl-ref/source/spl-book.cc
@@ -514,7 +514,9 @@ bool library_add_spells(vector<spell_type> spells)
     {
         vector<string> spellnames(new_spells.size());
         transform(new_spells.begin(), new_spells.end(), spellnames.begin(), spell_title);
-        mprf("You add the spell%s %s to your library.",
+        mprf(you.num_turns < 1
+             ? "You begin with the spell%s %s in your library."
+             : "You add the spell%s %s to your library.",
              spellnames.size() > 1 ? "s" : "",
              comma_separated_line(spellnames.begin(),
                                   spellnames.end()).c_str());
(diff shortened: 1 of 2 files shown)


joshgelbard:
To clarify the point of the message change: "You add the spells" is a real force_more_message I actually use (to make it harder to miss an important spell pickup while mindlessly autoexploring) and it'll be nice if that doesn't mean an extra keypress when I start a new game.

Printing nothing at all instead of "You begin with..." would also work and might even make more sense; I was just wary of restoring anything too much like the code so recently eliminated in b274e46, and wasn't sure whether there's a "only silence things by passing a `quiet` bool" rule or something like that.

crawl-ref/source/spl-book.cc
```diff
@@ -514,7 +514,9 @@ bool library_add_spells(vector<spell_type> spells)
     {
         vector<string> spellnames(new_spells.size());
         transform(new_spells.begin(), new_spells.end(), spellnames.begin(), spell_title);
-        mprf("You add the spell%s %s to your library.",
+        mprf(you.num_turns < 1
```
ebering, line 517: While not re-adding the quiet bool is good, this should simply not print a message at game start. The only case where we want to announce the starting library is for Wanderers, which are already handled elsewhere.

rawlins:
This wasn't an intended change so I'd probably rather add back in the `quiet` param for this case.

rawlins:
I'm actually going to push a different fix altogether than what's been discussed, sorry not to use your work directly but this PR + the patch were extremely helpful to me in figuring out what's going on! I am also about to push a more direct fix for the EE crash, and for the blank screen force more.

PR #1509, opened 2020-08-05
Update monster creation documentation

Implojin:
`monster_creation.txt` had previously contained several references to enum.h,
which was split into several files in 895272c. This commit updates the
documentation accordingly.


ebering:
Thanks for the mixups!

PR #1511, opened 2020-08-06
Door reachable fix

Naruni-Crawl:
Add feat_is_closed_door to catch runed translucent doors from having tornado blow through them.  Also try to adhere to coding conventions.

### DIFF
--- a/crawl-ref/docs/develop/monster_creation.txt
+++ b/crawl-ref/docs/develop/monster_creation.txt
@@ -10,8 +10,8 @@ monsters of similar type. Bolt yaks should be added after yaks and before death
 yaks, for example. Certain files list their entries in strictly alphabetical
 order. Whenever possible follow present convention.
 
-enum.h is a VERY important file! Many things you create will need to be declared
-there or your build will not compile.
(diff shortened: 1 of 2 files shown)


rawlins:
Thanks for the PR! Code-wise this looks good (and we already discussed it in ##crawl-dev). We usually request that the commit title/message itself contains the description of the change, not just the PR description (since that gets lost once the commits are merged). It's also usually a good idea to reference the bug number and description in a way that is interpretable on its own.  So here's an example of how the commit title and message could look using your PR description; I've also added a linebreak at 72 chars:

```
Stop tornado from going through transparent doors (12330)

Add feat_is_closed_door to catch runed translucent doors from having 
tornado blow through them.
```

To edit this now that you have pushed, you can use `git rebase` (I usually just do something like `git rebase -i HEAD^^` and use the `r` option to reword a commit message, but there might be other ways), and then `git push --force` to replace the commit.

ebering:

Naruni-Crawl:
Just "Naruni" is fine, thanks.

ebering:
But of course, you're already in the credits from your old mantis patches!

PR #1515, opened 2020-08-08
Don't crumble acquirement scrolls (12336)

Goratrix:
Unidentified acquirements scrolls are now cancellable after they are
read, so they should not crumble to dust.


Goratrix:
Messed up braces, will resubmit, sorry.

PR #1518, opened 2020-08-09
Don't repeat acquirement identification message (dinky)

Goratrix:
Currently the message sequence when reading un-id acquirement is:

 This is a scroll of acquirement!
_Something appears at your feet! It was a scroll of acquirement.

The second message is not necessary as it only applies to scrolls
which produce an immediate effect without a prompt/menu.


ebering:
Thanks for the patch!

PR #1519, opened 2020-08-10
Allow Rampaging on randart armour

Implojin:
Rampaging is currently pretty rare in practice:

* Ego boots are extremely rare spawns at itemgen.
* Many species cannot equip boots.
* Some characters end up with melded boots due to build choices.
* The positioning choices and skilling choices offered by rampaging
  are more impactful when presented earlier in the game.

This commit aims to make rampaging available to a wider range of
characters, by adding it as a potential randart armour property.

ARTP_RAMPAGING generates as 'potentially_good', similar to Harm.

-----
As another way to address rampaging rarity, @PleasingFungus had previously suggested adding rampaging to a new or existing unique monster. That still seems like a good idea to me, but my own attempts at writing a branch to that effect have been uninspired so far. 

In the meantime, here's an artprop.

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -298,6 +298,7 @@ static vector<string> _randart_propnames(const item_def& item,
         { ARTP_CAUSE_TELEPORTATION,   prop_note::plain },
         { ARTP_NOISE,                 prop_note::plain },
         { ARTP_HARM,                  prop_note::plain },
+        { ARTP_RAMPAGING,             prop_note::plain },
         { ARTP_CORRODE,               prop_note::plain },
         { ARTP_DRAIN,                 prop_note::plain },
         { ARTP_SLOW,                  prop_note::plain },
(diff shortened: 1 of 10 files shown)


crawl-ref/source/describe.cc
```diff
@@ -569,6 +570,8 @@ static string _randart_descrip(const item_def &item)
         { ARTP_FRAGILE, "It will be destroyed if unequipped.", false },
         { ARTP_SHIELDING, "It affects your SH (%d).", false},
         { ARTP_HARM, "It increases damage dealt and taken.", false},
+        { ARTP_RAMPAGING, "Its wearer goes twice as far when moving towards enemies.",
```
ebering, line 573:
```suggestion
        { ARTP_RAMPAGING, "It makes you to go goes twice as far when moving towards enemies.",
```

These are all second person or passive voice, no third-person wearers.

Implojin: Yeah, that wording was a copy of the rampage description added in 50c865f. If we're updating it to match the phrasing of the rest of the artprops, I'd prefer to rephrase both rampage descriptions slightly to remove some potential spoilers. Something like:

SPARM: "Its wearer takes one free step when moving towards enemies."
ARTP: "It bestows one free step when moving towards enemies."

Implojin: I've tweaked both of these descriptions, squashed to reword the commit message, and force pushed with the updated descriptions.

ebering:
Great! Thanks for the patch!

PR #1526, opened 2020-08-20
Correct phantom mirror message

Goratrix:
Phantom mirror no longer shatters after use with 1d127b8.


ebering:
Thanks for the patch!

PR #1533, opened 2020-08-23
Revamp Swamp [WIP]

PleasingFungus:
This is an attempt to revamp and revitalize Swamp by reducing total monster count (especially 'popcorn' monsters), reworking some existing monsters, removing overused monsters, and adding a few new ones.

Notable changes:
- Remove slime creatures, shadows, ugly things, cane toads, insubstantial wisps (sort of), hornets, komodo dragons, blink frogs, and very ugly things from normal spawns. Some of them were far too weak, others were just used in way too many branches already. (I'm planning to follow up to revise most Swamp vaults.)
- Add new monsters: bloated husks, kitsune, eleionomae, bunyips, goliath frogs, devouring swarms, and fenstrider witches. There are roughly as many types of monsters in Swamp as there were before.
- Increase bog body HD from 6 to 10 and merge their cold touch attack (4 damage!) with their main attack (hellmonk), to give them a chance of doing something.
- Make insubstantial wisps passively ignite poison/meph clouds and band them with swamp dragons. This is fairly sketchy but it seemed sort of neat when I tried it - definitely one of the ideas most likely to be scrapped, though.
- Rebalance Swamp spawns in general, per [this spreadsheet](https://docs.google.com/spreadsheets/d/1zJX3Me4kqdt8U3dFRrAs0tHUAKV_LILmY9vB6T7ZmTU/edit?usp=sharing).
- Remove deep water. It interacts poorly with swamp worm harpoon shot and it makes less of a distinction between Swamp and Shoals. I'm not sure if this is a good idea, but it's something to try!
- Reduce total Swamp non-vault spawns by about 33% to keep XP (and difficulty) in check. (Perhaps the levels should also be made smaller?)
- Remove chimerae.

I haven't touched vaults yet, and I don't think the balance is right in general, but in basic playtesting it feels decent.

Further testing would involve putting this on a branch somewhere after people have initial says and vaults are fixed. Swamp is forced to always generate in this branch, with an equal chance of spider, snake, or shoals appearing opposite. (I think this is how Spider was originally tested...?)

Some notes on the new monsters:
- *Bloated Husk*: Charge at you with a weakish AF_ROT attack and explode in a radius-2 blast when killed. Sort of like much simpler lurking horrors, or like ballistomycetes without the whole spore thing.
- *Kitsune*: smite, minor healing, foxfire! Monster foxfire is cool imo - all the visual pizzaz of orb of destruction without the dodging minigame.
- *Eleionoma*: It's a swamp nymph. Teleports into trees, heals while next to trees, shoots wood splinters with BEAM_FRAG.
- *Bunyip*: It's a swamp juggernaut crossed with a howler monkey. Not sure this is the right gimmick for it but it's an australian cryptid - we had to include it!
- *Goliath Frog*: Reaching acid tongue attack. In this scenario you're a fly.
- *Devouring Swarm*: Batty group attackers that implant eggs in injured monsters nearby. If you kill those monsters, more (durable summoned) swarms pop out. Gross!
-  *Fenstrider Witch*: Fast casters with paralyse, agony (melee), and a ranged toxic sludge attack that plops toxic bog at the victim's feet. (Short-duration, to avoid too much abuse against other monsters.)

(alternate title: 'get shrekt')


ByrelMitchell:
Quick thought; all other frogs in the game currently are EV tanks (and indeed, one of the reasons I suggested renaming spiny frogs to cane toads was because they were the odd man out on that.) Perhaps Goliath Frogs should have 0/20 or so defenses?

amcnicky:
Digging the new monsters in this. If it saves you time, the current build failure is in mon-data.h due to "kitsune has a corpse but no corpse tile". I had a similar build fail recently and took me a while to find the relevant line in the build log!

PleasingFungus:
@amcnicky thanks! I'll fix that the next time I have a chance to work on this.

@ByrelMitchell reasonable. I'll probably switch to EV for the frog as you suggest, especially given the size joke.

PleasingFungus:
For @gammafunk's benefit, planned changes that are relevant to vaults:
- Retheme kitsune to will-o-the-wisps (remove smite, change resists)
- Think about maybe removing with husks (and replacing with some other low-tier enemy?)
- Think about swarms more

PR #1535, opened 2020-08-26
Fix a rampaging bug (12353, Flugkiller)

Implojin:
This fixes a rampage interaction with deep water / lava.

Previously, the rampaging tracer wasn't properly failing on dangerous
terrain. This was causing the rampage to abort at its moveto check instead,
which erroneously blocked the second half of the move.

-----
This PR fixes [mantis issue 12353](https://crawl.develz.org/mantis/view.php?id=12353).


ebering:
Thanks for the patch!

PR #1536, opened 2020-08-26
Enable cursor in map-view for console

bjobae:
08ff59a6b337511f8596aeb26a0c06b099ee3b7f made the map cursor disappear, at least for me on Windows 10, console.

### DIFF
--- a/crawl-ref/source/viewmap.cc
+++ b/crawl-ref/source/viewmap.cc
@@ -799,6 +799,7 @@ bool show_map(level_pos &lpos, bool travel_mode, bool allow_offlevel)
     if (!lpos.id.is_valid() || !allow_offlevel)
         lpos.id = level_id::current();
 
+    cursor_control ccon(true);
     levelview_excursion le(travel_mode);
     auto map_view = make_shared<UIMapView>(lpos, le, travel_mode, allow_offlevel);
 


crawl-ref/source/viewmap.cc
```diff
@@ -799,6 +799,7 @@ bool show_map(level_pos &lpos, bool travel_mode, bool allow_offlevel)
     if (!lpos.id.is_valid() || !allow_offlevel)
         lpos.id = level_id::current();
 
+    cursor_control ccon(true);
```
ebering, line 802:
```suggestion
    cursor_control ccon(!Options.use_fake_cursor);
```

I believe, as in the direction chooser and the cursor_control in the offending commit, the fake cursor option needs to be checked here but I'm not super familiar with this part of the code and don't have access to a Windows system to test on. @aidanholm can you assist?

aidanholm: I don't have a windows system to test on either unfortunately. My understanding is that `use_fake_cursor` instructs crawl to disable the terminal's cursor, and manually flip fg and bg on the cell where the cursor would have done so. `cursor_control`, meanwhile, controls only whether the terminal cursor is enabled, so it only makes sense to use them together like this if you then change your painting behaviour (by flipping fg/bg for the correct cell).

Separately, this change is in the wrong location. This `cursor_control` affects the entire view-map context, including any popups opened within (e.g. help screens). The code that actually draws the map view alone is in `UIMapView::_render()`. There's a problem though: that's called by the UI framework, but there could easily be a popup drawn on top of the map view widget, and you only really want to enable the cursor for the frame if and only if there's nothing on top.

`cursor_control` doesn't let you do this, because it is a RAII style object, and so if you use it in `_render()` the old cursor state will be immediately restored again as soon as you return. The "correct" way to do this is to use `ui::show_cursor_at` inside `_render()`, which instructs the UI framework to show the cursor at the correct location after everything is drawn. However, you only want to show the cursor if there's nothing on top, and there's no way to get that information... `show_cursor_at` might need to be modified to only respect requests from the top layer.

Apart from that, this PR's code should work, but only because it's not overrided by another `cursor_control` before the `ui::pump_events()` call further down in `show_map()`. Passing `true` is correct, but ignores the `use_fake_cursor` option. Passing `!Options.use_fake_cursor` respects the option but requires additional work in `_draw_level_map`. Hope that helps :)

bjobae: I tested the PR by comparing cursor behavior with that of 0.25.1 using the various map commands like opening the help window, and everything seems to work as intended.

This was the best "solution" I could find without spending too much time and messing everything up with bigger changes, but I'm glad the issue caught your attention because it practically makes the game unplayable for those who like ascii.

wheals:
See the linked commit for more details of the fix I went with.

PR #1539, opened 2020-08-28
Wand description shows noise level

Naruni-Crawl:
Inspecting a wand will tell the player how much noise is generated when
evoking the wand.


ebering:
Thanks for the patch!

PR #1541, opened 2020-08-28
Two fixes for the spectral and protection brands

nlavsky:
This PR allows randarts to generate with the spectral ego, as was intended.

The removal of the protection brand from lajatangs and quarterstaves looks unintentional
because it wasn't mentioned in 65aa84ea, so the second commit fixes that.


ebering:
The brand table change was intentional, I should have mentioned it in the commit.

ebering:
Thanks for the other part of this patch!

PR #1543, opened 2020-08-28
Following PR#1538, resubmit new floating vaults

amcnicky:
The third PR in the series, submitting amended floating vaults following
a first round review by ebering. Vaults in this collection:

<float.des>
amcnicky_nick_rare_triplestair_openwater
A rare vault which groups all downstairs onto a platform in open water.

amcnicky_float_guardroom
A guard room staffed by more difficult monsters (and some potentially
good loot) is presented along an L-corridor layout with the potential
door or open connections.

amcnicky_float_beach_large
FTILE used to present the player with an oasis hidden within the
dungeon! Lower weighting to ensure this remains a rarer and thus more
unique sight.

amcnicky_float_corridor_feature
A decorative vault used to give the dungeon a chance to generate
corridors containing an intentional patterns of one of many features.
Uses the extra tag.

amcnicky_float_rare_duel
A very rare (weight 2) vault in which vault walls and blood pools
pre-empt a quite challenging engagement.

amcnicky_float_rare_at_lavas_edge
Another rare vault, this time presenting the player with a glass-lined
platform out over lava. Is this an ornate lava pier? or perhaps the
glass holds the lava back and the player is walking within the body of
lava itself.

### DIFF
--- a/crawl-ref/source/dat/des/variable/float.des
+++ b/crawl-ref/source/dat/des/variable/float.des
@@ -835,6 +835,33 @@ Ellllllll.
   ..!...
 ENDMAP
 
+NAME:    amcnicky_rare_triplestair_openwater
+ORIENT:  float
+TAGS:    extra luniq_staircase decor
+DEPTH:   D:2-, Lair:2-, Depths:2-
+WEIGHT:  4
+MAP
+ xxxxxxxxxxxxxxxxxxxx
+xxWWWWWWWWWWWWWWWWWWxx
+xWWWWWWWWWWWWWWWWWWWWx
+xWWWWWwwwwwwwwwWWWWWWx
+xWWWWwww.....wwwWWWWWx
+xWWWWww.{...(.wwWWWWWx
+xWWWWww...[...wwWWWWWx
+xWWWWwww.....wwwWWWWWx
+xWWWWwwww...wwwwWWWWWx
+xWWWWWwww...wwwWWWWWWx
+xWWWWWWww...wwWWWWWWWx
+xWWWWWWww...wwWWWWWWWx
+xxxWWWWww...wwWWWWxxxx
+  xxxxWww...wwWxxxx
+     xxww...wwxx
+      xww...wwx
+      xxw...wxx
+       xw...wx
+       xw@@@wx
+ENDMAP
+
 ##############################################################################
 # Celtic spiral tile pattern (rotated 45 degrees)
 # http://en.wikipedia.org/wiki/Celtic_maze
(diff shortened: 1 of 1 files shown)


crawl-ref/source/dat/des/variable/float.des
```diff
@@ -835,6 +835,33 @@ Ellllllll.
...
+xWWWWWwww...wwwWWWWWWx
+xWWWWWWww...wwWWWWWWWx
+xWWWWWWww...wwWWWWWWWx
+xxxWWWWww...wwWWWWxxxx
+  xxxxWww...wwWxxxx
+     xxww...wwxx
+      xww...wwx
+      xxw...wxx
+       xw...wx
+       xw@@@wx
```
ebering, line 862: I think you've switched deep `w` and shallow `W` water here.

crawl-ref/source/dat/des/variable/float.des
```diff
@@ -9930,3 +9957,138 @@ x..x..x11xx...........xx11x..x..x
 x1.x.1x11*xxx...{...xxx*11x1.x.1x
 xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 ENDMAP
+
+NAME:   amcnicky_float_guardroom
```
ebering, line 9961: As far as vault authoring goes, the syntax and style is in a good place.

However, the tactics aren't that interesting: the player has plenty of corridors to handle the tougher fight from (and corners!), and while positional magic means some spell users will back off to the open space provided, that's still not great because the monsters will come out of the corridor one at a time.

crawl-ref/source/dat/des/variable/float.des
```diff
@@ -9930,3 +9957,138 @@ x..x..x11xx...........xx11x..x..x
...
+xx@xx
+xp.px
+xx.xx
+xx.xx
+xp.px
+xx.xx
+xx.xx
+xp.px
+xx@xx
+ENDMAP
```
ebering, line 10037:
```suggestion
```

This was merged with your minivault changes.

crawl-ref/source/dat/des/variable/float.des
```diff
@@ -9930,3 +9957,138 @@ x..x..x11xx...........xx11x..x..x
...
+
+# Vault walls and blood hint that something tougher awaits
+NAME:   amcnicky_float_rare_duel
```
ebering, line 10040: This would be the start of a good minivault storyline concept if its tactical options were more interesting than a corridor with door.

crawl-ref/source/dat/des/variable/float.des
```diff
@@ -9930,3 +9957,138 @@ x..x..x11xx...........xx11x..x..x
...
+@+$pp+..8.v
+ xxvvvG.E.v
+     vvG.vv
+      vv%v
+       vvv
+ENDMAP
+
+# Involves lava, so using thematic weightings to
+# lean towards where lava seems most common in Depths
+NAME: amcnicky_float_rare_at_lavas_edge
```
ebering, line 10062: This gets away from corridor-tactics, but the terrain again doesn't provide much challenge: the player can lure the threats out of the vault easily without fighting too many at once. There's no hook to draw them into the fight.

Lava lakes are best avoided if the player can fly over them: we have them in old vaults but they encourage flight cheese or (if the player can't see the edge) flight to hunt for hidden areas/spoiler checking, so it's best not to add new ones.

ebering (changes requested):
Your other vaults were (with a few tweaks) well improved by the feedback, and I've merged them! Thanks! 

This batch I haven't merged yet (except for corridor_feature as a minivault, per the previous review), because while they address the line-by-line feedback in my previous review, each still suffers from the general remark I left:

> Your tendency to use small single-tile corridors as vault entrances, while mostly harmless in arrival vaults, is really not good for float vaults that place monsters. The goal of vaults is to provide interesting terrain, challenge, and encounters that the random level generator can't produce, while building flavor. The level generator is very good at creating single-tile corridors, and all of your float vaults that place monsters except the lava pool are accessed via single-tile corridors. Some of them are somewhat workable with the corridors trimmed but at that point they are mostly square rooms, which again the level generator can produce on its own.

They each to a degree build some flavor, but the tactics presented by the ones with monsters are single-tile-wide (door or corridor) fights or just a big mob scene.

Regarding the beaches. I don't want to merge either beach vault (the decorative one from mini vaults or this one). Ideally vaults will build flavour without recolour/retiling (the exception being Lugonu type altar vaults); these beach vaults aren't connected to Shoals and are square rooms without much in the way of tactics or story.

ebering:
There aren't super strong vault concepts in this batch, I hope the feedback was helpful and when you find crawl time again you'll send us some new stuff, or at least enjoy those vaults from your submissions that have been merged!

PR #1545, opened 2020-08-28
Add missing "info" command to --edit-save help

rosche:
The "info" command is missing from the --edit-save usage message.

### DIFF
--- a/crawl-ref/source/initfile.cc
+++ b/crawl-ref/source/initfile.cc
@@ -4048,6 +4048,7 @@ static void _edit_save(int argc, char **argv)
     {
         printf("Usage: crawl --edit-save <name> <command>, where <command> may be:\n"
                "  ls                          list the chunks\n"
+               "  info                        list detailed chunk and frag info\n"
                "  get <chunk> [<chunkfile>]   extract a chunk into <chunkfile>\n"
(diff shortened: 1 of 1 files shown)


ebering:
Thank you for the patch, how would you like to be listed in the CREDITS?

rosche:
I can't tell if you're joking due to the earth-shattering complexity of the patch, but I'll note that my previously merged commit was done using my work email address by accident; please use roderick@argon.org.  Thanks.

ebering:
I am not joking! All contributions, large and small, merit a listing in the CREDITS. Looking back at our previous conversation I see I didn't ask you then, but you contributed in 2012 so I hope that you are already listed in the credits under an alias: there isn't an entry matching the name or alias of your GitHub account.

Regarding your previous patch, when merging commits I don't edit the Author information provided by a git commit. You can use local git configuration to ensure that it is always set to your preferred identity.

rosche:
Thanks, I knew it was my fault about my email address.

I've looked and I'm not in the credits, please list me as Roderick Schertler.  My contributions have always been small but I enjoy doing them.  One of my changes became default behavior at some point (displaying the time the previous command took) and this pleases me greatly.

PR #1546, opened 2020-08-28
Add Mlioglotl, a new Lugonu-themed unique.

amcnicky:
The vision for Mlioglotl is to add a new unique with a strong
flavour themed around Lugonu and the Abyss. 

In-game monster description:
"A demon prince once betrayed his realm through a dark ritual to Lugonu.
Becoming the twisted mass of corruption known as Mlioglotl was his
reward.

Seeing such a powerful acolyte of Lugonu within the mortal plane is a
rare sight indeed, and one should fear the denziens and horrors of the
Abyss that are never far behind."

Gameplay flavour vision: A hulking mass ("unique thrashing horror"
wouldn't be far off the mark, visually speaking) charging after the
player whilst tearing open rifts between the dungeon and the abyss.

Gameplay design vision: The player is torn between wanting to focus
damage onto Mlioglotl to stop him casting further instances of corruption,
and needing to deal with anything too nasty that is spawned. The spawns
are durable, which adds a relatively novel additional angle to the
interplay between player and 'summoner'. Of course Mlioglotl is no
pushover himself either, and is positioned via depth marker to show up
only later in the game. During combat, Mlioglotl's corruption ability will
disrupt the terrain, allowing for unexpected tactical situations to arise if
vaults are breach or further monsters alerted. Additionally, Mlioglotl can
blink away as an emergency spell (to try and push at least some interaction
with corruption-spawned monsters).

The mechanics of Mlioglotl corrupting the dungeon are handled though a
new monster spell 'corrupt_locale'. This effectively triggers a
customised version of the Lugonu corruption player ability with a
toned down duration and toned down monster set.

One concern with any kind of level-altering effect of course would be
that a player could kite the monster around and use them to 'shatter'
open difficult geometry. Mlioglotl is fast, and hits relatively hard,
which perhaps is enough to deter this tedious behaviour, however I'd
welcome feedback on other potential methods to disincentivise this such
as preventing Mlioglotl from using stairs (he has a large tile
currently, so this would fit the visual-language-precedence set by the
Royal Jelly).

Anyway, this is my first time digging into the source code of crawl
beyond vaults and tiles, so apologies in advance for any convention
breaking that I'm not aware of. Always happy to tidy up, make suggested
changes, and generally take my submissions in a direction that others
feel is most healthy for the game.

Thanks,

aMcNicky


PleasingFungus:
> One concern with any kind of level-altering effect of course would be that a player could kite the monster around and use them to 'shatter' open difficult geometry.

I agree with this concern! It seems like the simplest solution here would be to make Milo's corrupt effect have visual impact (changing tiles/colours) and spawn monsters, but *not* destroy stone or metal, unlike players' Corrupt. What do you think?

crawl-ref/source/abyss.cc
```diff
@@ -1766,25 +1766,40 @@ struct corrupt_env
...
-static void _initialise_level_corrupt_seeds(int power)
+static void _initialise_level_corrupt_seeds(int power, monster* mons = nullptr, bool is_monster_version = false)
 {
     const int low = power * 40 / 100, high = power * 140 / 100;
     const int nseeds = random_range(-1, min(2 + power / 110, 4), 2);
 
     const int aux_seed_radius = 4;
 
-    dprf("Placing %d corruption seeds (power: %d)", nseeds, power);
+    // The corruption centred on the player is free.
```
PleasingFungus, line 1792: This comment is in the wrong place, isn't it?

crawl-ref/source/abyss.cc
```diff
@@ -1808,14 +1831,30 @@ static bool _incorruptible(monster_type mt)
...
+    || mons_demon_tier(mt) == 1     || mons_demon_tier(mt) == 2
+    || mt == MONS_TENTACLED_STARSPAWN;
+    // no tier 1 or 2 demons, or starspawn for the monster version
+}
+
 // Create a corruption spawn at the given position. Returns false if further
 // monsters should not be placed near this spot (overcrowding), true if
 // more monsters can fit in.
-static bool _spawn_corrupted_servant_near(const coord_def &pos)
+static bool _spawn_corrupted_servant_near(const coord_def &pos, bool is_monster_version)
```
PleasingFungus, line 1845: Weaving this `is_monster_version` flag through everywhere makes me nervous. I'd really consider splitting the function entirely, especially if you take my suggestion of making the monster version have fewer effects on terrain (or even only cosmetic effects).

crawl-ref/source/dat/descript/monsters.txt
```diff
@@ -336,6 +336,16 @@ when Mennas remained silent they cut out his tongue.
 Whether or not this angel is that same Mennas is something of a mystery, as he
 seems to have difficulty speaking.
 %%%%
+Mlioglotl
+
+A demon prince once betrayed his realm through a dark ritual to Lugonu.
+Becoming the twisted mass of corruption known as Mlioglotl was his
+reward.
```
PleasingFungus, line 343: Fun :)

crawl-ref/source/dat/descript/spells.txt
```diff
@@ -345,6 +345,12 @@ Corrosive Bolt spell
 
 Fires a penetrating bolt of acid.
 %%%%
+Corrupt Locale spell
+
+Tears apart the reality of the dungeon in the immediate vicinity, allowing
+the denziens and horrors of the Abyss to permanently seep through. Followers
+of Lugonu will find such abominations neutral to their cause.
```
PleasingFungus, line 352: Is this a fun mechanic? As with Vaud, having gods give you protection against specific monsters does not seem like a particularly good idea to me, and it's not like Lugonu gives you protection against abyss monsters in general. Are you sure this is a good idea?

crawl-ref/source/mon-cast.cc
```diff
@@ -472,6 +474,11 @@ static const map<spell_type, mons_spell_logic> spell_to_logic = {
         [] (const monster &caster, mon_spell_slot /*slot*/, bolt& /*beem*/) {
             _monster_abjuration(caster, true);
         }, nullptr, MSPELL_LOGIC_NONE, 20, } },
+    { SPELL_CORRUPT_LOCALE, {
+        _always_worthwhile
```
PleasingFungus, line 478: Perhaps consider banning it in abyss, unless you really enjoy the cantrip behaviour :)

crawl-ref/source/mon-spell.h
```diff
@@ -1412,6 +1412,15 @@ static const mon_spellbook mspell_list[] =
       }
     },
 
+    {  MST_MLIOGLOTL,
+      {
+       { SPELL_MIGHT, 8, MON_SPELL_NATURAL },
+       { SPELL_CAUSE_FEAR, 10, MON_SPELL_MAGICAL | MON_SPELL_EMERGENCY },
+       { SPELL_BLINK, 5, MON_SPELL_MAGICAL | MON_SPELL_EMERGENCY },
```
PleasingFungus, line 1419: This seems quite annoying. Are you sure you want it?

crawl-ref/source/mon-spell.h
```diff
@@ -1412,6 +1412,15 @@ static const mon_spellbook mspell_list[] =
       }
     },
 
+    {  MST_MLIOGLOTL,
+      {
+       { SPELL_MIGHT, 8, MON_SPELL_NATURAL },
+       { SPELL_CAUSE_FEAR, 10, MON_SPELL_MAGICAL | MON_SPELL_EMERGENCY },
```
PleasingFungus, line 1418: This is funny and thematic. I wonder how well it plays.

PleasingFungus (changes requested):
Another very cool idea! Thanks again for submitting it, sorry again for the belated feedback.

crawl-ref/source/abyss.cc
amcnicky: Correct place but just unhelpful/legacy. I've updated it to be more relevant and useful, though this may be changed further as I split the function out following the next comment.

crawl-ref/source/abyss.cc
amcnicky: Yes agreed - it's a little bit more code duplication that way but easier to follow/update in future and does fit better with the two effects being distinct in nature. I've split this out now.

crawl-ref/source/dat/descript/spells.txt
amcnicky: Yeah fair challenge on that, theme may well > fun on that one. Angry monsters for everyone!

amcnicky:
> > One concern with any kind of level-altering effect of course would be that a player could kite the monster around and use them to 'shatter' open difficult geometry.
> 
> I agree with this concern! It seems like the simplest solution here would be to make Milo's corrupt effect have visual impact (changing tiles/colours) and spawn monsters, but _not_ destroy stone or metal, unlike players' Corrupt. What do you think?

Agreed, that is a sensible solution - the monster version should now avoid destroying stone and metal, but should still re-colour them, which is nice.

crawl-ref/source/dat/descript/monsters.txt
amcnicky: Thanks! I was thinking it might be fun to do something in future like Maggie/Margery so that players can run into a scheming demon prince who's looking for a good place to conduct a dark ritual i.e. pre-corruption mlioglotl, as a mutually exclusive unique. No actual design for that in mind yet though.

crawl-ref/source/mon-cast.cc
amcnicky: I've made it a bad() idea when in the Abyss. Not quite banning it, as I assume even bad ideas are picked occasionally. The idea of Mlio pausing to bask in the beautiful corruption of the Abyss occasionally on the off chance he ends up back there seems nice.

crawl-ref/source/mon-spell.h
amcnicky: From some more playtesting I'm inclined to agree and have removed blink from Mlio

crawl-ref/source/mon-spell.h
amcnicky: From the playtesting I've done, either you're trying to run already, in which case fear won't stop you, or you're fighting Mlio, in which case he'll close distance to melee and you can still continue under cause fear. Therefore, a cast of this spell impacts mainly by 1. scaring away player summons (a nice interaction to see occasionally) and 2. reducing player melee effectiveness with a 33% flinch chance. As an emergency spell then it feels like a thematic thing for Mlio to try, which does have an impact, whilst not being overbearing, so I'm inclined to keep it for now unless you foresee any other issues and/or players give poor feedback on that area of the design.

gammafunk:
You've done a nice job of writing up many details an the overarching idea and motivation for this unique, so thanks for doing that. You did leave out some important gameplay details. A general recommendation for gameplay-related PRs is to give priority to gameplay aspects over listing flavor text. It's certainly fine to include the monster description, but I had to go digging through the commits to see basic details like where this unique is placed, what its speed, defenses, and attacks are etc.

Overall it's a unique with nice theme but has some problematic aspects. Before we get to those, let's talk balance. You've made this unique significantly nastier than Mennas. It has similar speed (14), two 1d15 attacks, one of which is chaos, and a 1d40 attack. Furthermore it has 240 HP (a lot more than Mennas), 20 AC and 5 EV. Mennas does have a shield, a weapon, and significantly more EV, but in any case your unique doesn't belong as early as late dungeon. I discuss placement a bit below, but overall this unique seems to be trying to be overly difficult. It's built like a very late game or extended unique, even thematically as well, but you seem to want it to place much earlier. If that's the case, I think you'll have to tone it down a lot. One thing to note in particular is that use of af_chaos melee brings the monster into "never let this melee you" territory. This is mostly because of paralysis to a lesser extent due to banishment. Non-formicid players can't stand next to a fast monster that can unconditionally paralyze them and just hope it doesn't happen if they're trying to win; certainly not one with a profile anything like this unique.

Regarding appropriate depth, think of D:15 as being "before (non-Slime) Lair branches", and the "absolute depth" value of a branch is a good reference for what area comes before what. I assume omitting Depths:1 when you have Depths:2- was to avoid "ambushing" players, but it's odd to do this when you place the monster on D:15. Ambushing players on Depths:1 isn't a concern, since players have so many resources by that point. I would also apply this logic to Vaults:1. It's fine to limit placement to only later levels of a branch for general balance, and ambushes in earlier areas are a valid concern. However for late game, it's much harder to kill the player, and it's fine to throw the nastier effects their way (provided that the effects are interesting). Here is a workbook of unique placement by level that may be helpful: 

https://docs.google.com/spreadsheets/d/1r5aE9AEz2f0t5f5HrWxVS4xxIg96-D9BaBPHbbpEh1k/edit?usp=sharing

Swamp, Early Vaults, and early Depths could all use a bit more unique coverage, which are places you could target balance-wise, but if it turns out this unique works best in Late Vaults+Depths or even extended, don't let those coverage numbers hold you back from that.

I don't think making the summons durable is a good idea at all. You state in the rationale that this is a novel type of summon, but tellingly you don't elaborate as to why it's useful from a gameplay perspective. I'd caution against use of "it's different" as a justification for why you're not using the existing mechanic that already does what you want. We use durable summons for things like Zot traps and God wrath, where there's a legitimate concern of "resting off" an effect in a context where the player shouldn't be able to do this, since the effect is meant to be an unavoidable penalty for a player decision (e.g. stepping on or allowing a monster to step on a trap or abandoning a god). But there's no player decision relevant to having a unique spawn, and you can't really stop this unique from using its summon spell. Ordinary summoning seems just fine for this unique.

Regarding the permanent terrain changing, that's a problematic aspect that we'd be reluctant to add more of to the connected branches. The profile of this monster is such that luring it to break into things is pretty tough, but I think that either the profile or the placement of this monster will need to change significantly before we'd consider a merge. Its bringing in of "patches" of the abyss to non-abyss areas does seem to be the main reason why the unique exists at all, since that's fun thematically. If it just ends up being tuned as a late game or extended unique, this vault breaking shouldn't cause too many problems. I'm not sure if there's a clever solution to this involving something like temporary terrain; that might be more trouble than its worth.

For use of the large tile, I have a bit of concern if this unique ends up showing in mid game. We've previously only used these types of tiles for extended branch monsters (which Slime counting as extended even if it is the easiest extended area). So we've previously used these tiles to telegraph "this is among the hardest monsters you can fight in the game". If Mlioglotl ends up toned down for earlier areas, I don't think it should use a large tile.

amcnicky:
@gammafunk Thanks for the comments! Helpful to have another perspective on the design. I've made changes which I think address the points raised so far. I would welcome a follow-up look over those changes if you get time.

To all keeping an eye on the PR, a few things to call out, 2 of which are mentioned in the most recent commit message and 2 of which just presented themselves from reviewing the build checks:
1. There is a known visual bug involving casting corrupt_locale, killing Mlioglotl, saving and reloading.
2. Mlioglotl potentially doesn't summon enough at the moment.
3. _corrupt_square_monster contains a lot of redundant/unused logic at the moment which I will deal with.
4. Do any git-fu masters out there have advice on the most efficient way to address the build failures? Looks like it's failing because this code was branched prior to the grd/igrd helper macro removals - I assume there's a smarter way to address that than going through and typing env.grid all over the codebase to catch the code back up...

amcnicky:
Short update, and sorry for the slow progress, not getting much crawl time the last few weeks. I've removed the redundant code from _corrupt_square. Point 2 in the above also feels like it isn't really an issue from some more play testing. This leaves the visual bug which I'll look into, and an open request for advice on how best to incorporate the recent master branch changes to grd/igrd helper macros into this PR in an efficient way. Responding to the Vaud PR on pause whilst I work through that.

ebering:
@amcnicky I know you haven't had much crawl time these past two months, but if your patches are something you're interested in finishing please do drop us a line. I've tagged these as 0.27 for now, since we've entered feature freeze for 0.26.

amcnicky:
Hello, apologies for my extended time away from crawl causing this PR to go stale. Since many of the files it touches are now out of date, I'm planning to rebuild the PR from the latest master. As such, I think it makes sense to close this particular PR, allowing for some fresh code and a fresh explanation of why Milo is a good unique to add. See you in the new PR soon :) 

(if that's not in keeping with good github practices, let me know)

PR #1547, opened 2020-08-31
Give x-v a selection menu when there are many describable things

ufshaikh:
This expands 2a96157, which lets you select from an item in a stack when x-v-ing, to list monsters and features as well. Now you are not blocked from examining items when there is a monster on the tile or from examining features when there is a monster or item.

Unfortunately if you use ctrl-x or ctrl-f with the menu in travel mode to see where something is and then want to check the description, you can end up needing to select the same item twice from similar menus. This isn't new with this commit but this does sort of double down on that. I'm open to suggestions.


ebering:
`_full_describe_menu' probably deserves to have its own menu class implemented instead of being hacked into `InvMenu` at this point, but that's a refactor for a different patch.

Thanks for this one!

PR #1548, opened 2020-08-31
Stop acquirement of +4 stat rings

ufshaikh:
These rings aren't supposed to exist anymore as of 2607d5cfa1. However
acquirement was making them by producing -4 bad stat rings and then
setting the rings' plus value to abs(the old plus).

This commit has acquirement directly set the plus for acquired rings to
the "good" value for that base type. This does not affect unrandarts but
may modify randarts.


ebering:
Thanks for the patch!

PR #1549, opened 2020-09-01
Remove bad position clamp for peeking through stairs

ufshaikh:
This should fix the recent bug where using X and then [ or ] while on the stairs often did not take you to the other end of the stairs.

(see the commit message for details)


ebering:
Thanks for the patch!

Issue #1451, opened 2020-06-11
lua item.subtype fails if weapon not wielded


mgdelmonte:
platform: local tiles
dcss version: 0.25-a0-1136-g2eff2d5ea4

Calling lua function it.subtype() on a weapon that you have not yet wielded returns nil.

I've only seen it happen for ranged weapons (eg, hunting slings) so far, but it may happen on others.

This seems caused by the underlying `item_type_known` function.

This result seems broken as the item type is completely known regardless of whether you wield it or not.

PR #1551, opened 2020-09-02
fixing weapon subtype

mgdelmonte:
Fixing #1451.  Weapon subtype should always be known.  You can tell what it is by looking at it -- identify or wielding does not change subtype.


ebering:
Thanks for the patch!

PR #1554, opened 2020-09-04
Improve docs for running DCSS after compilation

alexjurkiewicz:


erf:
I tried to build using "make -j4 mac-app-tiles". This builds the zips, but the application does not run.

For me to build on mac i have to run "make mac-app-tiles APPLE_GCC=y NO_PKGCONFIG=y CONTRIB_SDL=y TILES=y"

alexjurkiewicz:
What version of macos are you running? Do you have any third party compiler
installed? What about system copies of any of the contribs?

erf:
I run macOS Catalina 10.15.6.

Yes to both.

Here is the list from homebrew:
```
adns			gnupg			libtasn1		pcre
autoconf		gnutls			libtermkey		pcre2
automake		gobject-introspection	libtool			pinentry
bash			graphite2		libunistring		pixman
bdw-gc			guile			libusb			pkg-config
cairo			harfbuzz		libyaml			python@3.8
cmake			icu4c			llvm			qemu
coreutils		isl			llvm@7			ragel
ctls			jpeg			lua			readline
dart			libassuan		lua@5.1			ripgrep
e2fsprogs		libevent		luarocks		snappy
fd			libffi			lzo			sqlite
flock			libgcrypt		m4			swig
fontconfig		libgpg-error		mpfr			tre
freetype		libidn2			ncurses			unbound
gcc			libksba			nettle			unibilium
gdbm			libmpc			ninja			vapor
gettext			libpng			node			vde
git			libressl		npth			vttest
glib			libssh			openssl@1.1		wget
gmp			libssh2			p11-kit			xz
```

alexjurkiewicz:
I'm OK with the default instructions not working for people with an old
LLVM installed, so I don't think we need to mention APPLE_GCC. And
`CONTRIB_SDL=y` doesn't seem to do anything based on a grep of the
codebase. But `NO_PKGCONFIG=y` seems like the best way to tell the build
script "use contribs", and it might be worth mentioning.

I'm not going to investigate if that's true and update this PR though, I
think it's a minimal improvement as-is, and I'm not enough of a build
wizard to make a call re NO_PKGCONFIG.

erf:
I uninstalled all brew packages now and built again using "make -j4 mac-app-tiles", and i still could not run the mac app built. Have you tested it to run ok?

alexjurkiewicz:
Oh, good catch. I think you need `TILES=y` as well when building the tiles
app. Weird buglet in the Makefile.

erf:
I think you also need to have installed python and the pyyaml pip module (e.g. via brew). Is this mentioned anywhere?

erf:
Ok, it seem i can build for macOS with the following: `make mac-app-tiles TILES=y`

Can you change the document to that, and also mention the `python` and `pyyaml` dependency?

The latter can be installed with `brew install python` and `pip install pyyaml`.

erf:
Could you also mention to install `python` vis `brew install python` before `pyyaml`?

erf:
Not everybody has python on their system.

alexjurkiewicz:
macOS comes with Python preinstalled 😶

erf:
Ah, you are right. Its an older version 2.7, but i guess its fine as long as you also have pip installed.

PR #1556, opened 2020-09-05
Fix the divider position when tile_force_overlay is set

mhcerri:
When tile_force_overlay is set the divider should always be set to the
end of the of the screen. Move the code that sets that to later in the
code so it doesn't get overwritten.

Without this fix, the space reserved for the message window is still
left unused at the bottom of the screen while the message overlay is
moved to the top of the screen.


ebering:
Thanks for the patch, and welcome to contributing to crawl! How would you like to appear in the CREDITS?

mhcerri:
Hi, thanks! Marcelo Cerri works fine.

PR #1557, opened 2020-09-07
Add options to control background color and alpha of the message window overlay

mhcerri:
Add two new options, tile_overlay_col and tile_overlay_alpha_percent,
to control the background of the message window overlay.

The default color follows the same hard coded value used before while
alpha was rounded to the closest round percentage number. The previous
alpha value was 100 (in the 0-255 range), while the 40% new default
should be equivalent to 102.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -107,9 +107,10 @@ The contents of this text are:
                 tile_tag_pref, tile_window_width, tile_window_height,
                 tile_map_pixels, tile_viewport_scale, tile_map_scale,
                 tile_cell_pixels, tile_filter_scaling, tile_force_overlay,
-                tile_full_screen, tile_single_column_menus,
-                tile_font_crt_file, tile_font_stat_file, tile_font_msg_file,
-                tile_font_tip_file, tile_font_lbl_file, tile_font_crt_family,
+                tile_overlay_col, tile_overlay_alpha_percent, tile_full_screen,
+                tile_single_column_menus, tile_font_crt_file,
+                tile_font_stat_file, tile_font_msg_file, tile_font_tip_file,
+                tile_font_lbl_file, tile_font_crt_family,
                 tile_font_stat_family, tile_font_msg_family,
                 tile_font_lbl_family, tile_font_crt_size, tile_font_stat_size,
                 tile_font_msg_size, tile_font_tip_size, tile_font_lbl_size,
@@ -120,7 +121,8 @@ The contents of this text are:
                 tile_show_threat_levels, tile_layout_priority,
                 tile_display_mode, tile_level_map_hide_messages,
                 tile_level_map_hide_sidebar, tile_player_tile,
-                tile_weapon_offsets, tile_shield_offsets, tile_web_mouse_control
+                tile_weapon_offsets, tile_shield_offsets,
+                tile_web_mouse_control
 4-  Character Dump.
 4-a     Saving.
                 dump_on_save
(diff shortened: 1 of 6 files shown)


crawl-ref/source/options.h
```diff
@@ -561,6 +561,8 @@ struct game_options
     int         tile_map_pixels;
 
     bool        tile_force_overlay;
+    VColour     tile_overlay_col;           // Background color for message overlay
```
ebering, line 564:
```suggestion
    VColour     tile_overlay_colour;           // Background color for message overlay
```

crawl-ref/docs/options_guide.txt
```diff
@@ -2021,6 +2023,18 @@ tile_force_overlay = false
         the same number of lines. Increase the view_max_height option if you
         find yourself with unused screen estate.
 
+tile_overlay_col = #646464
```
ebering, line 2026:
```suggestion
tile_overlay_colour = #646464
```

Please give this option a verbose name, since it is adjacent to other options that involve the word column.

crawl-ref/source/tilesdl.cc
```diff
@@ -979,7 +979,9 @@ void TilesFramework::do_layout()
     m_region_tile->tile_ih = tile_ih;
 
     // Resize and place the message window.
-    m_region_msg->set_overlay(message_overlay);
+    VColour overlay_col = Options.tile_overlay_col;
```
ebering, line 982:
```suggestion
    VColour overlay_col = Options.tile_overlay_colour;
```

crawl-ref/docs/options_guide.txt
```diff
@@ -107,9 +107,10 @@ The contents of this text are:
                 tile_tag_pref, tile_window_width, tile_window_height,
                 tile_map_pixels, tile_viewport_scale, tile_map_scale,
                 tile_cell_pixels, tile_filter_scaling, tile_force_overlay,
-                tile_full_screen, tile_single_column_menus,
-                tile_font_crt_file, tile_font_stat_file, tile_font_msg_file,
-                tile_font_tip_file, tile_font_lbl_file, tile_font_crt_family,
+                tile_overlay_col, tile_overlay_alpha_percent, tile_full_screen,
```
ebering, line 110:
```suggestion
                tile_overlay_colour, tile_overlay_alpha_percent, tile_full_screen,
```

ebering (commented):
This patch looks good, except for the abbreviation in the user-facing name. I'll go ahead and fix that up and merge.

crawl-ref/source/initfile.cc
```diff
@@ -319,6 +319,8 @@ const vector<GameOption*> game_options::build_options_list()
         new BoolGameOption(SIMPLE_NAME(tile_menu_icons), true),
         new BoolGameOption(SIMPLE_NAME(tile_filter_scaling), false),
         new BoolGameOption(SIMPLE_NAME(tile_force_overlay), false),
+        new TileColGameOption(SIMPLE_NAME(tile_overlay_col), "#646464"),
```
ebering, line 322:
```suggestion
        new TileColGameOption(SIMPLE_NAME(tile_overlay_colour), "#646464"),
```

mhcerri: Hi, I don't have problem with the change. But most of the other color options seems to use `_col` instead, that's why I ended using it. Thanks!

mhcerri:
Are you sure using colour won't make it inconsistent with the existing options? They seem to use col instead

ebering:
So it would, the tile options use _col while the console options use _colour.

Issue #1234, opened 2019-12-17
Zot 5 vault needs no_tele_into


JayH2971:
Webtiles 0.25-a0-294-gea9b1ad ([morgue](http://crawl.akrasiac.org/rawdata/djehuti/morgue-djehuti-20191216-220440.txt))

**DESCRIPTION:**
Entered a permanent teleport trap in Zot 5 and was sent into a pond behind hardened glass. The only way out was to teleport again.

**SCREENSHOT:**
https://i.redd.it/z0l6mjycj2541.png

theJollySin:
In the morgue above it shows the unique vaults in Zot:5 are: 

> layout_gridlike, ebering_ghost_xom, st_stairs_1

But when I look at the last two, neither seems like it could possibly be the vault in question.

Unfortunately, I don't grok the level-building logic well enough to be able to track this down.  I will keep reading to see if I can find it.  BUT if anyone has any helpful directions for where to look, I'll happily add the `w = no_tele_into` into the appropriate place.

rawlins:
I'm not going to have a chance to look closely right now but this is probably caused by `layout_gridlike`, which has a longstanding issue with tele closets ([see](https://crawl.develz.org/mantis/view.php?id=7563&nbn=10#bugnotes)). It should be possible to reproduce via the seed + version.

theJollySin:
@rawlins   I was able to reproduce the bug offline with seed + version.  Indeed.

I am still trying to understand the `layout_gridlike` logic.  Though it looks like it would be enough to say `w = no_tele_into` on all of the various `layout_***` routines. BUT I have been hearing crickets on the suggestion, so I will keep looking.

gammafunk:
Putting `no_tele_into` on closets generated by the layout would be fine. Alternately we could just not generate decorative closets as part of the layout. That sort of thing can be handled by decor vaults, just not with quite as much smooth integration into the layout itself.

theJollySin:
@gammafunk I had dropped this task, as some conversation I had on Freenode made me think no one agreed on an approach to fix it.

I can try to pick it back up again.  I found the `layout_gridlike` code to be a little opaque, perhaps it's is a lack of experience with the codebase.  But I will try to grok it again and see if I can implement your change.

Thanks for the response!

theJollySin:
@gammafunk 
Okay, for the purposes of this discussion, let's talk about "layout_gridlike" in this file:

    crawl-ref/source/dat/des/builder/layout_cellular.des

[...]

PR #1561, opened 2020-09-11
Layout edits

Skrybe:
A few adjustments to layouts in Lair and Snake.  These are largely cosmetic:

layout_chaotic_city
layout_loops_ring
These layouts now only place rock walls when generated in Lair.  Previously they could also place stone and metal.

layout_cave_pods
Add a distortion effect to the layout when it generates in Snake, giving the walls a wavy appearance.  This same effect is already used by the onion and onion_interference layouts in Snake.


Skrybe:
Comparison of the original version of cave_pods with the new Snake variant:
[Original cave_pods, no distortion](https://www.skrybe.net/stuff/podsold.jpg)
[New cave_pods, with distortion](https://www.skrybe.net/stuff/podsnew.jpg)

ebering:
Thank you for the patch! As you've probably noticed the automatic checks have a whitespace issue, which I'll fix when merging. Since you've submitted several layout tweaks and vaults, would you be interested in checking out bug #1234 ?

PR #1567, opened 2020-09-14
Adjust tier 2 mutations

hellmonk2:
This pr adjusts several tier 2 demonspawn mutations and adds one new facet.

Modified:
Ignite Blood facet gives ignite blood/ignite blood/ignite blood. Mutlevel/3 chance to ignite, flame cloud immunity provided gratis with mutlevel 1.
Icemail facet gives condensation shield/icemail/icemail. 4 SH/4 AC/4 AC for a total of 4SH/8AC. No longer melts when flying above lava.
Demonic Guardian triggers when hit or at low HP rather than based on tension. Demon spawns adjusted and scaled partially with experience level. Mutation scheduling now follows the same pattern as other tier 2 mutations.
Spiny damage formula adjusted. Scales a little better with mutation level and a little worse with xl.
Foul Stench facet gives foul stench/foul stench/foul stench. Increasing chance to spawn miasma clouds; rank 3 of the facet nominally has the same chance as before. No longer causes sickness. Provides free rRot with rank 1.
Mana Link facet gives mp regen first, then guardian spirit. The mutations themselves are unchanged.

New:
Demonic/malevolent/whatever casting: new casting-oriented facet that tries to paralyze enemies in a radius equal to the rank when casting a spell (and at rank 3, using an evocable). The effect's power is dependent on the level of the spell cast and checks magic resistance. Very unsure about the balance here, but I think a sort of add-on effect for spells could be fun tactically and strategically.

Unchanged:
Powered by Pain, Powered by Death

The scales facet changes from my last pr are showing up here, sorry about that!


semitonal:
It looks like malevolent magic will only apply to wands, not other evokables - that seems fine to me in terms of design (since applying it to all evokers would probably get messy on the code side of things with unrands etc, and it matches MP-powered wands), so maybe just needs the description changing?

ebering:
Unfortunately I didn't get to this before feature freeze (corona'd rip) but I'll take a crack at it at 0.27. Probably the modified facets can merge right away. The demonic casting facet feeling bad for Be is... hmmmmmm. Big Brain also does. Having your Ds muts push you strongly away from Trog is possibly Cool and Exciting and not bad.

Plenty of time to bike shed it tho.

wheals:
Merged, thanks!

Issue #1453, opened 2020-06-13
The tile of disarmed Edmund still shows a flail


LarsH:
**PLATFORM and OS or BROWSER:**
local tiles

**DCSS VERSION:**
0.24

**DESCRIPTION:**
Casting Tukima's Dance on Edmund still shows him wielding a flail.

![2019-12-17-221106_1366x768_scrot](https://user-images.githubusercontent.com/1294050/84579931-7f60a900-add2-11ea-9298-e917ea93ed9d.png)

Solutions I could think of:
* just accept this visual quirk
* make Edmund immune to Tukimas dance
* replace Edmunds tile with a normal human tile
* add a new tile for disarmed Edmund

wheals:
Fixed in #1568

PR #1568, opened 2020-09-15
Give Edmund a weaponless tile, fixing #1453

amcnicky:
Following the template implementation of other uniques who can lose their primary weapon, such as Agnes (who has a 'staveless' tile). There are probably a few more who could use an implementation of such a tile, but for they're out of scope for this PR, which seeks to specifically respond to the issue raised at #1453.


amcnicky:
crawl-ref/source/rltiles/mon/unique/edmund_weaponless.png

aidanholm:
Nice; thanks!

PR #1571, opened 2020-09-16
MiAK lost his recommended species tag

smopucilowski:
Selecting Minotaur as a special lists Abyssal Knight as a recommended job, but selecting Abyssal Knight does not list Minotaur as a recommended species. It looks like this was dropped during the Centaur/Palentonga merge. Readded.


semitonal:
I think this is probably fine as-is - combo recommendations aren't necessary symmetrical, they can vary depending on whether species or background is selected first, since the aim is to suggest a few suitable combos given your initial choice. It looks like when Palentonga were added, the recommendations for Abyssal Knight were changed to replace AK->Sp with AK->Pa (since the number of recommendations is kept to around 5-6 for each).

PR #1572, opened 2020-09-17
Make Cloak of Starlight more interesting

amcnicky:
This commit adds an additional property {Radiance} to the cloak of
starlight following a brief discussion in ##crawl-dev. The new effect
works as follows:

Each turn, all blindable monsters in LOS are subject to a chance to be
blinded. This chance consists of two parts.
1. 50% chance to do nothing.
2. If [1] is successful, make a roll which is inversely
   proportional to both distance from the player and monster hit die.
If both rolls are successful, temporarily blind the monster.

Example success rates for this effect will be commented below and no
doubt should be subject to further tweaks based on playtesting. The
feeling for the current rates is that this is a powerful item when found
early (maybe very powerful), falling off a little to a modest, but
noticeable, effect in the later game.

EV bonus is reduced to 2, though this of course overall is a buff to the
item.

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -924,7 +924,8 @@ PLUS:     +1
 COLOUR:   ETC_ICE
 TILE:     urand_starlight
 TILE_EQ:  white
-EV:       4
+INSCRIP:  Radiant
+EV:       2
 COLD:     1
 STEALTH:  -2
 BOOL:     elec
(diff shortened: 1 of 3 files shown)


amcnicky:
Range\HD 
   | 1         | 7         | 14      | 30
1 | 8.33% | 4.17% | 2.63% | 1.43%
3 | 6.25% | 1.92% | 1.06% | 0.53%
7 | 4.17% | 0.93% | 0.49% | 0.23%

PR #1574, opened 2020-09-20
Update INSTALL.md

elguitar:
Instructions for Void Linux were removed in f485615. The table of contents should not have dead links.


PleasingFungus:
Thanks for the catch!

PR #1576, opened 2020-09-23
Cancel Heroism and Finesse on angering or abandoning Okawaru

nlavsky:
Currently, only Heroism, Finesse, and Elyvilon's Divine Protection
don't expire immediately when the player is placed under penance. It
seems thematically appropriate for Divine Protection, but even Elyvilon
removes all divine effects on abandonment.

This commit puts Okawaru in line with other gods.


PleasingFungus:
Seems reasonable. Does make the Oka conduct a little more annoying, but that's probably a separate problem.

Thanks for the PR!

PR #1577, opened 2020-09-23
Dynamic make jobs for mac install

erf:
I don't think the `-j4` argument is needed to build the mac app ( i don't seem to need it, but not sure what it does ).

### DIFF
--- a/crawl-ref/INSTALL.md
+++ b/crawl-ref/INSTALL.md
@@ -152,8 +152,8 @@ Then follow [the above compilation steps](#compiling).
     ```
 
 3. If you want to build a macOS application, add `mac-app-tiles` to your make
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
It compiles in parallel. Should really be replaced with `-j $(sysctl -n
hw.ncpu)` or whatever the macOS command is to detect number of CPUs.

erf:
Ah i see. I looked at the command now, and it seem you can limit the number of cores with `-j n`. I wonder if you just say `j` that it will use max number of cores..? but your solution should also seem to work. I tried also with and without `-j` and various numbers, and it always use around 5.9s, but i've already build the project so.. I can change to your solution and push a new fix for that if you'd like.

erf:
Forced pushed a new commit with your changes.

PR #1579, opened 2020-09-27
-4 dex to each background; +4 dex to each species

anderspapitto:
This has no effect on balance, as the changes exactly cancel out.

The background contributions for str and int span reasonable, intuitive
ranges. Many backgrounds provide zero additional str, and many
backgrounds provide zero additional int. However every background
provides at least four dex. This rescaling makes background stat
contributions more clearly reflect "discretionary" assignment.

Note that it also makes the weird role of Wanderer clearer, in that
Wanderer has an outside chance to have even less dex than Conjurer.

### DIFF
--- a/crawl-ref/source/dat/species/barachi.yaml
+++ b/crawl-ref/source/dat/species/barachi.yaml
@@ -37,7 +37,7 @@ aptitudes:
 can_swim: true
 str: 9
 int: 8
-dex: 7
+dex: 11
 levelup_stat_frequency: 4
 mutations:
   1:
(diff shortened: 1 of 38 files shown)


anderspapitto:
I have no idea if the dev team is interested in something like this, but I thought I'd throw it together. Feel free to close if you prefer the status quo.

PleasingFungus:
Interesting. I think if people were going to object to this, it'd be on the basis of "human" no longer having even stats, whereas normally we expect human to be the absolute baseline. Human has uneven skill aptitudes for the same reason, though, so maybe it's fine?

anderspapitto:
eh, I guess I've come around to the status quo on this. Dex from backgrounds also has a tighter range than str/int, so it would look like a distribution even if shifted to have a similar baseline value of 0

PleasingFungus:
Sorry for the long lack of response on this! I appreciate the thought, even if you decided against the change.

PR #1582, opened 2020-10-11
Check right god for life-saving in god description

ufshaikh:
Currently describing a god checks your passives when deciding whether to
print something for the protect_from_harm passive. This leads to errors
in both directions: the passive is spuriously indicated when describing
gods that don't give it if you happen to be worshipping a god that gives
it, and is not printed for gods that do give it if you're not
worshipping them.

After this commit the god description checks whether the god being
described gives that passive.


PleasingFungus:
I've pushed this with a bit of refactoring. Thanks for the fix!

PR #1584, opened 2020-10-13
docs: Document invalidate_agrid

CrawlCycle:
The invalidate_agrid function invaldiates the area effect
cache. The function has about 40 usages in the codebase.
This commit adds doxygen comments to explain its effect.

### DIFF
--- a/crawl-ref/source/areas.cc
+++ b/crawl-ref/source/areas.cc
@@ -28,6 +28,7 @@
 #include "traps.h"
 #include "travel.h"
 
+/// Bitmasks for area properties
 enum class areaprop
 {
     sanctuary_1   = (1 << 0),
@@ -36,14 +37,16 @@ enum class areaprop
     halo          = (1 << 3),
     liquid        = (1 << 4),
     actual_liquid = (1 << 5),
-    orb           = (1 << 6),
+    orb           = (1 << 6), ///< The glow of the Orb of Zot
     umbra         = (1 << 7),
-    quad          = (1 << 8),
+    quad          = (1 << 8), ///< Multiplies damage you do by 4
     disjunction   = (1 << 9),
     soul_aura     = (1 << 10),
 };
+/// Bit field for the area properties
 DEF_BITFIELD(areaprops, areaprop);
 
+/// Center of an area effect
 struct area_centre
 {
     area_centre_type type;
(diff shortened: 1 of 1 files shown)


crawl-ref/source/areas.cc
```diff
@@ -91,6 +111,11 @@ void areas_actor_moved(const actor* act, const coord_def& oldpos)
     }
 }
 
+/// \brief Add some of the actor's area effects to the grid and center caches
+/// \param actor The actor
+/// \details Adds an actor's silence, halo, liquidation, and umbra effect to
+/// the area grid (\ref _agrid) and center (\ref _agrid_centres) caches.
```
PleasingFungus, line 117: In a few places (in this & other commits), you go into this level of detail, listing every current possible value for something. That's fragile, since if we add or remove area effects, we're quite unlikely to update this comment. Better to just talk about the category with maybe one example - "area effects (e.g. halos)" - and leave it at that.

crawl-ref/source/areas.cc
```diff
@@ -156,8 +181,8 @@ static void _update_agrid()
         return;
     }
 
-    _agrid.init(areaprops());
-    _agrid_centres.clear();
+    _agrid.init(areaprops());  // Set all cells of _agrid to 0
+    _agrid_centres.clear();  // Clear the vector of centers of area effects
```
PleasingFungus, line 185: Do these comments make the code clearer?

crawl-ref/source/areas.cc
```diff
@@ -71,6 +79,18 @@ static bool _check_agrid_flag(const coord_def& p, areaprop f)
...
+/// area effects) may trigger a re-calculation (of the caches by checking the
+/// player and all monsters for area effects). If \p recheck_new is true,
+/// next check will always trigger the re-calculation. Otherwise, next check
+/// will trigger the re-calculation only if the cache believes the level has
+/// existing area effects.
+/// \param recheck_new
+/// - If true, next check will always trigger a re-calculation. Use this if we
+///   have just created a new area effect.
+/// - If false, next check will trigger a re-calculation only if the level
+///   has existing area effects.
```
PleasingFungus, line 93: This is a useful comment, though it could use a little work on phrasing and grammar. I'd probably go with something like:
```
// Invalidates the area effect cache, causing the next request for
// area effects to re-calculate which locations are covered by halos, etc.
// If recheck_new is false, the cache will only be invalidated if the level
// had existing area effects.
```

We can do the whole doxygen style comments if you want (though it's very unclear to me that anyone is ever generating or reading doxygen comments?). The key differences that I'm aiming for here are that we're explaining what's going in a shorter way, which is simpler for the reader to understand, and that we aren't telling the user what to do. It's rarely correct to say "use this if we've just created a new area effect" or things along that line. In most cases, we'd rather just explain what the options do and let the reader figure out how that applies to their use case.

crawl-ref/source/areas.cc
```diff
@@ -36,14 +37,16 @@ enum class areaprop
     halo          = (1 << 3),
     liquid        = (1 << 4),
     actual_liquid = (1 << 5),
-    orb           = (1 << 6),
+    orb           = (1 << 6), ///< The glow of the Orb of Zot
     umbra         = (1 << 7),
-    quad          = (1 << 8),
+    quad          = (1 << 8), ///< Multiplies damage you do by 4
```
PleasingFungus, line 42: Is this comment helpful here?

PR #1585, opened 2020-10-14
docs: Document artefact_property

CrawlCycle:
The artefact_property function looks up the value of one
artefact property of an artefact. The function has about
65 usages outside of artefact.cc but it lacks doxygen
documentation. I also documented the artefact_properties
function, which serves a similar purpose.

### DIFF
--- a/crawl-ref/source/artefact.cc
+++ b/crawl-ref/source/artefact.cc
@@ -1014,6 +1014,13 @@ void artefact_known_properties(const item_def &item,
     }
 }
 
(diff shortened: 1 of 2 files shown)


PleasingFungus:
I like the `artp_value_type` documentation in this PR and would merge it if you re-opened the PR. (I think the function documentation isn't quite up to the same standard - that code is more self-explanatory, and the 'run time error' mentions are an internal detail that doesn't need to be mentioned here.)

PR #1586, opened 2020-10-14
build: Make the catch2-tests target look for tests

CrawlCycle:
Currently, adding a new set of catch2 tests requires:

  1. Add a new test_xyz.cc file in the source/catch2-tests
  2. Add test_xyz.o to TEST_OBJECTS in Makefile.obj

This commit removes the need to perform step 2 by making
the TEST_OBJECTS automatically search for the tests in
the source/catch2-tests/ directory. The change would
streamline the process and help to avoid potential merge
conflicts that arise when multiple developers try to
commit new sets of catch2 tests simultaneously.

### DIFF
--- a/crawl-ref/source/Makefile.obj
+++ b/crawl-ref/source/Makefile.obj
@@ -303,22 +303,9 @@ windowmanager-sdl.o \
 glwrapper-ogl.o \
 fontwrapper-ft.o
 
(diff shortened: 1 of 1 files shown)


CrawlCycle:
i don't know about this one. hmm...

PR #1587, opened 2020-10-18
test: Add catch2 fixture for running dLua/cLua

CrawlCycle:
Writing tests for an entity in crawl's main source requires:

    1. Setup a test scenario
    2. Call the entity

Setting up the scenario with dungeon builder lua (dlua) is
considerably easier than with the C++ functions. However,
choosing to use Lua to test the entity requires calling the
entity in Lua. That means adding C++-Lua binding in crawl's
main source code.

However, sometimes writing the binding is difficult. For
example, while Lua has iterator over list, Lua does not have
the direct equivalent of a C++ iterator over more complex
data structures. That means writing a binding for the
iterator requires also writing a binding for the data
structure to bring the concepts of both the iterator and
the data structure to Lua. Without writing the direct
binding, a possible alternative would be to write C++
functions for all test cases for the iterator, and then
bind all the C++ functions to Lua. The problems of this
approach are: 1. The lua bindings for those particular tests
are not generally reusable for the other purposes. 2.
Implementing all the tests in the main source of crawl
needlessly bloats the main source. 3. Adding lua binding to
C++ function is a new feature. The bindings may warrant
new tests.

This commit adds a catch2 fixture that enables running dlua
commands and scripts in catch2 tests, which are written in
C++. The fixture setup a cut-down initialization of crawl,
up to the point of the execution of Lua tests (such as
test/mutation.lua). The teardown of the fixture is an
adaptation of the reset game function in main.cc. Some
initialization steps of crawl has no corresponding clean-up
function and running these steps twice crashes the tests.
Therefore, this commit adds an execution guard to ensure
only one execuation of these steps across all tests. This
commit also includes catch2 tests of some features of the
fixture.

The limitations of this fixture includes: no screen (can't
run abyss_shift.lua), tile, android, or hint mode. Can't run
the tests concurrently. Future modifications of crawl's
initialization or definition of game state may bring
uncontrolled variables into tests based on this fixture.
Setting up reusable and more reliable test scenarios require
verifying the assumptions of the scenario.


CrawlCycle:
The current catch2 tests use fake-main.cc, which creates the global variables and global Lua interpreters for the game state (`you`, `env`, `crawl_state`, `clua` and `dlua`). 

Nearly everything in crawl's source code modify these global variables. That means setting up test scenarios with crawl's source code use many crawl functions that all modify the global variables. These many functions possibly have unexpected and untested interaction with the entity to be tested.

True unit testing of an entity require isolation of the entity. Replacing the global variables by mocks would cut the interaction of the entity with the other untested crawl functions. The mock would ideally only implement the minimal requirement of the test.

However, manually setting up mocked global game states has high maintenance cost. Change of crawl's definition of the game state (e.g. modification of any class used by `env`) would require corresponding manual change of the mocks. Using mock frameworks can make writing, using, and updating mocks easier.

The approach implemented in this commit is not true unit testing because it doesn't mock the global game state. But writing unit tests with this approach is immediately easy. Implementing checks within test scenarios would help to check for unforeseen change of the source code. For example, orcs may see invisible in the future and that violates the assumption in a test scenario that it can't see the invisible player an bat. Those check can be built into the reusable test scenarios. 

Which way is better?

By the way, I would think that naming specification of test scenarios with "scenario_xyz.cc" and naming fixture file as "fixture_xyz" is a good idea.

CrawlCycle:
Probably this pull request is too big.

PR #1594, opened 2020-10-24
Appreciate the great tradition of curling

PleasingFungus:
To make palentongas a bit more interesting, and to synergize with their
'roll into a bunch of enemies' power, give palentongas a passive
'defensive curl' mutation at XL 7. This gives them +7 AC after they've
been hit, lasting until the *start* of their next turn. In short, it
makes them tougher against multiple enemies or against multi-hit enemies
like hydras, somewhat like the inverse of how shields work.

The main issue here is communication. I'm not sure how best to demonstrate
to the player that this is happening on subsequent hits without spamming
'you curl/you uncurl' messages after every hit. The behaviour in this PR
has no spam but is also super easy to miss entirely. Suggestions welcomed!


crawl-ref/source/ouch.cc
```diff
@@ -796,6 +796,40 @@ static bool _is_damage_threatening (int damage_fraction_of_hp)
...
 }
 
+// Palentongas curl up after the first time they've been hit in a round.
+static void _consider_curling(kill_method_type death_type)
+{
+    // can't curl
+    if (!you.has_mutation(MUT_CURL))
+        return;
+
+    // already curled
```
aidanholm, line 806: Nit: I think these two comments are probably not needed; the code is clear enough.

crawl-ref/source/main.cc
```diff
@@ -2331,6 +2331,12 @@ void world_reacts()
     // the loudest noise tracking for the next world_reacts cycle.
     you.los_noise_last_turn = you.los_noise_level;
     you.los_noise_level = 0;
+    // palentongas uncurl at the end of the turn
+    if (you.props[PALENTONGA_CURL_KEY].get_int() > 0)
+    {
+        you.props[PALENTONGA_CURL_KEY] = 0;
```
aidanholm, line 2337: What about making it boolean? Seems this code only ever uses 0/1.

PleasingFungus: For some reason I thought we didn't support bools. Good call.

amcnicky:
Do you think having a separate tile for the player doll, to be displayed briefly during monster/spell animations, once defensive curl has been activated, would be helpful with respect to communication of the effect? So, for example, having the player doll look a little more 'hunkered down'/'curled up' whilst all those yaktaur bolts are zapping around the screen.

PleasingFungus:
@amcnicky absolutely! I was considering exactly that. (it'd be nice if the centaur player tiles were redrawn or at least recoloured to be less half-human and more all-palentonga at some point, too...)

crawl-ref/source/player.cc
```diff
@@ -5964,6 +5964,12 @@ int player::armour_class_with_specific_items(vector<const item_def *> items) con
 
     AC += sanguine_armour_bonus();
 
+    if (you.has_mutation(MUT_CURL)
+        && you.props[PALENTONGA_CURL_KEY].get_bool())
+    {
+        AC += 7 * scale;
+    }
+
```
aidanholm, line 5972: Refactoring idea: rather than toggling a bool and having this method check all bools, add/remove an "AC effector" instance on a vector. Calculating AC would iterate over all instances, get each AC component, and tally them up. This works because AC is additive.

Advantages:
 - You don't have to set `you.redraw_armour_class`; that can be done when pushing/popping an AC affector.
 - Avoids another long function that has lots of dependencies / does a lot of stuff. Keeps related code together.
 - More composable, more easily testable.

Possible disadvantages:
 - Can't see all AC calculation code together. Not sure how much this would matter.

PleasingFungus: I do like the idea of refactoring! I've never liked the way that `redraw_armour_class` works.

One note is that `redraw_armour_class` is *also* responsible for updating AC colour *and* SH value & colour, but your suggestion shouldn't conflict with that, if I understand it correctly.

PR #1598, opened 2020-10-26
Change the brand of Glaive of Guard to spect

sdynet:
I saw the development idea and worked on it. However, I don't think we need to turn this weapon into a double brand weapon because we already have a thermic engine.
https://github.com/crawl/crawl/wiki/Unrand-Review

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -449,7 +449,7 @@ PLUS:     +8
 COLOUR:   ETC_ELECTRICITY
 TILE:     urand_guard
 TILE_EQ:  glaive_of_the_guard
-BRAND:    SPWPN_ELECTROCUTION
+BRAND:    SPWPN_SPECTRAL
 AC:       5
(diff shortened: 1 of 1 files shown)


sdynet:
I've heard a funny idea, 'What do you think about creating five Spectral Weapon?' but it seems hard to apply. But creating two seems fine. :) By the way, what code do I need to modify to make two Spectral Weapon appear when player attack the enemy?

ebering:
The spectral weapon code is not really set up for there to be more than one spectral weapon, but it wouldn't be hard to generalize it. The relevant functions are in the following in spl-summoning.cc

`find_spectral_weapon` would need to return a vector instead of a single monster pointer
`cast_spectral_weapon`, `end_spectral_weapon`, `trigger_spectral_weapon`, would all need to handle this vector. Finally in `handle_spectral_weapon` in player.cc would need to be modified.

I think having GoG create multiple spectral weapons is a cute idea; 3 is a good number (subject to balance concerns...).

PR #1603, opened 2020-11-04
Address issue #0012378

dragonsploder:
This is my first time doing this, so this is just a very simple bug fix to get the feel of things.

### DIFF
--- a/crawl-ref/source/dat/species/kobold.yaml
+++ b/crawl-ref/source/dat/species/kobold.yaml
@@ -34,6 +34,7 @@ mutations:
     MUT_NIGHTSTALKER: 3
 recommended_jobs:
   - JOB_HUNTER
+  - JOB_BRIGAND
   - JOB_BERSERKER
   - JOB_ARCANE_MARKSMAN
   - JOB_ENCHANTER


PleasingFungus:
Hey, thanks for submitting this! But I'm quite confused by this (both the PR and the bug report), sorry. Why do you think there is a bug here?

The goal of the recommended backgrounds/species lists isn't to list every single strong combination. It's to list a specific set of combinations that we think that players will enjoy. It's not a big deal to add brigand to this list, but I don't see any particular reason why we should. Could you please explain?

dragonsploder:
Thanks for the quick reply. I don't have a great understanding of crawl balance, so I just assumed whoever made the bug report knew more than me. I guess we could add it because there are no other really "stealth" roles recommended for kobolds, and they have high aptitude in stealth, but I really don't know.

PleasingFungus:
Gotcha!

To clarify, brigand isn't really a "stealth role" - they're a melee background that starts with some poison to deal with early tough enemies. And Stealth aptitudes aren't really very important - it's sometimes nice to be able to avoid fights, but if you're looking for a 'stealth playstyle', you need the ability to incapacitate enemies, with Stealth skill doesn't really help with.

For that reason, Enchanter actually is a 'stealth background' - it's the stealthiest background in the game. Arkane Marksperson also has some hexes, along with playing into kobolds' good ranged apts.

dragonsploder:
Thanks. That makes sense. I can close the request, but I don't know how to deal with the bug report on the Mantis tracker.

PleasingFungus:
Thanks, sorry for the confusion! I'll ask a Mantis admin to close your bug report.

PR #1604, opened 2020-11-07
Fix display of hex chances on the xv screen

nlavsky:
After 424c9a47, xv displays hex chances only for spells of friendly
monsters.


PleasingFungus:
Good catch, thanks!

PR #1605, opened 2020-11-08
Show monster spell damage on xv in WebTiles

nlavsky:
Commit 9d7e6eef added this functionality to console and Local Tiles.
Also, since `_effect_string()` already returns a hex chance, deduplicate
some code.


PleasingFungus:
Thanks for the catch & the fix!

PR #1607, opened 2020-11-10
Properly Punctuate small damage distortion

Midn8:
I don't even know how long this was broken, attack.cc has had a comment claiming that it was fine since disto can't do more than 7 damage, but 7 damage is one ! already. Ideally someone will also fix harm to work properly with attack_strength_punctuation, but that's above my pay grade.


wheals:
Manually merged, thanks for the fix!

PR #1608, opened 2020-11-10
travel_open_doors extra setting

Aliscans:
Add an option for explore & travel to treat closed doors more like walls than floor.

Previously, travel_open_doors had two values.
If true, the travel command would open a door if it was quicker than going around. The explore command would open the door if this was the quickest way to get to something interesting.

I have retained these, and added a third in which it treats it in the same way it treats a runed door.

There are two changes to the way explore behaves regardless of travel_open_doors. If autopickup is on, and there is a known object behind a runed door, it used to give a "Partly explored" message. It now moves to the door and stops there (it already did this if the area behind the runed door was partly explored). And, if you try to explore or autotravel, but don't move for any amount of time, it gives a message to tell you.

I've added the code to set "set option" code to game_options::read_option_line in the section "easy_confirm" is in. As it's exposed to lua, I've also added code in l-option.cc. 

An alternative would be to make a GameOption-style class for it. There are already a few options which accept one of a fixed list of values, so this feels more like a separate enhancement to me.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -985,8 +985,8 @@ auto_exclude += <monster name>, <monster name>, ...
 
 auto_switch = false
         This option will allow you to automatically switch to an appropriate
-        weapon when attacking in melee, as long as the one you are wielding
-        and the one you switch to are both in slot 'a' or 'b'.
+        weapon when firing or attacking in melee, as long as the one you are
+        wielding and the one you switch to are both in slot 'a' or 'b'.
 
 travel_open_doors = (avoid | approach | open)
         Configure how autoexplore/travel interacts with doors.
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -988,9 +988,15 @@ auto_switch = false
         weapon when attacking in melee, as long as the one you are wielding
         and the one you switch to are both in slot 'a' or 'b'.
 
-travel_open_doors = true
-        If this is set to false, autoexplore/travel will not open doors,
-        instead stopping in front of them.
+travel_open_doors = (avoid | approach | open)
+        Change how autoexplore/travel interact with closed doors.
+           avoid = If they can explore (or reach their destination) without
+                   opening doors, they will do. If not, the stop by a closed
+                   door.
+        approach = They walk up to a door if it's on their route, and stop in
+                   front of it.
+            open = They open any closed doors in their path. (default)
+        This does not affect runed door, which they always avoid.
 
 easy_unequip = true
         Allows auto removal of armour and jewellery when dropping it.
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -989,13 +989,13 @@ auto_switch = false
         and the one you switch to are both in slot 'a' or 'b'.
 
 travel_open_doors = (avoid | approach | open)
-        Change how autoexplore/travel interact with closed doors.
-           avoid = If they can explore (or reach their destination) without
-                   opening doors, they will do. If not, the stop by a closed
-                   door.
-        approach = They walk up to a door if it's on their route, and stop in
-                   front of it.
-            open = They open any closed doors in their path. (default)
+        Configure how autoexplore/travel interacts with doors.
+           avoid = Autoexplore/travel will treat closed doors like walls. If
+                   the only way for exploration or travel to continue is through
+                   a closed door, it will stop by a closed door.
+        approach = Autoexplore/travel will not open doors, instead stopping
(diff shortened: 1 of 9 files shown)


crawl-ref/source/travel.cc
```diff
@@ -333,6 +336,38 @@ static bool _monster_blocks_travel(const monster_info *mons)
...
+    }
+
+    if (you.elapsed_time == you.elapsed_time_at_last_input)
+    {
+        string barrier;
+        if (feat_is_runed(grid)) barrier = "unopened runed door";
+        else barrier = "unopened door";
+        mpr("Could not " + you.running.runmode_name() + ", " + barrier + ".");
+    }
+    return true;
```
ebering, line 368: Small logic checks in static methods should when possible avoid side-effects like this, better to handle messaging from where this function is called from.

crawl-ref/docs/options_guide.txt
```diff
@@ -985,9 +985,15 @@ auto_switch = false
...
-        instead stopping in front of them.
+travel_open_doors = (avoid | approach | open)
+        Change how autoexplore/travel interact with closed doors.
+           avoid = If they can explore (or reach their destination) without
+                   opening doors, they will do. If not, the stop by a closed
+                   door.
+        approach = They walk up to a door if it's on their route, and stop in
+                   front of it.
+            open = They open any closed doors in their path. (default)
+        This does not affect runed door, which they always avoid.
```
ebering, line 996:
```suggestion
travel_open_doors = (avoid | approach | open)
        Configure how autoexplore/travel interacts with doors.
           avoid = Autoexplore/travel will treat closed doors like walls. If the only
           way for exploration or travel to continue is through a closed door, it will stop by
           a closed door.
        approach = Autoexplore/travel will not open doors, instead stopping in front of any
        door in its path.
            open = Autoexplore/travel will open doors. (default)
        This does not affect runed doors, which are always avoided.
```

I can't set the whitespace correctly in GitHub, but this is some improved wording.

ebering (changes requested):
Hello and welcome to contributing to crawl! Thank you for the submission. I'm not 100% clear on why someone would want to autoexplore this way, but it's clear you do since you went to the trouble of implementing it. Additionally, it looks like it improves some messaging when autoexplore gets stuck, which is a long desired implementable. Thank you for that!

I've gone through and annotated the specific files with details. In addition to the small things, the big change that needs to happen is: the `l-option.cc` lua interface should be removed and the `X` macro not used. See the detailed comments for why `l-option` isn't needed.

ebering:
Thank you for the changes. I'm not sure what happened with your commit history, it looks like some funny business with merges? Anyhow, for future reference the preferred crawl workflow for a PR is:

- keep working on the branch, pushing new changes in response to feedback (don't bother merging master for small changes)
- when finished, either you or the merging devteam member will rebase the patch to current master

ebering:
How would you like to be listed in `CREDITS.txt`?

Aliscans:
On Mon, 28 Dec 2020 11:08:50 +0000 (UTC)
"Edgar A. Bering IV" <notifications@github.com> wrote:

> Thank you for the changes. I'm not sure what happened with your commit
> history, it looks like some funny business with merges? Anyhow, for future
> reference the preferred crawl workflow for a PR is:

Hi Edgar

I only rebased it because the "check" page on github said it hadn't compiled.
The "problem" it reported was in game-option.h, where it seems to have tried to
moosh together the changes I'd made and changes someone else had made since I
created the branch, and then compile that version. It didn't work, but there's
no reason why it should have done.

It skipped most of the other checks, so I didn't find out if it actually built
correctly on github's system.

Rebasing the source seemed like the easiest way to make it compile. It still
didn't pass the checks, though, as github said a travis-ci check had failed. I
didn't know what that was testing.

Should I ignore the checks on github in future?

A couple of points in response to a previous e-mail:

1. You refer to an "important" check in movement.cc. I can't see anything in
the documentation to explain what this should be, so I'm at a bit of a loss.

From what I can see, the effect of the check you added is to ensure that moving
towards a door from one square away using (say) "L" does nothing unless
travel_open_doors is set to "open". Moving from further away always stops next
to the door, and moving using "l" always opens the door. It has no effect on
autotravel/explore, as _feat_is_blocking_door_strict() does the same thing.

If this is actually what it's intended to do, would _safe_move_player() be a
better place for it than move_player_action()? Then the game wouldn't run the
code that often, and it wouldn't be buried in a 400 line function.

2. With regard to lua, travel_open_doors seems to have replaced the easy_open
option in the options code at some point. easy_open had been exposed to lua, so
travel_open_doors was. easy_open was exposed to lua in the first version on
github.

I don't know much about the lua interface, so I don't know how useful it was to
have either option there.

Thanks.

Kieron

PR #1615, opened 2020-11-13
Colourize beam damage descriptions in WebTiles

nlavsky:
As a followup to 930a2b7a, this commit colourizes beam damage
descriptions on the xv screen:

spells


PleasingFungus:
Thank you for this. It brings me great happiness. Merged as 67df3f2acb .

PR #1616, opened 2020-11-14
+1 spellcasting apt to all species

anderspapitto:
Spellcasting has received some indirect nerfs. It no longer serves
to reduce hunger costs, and the "1 level spellcasting = 0.25
levels of every spell school at once" has been devalued by the
removal of the charms school. There are also fewer total spell
levels in the game since the charms removal and positional magic
overhaul, which reduces the value of spell slots. This also makes
0 the standard "human-level" aptitude for spellcasting, bringing
it in line with almost all other skills.

### DIFF
--- a/crawl-ref/source/dat/species/barachi.yaml
+++ b/crawl-ref/source/dat/species/barachi.yaml
@@ -22,6 +22,7 @@ aptitudes:
   dodging: 1
   shields: 1
   unarmed_combat: 1
+  spellcasting: 1
   conjurations: 1
   hexes: 1
   summoning: 2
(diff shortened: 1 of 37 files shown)


Midn8:
" There are also fewer total spell
levels in the game since the charms removal and positional magic
overhaul, which reduces the value of spell slots."

I'm not sure why there being fewer levels of spells you want to memorize is an argument for INCREASING everyone's spellcasting apts - rather, this would be an argument to decrease them to bring it more in line with previous changes. Similarly, spellcasting wasn't trained on most characters to reduce spell hunger or increase spell power - rather you train it to get more slots and MP, which is still does both of at a good rate.

ebering:
I agree with @Midn8 's comments. I'm not convinced that there's much out-of-balance about the spell casting skill. Especially considering that even with the reduction in the total number of spells there are still far more spells than a character can memorize in a normal game, and its contributions to maxMP and all spell skills are far more significant than hunger reduction.

I'll leave this open for a week or so in case another team member wants to weigh in.

Issue #1564, opened 2020-09-13
Grey Draconian poisoned by poison gas clouds


ardl:
**PLATFORM and OS or BROWSER:**
Webtiles trunk on cbro. Firefox 80

**DESCRIPTION:**
When standing in poison clouds grey draconians get poisoned even though they are unbreathing. Mephitic clouds have no effect as intended. I checked to make sure it wasn't the initial hit of the poison breath by swamp dragons. If this is intended then that's fine, the learndb says unbreathing so negate the effect of poison clouds and that might just refer to unbreathing enemies.

**EXTRA FILES:**

![image](https://user-images.githubusercontent.com/70080435/93022770-ea2fa500-f5b8-11ea-9df9-c3bda17da769.png)

*To upload a save file directly to github, compress it to a .zip archive, and
then github will accept the upload. (It will not take a .cs file directly.)*

*If this is an online game, see [these instructions](https://crawl.develz.org/wiki/doku.php?id=mantis:savebackup)
for more information on how to make a save backup; you can provide the
resulting link here.*

PR #1622, opened 2020-12-01
Fixed unbreathing bug (#1564)

sefie1:
The [issue](https://github.com/crawl/crawl/issues/1564) describes that a grey draconian can get poisoned by a poison cloud, even though it is unbreathing.

I verified this in WizMode.

This PR ensures that the cloud_immunity_check checks whether the affected actor is unbreathing.


rawlins:
Thanks for the fix! I manually merged the specific patch in https://github.com/crawl/crawl/commit/e1f20c3b445d4e6f4d869a2e080ec252b498914c, because something went odd with the commit sequence in this PR apparently due to a merge. How would you like to appear in the credits?

sefie1:
The odd sequence was my error I believe... Thanks for going through the
trouble..

"Heinrich Ody" (without quotation marks) would be nice.



Am Fr., 11. Dez. 2020 um 21:57 Uhr schrieb Kyle Rawlins <
notifications@github.com>:

PR #1624, opened 2020-12-04
Fix autofight not prioritizing petrified enemies

nlavsky:
After 9c76b8a6, there are only two stabbing tiers. Also, the player
gets a top-tier stab bonus when attacking petrified enemies.

This commit updates autofight rules so petrified enemies get the same
priority as sleeping or paralysed ones.


ebering:
Thanks for the patch!

PR #1627, opened 2020-12-06
add a new overflow vault for gozag and zin

pdpol:
This vault is an idea inspired by zin_treasury that I thought could be an amusing play on the contrast between Zin and Gozag.
Gozag tempts the player with a single-charged wand of digging to steal a negligible amount of gold from Zin's treasury just for the sake of it.

### DIFF
--- a/crawl-ref/source/dat/des/altar/overflow.des
+++ b/crawl-ref/source/dat/des/altar/overflow.des
@@ -3454,6 +3454,30 @@ DD-n....n--E
    cc@@cEEE
 ENDMAP
 
+NAME:   pdpol_temple_overflow_zin_treasury_gozag_greed
+TAGS:   temple_overflow_2 temple_overflow_zin temple_overflow_gozag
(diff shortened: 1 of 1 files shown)


ebering:
Welcome to contributing to crawl! Cute idea. I toyed with trying to make it more obvious or more symmetric, but I really like the subtlety. If players don't always get it, well, oh well, because making the Gozag side able to see the gold feels too hamfisted.

I merged this manually to tweak the commit message title (we like to uppercase the first letter, and its nice to capitalize the gods as well), but otherwise didn't tweak it.

How would you like to appear in `CREDITS.txt`?

pdpol:
Hi @ebering, thank you! I would love to appear in `CREDITS.txt` and intend to stay involved. Good to know about the commit message format, I'll keep the next one in line with that.

ebering:
What name would you like to have entered into the credits?

pdpol:
Paul Pollack would be great, thanks!

PR #1629, opened 2020-12-07
Don't show a bigger targeter for Hailstorm when worshipping Vehumet

nlavsky:
`spell_range()` can return 4 instead of 3 for Hailstorm, because
Vehumet supports the use of this spell. But Hailstorm's range never
changes, since it's hardcoded as 3 in `cast_hailstorm()`.


rawlins:
Thanks!

Issue #1408, opened 2020-05-15
Regexp Ignores String Anchors


Bartholomue:
DCSS doesn't respect line anchors like \A and \Z, making specific string matches very hard.
Consider the following sound triggers:
sound += (A|An|The) (arrow|(damnation )?bolt|(silver )?javelin) pierces through ((?!you).+)(\!\!|\!):combat/penetrate/2.wav
sound += (A|An|The) (arrow|(damnation )?bolt|(silver )?javelin) pierces through ((?!you).+)((?<!does no damage)\.):combat/penetrate/1.wav

The second option plays a certain sound if the text contains a single period or a single exclamation mark to denote damage level. The first option string plays another sound if it matches two or three exclamation marks, and so on.
The problem is that there is no way to anchor it. Even if you're looking for a single !, a string containing !! will match. If I could at least place a \A or \Z in the string, it would at least let the player create much more exacting patterns.

sefie1:
While it may not completely solve your problem, a tempory solution might be to reorder your regular expressions.

Apparently, Crawl always plays the sound for the last matching regular expression. Hence, putting the most restrictive expression last will help for cases where you want to distinguish the number of exclamation marks.

PR #1630, opened 2020-12-08
Improve sound option description

sefie1:
Added details for playing sounds via regular expressions

For sounds, added a sentence that only the last matching
 regular expression is considered.

This is related to #1408, as there it seems unknown how to distinguish messages with different amount of exclamation marks.

The extended description of the sound option alleviates this.


ebering:
Thanks for the patch! I merged it manually with a rebase. How would you like to appear in the CREDITS.txt ?

sefie1:
> How would you like to appear in the CREDITS.txt ?
"Heinrich Ody" (without quotation marks) would be nice.

PR #1660, opened 2020-12-28
Add a flashy Wu Jian overflow altar vault

pdpol:
This vault is a nod to a classic 70's Kung Fu flick called Five Deadly Venoms.
The titular characters of the movie have nicknames and martial arts
styles that are reasonably well-represented by existing monsters and
tiles in DCSS and thought Wu Jian deserves a fancy Lua'd up altar like
some other deities!

[Movie clip for reference](https://www.youtube.com/watch?v=znPXlUgcMWA)

I figured I'd put this out there to see if you guys like it, I thought it was a fun idea but did wonder if the demonic crawler name replace and abjuration timer were maybe a little hackier than permissible.

### DIFF
--- a/crawl-ref/source/dat/des/altar/overflow.des
+++ b/crawl-ref/source/dat/des/altar/overflow.des
@@ -108,6 +108,16 @@ function callback.xom_greatest_gift(data, triggerable, triggerer, marker, ev)
   end
 end
 
+function callback.five_deadly_venoms(data, triggerable, triggerer, marker, ev)
+  crawl.god_speaks("Wu Jian", "For a moment, you catch a glimpse of warriors who walked this path long ago.")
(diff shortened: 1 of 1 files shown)


crawl-ref/source/dat/des/altar/overflow.des
```diff
@@ -2354,6 +2364,44 @@ xxxxxx.xxxxxx
      x@x
 ENDMAP
 
+NAME:  pdpol_wu_jian_five_deadly_venoms
+DEPTH: D:2-, Lair
```
ebering, line 2368:
```suggestion
DEPTH: D:2-9
```

Overflow vaults are restricted to this specific depth, they get special handling in the level builder.

ebering:
Thanks for the vault! Great flavour!

PR #1663, opened 2020-12-29
Small update in options guide

Goratrix:
Eringya's Noxious Bog has a default force_targeter now,
so update options guide accordingly. Also fix two stray rows
which were longer than 80 characters.


ebering:
Thanks!

Issue #1639, opened 2020-12-13
Inaccessible vault generation


ukdong90:
![캡처](https://user-images.githubusercontent.com/65288602/102005741-a8043400-3d5e-11eb-958d-31f0e3fab743.PNG)

![1607807518](https://user-images.githubusercontent.com/65288602/102005758-d71aa580-3d5e-11eb-8ab9-1bc26905e9da.jpg)

Played trunk version on the CWZ server.

rawlins:
Thanks for the report; when this game is finished could you post the seed or just a link to the morgue on CWZ? (And also the level that this was found on.)

ukdong90:
morgue of the first picture 
-> https://webzook.net/soup/morgue/trunk/dilly/morgue-dilly-20201213-082507.txt

D:5:      cheibrodos_thug_alley, layout_basic,
          nicolae_ossuary_entry_false_entrances


The second picture was played by other person, so I cannot give you one. I think it's 'nicolae_ossuary_entry_catacombs' in ossuary.des file

I think the code in ossuary.des has problem with entry point.

PR #1665, opened 2020-12-30
Mark an ossuary enterance as transparent 2

ukdong99:
Resolves another vault which has the same issue in #1639


ebering:
Thanks for the patch!

PR #1667, opened 2020-12-30
Have Vehumet support Eringya's Noxious Bog

nikheizen:
Though Bog doesn't do its damage by hurling something, given precedent of
both Frozen Ramparts and Toxic Radiance: it does do direct damage,
albeit over time, as its main purpose. Thus, Vehumet should support it.

This commit follows the logic in `c6dd0926df0a3b2a1a04d66e495a1bbcbd61f404` of Noxious Bog being a damage spell, first and foremost.


gammafunk:
Fixed in 135aecc. Thanks for the PR, but Lici brought it to my attention in discord first, and I always do what discord wants, so...

PR #1668, opened 2020-12-31
some bugfixes for stone_soup-0.16

krator44:
here's a few compatibility/maintenance bugfixes for stone-soup-0.16.2
maybe we could merge them into this branch


ebering:
Hi, welcome to crawl contributing. We're in the midst of getting ready for the release of 0.26, so it'll be a bit before someone has a chance to look at this.

0.16 is also 10 stable versions ago, while we don't have a formal maintenance window (and have back ported serious security fixes before as well as build fixing tweaks), we typically don't continue to add ui improving features and tweaks beyond one or two stable versions in the past. If this is a one-off I don't see much harm in merging it, but in general ancient versions don't receive indefinite continued development, so if you're interested in being involved with the project long term you might consider working with the latest trunk.

Finally a quick code review remark: style-wise we prefer our commit messages to start with a capital letter; if the commit body is a continuation of the sentence in the commit title, use `...` (but try not to do this).

rawlins:
Just to add to @ebering's note, I agree that most of what's here could be worth merging at least as a one-off. However, I don't think it would be ideal to merge the rng commit. I understand why one might do  this, in that ASG was pretty bad, but (i) this is a substantial divergence from how the project actually went (it switched to PCG in 0.17, see 415d666aa2f05; dcss is still using PCG with 128 bit seeds), (ii) it strikes me as something that would require non-trivial testing / maintenance burden to merge (especially given that 0.16 is still live on some servers), and (iii) there are a bunch of problems that would bias me against Mersenne Twister in principle, mainly to do with off-the-shelf prediction algorithms that work with a very small and constant number of draws. See (among others):

 * https://nitish.ch/notes/predictability-in-a-prng/
 * https://arxiv.org/abs/1910.06437
 * https://github.com/kmyk/mersenne-twister-predictor

Backporting 415d666aa2f05 to 0.16 (if that is possible; not sure) *might* be more viable in terms of getting someone on the devteam to merge it.

I realize that this may not be what you would want to hear, given that it appears that you may have written the MT implementation yourself, which I realize is a non-trivial project in and of itself, and I don't mean this as a review of your MT code. But hopefully this will at least provide some guidance as to how future commits to a mostly non-maintained branch like this might be received.

krator44:
agreed. removing changes and tweaks.

krator44:
in fact I completely understand the reluctance in maintaining a branch ten releases out of date

PR #1671, opened 2021-01-04
Ely rework

petercordia:
I'm sure this will have to become a playable branch before merging it in will be considered, but I thought it might still be worth making the pull request so that there is a space for discussion on Git.

Currently implemented:
* worshipping Ely gives a (permanent) mutation after which seeing monsters grants 50% of their xp, and killing monsters grants none.
* Ely wrath removed

Still planned as part of the Ely rework:
* update Ely descriptions
* re-introduce priest/pacifist background
* change Ely piety to give small penalty for killing living creatures
* make monsters go neutral when they are classed as "not dangerous" and the player has gained a significant amount of xp since first seeing them
* make neutral monsters drop their equipment when they leave

Planned because it'll be an experimental branch and why not:
* simplify weapon damage
* make weapon compare screen


ebering:
Hi, this draft PR hasn't been updated in two months, and also hasn't generated any discussion.

It seems like the goal is to make Ely support a pure pacifist platstyle.

We moved away from a pacification-heavy Ely for a reason: smite targeted "go away" plus lots of monster avoiding using a tiny subset of crawls skills (escape spells, stealth, nets, and atropa darts these days) isn't super good gameplay in the crawl world. Almost the entire skill and character building system is structured around fighting monsters (compared to some other roguelikes). A penalty for killing and the XP modifier being permanent (a harsher wrath than even keeping Ru sacrifices) strongly pushes the player into this bad gameplay space.

Since this PR has been idle and not attracted other team members attention I'm going to go ahead and close.

As an aside
> Planned because it'll be an experimental branch and why not

It's best to keep PRs to a single topic.

PR #1675, opened 2021-01-09
Remove Will-O-The-Wisps from early vault

ardl:
Removes Will-O-The-Wisps from pond vault


gammafunk:
This PR seems to be based on an outdated version of the DES from possibly 0.25, as it removes insubstantial wisps from the monster list, not will-o-the-wisp, which is the current monster in that list.

I will adjust this vault properly for depth in a commit I'll soon merge. It will still sometimes place a wisp, just properly adjusted for depth. The current map places monsters that are fine for D:10 but way too easy for late Lair and Dungeon.

PR #1689, opened 2021-01-15
Monster spell overhaul

ebering:
The genesis of this branch is from a couple of different places. First is some recent experiences of mine this tournament trying to keep track of just which dang deep elf mage was the one with lightning bolt (augh); next, various conversations with players about "why on earth do Ancient Champions cast Pain" and similar questions. Finally, and this will possibly appear in a separate branch, I've been thinking about crawl's use of demons (summon (greater) demon is on so many monsters). There's a lot of missed opportunity for more focused spell books to create greater flavor as well as better threats.

Out of this tumble comes this branch.

Goals
=====

- Make all monsters single spell book
- Remove low level spells from high level monsters
- Reduce reliance on overused spells
- Replace generic summons with focused summons for stronger flavor

Summary of changes
================

Pain is the most notable low-level spell on high level monsters, it has been cleaned up, along with a look at the other level 1 damage spells used by monsters. Some hexes might also need looking at, but hexes remain impactful throughout the game.

There are a few more small spell book tweaks aimed at making monster books tighter theme wise.

Psyche gets a chaotic new spell set to replace the orc wizard book she previously had, which should be significantly more memorable.

The major change is converting most monsters to single spell book monsters in d5e8436. The commit message is copied into this PR in the section book changes.

Book changes
===========

The general guiding principle I've followed when assembling the new
book is to try to follow the "feel" of the old book sets that a player
would get if they weren't checking each monster's book carefully.

Faun
----

The old Faun spellbooks were corona + another hex. There are a mix of
Will checking threats in shoals, but often players arrive with lots of
Will gear stacked up. Fauns show up in packs and their Corona+Hex book
was designed to make them amplify other ambient ranged threats. The new
book gives them Strip Willpower in place, for helping one another and other
ambient monsters hex the player.

New Faun book:

- Corona
- Strip Willpower

Deep Elf Knight
---------------

Deep Elf Knights had one book with Icicle, Haste, and Invisibility,
and a second which had Stone Arrow, Force Lance, and Invisibility.
The positional implications of Force Lance are great to keep around,
so it's added to the rest of the first book.

New Deep Elf Knight book:

- Throw Icicle
- Force Lance
- Haste
- Invisibility

Orc Wizard
----------

Orc Wizards had three(!) distinct books, but their most iconic behavior
was confusing the player and going invisible. This is book II, shared
with Sigmund, and this is made into the single Orc Wizard book.

New Orc Wizard book:

- Magic Dart
- Throw Flame
- Confuse
- Invisibility

Blork
-----

Blork the Orc used Orc Wizard book III, this is basiclly preserved but
emergency invisibility is removed from the book since it was quite long.

New Blork book:

- Cantrip
- Magic Dart
- Throw Flame
- Throw Frost
- Haste

Necromancer
-----------

Necromancers had two books which differed only in the choice of bolt
spell. In a large fight where Bind Souls has produced many simulacra,
the player could hide from either bolt by using the simulacra to block
LOF. The new book gives Necromancers Bolt of Draining, which works
better with their mechanic.

New Necromancer book:

- Bolt of Draining
- Agony Range
- Bind Souls
- Invisibility

Wizard
------

Wizards' book assortment conveyed a mix of magics, to imply a mastery.
Based on the D&D class as far as I can tell. To keep that flavor I
picked a range of spells from their books that produced the same "mixed
magic" feel. They share depth with Ogre Magi but don't come in packs, so
I went with fireball and banishment here to differentiate the two spell
sets.

New Wizard book:

- Fireball
- Banishment
- Blink
- Invisibility

Hell Knight
-----------

Hell Knights are one of a couple of late game monsters that cast Pain,
which at their depth isn't much of a threat even when it can beat the
player's Will. Bolt of Fire is good for them, they can just use that.
Perhaps blink should be made non-emergency, but because Hell Knights
come in packs an injured HK blinking away to let a fresh one attack the
player is a nice touch.

New Hell Knight book:

- Bolt of Fire
- Blink (emergency)
- Haste

Ancient Champion
----------------

Another late game monster with Pain. Since they often come in bands with
undead, retaining Bolt of Draining might be good but for now I went with
short and sweet.

New Ancient Champion book:

- Iron Shot
- Haste

Royal Mummy
-----------

Each of the four Royal Mummy books had Smite and Torment, the remaining
mixed Summon Demon, Summon Undead, Summon Mana Viper, Summon Emperor
Scorpions, and Summon Death Scarabs in some combination. Plain Summon
Demon doesn't do much for RMs, Vipers are also quite fragile at the time
Royal Mummies appear and vulnerable to torment, and Mummy Priests
already summon undead. Scorpions are vulnerable to torment, so they're
not the best here, but the flavour is good. Perhaps we could give
emperor scorps rTorm. This is the old Book IV, which has the strongest
Mummy flavour.

New Royal Mummy book:

- Smite
- Symbol of Torment
- Summon Emperor Scorpions
- Summon Death Scarabs

Ogre Mage
---------

Ogre Mages typically come in packs with other Ogres and are notable for
hasting them. Beyond that they have a sort of wizard flavour to their
book set; this consolodation picks Bolt of Magma and Parlayse to round
out the set to differentiate them from Wizards which appear at
overlapping depth.

New Ogre Mage book:

- Bolt of Magma
- Haste Other
- Paralyse
- Invisibility

Tengu Conjurer
--------------

The different Tengu spellcasters have a hodgepodge of "battlemage"
spells. Flavour wise Tengu are supposed to be good at Conjurations and
Air magic, so I settled on the air book from the four book choices.

New Tengu Conjurer book:

- Lightning Bolt
- Battlesphere
- Blink

Tengu Reaver
------------

In keeping with the air theme (though less so conjurations), the new
Tengu reaver book attempts to give them a pair of spells that force a
player trying to melee them to reposition. The monster AI change when it
comes to blocking ranged allies means that the reaver won't block its
lightning spire if it can step aside, so this should interact well. It
might need spell logic to prevent the reaver from summoning 10000 spires
since monsters don't have summon caps, let's see how it plays.

New Tengu Reaver book:

- Freezing Cloud
- Summon Lightning Spire

Liches
======

In merging spell books it was often easy to pick out the "iconic" spells and weave them together into a single book that still felt like the monster. For Liches, the list of "iconic" spells don't quite work together as a single book. Based on some of the input, I've gone with keeping Liches and Ancient lichee having the same book, a conjurations focussed book:

- Corrosive Bolt
- LCS
- Haste
- Slow
- Invisibility

To complement this, since as nikheizen pointed out there's plenty of space on the L glyph, this branch introduces a new monster, draft name "dread Lich" (sorry I didn't go for the more ambitious names suggested). Dread Liches have the same stats as Ancient Liches, and appear in the same places (except for a couple of vaults that were using ancient liches specifically one way or another), with the spell book:

- Haste
- Summon Greater Demon
- Paralyse
- Invisibility

I thought about possibly giving them Haunt or XXX, or some kind of damaging conjuration, but this spell set is rather tight and gets the point across well. The casting frequency total is still 60/200, perhaps they could move up to 80 or 100 since once they've cast their self-enchants there's not much to do. Another direction to take them, since ancient liches are pretty tanky, would be to buff their AF_DRAIN melee.

Deep Elf Mages
============

These monsters are split into two:

- deep elf pyromancers, who use the old fire book
- deep elf zephyrmancers, who use the old air book

With 6 books, reducing to a single monster was impossible. These two books were the closest to "iconic", based on some player conversations they're the most memorable, the ice and earth books get lost in the mix with knights and elementalists.


nikheizen:
> For Liches, the list of "iconic" spells don't quite work together as a single book:
> *snip*
> Because the demons get in the way of the conjurations. Giving both Liches and Ancient Liches the same book also seems like a lost opportunity. So this is another place where I'm looking for input.

I have a couple thoughts about this: I think Banishment, Petrify, and Haunt are another two fairly scary Lich spells and some I definitely associate with the monster type. The same is true to a lesser extend with Corrosive Bolt. I think Liches are in a good place as a hasting hexer that *doesn't* summon, but Ancient Lich definitely does want to summon. A possibility would be to split Ancient Liches up into two monsters (**L** doesn't have the space problems that **e** does), and have one focus on summoning with:

- Haunt
- Summon Greater Demon
- Haste
- Banishment
- Force Lance

This book emphasizes the summons by giving the Ancient Lich a distancing tool in their only Conjuration, but still letting them hit hard when they get into melee. I put Banishment here as I think it fits the theme of calling otherworldly beings into existence and the game just won't be the same without getting Banished to Abyss:4 in the middle of Zot:5. If greater demon summoning is too prevalent, Ancient Liches could be a good candidate for the fairly rare XXX monster summon.
And one focus on conjuration with:

- Corrosive Bolt
- Lehudib's Crystal Spear
- Haste
- Paralyze
- Invisibility

The spells in this package have a neat synergy, obviously paralysis prevents evasion, but invisibility makes it easier to hit the player and Corr. Bolt reduces AC for LCS and further Corr. Bolts.

At this point we could ask if we want to stratify Lich and Ancient Lich as separate Summoner/Conjurors. I'm not sold on the idea as I think players associate both of them with both of these things- I certainly associate Ancient Liches with both. But if that were the path, I think that giving Ancient Liches the first book and Liches the second book would be the better course of action. (In this case, cutting Corr. Bolt for OoD is probably correct).

As for regular Liches, I think a good package would be:

- Orb of Destruction
- Iron Shot
- Haste
- Petrify
- Invisibility

It's possible here that OoD could go up to the Ancient Lich Conjuration "book", giving regular Liches Corrosive Bolt and Iron Shot, or simply Corrosive bolt as their lone conjuration- weighted up so that they cast it fairly often.

**In summary**, my thoughts on Liches are that regular Liches fit the Conjuror mold much better than the summoning one. Greater Demon is definitely a spell I strongly associate with Ancient Liches and not so much regular ones (Ancient Liches stick around much longer for their summons to do good work). Ancient Liches definitely run into a problem of summons and conjurations not mixing too well, so either split their books into two separate monsters or make them terrifying summoners.

Possible names for splitting ancient Liches: Ancient/Timeless/Primeval/Deathless as the age descriptor, Eradicator/Disintegrator/Annihilator for the Conjurations descriptor, and Ritualist/Malefactor/Ba'al(it means Lord) for the Summonings descriptor. I like Ancient Eradicator and Primeval Malefactor myself.

I hope my rambling is of some help to you. I love this idea, as monster books have always been incredibly confusing and frustrating to deal with for me personally.

RojjaCebolla:
What if the first, 0-aut action for alert Deep Elf Mages in LoS was to "become" or reveal themselves as being a fire mage, smoke mage, wind mage, etc?  Then players don't have to x-v, the name and tile shows exactly which book the monster has.

PleasingFungus:
I'm planning to look this over this week - thanks for making it!

ebering:
> What if the first, 0-aut action for alert Deep Elf Mages in LoS was to "become" or reveal themselves as being a fire mage, smoke mage, wind mage, etc? Then players don't have to x-v, the name and tile shows exactly which book the monster has.

I don't like something like this; on-LoS effects tend to be buggy, and this would add a lot of complexity for something that's effectively still just creating multiple monsters.

crawl-ref/source/dat/descript/monsters.txt
```diff
@@ -900,21 +895,26 @@ deep elf knight
...
-wide array of harmful spells.
-%%%%
 deep elf master archer
 
 A deep elf archer who has devoted a lifetime to the mastery of archery,
 dismissing magic as a frivolous distraction.
 %%%%
+deep elf pyromancer
+
+A deep elf learned in the ways of firey magic. It is capable of casting a
```
PleasingFungus, line 905:
```suggestion
A deep elf learned in the ways of fiery magic. It is capable of casting a
```

crawl-ref/source/dat/descript/spells.txt
```diff
@@ -1449,6 +1449,10 @@ Summon Spectral Orcs spell
 Summons between one and three spectral orcs, specialised for melee combat
 and equipped with spectral weaponry and armour.
 %%%%
+Summon Tzitzimitl spell
+
+Summons a tzitzimitl; a demon of disease and despair.
```
PleasingFungus, line 1454:
```suggestion
Summons a tzitzimitl, a demon of disease and despair.
```

crawl-ref/source/mon-cast.cc
```diff
@@ -4676,7 +4675,8 @@ static const pop_entry _invitation_orc[] =
 
 static const pop_entry _invitation_elf[] =
 { // Elf enemies
-  {  1,   1,  100, FLAT, MONS_DEEP_ELF_MAGE },
+  {  1,   1,  50, FLAT, MONS_DEEP_ELF_AIR_MAGE },
+  {  1,   1,  50, FLAT, MONS_DEEP_ELF_FIRE_MAGE },
```
PleasingFungus, line 4679:
```suggestion
  {  1,   1,   50, FLAT, MONS_DEEP_ELF_AIR_MAGE },
  {  1,   1,   50, FLAT, MONS_DEEP_ELF_FIRE_MAGE },
```

crawl-ref/source/mon-cast.cc
```diff
@@ -7636,6 +7636,15 @@ static ai_action::goodness _monster_spell_goodness(monster* mon, mon_spell_slot
...
             return ai_action::good_or_bad(_trace_los(mon, _elec_vulnerable));
 
+    case SPELL_CHAIN_OF_CHAOS:
+        if (you.visible_to(mon) && friendly)
+            return ai_action::bad(); // don't zap player
+        else
+        {
+            return ai_action::good_or_bad(_trace_los(mon, [] (const actor * a)
+                                                          { return true; }));
+        }
```
PleasingFungus, line 7646:
```suggestion
        return ai_action::good_or_bad(_trace_los(mon, [] (const actor * a)
                                                          { return true; }));
```

crawl-ref/source/mon-util.cc
```diff
@@ -2602,156 +2602,71 @@ monster_type random_demonspawn_job()
...
 
-vector<mon_spellbook_type> get_spellbooks(const monster_info &mon)
+mon_spellbook_type get_spellbook(const monster_info &mon)
 {
     // special case for vault monsters: if they have a custom book,
     // treat it as MST_GHOST
     if (mon.props.exists(CUSTOM_SPELLS_KEY))
-        return { MST_GHOST };
+        return MST_GHOST;
     else
```
PleasingFungus, line 2616: Delete 'else'.

PleasingFungus (approved):
Nice cleanup! Very wide-reaching change - excited for dread liches to start killing players :)

PR #1698, opened 2021-01-25
Remove Evocation scaling from Spectral

PleasingFungus:
Currently, the spectral brand is the most complex weapon ego in the game.
It has two significant drawbacks:

1. It can significantly increase the damage the wielder takes, since the
spectral weapon shares damage with the player.
2. It requires Evocations training to use.

Both due to the overall complexity and the perceived weakness of this
brand, I'm inclined to simplify by removing one of these effects. Pain
sharing makes the brand a bit harder to evaluate, since most brands have
no downside at all vs an unbranded weapon (chaos being the notable exception).
Evocations scaling has several other problems, though:

- It confuses players on a regular basis. People miss the subtle messages about
  'ineptitude' and fail to understand why the spectral weapon isn't triggering -
  'is it broken?' Some of them ask other players and get the confusion sorted out
  but others, presumably, just continue thinking it's bugged!
- It's the only effect where Evocations training gives benefit to a passive effect,
  rather than an evocable item or ability. This isn't the end of the world, but it
  is an odd exception.
- Evocations skill is currently very useful for most characters even without this
  - most games present a large number of powerful, Evocations-scaling items that
  strongly incentivize Evocations training. It's somewhere between unuseful and
  counterproductive to add one more.

It's possible that the damage sharing effect should also go away or be reduced
in magnitude - the downsides of "need to mess with positioning" and "don't have
another brand" might be enough to balance spectral. We'll see!

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -1351,10 +1351,8 @@ static string _describe_weapon(const item_def &item, bool verbose)
                     "affected.";
             break;
         case SPWPN_SPECTRAL:
-            description += "It retains the spirit of the tree from which "
-                           "it was made. In the hands of one skilled in "
-                           "evocations this spirit is drawn out to fight "
-                           "along side the wielder.";
+            description += "When it strikes, its spirit leaps out and "
+                           " fights alongside the wielder.";
             break;
         case SPWPN_NORMAL:
             ASSERT(enchanted);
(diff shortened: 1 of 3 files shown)


ebering:
The Evo scaling is an artefact of the charms re-design thought process, which envisioned charm effects being put on items and those that needed scaling scaling with evo. In the final work, the only thing left with scaling was spectral weapon.

I will note that magic staff melee also scales with evo—perhaps we should change that to Spellcasting. But that doesn't affect this change, which I agree with so I'm merging.

PleasingFungus:
Fair point about magic staff meleé! I'd forgotten about that.

PR #1706, opened 2021-01-27
Prevent Ozocubu's Armour from breaking when equipping heavier armour

AdamPG:


ebering:
Thanks for the patch; though icy armour should break on any body armour change now, since the granted AC is dependent on the body armour ER at the time of casting.

AdamPG:
> Thanks for the patch; though icy armour should break on any body armour change now, since the granted AC is dependent on the body armour ER at the time of casting.

(1) Icy armour currently only breaks when equipping an armour that was too heavy for the former spell (it does not break when equipping light armours).

(2) The AC granted by Ozo’s correctly updates when switching armours after compiling this patch.

For these reasons, I think the patch is still appropriate.

ebering:
Oh, so it does; it only saves the spellpower at the time of casting.

ebering:
Thanks for the patch!

PR #1708, opened 2021-01-28
Ice armour bug fixes

AdamPG:
There were several inconsistencies in the types of movement that would end Ozocubu's armour and frozen ramparts and/or trigger barb damage: for example, palentonga's rolling charge and Wu Jian's wall jump would trigger these effects, but barachian's hop and Ru's power leap would not. Furthermore, the effect of frozen ramparts is tied to the player's position at the time of casting, making it intuitive for any change of position to end the effect. This PR resolves these issues by making Ozocubu's armour and frozen ramparts end whenever the player's position changes (providing a prompt when the player attempts to change their position voluntarily), and by making barb damage trigger whenever the player's position changes in a way that thematically requires physical movement.

### DIFF
--- a/crawl-ref/source/stairs.cc
+++ b/crawl-ref/source/stairs.cc
@@ -653,9 +653,7 @@ void floor_transition(dungeon_feature_type how,
 
     // We "stepped".
     if (!forced)
-    {
         apply_barbs_damage();
-    }
 
     // Magical level changes (which currently only exist "downwards") need this.
     clear_trapping_net();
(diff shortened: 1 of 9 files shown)


crawl-ref/source/stairs.cc
```diff
@@ -655,7 +655,6 @@ void floor_transition(dungeon_feature_type how,
     if (!forced)
     {
         apply_barbs_damage();
```
wheals, line 657: These brackets should be removed to fix the style code linting (you can do that automatically by running `util/unbrace`)

AdamPG:
I considered removing the prompts for entering/leaving the Abyss, but I decided to air on the side of being overzealous.

PR #1711, opened 2021-01-29
Simplify Zin's recite power calculation formula

sdynet:
As with the power of many gods, I hope Recite's power calculation method will be simple and clear.


ebering:
Per conversation on discord: A change like this needs more elaboration than just "simplify". The formula isn't displayed directly to the player. The description doesn't mention this, but I've just fixed that.

PR #1712, opened 2021-01-29
Add option to configure rest/wait duration

Goratrix:
Currently the time to wait (pressing '5') while at full
HP and MP is fixed at 100 turns, this makes it configurable.


Goratrix:
Hm, I just realized even though this works fine, it's probably incorrect, because RMODE_WAIT needs to be sent to runrest::initialise as parameter, so it shouldn't be simply replaced by the turns value...

PR #1714, opened 2021-01-31
Don't try to spawn spiny frogs in arena sprint

joshgelbard:
Saw an error (that I didn't write down) in someone else's game. Seemed to block progress.


ebering:
Thanks for the patch!

PR #1715, opened 2021-01-31
Stepdown removal

hellmonk3:
Removes two opaque stepdowns on evasion. I think knowledge of the dex stepdown has more or less filtered out to the playerbase, but the EV one is not well-known. The practical upshot is that players with more than 18 dex get a bit more EV if they have some dodging skill, and players with more than 30 "base" EV and a lot of dex can get significantly more EV. For most characters, this shouldn't be a huge deal: 25 dex robe human gets 3 more ev at 20 dodging compared to before, 7 if they take dodging all the way to 27. However, it does significantly buff characters that can get a very large amount of dexterity since dex and dodging skill are multiplicative. I went ahead and nerfed spriggans slightly since they're the most likely species to get out of control, but otherwise I've left everything alone. It's probably worth monitoring how big of a buff players get in practice and then adjusting accordingly, but I think it shouldn't be too hard to compensate for and improving the clarity of the formulas seems worthwhile.

### DIFF
--- a/crawl-ref/source/dat/species/spriggan.yaml
+++ b/crawl-ref/source/dat/species/spriggan.yaml
@@ -18,7 +18,7 @@ aptitudes:
   slings: 2
   bows: 2
   armour: -3
-  dodging: 4
+  dodging: 3
   stealth: 5
   shields: -3
   unarmed_combat: -2
@@ -37,7 +37,7 @@ aptitudes:
 size: little
 str: 4
 int: 9
-dex: 11
+dex: 10
 levelup_stat_frequency: 5
 levelup_stats:
   - int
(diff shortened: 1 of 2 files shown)


crawl-ref/source/dat/species/spriggan.yaml
```diff
@@ -37,7 +37,7 @@ aptitudes:
 size: little
 str: 4
 int: 9
-dex: 11
+dex: 10
```
RojjaCebolla, line 40: total is now not 24; bump int?

PleasingFungus: Why should the total be 24?

RojjaCebolla: i thought most races were 24, based on the human 8 8 8

PleasingFungus: Seems possible, but not very important if so. Not like races are so similar or so perfectly balanced that equal stat totals are meaningful.

PR #1716, opened 2021-01-31
Migrate to actions/cache v2 (was v1)

alexjurkiewicz:
No breaking changes. Several minor improvements to performance. Will hopefully reduce incidence of spurious errors.


wheals:
Thanks! Looks like any performance improvements are within random variance, but I expect it can't hurt...

PR #1717, opened 2021-02-02
Unrandart powers

Aliscans:
1, Describe the special powers the staff of Olgreb, the storm bow and Maxwell's Thermic Engine have in art-data.txt rather than describe.cc.
2. Expand the special power descriptions for the staff of Olgreb and Maxwell's Thermic Engine, so that they cover all of the items' special powers.
3. Add descriptions for the special powers of the salamander hide armour and the macabre finger necklace.

The text from unrand.txt does cover the same sort of power, but (a) it also includes flavour text, making the item's properties less clear, and (b) it isn't included in a character dump, so that they list some of an artefact's properties, but not others.

I added some armour and jewellery as these are in a different part of the code. There are dozens of artefacts which have properties like this.

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -1168,6 +1168,7 @@ TILE:     urand_finger
 COLOUR:   ETC_BONE
 LIFE:     1
 VALUE:    400
+DESCRIP:  It enables you to equip an extra ring.
 
 # start TAG_MAJOR_VERSION == 34
 ENUM:     SPIDER
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -222,6 +222,7 @@ VALUE:   800
 
 NAME:    staff of Olgreb
 INSCRIP: rPois∞
+DESCRIP: It grants immunity to poison.
 OBJ:     OBJ_WEAPONS/WPN_STAFF
 PLUS:    +9
 COLOUR:  ETC_POISON
@@ -451,6 +452,9 @@ TILE_EQ:  glaive_of_the_guard
 BRAND:    SPWPN_ELECTROCUTION
 AC:       5
 INSCRIP:  spectral,
+DESCRIP:  It retains the spirit of the tree from which it was made. In the hands
+ of one skilled in evocations this spirit is drawn out to fight along side the
+ wielder.
 
 ENUM:     ZEALOT_SWORD
 NAME:     zealot's sword
@@ -691,6 +695,8 @@ TILE:     urand_storm_bow
 TILE_EQ:  storm_bow
 BRAND:    SPWPN_ELECTROCUTION
 INSCRIP:  penet
+DESCRIP:  Ammo fired by it passes through the targets it hits, potentially
+ hitting all targets in its path until it reaches maximum range.
 
 NAME:    tower shield of Ignorance
 OBJ:     OBJ_ARMOUR/ARM_TOWER_SHIELD
@@ -1373,6 +1379,9 @@ ENUM:     THERMIC_ENGINE
 NAME:     Maxwell's thermic engine
 OBJ:      OBJ_WEAPONS/WPN_DOUBLE_SWORD
 INSCRIP:  freeze,
+DESCRIP:  It has been specially enchanted to freeze those struck by it, causing
+ extra injury to most foes and up to half again as much damage against
+ particularly susceptible opponents.
 PLUS:     +2
 COLOUR:   ETC_VEHUMET
 TILE:     urand_thermic_engine
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -222,7 +222,11 @@ VALUE:   800
 
 NAME:    staff of Olgreb
 INSCRIP: rPois∞
-DESCRIP: It grants immunity to poison.
+DESCRIP: It enhances your poison magic.\nIt
+ grants immunity to poison.\nIt
+ can be evoked to poison those around you for a time.\nIt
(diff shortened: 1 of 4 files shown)


crawl-ref/source/art-data.txt
```diff
@@ -451,6 +456,9 @@ TILE_EQ:  glaive_of_the_guard
 BRAND:    SPWPN_ELECTROCUTION
 AC:       5
 INSCRIP:  spectral,
+DESCRIP:  It retains the spirit of the tree from which it was made. In the hands
+ of one skilled in evocations this spirit is drawn out to fight along side the
+ wielder.
```
ebering, line 461: The description of the spectral has changed, this should match the spectral brand description.

crawl-ref/source/describe.cc
```diff
@@ -642,6 +642,34 @@ static string _randart_descrip(const item_def &item)
...
 #undef known_proprt
 
+// If item is an unrandart with a DESCRIP field, return its contents.
+// Otherwise, return "".
+static string _unrandart_descrip(const item_def &item)
+{
+    if (!is_unrandom_artefact(item)) return "";
+    auto entry = get_unrand_entry(item.unrand_idx);
+    if (!entry->descrip) return "";
+    return string("\n")+entry->descrip;
```
ebering, line 652: I'm concerned that you add a newline here and also at lines 1410, 1853, and 1960. (I haven't downloaded and built this for testing yet). The function `_artefact_descrip` shouldn't handle any newline padding before or after the description text, this should be left to the callers.

ebering (changes requested):
Thank you for this submission, it's a nice re-factor of an increasingly unwieldy part of the description code.

There's a small tweak with the padding that I've commented on.

I would also prefer that instead of using "\n" in art-data.txt, the format use multiple`DESCRIP:` entries and put in the newlines in art-data.pl's generation of art-data. For example change:

```
DESCRIP: It enhances your poison magic.\nIt
  grants immunity to poison.\nIt
  can be evoked to poison those around you for a time.\nIt
  may poison the flesh of those it strikes, and is more effective with a high
  Evocations skill.
```

to

```
DESCRIP: It enhances your poison magic.
DESCRIP: It grants immunity to poison.
DESCRP: It can be evoked to poison those around you for a time.
DESCRIP: It may poison the flesh of those it strikes, and is more effective with a high
  Evocations skill.
```

crawl-ref/source/art-data.txt
Aliscans: I've changed this now.

crawl-ref/source/describe.cc
Aliscans: _artefact_descrip() doesn't add any padding. There is a newline, but that's to ensure that the text appears at the start of a line. You get this throughout get_item_description().

1410, 1853 and 1960 do add a blank line, but this was already there.

Aliscans: The script doesn't let you repeat lines in an artefact definition at the moment. Rather than change this, would something like this be acceptable?

DESCRIP: It enhances your poison magic.
+ It grants immunity to poison.
+ It can be evoked to poison those around you for a time.
+ It may poison the flesh of those it strikes, and is more effective with a high
 Evocations skill.

i.e. make + continue the previous field, but with a \n at the start? Thanks.

ebering: > _artefact_descrip() doesn't add any padding. There is a newline, but that's to ensure that the text appears at the start of a line. You get this throughout get_item_description().

Right; but `_unrand_descrip()` (the line that I'm commenting on) adds this leading newline, while `_randart_descrip()` does not. The two functions should be consistent, and `_artefact_descrip()` should consistently return a string that starts with the first letter of the descrip text, not a newline.

ebering:
> The script doesn't let you repeat lines in an artefact definition at the moment. Rather than change this, would something like this be acceptable?
> 
> ...
> 
> i.e. make + continue the previous field, but with a \n at the start? Thanks.

Github displays messages as markdown in the web ui; if you're interacting via e-mail it will also read your e-mail parse the markdown and send html e-mails to the people subscribed to the pull request. I think this messed up what you were suggesting. 

My interpretation of your last sentence plus the mangled message is that you're suggesting the format:

```
DESCRIP: It enhances your poison magic.
+ It grants immunity to poison.
+ It can be evoked to poison those around you for a time.
+ It may poison the flesh of those it strikes, and is more effective with a high
   Evocations skill.
```

If this is what you were suggesting, then yes, please do this.

crawl-ref/source/describe.cc
Aliscans: I'm a bit confused. _randart_descrip() either returns "", or returns a string which starts with a \n. The parts of the code which add text to the string it creates are:

```
const char* type = _jewellery_base_ability_description(item.sub_type);
description += "\n";
description += type;
```
...
```
sdesc = (string based on something in propdescs[])
description += '\n';
description += sdesc;

```
...

```
snprintf(buf, sizeof buf, "\nIt makes you %s%s stealthy.", PARAMS)
description += buf;

```
_artefact_descrip() can return several lines of text, which would need the correct number of \n characters between them. While it's possible for the function to strip the initial \n before it returns the string, the caller would only have to add it back in.

I've made a few more changes to artefact_descrip() now (mainly to eliminate the need for \n in art-desc.txt), but it still returns a string which starts with \n.

Aliscans:
I've just had a look at what's in master at the moment, and I don't tthink the spacing is quite right (going by the rules that there should be a blank line before and after the list of powers, and a blank line before and after each brand).

The macabre finger necklace should have a blank line before the DESCRIP field.
The glaive of the Guard should have a blank line after the DBRAND field.

I don't know how you'd like to do this.

PR #1718, opened 2021-02-02
Stop art-data.pl from rewriting files which haven't changed.

Aliscans:
Stop art-data.pl from rewriting files which haven't changed.

Before writing to an output file, art-data.pl now reads the file, and only writes it if the new version is different.

This should reduce the number of files make rebuilds when art-data.txt is changed. In particular, many files depend on art-enum.h, but this only stores the names and ID numbers of artefacts, so need only be rewritten when they change.


ebering:
Thank you for the patch!

PR #1721, opened 2021-02-04
Remove 3 compiler warnings

Aliscans:
This PR removes compiler warnings I get with gcc 10.2.0 when compiling Crawl (in l-file.cc, quiver.cc and view.cc). I don't think it has any other effect.

### DIFF
--- a/crawl-ref/source/l-file.cc
+++ b/crawl-ref/source/l-file.cc
@@ -90,7 +90,7 @@ enum class lua_persist
 {
     none,
     number,
-    string,
+    text,
     // [ds] No longer supported for save portability:
     function,
     nil,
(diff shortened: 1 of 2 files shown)


crawl-ref/source/viewgeom.h
```diff
@@ -22,6 +22,7 @@ class crawl_view_buffer
 public:
     crawl_view_buffer();
     crawl_view_buffer(const coord_def &sz);
+    crawl_view_buffer(crawl_view_buffer const&) = default;
```
randomMesh, line 25: I don't like the placement of `const` here. The line above has it the other way round. Be consistent.

wheals:
I ended up doing a different fix for the third commit. Please rebase to remove that from the history and then I'll merge the first two.

wheals:
Thanks! Are you credited in CREDITS.txt yet?

Aliscans:
I am already, thanks.

PR #1724, opened 2021-02-05
Attraction by nearest

christopherlandry:
request by PF

### DIFF
--- a/crawl-ref/source/directn.h
+++ b/crawl-ref/source/directn.h
@@ -106,15 +106,6 @@ struct direction_chooser_args
 
 };
 
-struct dist_sorter
-{
-    coord_def pos;
-    bool operator()(const actor* a, const actor* b)
-    {
-        return a->pos().distance_from(pos) > b->pos().distance_from(pos);
-    }
-};
-
 class direction_chooser;
 
 class direction_chooser_renderer : public view_renderer
--- a/crawl-ref/source/directn.h
+++ b/crawl-ref/source/directn.h
@@ -106,6 +106,15 @@ struct direction_chooser_args
 
 };
 
+struct dist_sorter
+{
+    coord_def pos;
+    bool operator()(const actor* a, const actor* b)
+    {
+        return a->pos().distance_from(pos) > b->pos().distance_from(pos);
+    }
+};
+
 class direction_chooser;
 
 class direction_chooser_renderer : public view_renderer
--- a/crawl-ref/source/spl-transloc.cc
+++ b/crawl-ref/source/spl-transloc.cc
@@ -1472,9 +1472,7 @@ void attract_monsters()
     //get list of near monsters
     vector<monster *> targets;
     for (monster_near_iterator mi(you.pos(), LOS_NO_TRANS); mi; ++mi)
-    {
         targets.push_back(*mi);
-    }
     //sort them in order of distance to you
     dist_sorter sorter = {you.pos()};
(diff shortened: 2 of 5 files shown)


crawl-ref/source/directn.h
```diff
@@ -106,6 +106,15 @@ struct direction_chooser_args
...
 
+struct dist_sorter
+{
+    coord_def pos;
+    bool operator()(const actor* a, const actor* b)
+    {
+        return a->pos().distance_from(pos) > b->pos().distance_from(pos);
+    }
+};
+
```
ebering, line 117: directn.* is primarily for the direction_chooser and associated components, this struct belongs in spl-util.h since it is (presently) only used by spell-like things.

crawl-ref/source/directn.h
```diff
@@ -106,6 +106,15 @@ struct direction_chooser_args
 
 };
 
+struct dist_sorter
+{
+    coord_def pos;
+    bool operator()(const actor* a, const actor* b)
+    {
+        return a->pos().distance_from(pos) > b->pos().distance_from(pos);
```
ebering, line 114: Note the comparison `>`: this sorter sorts actors from farthest to nearest order. If you want to re-factor and generalize to also use this to have potions of attraction sort by nearest, you should do one of two things:

-  add a boolean field to the struct and make the comparison be nearest-first or farthest first
- create two structs, one for each behavior

Regardless of which one is chosen, also add a comment to explain the two fields and the structs purpose.

crawl-ref/source/spl-transloc.cc
```diff
@@ -1469,9 +1469,17 @@ static bool _can_move_mons_to(const monster &mons, coord_def pos)
  */
 void attract_monsters()
 {
+    //get list of near monsters
+    vector<monster *> targets;
     for (monster_near_iterator mi(you.pos(), LOS_NO_TRANS); mi; ++mi)
-    {
-        if (!_can_beckon(**mi) || mi->friendly())
+        targets.push_back(*mi);
```
ebering, line 1475:
```suggestion
        if (!_can_beckon(**mi) || mi->friendly())
            targets.push_back(*mi);
```

Filter valid targets now, before you sort.

crawl-ref/source/spl-transloc.cc
```diff
@@ -1469,9 +1469,17 @@ static bool _can_move_mons_to(const monster &mons, coord_def pos)
...
+    //sort them in order of distance to you
+    dist_sorter sorter = {you.pos()};
+    sort(targets.begin(), targets.end(), sorter);
+
+    //now operate on the sorted list
+    for (monster *mi : targets) {
+        if (!_can_beckon(*mi) || mi->friendly())
             continue;
```
ebering, line 1483:
```suggestion
```

ebering (changes requested):
Thanks for the submission!

The idea to move dist_sorter to a single place (instead of repeating the code) is good, but it should go in spl-util.h.

Also, beware! The dist_sorter that you're moving sorts actors from farthest to nearest, and for this you want nearest to farthest. I've commented in more detail about exactly where the changes need to happen to address this.

ebering (approved):
Looks good! I'm merging this manually since I changed my mind (the distance sorters should go in spl-transloc.h) and have a few other tweaks.

Are you familiar with `git rebase`? In the future if you keep sending patches it would be helpful (though not required) to learn how to use it to edit your commit history—in this case squash all four commits into a single one implementing the change.

PR #1725, opened 2021-02-05
Vampiric Draining checks susceptibility in targeter and quiver

Ebonnov:
While aborting Vampiric Draining on message "You can't drain life from that!" is not too bad (doesn't cost mana or time), it can feel knowledge-heavy/spoilery when it's not quite clear what monsters are not susceptible (e.g. gargoyles), and misleading when there's no susceptible targets in range while the quiver shows enabled.

For the quiver side change, I can see a few other approaches and not sure if they make it cleaner:
- turn vampiric draining into zap and beam so that the quiver check can use the tracer beam.
- refactor `get_dist_to_nearest_monster()` to reuse the logic to check if a monster makes sense to target (in itself as used to get `minRange`, in cloud spell treatment, in this PR)


ebering:

> turn vampiric draining into zap and beam so that the quiver check can use the tracer beam.

This is definitely the preferred solution. Vampiric Draining goes back to pre-stone-soup, which probably explains some of its implementation. If you did this it would also supersede 6fba9c5.

From a perspective of timing and merging this PR: would you be able to update the PR with the conversion of Vamp Drain to use a zap? If so I'll hold off on merging this.

Ebonnov:
> would you be able to update the PR with the conversion of Vamp Drain to use a zap?

Let me take a couple days and have a stab this direction. I wonder if the damage formula has to be touched a bit.

Ebonnov:
Until a few days in this I learned that malign offering is this close to vampiric draining, which kind of settles if the new beam should be enchantment or not.

At this point I'm mainly feeling uncomfortable about message flavor (the worst being monster casting on rN+++ player), and that autofight can fizzle.

Ebonnov:
Do you think this is in a merge-able state?

ebering:
Sorry, I haven't had a chance to review the updated PR yet—it hasn't fallen off my radar! 

In your comment you say that auto fire can fizzle, which would be a showstopper; perhaps @rawlins can take a quick look and see if there's an easy fix.

Ebonnov:
> it hasn't fallen off my radar!

That's reassuring to hear, thanks!

Having autofight pick from susceptible targets was a goal I wanted to have before trying zap-ify the spell, but it does seem to me that the target selection happens in lua autofight code and that goal likely needs a yet larger scope (coincidentally, casting quivered Pain seems to suffer from same target selection issue and can fizzle). Monster info is available there, but the nature of the action isn't, so it's not super clear what's the best way forward there.

ebering:
Merged this with some tweaks, thanks!

Big ones: I didn't like the gameplay checks split, and went with the beam only affecting susceptible to vampirism targets. To make that easier to understand (and remove the quasi-guessing in the desc string) I added a monster_info bit to indicate can't drain.

PR #1728, opened 2021-02-07
Deminogod "rework"

hellmonk3:
No, not the random god abilities on levelup one. This is a small commit that tries to lean into the current demigod gimmick of having very big, player-directed stats. Demigod starting stats are reduced by 2 each in exchange for a buff to their stat choice - instead of adding two points to a single stat every third level, they add four points. This should flatten out their difficulty curve compared to other species a little bit and, I hope, make them feel slightly more interesting. Total attributes break even with old demigod at xl 9, which seems about right - non-demigod characters will usually have divine help by that point. 

I also raised their xp apt to -1. I don't feel particularly strongly about this, but the -2 xp apt tended to "feel bad" despite not really being much worse than -1 past like, D:4, and I don't think it was central to their gimmick.

### DIFF
--- a/crawl-ref/source/dat/species/demigod.yaml
+++ b/crawl-ref/source/dat/species/demigod.yaml
@@ -6,7 +6,7 @@ adjective: "Divine"
 difficulty: Advanced
 difficulty_priority: 80
 aptitudes:
-  xp: -2
+  xp: -1
   hp: 1
   mp_mod: 2
   mr: 4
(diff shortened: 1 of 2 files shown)


sdynet:
I need that.

mainiacjoe:
I think rMut would be a good fit for Dg.

sdynet:
> 
> 
> I think rMut would be a good fit for Dg.

Oh... I agree with that opinion, but I'm worried that it's similar to Ha.

mainiacjoe:
Sturdy Frame for Ha.

christopherlandry:
Removal for Ha

ebering:
Please try to keep PR discussion focused to the PR at hand and use crawl-dev irc or another forum for more freewheeling discussions.

PleasingFungus:
I've merged the attribute change parts of this PR as 81277dbfc4c85b2ed19224276b53eb6cfeb2f090 - thanks for suggesting it!

As you said, the xp apt change was less essential, and others (e.g. kate-) thought it wasn't necessary, so i've cut it for now. We can revisit that separately.

PR #1732, opened 2021-02-09
Remove some water from mumra_spider_spiderweb

nlavsky:
The vault has several spiders trapped behind deep water, as they can
no longer cling to walls.

This commit frees the spiders and reduces the number of water and deep
water tiles in the vault. Also, it removes an escape hatch in the
inner area, since the vault is not a teleport closet anymore.

-----

A code archaeology fact of the day: when monster clinging was removed in 0.13, there was exactly the same problem with spiders trapped behind deep water. `mumra_spider_spiderweb` and other Spider vaults were fixed in bdc56382, but the changes were reverted in 40eb2104 when clinging was brought back.


PleasingFungus:
Ooh, good archaeology!

PR #1747, opened 2021-02-15
Update the des syntax doc

nlavsky:
This commit updates examples and removes mentions of obsolete/deprecated
item types, glyphs, and modifiers.

### DIFF
--- a/crawl-ref/docs/develop/levels/syntax.txt
+++ b/crawl-ref/docs/develop/levels/syntax.txt
@@ -65,18 +65,13 @@ Features
  >< - escape hatches - you can leave the level by these but will usually
       not land on stairs/hatches
 
- ^ - random trap. Deprecated.
- ~ - random trap suitable for the branch and depth the map is being used.
-     Deprecated.
-
  A - Stone arch.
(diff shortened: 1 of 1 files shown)


crawl-ref/docs/develop/levels/syntax.txt
```diff
@@ -496,18 +491,17 @@ ITEM:     (list of items, separated by comma)
...
 
           Modifiers:
           * "q:N" sets the item quantity to N (if N > 0). Does nothing
              if the item is not stackable.
           * "charges:N" sets the number of charges for wands.
-          * "plus:N" and "plus2:N" set the item pluses for weapons, armour,
-            jewellery (such as rings of intelligence or slaying), and
-            throwing nets.
+          * "plus:N" sets the item plus for weapons, armour, and jewellery
+            (such as rings of intelligence or slaying).
```
ebering, line 504: Jewellery has fixed pluses now, but maybe vaults can still override that? Haven't checked carefully.

nlavsky: It's still possible to place custom jewellery in a vault. Stat rings and rings of evasion/protection work fine with ``plus:100``. Also, meatsprint has a lot of custom rings of slaying.

PR #1748, opened 2021-02-15
Delete some unnecessary things from travel_pathfind.

Aliscans:
Make travel_pathfind::points a local variable in pathfind().

Move _find_explore_status() into travel_pathfind.

Delete set_annotate_map(), set_distance_grid(), greedy_square(), travel_move(), unexplored_square() and get_unreachables(). The functions which had referred to them (if any) now refer to the underlying variables.

### DIFF
--- a/crawl-ref/source/travel.cc
+++ b/crawl-ref/source/travel.cc
@@ -1115,7 +1115,7 @@ travel_pathfind::travel_pathfind()
       need_for_greed(false), autopickup(false),
       unexplored_place(), greedy_place(), unexplored_dist(0), greedy_dist(0),
       refdist(nullptr), reseed_points(), features(nullptr), unreachables(),
-      point_distance(travel_point_distance), points(0), next_iter_points(0),
+      point_distance(travel_point_distance), next_iter_points(0),
       traveled_distance(0), circ_index(0)
 {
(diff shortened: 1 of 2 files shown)


ebering:
It's possible that this history isn't available, but could you use `git blame` and shed some insight on the original design of these functions? If they exist to support a hypothetical functionality that sounds like a good idea I don't want to rip them out.

If the history isn't available then this is ok to merge.

Aliscans:
get_unreachables() was created in 2011 in 8d595aab95. The commit says "it could be possible to propose seeing [unreachable places] in map mode to the player". is_explore_horizon() does something similar, so I don't think this will ever be used.

The other functions are as old as travel_pathfind (commit a0d48b0186 from 2007). The commit message doesn't say anything about them, but they all seem to have been used in much the same way they are used now: set_annotate_map() and set_distance_grid() being unused; travel_move() used by pathfind(); and greedy_square() and unexplored_square() by find_explore_status().

These are all simple functions. If someone has an idea that requires (say) a non-default point_distance array, setting the variable will only be a tiny part of the overall change.

I hope this helps.

ebering:
Thanks for the archaeology and the clean-up patches!

PR #1749, opened 2021-02-15
Split find_travel_pos().

Aliscans:
Split find_travel_pos() into two functions (one with move_x and move_y, one with features).

Remove some unnecessary code from the latter version.


ebering:
Thanks for this! In the future if your refactor is cleaning up a long-standing XXX or TODO it helps with reviewing to mention that in the commit message.

Aliscans:
Thanks; I'll keep it in mind.

PR #1751, opened 2021-02-16
Simplify the Elixir

sdynet:
The complexity of card effects is terribles enough to drown beginners in the sea of information. Even if it takes a long time, we have to solve this complicated thread step by step.
This is the first button at 0.27.

### DIFF
--- a/crawl-ref/source/decks.cc
+++ b/crawl-ref/source/decks.cc
@@ -1173,16 +1173,12 @@ static void _elixir_card(int power)
     switch (power_level)
     {
     case 0:
-        if (coinflip())
-            you.set_duration(DUR_ELIXIR_HEALTH, 1 + random2(3));
-        else
-            you.set_duration(DUR_ELIXIR_MAGIC, 3 + random2(5));
+        you.set_duration(DUR_ELIXIR_HEALTH, 1 + random2(3));
+        you.set_duration(DUR_ELIXIR_MAGIC, 1 + random2(3));
         break;
     case 1:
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Silly question: if Elixir now always regenerates both HP and MP, why not merge the statuses together?

Also, don't we need to update `dat/descript/cards.txt`? It currently reads

```
the Elixir card

The card depicts a smug smile formed out of nine chalices.

Drawing this card rapidly restores the user's health or magic, or both at high power
levels. It will also heal their allies.
```

sdynet:
> Silly question: if Elixir now always regenerates both HP and MP, why not merge the statuses together?
> 
> Also, don't we need to update `dat/descript/cards.txt`? It currently reads
> 
> ```
> the Elixir card
> 
> The card depicts a smug smile formed out of nine chalices.
> 
> Drawing this card rapidly restores the user's health or magic, or both at high power
> levels. It will also heal their allies.
> ```

oops. that's right.

ebering:
The biggest obstacle to card clarity is the `power_level` parameter, an artefact of the Nemelex revision I did ( #916 ) where I tried to match the old power system of the card decks. Until someone finds the energy to pick over the whole set of cards (I know you're familiar with [this old thread](https://crawl.develz.org/tavern/viewtopic.php?f=8&t=25938) ) it's going to be tough going.

This change does a good job of smoothing over the `power_level` but if you continue your simplification project it'll eventually reach a point where removing this hidden `power_level` is necessary.

PR #1752, opened 2021-02-16
Don't assault firewood

nlavsky:
Also, fix an overlap between ``wont_attack()`` and ``neutral()`` in a condition.

The former means ``ATT_FRIENDLY || ATT_GOOD_NEUTRAL || ATT_STRICT_NEUTRAL``,
the latter means ``ATT_NEUTRAL || ATT_GOOD_NEUTRAL || ATT_STRICT_NEUTRAL``.


PleasingFungus:
Good call on both points, thanks!

PR #1753, opened 2021-02-16
Replace running with rampaging in wizlabs

shummie:
Two vaults - Cloud Mages' Chambers and Roulette of Golubria have boots of running as possible rewards. With the removal of boots of running, these were missed. Replace these with boots of rampaging since they still fit the intended theme/reward of the wizlab.


PleasingFungus:
Nice catch, thanks!

PR #1756, opened 2021-02-18
Describe Ozocubu's Refrigeration and Drain Life in monster

nlavsky:
The spells weren't handled by ``mons_spell_beam()``, so the monster
utility wasn't showing their damage.

Unlike monster, xv can already show the damage for these spells, since
it gets the numbers in a different way.


PleasingFungus:
Merged as 905266d83c, thanks! I tweaked the implementation a little bit to use the nicer struct that i added in 2016 or w/e and then forgot about.

PR #1758, opened 2021-02-18
Fix a couple of compiler warnings from GCC.

Aliscans:
Make _monster_currently_description() static.
Make SpellLibraryMenu::entry_colour() return int.

The second is a bit weird, as it doesn't change the thing GCC warns about. The warning goes away when both alternatives are signed values, but it seems a bit silly when either alternative will fit in an int.


ebering:
Thanks for pointing this out; the non-static function was staticified by 479c5223a71 and I closed the entry_colour warning a different way.

Aliscans:
Thanks.

PR #1762, opened 2021-02-20
Remove verbage for extra airstrike damage

shummie:
8719b9d removed Airstrike bonus damage, but left in the descriptor for the monster Airstrike spell. This should reword the phrase: "The air twists around and violently strikes you in flight!" to "The air twists around and strikes you!"


semitonal:
Thanks for the catch! This would need to also remove the corresponding %s entries from the mprf to work properly - have committed a fixed version.

PR #1764, opened 2021-02-20
update resistances when both rings worn, add messages

elliottbernstein:
When ring of ice and ring of fire are worn at the same time, the negative effects cancel out, and the result is rF+ rC+. Additionally, the ring that is already equipped has a small reaction.

### DIFF
--- a/crawl-ref/source/player-equip.cc
+++ b/crawl-ref/source/player-equip.cc
@@ -1171,10 +1171,14 @@ static void _equip_jewellery_effect(item_def &item, bool unmeld,
     {
     case RING_FIRE:
         mpr("You feel more attuned to fire.");
+        if (you.wearing(EQ_RINGS, RING_ICE) > 0)
+            mpr("Several drops of water fall from your ring of ice.");
         break;
 
     case RING_ICE:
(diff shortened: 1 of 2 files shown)


ebering:
Hello! and welcome to contributing to crawl. For gameplay changes like this, we like to see some explanation of why the change is good gameplay. That explanation doesn't have to happen in the pull request (there are lots of places where crawl gets discussed, the official dev irc channel  ##crawl-dev on the freenode irc network being the best place to get ahold of the development team), but it does need to happen somewhere.

So drop us a line or swing by!

My take on this change is that it's the kind of special case we try to avoid in crawl. At present, a player's net resistance to an element is the count of all the plusses they have (from equipment, mutations, innate traits, the potion, and forms) minus the count of all the minuses. With your change this now gains a special case, and it's not clear to me there's any gameplay benefit to the added complexity other than "it's cool flavour"—a justification that alone isn't enough.

ebering:
Since you (the submitter) and also no other devteam members have chimed in I'm going to close this.

If you've got a compelling gameplay reason to add this complication, feel free to re-open and let me know!

PR #1765, opened 2021-02-20
Inner flame message fix

christopherlandry:
@PleasingFungus fix inner flame death message


crawl-ref/source/beam.cc
```diff
@@ -2713,7 +2713,7 @@ void bolt::internal_ouch(int dam)
              aux_source.empty() ? nullptr : aux_source.c_str(), true,
              source_name.empty() ? nullptr : source_name.c_str());
     }
-    else if (is_death_effect)
+    else if (is_death_effect || aux_source == "ignited by their inner flame")
```
wheals, line 2716: Rather than checking the `aux_source`, it would make more sense for inner flame to set `is_death_effect = true` on the explosion beam as well (it is a death effect too, after all)

PleasingFungus: It already does set that. I suspect this extra check is not needed.

PleasingFungus:
Thanks for hunting this one down! This was a very hairy little issue - great work fixing it.

PR #1770, opened 2021-02-22
Turn wear-id into floor-id

Implojin:
Following much the same reasoning as in f6d4cf81.

With the removal of curses, players are currently incentivized to wear-id
(almost) everything they can. This commit is targeted as a tedium reduction:

Every item you could equip-id before, you can now identify by stepping on.

Dev notes: The equip-id code was spread out and a little hairy.

In the course of moving the logic over, I've unified the floor-id codepath
for identifying OBJ_WEAPONS, OBJ_STAVES, OBJ_ARMOUR, and OBJ_JEWELLERY.

I've tried to preserve the old wear-id behavior as closely as possible
while remaining reasonable here, but there was some spurious logic
(effective noops and duplicate flags being set) that I've removed in the
course of refactoring this.

That "newly_identified" prop is not well tested, but hey, that's what
trunk is for.

### DIFF
--- a/crawl-ref/source/player-equip.cc
+++ b/crawl-ref/source/player-equip.cc
@@ -176,6 +176,9 @@ void equip_effect(equipment_type slot, int item_slot, bool unmeld, bool msg)
         _equip_armour_effect(item, unmeld, slot);
     else if (slot >= EQ_FIRST_JEWELLERY && slot <= EQ_LAST_JEWELLERY)
         _equip_jewellery_effect(item, unmeld, slot);
+
+    if (item.props["newly_identified"].get_bool())
+        item.props.erase("newly_identified");
 }
 
 void unequip_effect(equipment_type slot, int item_slot, bool meld, bool msg)
@@ -220,7 +223,6 @@ static void _equip_artefact_effect(item_def &item, bool *show_msgs, bool unmeld,
             you.unrand_reacts.set(slot);
     }
 
-    const bool alreadyknown = item_type_known(item);
     const bool dangerous    = player_in_a_dangerous_place();
     const bool msg          = !show_msgs || *show_msgs;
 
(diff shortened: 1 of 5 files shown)


Implojin:
Oh, I suppose it might be good to include save-upgrade code here, as a convenience thing for players with ongoing games?

I've never written a TAG_MINOR before. Would that be required here?

crawl-ref/source/player-equip.cc
```diff
@@ -604,9 +583,9 @@ static void _equip_weapon_effect(item_def& item, bool showMsgs, bool unmeld)
             switch (special)
             {
             case SPWPN_DISTORTION:
-                if (!was_known)
+                if (item.props["newly_identified"].get_bool())
```
PleasingFungus, line 586: This doesn't really make any sense - there's no "surprise, you accidentally wielded a distortion weapon!". We should just toss this whole branch.

crawl-ref/source/player-equip.cc
```diff
@@ -265,9 +267,9 @@ static void _equip_artefact_effect(item_def &item, bool *show_msgs, bool unmeld,
     if (proprt[ARTP_RAMPAGING] && msg && !unmeld)
         mpr("You feel ready to rampage towards enemies.");
 
-    if (!alreadyknown && dangerous)
+    if (item.props["newly_identified"].get_bool() && dangerous)
```
PleasingFungus, line 270: Likewise, delete.

crawl-ref/source/player-equip.cc
```diff
@@ -1277,7 +1246,7 @@ static void _equip_jewellery_effect(item_def &item, bool unmeld,
 
     if (!unmeld)
         mprf_nocap("%s", item.name(DESC_INVENTORY_EQUIP).c_str());
-    if (new_ident)
+    if (item.props["newly_identified"].get_bool())
         auto_assign_item_slot(item);
```
PleasingFungus, line 1250: I believe this should happen on pickup, so I think we can delete this (and the "newly_identified" prop in general). Please test that, though.

PleasingFungus:
Code changes mostly LGTM, but I'm not in a position to test right now. Left some small comments.

crawl-ref/source/player-equip.cc
Implojin: All of this sounds good to me, I'll have to look at the item slot stuff since there were some rcfile things being called in there.

I probably won't get around to updating this until tomorrow, though.

Thanks PF~

Implojin:
Oh! Thanks for cleaning this up, I didn't have time to get to it last night.

PleasingFungus:
No problem. Thanks for implementing it!

PR #1771, opened 2021-02-22
Increase regeneration rate for alive vampires

nlavsky:
to match their old regeneration bonus (half a REGEN_PIP).

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -1162,7 +1162,7 @@ int player_regen()
 
     // Bonus regeneration for alive vampires.
     if (you.species == SP_VAMPIRE && you.vampire_alive)
-        rr += 20;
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Why?

nlavsky:
It didn't became less valuable, but it's less noticeable now. The manual says that "a Vampire full with blood will regenerate very quickly.", and +0.2 HP/turn is definitely not "very quick" now. I was sure you simply missed them in c8f691b2. Please feel free to close this PR otherwise.

PleasingFungus:
Nope, I'd intentionally left this untouched. Good point about it being less noticeable, but I'm not convinced that vp need more reason to be alive instead of dead. Perhaps worth revisiting this in a wider discussion of vampire design.

Thanks for pointing it out!

PR #1773, opened 2021-02-24
cut down hell floors to 2; increase hell spawn rate

pig005:
### Summary
- hell floors cut down from 7 to 2
- hell monster spawn rate increased from 23 to 60;

### Why?
most of dcss players do 3-rune run. not much tries 15-rune run. one of the main reasons is that end-game contents are long and tedious.
- there's not much left for player to grow;
- pandemonium and hell are too long;
  - endlessly wandering about in pande, same scenery, same monsters except rune floor
  - each hell hosts 7 floor, and you are already tired from all that quest from main dungeon to zot

### Conclusion
so, some accessibility must be provided before adding end-game contents, which would not be done neartime; since adding contents take long time to develop.
this PR tries to add a bit of accessibility to end-game contents by shotening hell first. shortening pande is planned too;

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .cquery_cache/
 .DS_Store
 .vscode/
+.ssh/
 
 *.sublime-workspace
 *.sublime-project
--- a/crawl-ref/source/branch-data.h
+++ b/crawl-ref/source/branch-data.h
@@ -157,7 +157,7 @@ const Branch branches[NUM_BRANCHES] =
       LIGHTGREY, LIGHTRED,
       'H', {}, branch_noise::normal },
 
-    { BRANCH_DIS, BRANCH_VESTIBULE, 1, 1, 7, 28,
+    { BRANCH_DIS, BRANCH_VESTIBULE, 1, 1, 2, 28,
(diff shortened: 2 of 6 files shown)


ebering:
Per discussion on discord:

The "dive" aspect of hell is one that many on the devteam would like to keep. Even as a temporary measure, cutting the hells to being very short and very dense with trash monsters isn't the direction we'd like to go in, so I'm closing this.

If you're interested in improving end-game contents, a much better "quick fix" would be to shrink the hells level size (a lot), so that the "dive" is emphasized.

In that conversation I also sketched some other hell improving directions (improving the monster list threat levels, improving the timed effect hostility, and implementing branch-wide effects) if you're interested in further work.

pig005:
was very busy and finally seeing this of now.

> Even as a temporary measure, cutting the hells to being very short and very dense with trash monsters isn't the direction we'd like to go in, so I'm closing this.

I'm having an impression that as if I was the one filling hell with trash monsters but nevermind; just glad that devteams are interested on improving hell. Can't really make time to help but that 'diving' mechanics would be great to have, honestly.

PR #1774, opened 2021-02-24
Don't waste time casting Sublimation of Blood when at max MP

nlavsky:
After 4502bb1e, the player can cast Sublimation when at max MP. The description still has the "This spell would have no effect right now" warning, and the spell is correctly greyed out on the quiver in this situation, though.

Fix this by moving Sublimation's MP check into a function that is called before subtracting the MP cost. This adds a spell-specific check to casting-specific ones, but it's good to have all MP-related checks in one place. Also, it's clearer than temporarily refunding the spell cost, as it was before 4502bb1e.


ebering:
Thanks for cleaning up an old XXX along the way!

PR #1778, opened 2021-02-25
New Trog-themed trove

Skrybe:
New treasure trove with some Trog acquirements mixed in with the usual weapon trove loot.

Also a minor update to a Vaults entry of mine to vary the wall types used.


ebering:
Flavourful! As always, thanks for the submission!

PR #1784, opened 2021-02-28
Nerf the dex meta.

hellmonk3:
I think this is the best way to retain a linear formula without impacting the rest of
the game too much; slope and intercept can be adjusted further if necessary.

Ended up going with 20% reduction after aliscans pointed out that I messed up
the intercept value (whoops); 20% should be reasonable after all.


hellmonk3:
Ah, yeah, my bad. Corrected that and decided to drop dex * dodging 20% after all.

PR #1785, opened 2021-02-28
Make *slow stack in the same way as *corrode.

hellmonk3:
Instead of being slightly different for no reason.


PleasingFungus:
Some slightly less different dreams come true (:

Issue #1766, opened 2021-02-22
Travel says "Done exploring." prematurely if there are exclusions by the unexplored area.


Aliscans:
**PLATFORM and OS or BROWSER:**
Linux 3.16.0 console using XTerm(358)

**DCSS VERSION AND GAME INFORMATION:**
0.27-a0-404-g4390dbc5be

**DESCRIPTION:**
A couple of problems with explore and excluded squares:

1 (see Adwydric_Done.cs): There are unexplored parts of the level, but I would need to pass through an excluded area to reach them. If I press o to explore after loading this save, however, it says "Done exploring.", when I haven't explored everywhere.

If I change it _is_travelsafe_square() so that it doesn't reject excluded squares if ignore_hostile is set, it says "Partly explored, can't reach some places". This is better, but it would be nice if it also mentioned the items. 

The change also means that non-excluded floor squares which can't be reached because of the exclusions are highlighted, but I'm ambivalent about that.

2 (see Adwydric_Partly_explored.cs): There are no unexplored parts of the level, but there is an excluded area which reaches the edge of the known area (or possibly stretch beyond it - I haven't checked the range). If I press o to explore after loading the save (with autopickup off), however, it says "Partly explored, can't reach some places.", when I have explored everywhere.

I've edited this after looking into it further.

**EXTRA FILES:**
[dif.txt](https://github.com/crawl/crawl/files/6018685/dif.txt)
[save.zip](https://github.com/crawl/crawl/files/6019041/save.zip)

ebering:
Fixed in #1787

PR #1787, opened 2021-03-01
Make explore behave better with exclusions

Aliscans:
Change how auto-explore responds to excluded areas of the level (to address the issues I raised in #1766).

1. Walls and trees are never considered reseedable (so you can't change a level from being explored to not being explored by adding a travel exclusion).
2. Excluded squares may be considered travelsafe when ignore_hostile is set (so the game will look beyond the start of the excluded area to find out if the level has been explored).
3. Explore will consider putting the explore_target somewhere where point_distance < 0 (so that it reports when it can't reach items in an excluded area).

3 is actually achieved by reversing e99fe26798. I added a similar check in _explore_find_target_square(), so that the explore command continues to go up to a runed door which is preventing you from exploring everywhere.

### DIFF
--- a/crawl-ref/source/travel.cc
+++ b/crawl-ref/source/travel.cc
@@ -385,14 +385,14 @@ static pair<bool, string> _feat_is_blocking_door_strict(
  */
 static bool _is_reseedable(const coord_def& c, bool ignore_danger = false)
 {
-    if (!ignore_danger && is_excluded(c))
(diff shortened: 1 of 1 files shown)


ebering:
Thanks for the patch! And thanks for the work you've been doing on autoexplore recently!

Aliscans:
Thanks.

PR #1790, opened 2021-03-01
Foxfire tweak

AdamPG:
Some foxfire tweaks to make the spell more user friendly.

### DIFF
--- a/crawl-ref/source/spl-summoning.cc
+++ b/crawl-ref/source/spl-summoning.cc
@@ -2994,7 +2994,10 @@ spret cast_foxfire(actor &agent, int pow, god_type god, bool fail)
              agent.is_monster() ? "s" : "");
     }
     else if (agent.is_player())
-        canned_msg(MSG_NOTHING_HAPPENS);
+    {
+        mpr("There is not enough space to conjure foxfire!");
+        return spret::abort;
+    }
 
     return spret::success;
 }
--- a/crawl-ref/source/spl-summoning.cc
+++ b/crawl-ref/source/spl-summoning.cc
@@ -2964,22 +2964,17 @@ spret fedhas_grow_oklob(bool fail)
 spret cast_foxfire(actor &agent, int pow, god_type god, bool fail)
 {
     fail_check();
-    int created = 0;
+    bool created = false;
 
-    for (fair_adjacent_iterator ai(agent.pos()); ai; ++ai)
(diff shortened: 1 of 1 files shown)


PleasingFungus:
The 'not enough space' commit is an information leak if there are invisible monsters nearby, right?

AdamPG:
> 
> 
> The 'not enough space' commit is an information leak if there are invisible monsters nearby, right?

Yes, but no more information is leaked than before, since a savvy player knows that the "nothing happens" message means that the spell lacked the space to place the foxfire.

PleasingFungus:
That's not what I meant, sorry. The key thing here is that it gives the player information *without taking any time*, which in some circumstances can encourage unfun play patterns (though maybe not so much here).

I'm fine with improving the messaging, I'm just a bit worried about aborting the spell.

AdamPG:
Fair enough. Like I said in the commit message, the better solution is to make a helper function that checks for sufficient room before casting any summoning spell. With that implementation, one could easily make the player still waste a turn if there is insufficient space due to an invisible monster. I plan to eventually make this helper function, so perhaps you could accept the first commit and reject the second (I'm not super familiar with managing git repositories but I assume that this is possible).

PleasingFungus:
Yep, that's possible!

There was some discussion of this PR on ##crawl-dev today - if you go to https://s-z.org/crawl-dev/%23%23crawl-dev-20210301.lg and search for 'foxfire', you'll find it.

ebering:
Merged a check in the commit that closed this PR.

Replying to IRC: I'm aware that increasing the radius of foxfire to 2 would be a buff and make it work better in corridors. It would also significantly alter its mechanics around corridor endpoints. Per some of our previous discussions, the very-early FE gameplay is a very different experience from other book backgrounds and the better foxfire gets in a corridor the less distinct FE becomes. If FE were painfully weaker than other book backgrounds it would be different, but the broad gameplay stats don't support that position.

From broad gameplay statistics FE is not noticeably weaker than the flame tongue / throw flame version based on how well players do with it: win rates are within statistical noise (drowned out by other changes to the game), death rates on early floors are similar to other book backgrounds, and these observations persist when restricting to various "good player" nicks.

The times when an early FE can do absolutely nothing but punch with untrained fists while trapped in a corridor and this wasn't (at least exacerbated by) player error seem to be within our "acceptable unavoidable deaths" threshold.

PR #1791, opened 2021-03-01
Depths vault cleanup

kippig:
Updates the Depths vaults to match the new depths monster table.

Boggarts, Centaurs, Yaktaurs, Ogres, Imps, wraiths, shadows, shapeshifters, death yaks, hell hounds, red devils are changed. Either removed, or made more dangerous. Some of these vaults might need adjustment because they are giving more xp.

eleionoma have been added to forest vaults instead of boggarts


gammafunk:
Thanks, will clean things up and do a review before merging.

kippig:
Just played the new depths and I think I was too conservative with some of the removals. Going for a second pass getting rid of ice devils and similar ilk. Going to also try to keep each vault at similar XP.

PR #1792, opened 2021-03-01
Rename some local variable in spectral_weapon_fineff::fire().

Aliscans:
The following compiler warning was given by gcc 10.2.0. Renaming def removes the warning. I also renamed atk, but only for consistency.

fineff.cc: In member function ?virtual void spectral_weapon_fineff::fire()?:
fineff.cc:721:12: warning: declaration of ?def? shadows a member of ?spectral_weapon_fineff? [-Wshadow]
  721 |     actor *def = defender();
      |            ^~~
In file included from fineff.cc:9:
fineff.h:30:16: note: shadowed declaration is here
   30 |     mid_t att, def;

Looking at this, though, I wonder if using mid_t here does anything more than let you cast a const actor* to an actor*.

### DIFF
--- a/crawl-ref/source/fineff.cc
+++ b/crawl-ref/source/fineff.cc
@@ -717,15 +717,15 @@ void summon_dismissal_fineff::fire()
 
 void spectral_weapon_fineff::fire()
 {
-    actor *atk = attacker();
-    actor *def = defender();
-    if (!def || !atk || !def->alive() || !atk->alive())
+    actor *attack = attacker();
+    actor *defend = defender();
+    if (!defend || !attack || !defend->alive() || !attack->alive())
         return;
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Feels a little awkward that `attack` shadows the class in `attack.h`. WDYT about just changing `atk` to `atkr` or something silly like that? (`atk_act`?)

PleasingFungus:
> I don't have any strong feelings about any name; I only chose "attack" because that was what functions such as mirror_damage_fineff::fire() use.

Ah, I hadn't realized that! That's reasonable. Probably we should standardize at some point.

Thanks for the change!

Aliscans:
Thanks.

PR #1793, opened 2021-03-01
Improve Vaults monster sets

PleasingFungus:
The Vaults branch has a few long, well-known problems:

- It lacks a distinct identity. The overlap between Vaults' monsters and other branches the player is likely to do shortly beforehand or after (Elven Halls, the Depths, etc) is very high.
- There's too much chaff. It's good for branches to have a mix of weaker and stronger enemies, and to recontextualize enemies that were strong earlier in the game once the player has become more powerful. However, enemies like ogres, hornets, and centaurs are just extremely unlikely to pose any kind of threat to a player in Vaults.

As a result, Vault 1-4 often feel a bit of a slog, midgame doldrums. There are many things about the branch that work well - the distinctive layouts, the cool human enemies, the iconic yaktaur herds - so we should focus on them.

This branch currently makes the following changes:
- Many enemies have been removed from the Vaults spawn tables, generally because they posed no threat to Vaults characters or because they overlapped with Depths, Elf, or Crypt. (I don't feel strongly about the Crypt monsters - we could certainly add some of those back in if people want.)
- A few existing enemies are brought into the Vaults. Lindwurms and dire elephant bands come visiting from Lair, and peacekeepers visit from the Desolation of Salt. The former are top-tier Lair threats that pose a much lower threat in Vaults, and the latter is a standard Desolation threat that's scarier in Vaults. These enemies aren't heavily used anywhere, so it should be fun to see them in a new context.
- A new human monster, the Icegrip Stalwart. Casts a Frozen Ramparts variant that does instant cold damage + flash freeze to non-allied creatures standing next to walls. (Playing with the Vaults terrain.)
- A new ogre, Ironbound Thunderhulk. Fires a smitey lightning bolt that can only hit targets at least 3 cells away. (Playing with minimum range mechanics and with the big Vaults mobs that tend to clog doorways and hallways.)

TODO:
- Consider whether or not we want to add more new monsters. Suggestions include ebering's Ironwhip Beastmaster and new super-orcs, Orc Hierophants and Orc Occultists.
- Take another pass at the spawn tables. We should either split apart Vaults:5's spawn tables from Vaults 1:4 completely or (more likely) unify them in some way. We should double check that XP values are reasonable.
- Play through some levels with a few characters and think about whether we'll need to shrink level gen to avoid levels feeling empty with fewer but nastier monsters, similar to Swamp.
- Add some sprites for the new monsters.


PleasingFungus:
Notes on proposed new monsters (low confidence in these):

- The proposed orcs would also make Beoghite orc sorcerers and high priests scale a bit better into the late game. I consider this a plus. Beoghite players love evolving their pokemon into better pokemon, and it feels bad that orc knight -> warlord is the only path worth pursuing later.
- Orc Hierophants would smite, summon something nasty, and cast Beogh's Grasp, a touch-range spell that hurts the victim and heals all allies in sight.
- Orc Occultists would maybe cast paralysis, bolt of acid, and summon something nasty? (Overlap between them & Hierophants is intentional, echoing high priest / sorcerer summon overlap.)
-  Ironwhip Beastmasters would spawn with bands of beasts (echoing the current humans + exotic beasts mini-theme of Vaults) and synergize with them in some way? (TBD.)

crawl-ref/source/dat/descript/monsters.txt
```diff
@@ -1405,6 +1405,12 @@ howler monkey
 An aggressively territorial primate, known for the astonishing variety and
 volume of noise it can produce.
 %%%%
+hulking thunderguard
```
kippig, line 1408: Does this not have a species? maybe make it an ironclad or a palentonga or something

kippig: I see its a human below, why not name it similar to other vaults monsters. Ironheart?

PleasingFungus: Sure, could do. No strong opinion.

semitonal:
Not having generic "beast" monsters feels like it's one of the main distinguishing features of the Vaults monster set to me so I wouldn't be super enthused about bringing in dire elephants in particular. They seem fine just being a rare high-level Lair threat where their main feature of just being occasional huge damage sponges can be relevant, I don't see that being a particularly fun property in more frequent packs.

I'd propose sticking with the simpler naming scheme of "ogre beastmaster/thunderhulk/<jobname>" at least for the new 'O'-glyph monsters, rather than add lots more of the somewhat flowery newvaults-style names.

PleasingFungus:
> I'd propose sticking with the simpler naming scheme of "ogre beastmaster/thunderhulk/<jobname>" at least for the new 'O'-glyph monsters, rather than add lots more of the somewhat flowery newvaults-style names.

I'd been trying to specifically associate the new ogre monsters with the existing human ones, to make the theme a bit more consistent/cohesive. I'm open to changing that, though!

Fair point re dire elephants - I'd eyed them specifically as a monster that wasn't much used elsewhere and that would feel a bit different than it did in lair, but agree they aren't the most fascinating enemy without support.

ebering:
Part of the beast master genesis was my sense that more unattended beasts was un vaults like, so I agree with Kate there.

Re the new O glyph monsters: they're on O because we ran out of p. I think they'd be better as "vaults humans" with new vaults style names and sprites. In this direction I stuck them in warden bands.

We'd have enough space on p if we moved the desolation ps (other than convokers) to somewhere else. In irc I suggested dwarves (g), who are an ok fit for the ancient and evil empire remnant, though PF disagrees with g. 

Other options are 'a' for ancient (totally free) or 'R' with a demonic twist. 'H' also has enough room if we make them say minotaurs or some other hybrid.

PR #1795, opened 2021-03-02
Don't make a note when you find an altar you can't use.

Aliscans:
Add a "temp" parameter to player_can_join_god() and player::undead_or_demonic(). If false, the functions ignore temporary effects from things like mutations and spells, and only consider permanent barriers such as the player's species and Ru sacrifices.

Use this to stop the game from making a note on finding an altar for a god the character will never be able to worship.


ebering:
Nifty, thanks for the patch!

PR #1806, opened 2021-03-08
Add clua documentation for spells.fail_severity()

Implojin:
This series of commits:
* Adds clua documentation for spells.fail_severity()
* Replaces a miscast threshold magic number in spl-cast.cc with MISCAST_THRESHOLD
* Staticifies MISCAST_THRESHOLD and MISCAST_DIVISOR

### DIFF
--- a/crawl-ref/source/l-spells.cc
+++ b/crawl-ref/source/l-spells.cc
@@ -173,8 +173,13 @@ LUAFN(l_spells_fail)
     PLUARET(number, failure_rate_to_int(raw_spell_fail(spell)));
 }
 
-/*** The failure severity of the spell.
- * TODO: Document these numbers
+/*** The miscast severity of the spell as a number in [0,5].
+ * 0: light grey, no chance of damaging miscast
+ * 1: white, <= 10% max HP damage
+ * 2: yellow, <= 30% max HP damage
+ * 3: light red, <= 50% max HP damage
+ * 4: red, <= 70% max HP damage
+ * 5: magenta, > 70% max HP damage (potentially lethal)
  * @tparam string name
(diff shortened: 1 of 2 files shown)


crawl-ref/source/spl-miscast.h
```diff
@@ -8,10 +8,10 @@
 #include "spl-util.h"
 
 // scale miscast severity
-const int MISCAST_DIVISOR = 9;
+static const int MISCAST_DIVISOR = 9;
 
 // cutoff for getting a miscast
-const int MISCAST_THRESHOLD = 150;
+static const int MISCAST_THRESHOLD = 150;
```
ebering, line 14: For your future reference `static const` and `const` mean the same thing in C++; crawl code is inconsistent about which one we use (of course).

If you could rebase this dropping the staticify commit (so `git blame` is a bit easier for these lines) this looks good to go.

ebering (changes requested):

crawl-ref/source/spl-miscast.h
Implojin: Thanks ebering.

Unlike many of the folks in -dev I don't have formal CS experience, and I sometimes realize I've been cargo culting something or other, so I appreciate your taking the time.

Implojin: Okay, I've rebased to master and dropped the staticify commit.

ebering: > Unlike many of the folks in -dev I don't have formal CS experience

Even with formal experience C++ is widely known to be a monster.

ebering (approved):

PR #1809, opened 2021-03-08
Fix shop purchase potion/scroll stacking

Implojin:
This was broken in 83025b5. It turns out that items_stack() has some fairly
particular conditions that don't necessarily match ISFLAG_IDENT_MASK, due
to a potential lua info leak. For some background on this, see:
crawl-dev-20151123.lg

If this commit somehow breaks identify again, I apologize in advance.


ebering:
Thanks for this fix!

PR #1813, opened 2021-03-09
add spells.describe lua function

mgdelmonte:
This PR would add the `spells.describe` function, which provides the same description one can find by accessing a particular spell in the memorised spells list.  Useful for evaluating a spell's current damage, accuracy, and other attributes, within a script context.


ebering:
Hi, thanks for the patch. We strongly prefer rebases to merges, especially when the change fits in a single commit.

Also, as I've written before, we'd really appreciate it if you included your nice pull-request text in the commit messages!

PR #1816, opened 2021-03-09
Clarify that amulet of the acrobat doesn't work while taking stairs

unforgettableid:
I would suggest that, when updating the changelog, you add this entry:

"The amulet of the acrobat now makes it explicit in its description that it
doesn't work while taking stairs."

Many players still don't know that the amulet of the acrobat doesn't work while
taking stairs. If you create my suggested changelog entry, at least those
players who read the changelog will learn about this fact. In the end, the
changelog addition may prevent quite a few needless character deaths.

### DIFF
--- a/crawl-ref/source/dat/descript/items.txt
+++ b/crawl-ref/source/dat/descript/items.txt
@@ -40,9 +40,12 @@ function, it must first attune itself to the wearer's body at full health.
 %%%%
 amulet of the acrobat
 
-An amulet that allows the user to tumble and roll to evade the blows of their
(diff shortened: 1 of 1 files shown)


unforgettableid:
The idea for this pull request came from @RoGGa-loves-DCSS, who mentioned it in an old [Tavern post.](https://crawl.develz.org/tavern/viewtopic.php?p=330218#p330218)

RoGGa, I thank you for your idea!

ebering:
Thanks for the patch! I went with a more concise clarification.

PR #1818, opened 2021-03-09
Make Poison vault always accessible

kippig:
gotta dig to this one sometimes


ebering:
Thanks for the patch! I resolved this a different way: in general we prefer to set the `transparent` tag and use `KMASK:` to indicate `opaque` inaccessible sections. Someday we might be able to shift to default transparent vaults but that's a long ways off.

PR #1821, opened 2021-03-11
Add Hepliaklqana Heroes Vault

kippig:
Add's a vault dedicated to the three paths Hepliakqana heroes can take. The monsters in this vault are equipped similarly to the ancestors. The monster's ancestors are permanently hasted at later levels (or they are liches in the case of spellcasters). Some of the Heroes (for Knight) don't quite match the ancestor because a guaranteed broad axe of speed seems too good. Instead players have a 9% chance of that drop in the case where the knight spawns in depths.

Might need some balancing (maybe don't spawn orc knight on d:10). Not entirely sure about how overflow altars work in later stages like depths.

### DIFF
--- a/crawl-ref/source/dat/des/altar/overflow.des
+++ b/crawl-ref/source/dat/des/altar/overflow.des
@@ -1238,6 +1238,84 @@ xxwwwwwxx
 xxxxxxxxx
 ENDMAP
 
+NAME:   beargit_hepliaklqana_dungeon_heroes
+TAGS:   temple_overflow_hepliaklqana temple_overflow_1
+TAGS:   no_monster_gen patrolling no_trap_gen
+DEPTH:  D:4-, Depths
+SHUFFLE: 12d
+# Battlemage
+: if crawl.one_chance_in(3) then
+:   if you.in_branch("D") and you.depth() < 10 then
+KMONS:  1 = blork the orc; quarterstaff, orc wizard; quarterstaff . robe
+KMONS:  2 = spectral orc wizard name:ancestral_orc_wizard name_replace
+ITEM:   randbook owner:Hepliaklqana spells:stone_arrow numspells:1
+:   elseif you.in_branch("D") then
+MONS:   ogre mage; lajatang . potion of haste
+MONS:   spectral ogre mage name:ancestral_ogre_mage name_replace
+ITEM:   randbook owner:Hepliaklqana spells:bolt_of_magma numspells:1
+:   else
+MONS:   deep elf annihilator; lajatang ego:freezing . potion of haste
+MONS:   ancient lich name:ancestral_lich name_replace
+ITEM:   randbook owner:Hepliaklqana spells:lehudib's_crystal_spear numspells:1
+:   end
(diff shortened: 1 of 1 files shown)


gammafunk:
Looking over this vault, like the concept of fighting the ancestor archetypes, but the spectral versions don't really work. The renamed spectral are going to generate confusion as vault-defines (e.g. perma-haste) and would need a custom descriptions, yet for doing all that all you have is at best a hasted spectral, which is not terribly interesting.

Here's an alternate proposal: make the vault about fighting just one, two, or even all three ancestral monster archetypes. It's easier to create those archetype monsters by using conventional monsters with appropriate equipment, especially in Dungeon and Depths where you have a big pool of monsters to choose from. You could give e.g. the wand of paralysis to the hexer monster, and along with a battle mage or appropriate knight monster, this could be a pretty nasty fight that would be thematic without requiring vault-defined monsters.

Regarding balance, a D:10 orc knight is fairly nasty, but you can control its gear by conditioning it based on `you.depth()`. I'd recommend treating using breakpoints of D:2-D:6 for first 'tier', D:7-D:11 for second, and D:12- for third, with the final tier happening in Depths.. So for D:2-D:6 it's the orc warrior flail, kite shield, and chain, for D:7-D:11, you can use an orc knight with a flaming broad axe and kite shield, and D:12- you can give it tower reflect shield and haste potion(s) (multiple should be ok here). If you use this structure for all three archetypes, it should be thematic in that it's consistent with the actual ancestor upgrades. 

It might actually work best to have the player fight just the solo "ancestor" monster so you have the freedom to upgrade monster types; you could use orc warlord for the final tier in D:12-, for example. Alternately you can tone down the base types, but have the player fight two archetypes, so maybe just orc warrior for D:7-11 with appropriate gear, but there's also a battlemage or hexer (this would be randomized).

Let me know what you think, if you have any questions, etc.

kippig:
Yeah, this one I didn't expect to get past review on the first try. Agreed on dropping the spectrals. Since this is an overflow altar I think the first tier shouldn't be overwhelmingly dangerous. But as you get later you'll get ganged up on by all 3.

I have a number of balance changes in mind too:
broad axe speed -> arga (with low probability) because even arga in depths shouldn't happen every time the vault spawns.
I'll redo the map a bit so there's three "cubbies" that can see the central area so you get mobbed in 3 directions at once (and can retreat). Cubbies will have random Hep themes (randomization in the rotation). Hexers will have wand of paralyze so they can be more types of monsters. Adding salamander mystic since it has bolt of magma and can extend haste on the others if their potions run out. probably other things will happen when I'm editing.

kippig:
Take a look. I set it to 1 monster (early early game) and then you get mobbed by 3 later. Also made more species show up (did all the reasonable ones, nothing weird like Formicid). Notable missing species that are normal dungeon dwellers are ghouls and demonspawn. You might want to remove the hexer merfolk siren since its a bit out of place too. I think hasted spriggan with a wand of paralyze is pretty hilarious but not sure that'll pass standards either.

PR #1822, opened 2021-03-11
Add diablo church arrival vault

kippig:
adds a church based on the tristram cathedral from the first
diablo game. Hell has desecrated the altar, but a player must
first contend with the Imp guarding it if they dare. a hatch is
provided so you can worship and run or hatch if you get an
unlucky teleport. To the sides of the church there is a crypt and
through the windows the player can see the lust forest they leave
behind. There's also a baptism font (usually dry) because why not.

### DIFF
--- a/crawl-ref/source/dat/des/arrival/large.des
+++ b/crawl-ref/source/dat/des/arrival/large.des
@@ -1115,3 +1115,55 @@ MAP
       xxxxxxGxxxxxx
         xxxxxxxxx
 ENDMAP
+
+##############################################################################
+# 37x22
+NAME:   beargit_arrival_diablo_church
+TAGS:   arrival no_trap_gen no_item_gen no_monster_gen
+ORIENT: float
+MONS:   rat / nothing w:50
+FTILE:  `t = floor_grass, 1'G_> = floor_limestone
+COLOUR: ` = green, 1'G = white
+KPROP:  ` = no_tele_into
+SUBST:  V = VVVT, H = ....H
+: dgn.delayed_decay(_G, 'H', 'human skeleton / elf skeleton')
+MAP
+       ccccccc
+       ctttttc
+       ct```tc
+       ccnnncc
+        c.{.c
+        c...c
+  ccccccc+++ccccccccc
+  cV............+...+
+  cV............ccccc
+  ccc..cc...cc..ccc
+  ccc..cc...cc..ccc
+  ctc...........ctc
+  ctn...........ntc
+ cctc...........ctc
+ cttc...........ctc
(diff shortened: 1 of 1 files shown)


semitonal:
Arrival vaults aren't really an appropriate place for runed door challenges and guaranteed monsters, something like this should just be a regular vault elsewhere or just be purely decorative without any monsters/reward.

kippig:
> Arrival vaults aren't really an appropriate place for runed door challenges and guaranteed monsters, something like this should just be a regular vault elsewhere or just be purely decorative without any monsters/reward.

I'll just remove the imps/runed door.

crawl-ref/source/dat/des/arrival/large.des
```diff
@@ -1115,3 +1115,57 @@ MAP
...
+
+##############################################################################
+# 39x22
+NAME:   beargit_arrival_diablo_church
+TAGS:   arrival no_trap_gen no_item_gen no_monster_gen
+ORIENT: float
+MONS:   rat / nothing w:50
+FTILE:  `t = floor_grass, 1'G_> = floor_limestone
+KPROP:  ` = no_tele_into
+KFEAT:  _ = altar_ecumenical
```
ebering, line 1127: This doesn't look like it places anywhere?

Unless its placing behind the glass, I would remove this line; while ecumenical altars do spawn on D:1 quite regularly after the recent change, putting them guaranteed in a vault makes start scamming too easy.

crawl-ref/source/dat/des/arrival/large.des
```diff
@@ -1115,3 +1115,57 @@ MAP
...
         xxxxxxxxx
 ENDMAP
+
+##############################################################################
+# 39x22
+NAME:   beargit_arrival_diablo_church
+TAGS:   arrival no_trap_gen no_item_gen no_monster_gen
+ORIENT: float
+MONS:   rat / nothing w:50
+FTILE:  `t = floor_grass, 1'G_> = floor_limestone
```
ebering, line 1125: You should correspondingly have a line:
`COLOUR:  ` = green, 1'G = white` so that console players get a good experience.

ebering (changes requested):
Looks possibly a bit too tall (probably fine, but maybe remove two rows?) and needs a small tweak plus console colouring.

crawl-ref/source/dat/des/arrival/large.des
kippig: yeah, that was missed when I removed the imps

ebering:
Thanks for the vault!

PR #1826, opened 2021-03-14
Sort items in a shop in a more predictable order using ORDER_DEFAULT

Aliscans:
As it was, ORDER_DEFAULT sorted items according to the order in which they were originally created by the game. This seems like an unhelpful default to me, and there is no option to control it.

This change means that, when you go into a shop, items are sorted by DESC_DBNAME, and by DESC_PLAIN where this doesn't distinguish them.

Similar items are together (all robes, all +0 robes, all wands, etc.), and most things are in alphabetical order. If you switch to "sort by price", this order is used for items which cost the same amount.

I find that this makes it easier to scan through the contents of a shop for anything I might like.

I have not changed the order in which items in shops appear in character dumps. Items on the floor already appeared in an order the player is never shown in-game, so this is at least consistent.

Artefact jewellery and spellbooks appear in an odd position. They're rare, and are highlighted, so this shouldn't cause any problems.

### DIFF
--- a/crawl-ref/source/shopping.cc
+++ b/crawl-ref/source/shopping.cc
@@ -1171,28 +1171,26 @@ void ShopMenu::resort()
     switch (order)
     {
     case ORDER_DEFAULT:
-        {
+    {
+        // Using a map to sort reduces the number of item->name() calls.
         const bool id = shoptype_identifies_stock(shop.type);
-        sort(begin(items), end(items),
-             [id](MenuEntry* a, MenuEntry* b)
-             {
-                auto a_item = dynamic_cast<ShopEntry*>(a)->item;
-                auto b_item = dynamic_cast<ShopEntry*>(b)->item;
-
-                // "fizzy potion" is "potion", "potion of curing" is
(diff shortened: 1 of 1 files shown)


wheals:
I left ORDER_DEFAULT around when rewriting the shop code a while back since I thought it was a nice flavor touch of having the items in a sort of random jumble. But I'm not super attached to it, and this PR is reasonable.

That said, since this new ordering is sort of a combination of alphabetical and by-type, I think it should be merged into one of them. My suggestion is to remove sort by type since it's mostly redundant now (the one thing lost is that sort by type sorted all weapons together, all armour together, etc., but I'm not sure how often this functionality was  even used), and rename this default sort to sort by type.

Aliscans:
I've removed ORDER_TYPE, as you suggested. I've also changed the way artefacts are shown, and now everything should be sorted based on part of the displayed name.

With weapons and body armour, there's nothing in the name to say that it's a weapon or armour. If they were put together in a general shop, I don't know where they would go in the listing.

PR #1827, opened 2021-03-14
Simplify + bugfix monster weapon damage

hellmonk3:
Removes the opaque effect of weapon base delay on monster attack delay. Fixes the monster weapon enchantment roll. Removes the 1 + random2(3) damage reduction for weapon wielding monsters. This is a pretty sizable player nerf (or rather, monster buff) overall.


PleasingFungus:
I'm wondering about nerfs to 2h ogres, ettins, and reapers. I don't want to disrupt the purity of the branch name, though.

PR #1828, opened 2021-03-15
Fix ?/m entry for rockslimes

nlavsky:
Because of a missing delimiter, `?/mrockslime` showed the river rat's
description.


PleasingFungus:
Merged as 888f4d52c2, thanks!

PR #1830, opened 2021-03-15
Enable fullscreen PWA for WebTiles on iOS

Widdershin:
Currently, adding the webtiles interface to the desktop on iOS or iPadOS creates what is effectively a bookmark. It opens in the browser, with the domain displaying, and tabs available.

By adding these meta tags, if a user creates a PWA for WebTiles on iOS, it will launch in full screen as a separate application to the browser, without the standard tabs interface.

This should be a fairly low risk change, as users have to opt in to creating a PWA. The biggest issue is that this hides the browser navigation if you’re using a PWA, so you might theoretically be able to navigate to part of the site that you couldn’t easily return from.

This might seem like an edge case, but it would be awesome to be able to play DCSS full screen on an iPad with a keyboard.

### DIFF
--- a/crawl-ref/source/webserver/templates/client.html
+++ b/crawl-ref/source/webserver/templates/client.html
@@ -8,6 +8,8 @@
     </script>
     <script src="/static/scripts/contrib/require.js" data-main="/static/scripts/app"></script>
     <link rel="stylesheet" type="text/css" href="/static/style.css">
+    <meta name="apple-mobile-web-app-capable" content="yes">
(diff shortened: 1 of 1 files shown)


ebering:
Welcome to contributing to crawl and thanks for the patch!

How would you like to be listed in CREDITS.txt?

Widdershin:
Thanks @ebering 😄 

Happy to just be listed as "Nick Johnstone" for my credits entry!

PR #1831, opened 2021-03-16
Make early dungeon spicier

hellmonk3:
Based on this minmay thread: https://crawl.develz.org/tavern/viewtopic.php?f=8&t=26580. Add some more variety and interest to D:1 and other early floors while trying to avoid increases to player power. Things I implemented: 

- let shops generate on D:1-3
- allow early rare scroll generation
- allow books, manuals, and staves to generate more often in early dungeon
- modify weapon subtype selection early to increase variety
- allow ecumenical altars on D:1
- buff ball pythons slightly and allow D:1 quokkas
- reduce the number of items generated on the first couple floors

Things from the thread that I did not implement:

- allowing early artifacts. I don't oppose doing this in principle if unrands are excluded, though.
- allowing D:1 adders. They were ok as ood spawns when those generated but are not reasonable for a character to fight at xl 1 in my opinion.
- Increasing the frequency of non-ecumenical D:1 altars. God power levels have tended to rise on average, even if specific outliers have been nerfed. I think adding more specific altars to D:1 would give out too many easy wins; ecumenical altars are at least a gamble.
- allowing early misc items. I think it's fine to keep those as a wanderer-only thing; they're pretty strong.


PleasingFungus:
Mostly merged, thanks! Held off on the ecu altar change for now.

PR #1835, opened 2021-03-17
Improve ancestor management

sdynet:
1. Allow projectiles to pass through ancestors
This change will slightly break Hep's balance, but instead I expect that the player's management QoL will be much better. (Thank PleasingFungus)

2. Respawn of ancestor faster.
It's tiring to take care of ancestor so that they don't die. Therefore, it will be respawn faster to reduce the management burden of the player. Although respawn has become very fast, the absence of an ancestor for 5-15 turns in urgent combat situations is still threat to the player. Therefore, I expect that the value of 'idealise' will not be completely tarnished.

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -6597,7 +6597,7 @@ int _ench_pow_to_dur(int pow)
 bool always_shoot_through_monster(const actor *originator, const monster &victim)
 {
     return mons_is_projectile(victim)
-        || (mons_is_avatar(victim.type)
+        || ((mons_is_avatar(victim.type) || mons_is_hepliaklqana_ancestor(victim.type))
             && originator && mons_aligned(originator, &victim));
 }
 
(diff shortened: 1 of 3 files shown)


PleasingFungus:
Allowing projectiles to pass through seems reasonable, as long as it's communicated to the player. I can take a look at merging that in a bit (I have another hep change I want to make in the process.)

I'd rather not speed up ancestor respawns significantly. I think it's better for the ancestor to be fairly unlikely to respawn within a given fight.

sdynet:
> I'd rather not speed up ancestor respawns significantly. I think it's better for the ancestor to be fairly unlikely to respawn within a given fight.

Well, but in later games, ancestors die too easily. I can heal them, but I can't overdo it because it costs too expensive...

PleasingFungus:
Very sorry for slow follow up here. I still want to at least merge the first part of this PR, maybe along with another change to let the ancestor also fire through the player (friends!). However, I'd like to move the ancestor to 1* first, and I've had very little time to work on crawl for the last few weeks.

Thank you for submitting this! I promise I will come back to this when I do have more time for Crawl. (And if another dev wants to take over or just merge the firing-through-your-ancestor part of this PR, I'm totally fine with that.)

ebering:
Have a patch sequence picking up this thread of ideas that I'll push in the next few hours.

PR #1836, opened 2021-03-19
Freshen up the Snake Pit

ebering:
"Poison" as a theme for a branch is somewhat rough, though unlike water its harder for "poison" to exert good tactical pressure because of how slow poison ticks. This PR accelerates poison ticks, so that when severely poisoned a player must face some real urgency. This should improve both poison branches.

The Snake Pit has an additional issue: slow monsters. The difference between "fast fragile swarms" and "slow tanky hard hitters" does make for a nice distinction between Snake and Spider. However, for slow monsters to be really threatening (and not tedious) they need to hit hard (consequences for making a mistake and letting one get close) and have some tools for catching at least some players. At present, other than guardian serpents spells and constriction, Naga monsters aren't good at either category. So let's give them player-catching tools and harder hits.

Specific catching tools and harder hits:
- Damage of plain naga and naga warriors is increased
- Naga Warriors gain battlecry
- Nagaraja trade Teleport Other for Dimension Anchor, and drop force lance
- Naga Ritualists lose force lance (and become Priests for flavour reasons)
- Naga Mages get a new spell Concentrate Venom, which enchants an allied N or S genus monster to make their poison breath or AF_POISON melee into curare.
- Naga Mages lose most attack spells other than venom bolt, and Gain Haste other as well.


RojjaCebolla:
would it be too spoilery for curing pots to not drop in Snake?

PR #1837, opened 2021-03-19
Update to Blade Hands description

WhiteZhark:
The old Blade Hands description used the word "hands" regardless of
the manipulator type for the species. In addition, it failed to mention
that it would meld shields and gloves. A function was added to l-you.cc
to get the hand type of the player species, and the Blade Hands
description now checks hand type to determine the noun used for the
player's manipulators. The description was also changed to say that
shields and gloves are also melded.

### DIFF
--- a/crawl-ref/source/dat/descript/spells.txt
+++ b/crawl-ref/source/dat/descript/spells.txt
@@ -141,11 +141,23 @@ deplete magical power.
 %%%%
 Blade Hands spell
 
-Causes long, scythe-shaped blades to grow from the caster's hands, increasing
-melee damage significantly.
+{{
+    local hands = you.hand() .. "s"
+    local description =
+               "Causes long, scythe-shaped blades to grow from the " ..
+               "caster's " .. hands .. ", increasing melee damage " ..
+               "significantly.\n\nWhile transformed, equipped weapons, " ..
+               "shields, and gloves are melded, and casting spells " ..
+               "becomes somewhat more difficult."
 
-While transformed, any equipped weapons are melded, and casting spells becomes
-somewhat more difficult.
+    if hands == "paws" then
+        return description .. " In addition, the clacking of blades " ..
+               "on the ground makes the caster slightly less stealthy."
+    else
+        return description
+    end
+}}
+
 %%%%
 Blink Allies Away spell
 
--- a/crawl-ref/source/l-you.cc
+++ b/crawl-ref/source/l-you.cc
@@ -69,7 +69,12 @@ LUARET1(you_race, string, species_name(you.species).c_str())
  * @treturn string
  * @function class
  */
-LUARET1(you_class, string, get_job_name(you.char_class))
+ LUARET1(you_class, string, get_job_name(you.char_class))
+/*** Get noun for player's hands.
+ * @treturn string
+ * @function hand
+ */
+ LUARET1(you_hand, string, species_hand_name(you.species).c_str())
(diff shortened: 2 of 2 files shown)


crawl-ref/source/l-you.cc
```diff
@@ -69,6 +69,11 @@ LUARET1(you_race, string, species_name(you.species).c_str())
  * @treturn string
  * @function class
  */
+LUARET1(you_hand, string, species_hand_name(you.species).c_str())
+/*** Get noun for player's hands.
+ * @treturn string
+ * @function hand
```
sam-oberg-dev, line 75: your linting issue is due to whitespace at the ends of each of these lines

crawl-ref/source/l-you.cc
```diff
@@ -69,6 +69,11 @@ LUARET1(you_race, string, species_name(you.species).c_str())
  * @treturn string
  * @function class
  */
+LUARET1(you_hand, string, species_hand_name(you.species).c_str())
+/*** Get noun for player's hands.
+ * @treturn string
+ * @function hand
+ */
```
ebering, line 76:
```suggestion
/*** Get noun for player's hands.
 * @treturn string
 * @function hand
 */
LUARET1(you_hand, string, species_hand_name(you.species).c_str())
```

ldoc comments appear before the function being documented, not after. Github won't let me suggest it, but you also need to make sure the doc comment for you_class remains in the right place.

crawl-ref/source/dat/descript/spells.txt
```diff
@@ -141,11 +141,36 @@ deplete magical power.
...
+               "are melded, and casting spells becomes somewhat more " ..
+               "difficult."
+    else
+        return "Causes long, scythe-shaped blades to grow from the " ..
+               "caster's hands, increasing melee damage significantly.\n\n" ..
+               "While transformed, equipped weapons, shields, and gloves " ..
+               "are melded, and casting spells becomes somewhat more " ..
+               "difficult."
+    end
+}}
```
ebering, line 171:
```suggestion
{{
    local hands = you.hand() .. "s"

    return "Causes long, scythe-shaped blades to grow from the " ..
           "caster's " .. hands .. ", increasing melee damage significantly." ..
           "\n\nWhile transformed, equipped weapons, shields, and " ..
           "gloves are melded, and casting spells becomes somewhat " ..
           "more difficult."
}}
```

No need for a case statement. If a species with an irregular-plural hand noun gets added this will break, but its better to minimize duplication.

ebering (changes requested):
Hi and welcome to contributing to crawl!

Thanks for submitting this patch, I've left two suggested changes.

crawl-ref/source/dat/descript/spells.txt
WhiteZhark: Thanks for the suggestion. I got rid of the duplication but left a special case for paws because I wanted to include the info about Felids having reduced stealth while using Blade Hands.

crawl-ref/source/l-you.cc
WhiteZhark: Fixed. I think the text highlighting in my editor confused me so I didn't realize the proper ordering. Thanks.

ebering (approved):

ebering:
Thanks for the patch! How would you like to be listed in CREDITS.txt?

WhiteZhark:
Sure, you can put me in as WhiteShark

PR #1838, opened 2021-03-19
Compiler warnings

Aliscans:
Make a couple of functions static. This removes warnings GCC gives for them with -Wmissing-declarations.


ebering:
@semitonal staticified these in 0b40708614a

For future reference, we really prefer commit messages to have a 72 character title line whenever possible.

PR #1839, opened 2021-03-20
Don't nerf the warlock's mirror. Improve debugging messages & comments

Aliscans:
Add a debug statement to give the odds of blocking an AUTOMATIC_HIT bolt with the warlock's mirror.

Reduce the chance to block AUTOMATIC_HIT bolts with a displayed SH of 20 from 2/3 to 1/2.

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -3086,20 +3086,30 @@ bool bolt::misses_player()
         && !aimed_at_feet
         && SH > 0)
     {
-        // We use the original to-hit here.
-        // (so that effects increasing dodge chance don't increase block...?)
-        const int testhit = random2(hit * 130 / 100
-                                    + you.shield_block_penalty());
+        bool blocked = false;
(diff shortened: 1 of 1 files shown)


ebering:
What is the reason for the gameplay change? The Warlocks mirror isn't well renowned as an overly strong unrand...

Aliscans:
As it was, a comment in bolt::misses_player() said it gave a 50% chance of reflecting AUTOMATIC_HIT attacks at 20 SH, but the code gave a 50% chance of reflecting them at 10 SH.  While the warlock's mirror seems to have changed quite a lot in power over the years, I can't see anything in the log to suggest that anyone wanted it to reflect AUTOMATIC_HIT things more often.

I only chose to change the code because it seemed more in keeping with the intention behind the artefact. I'd be just as happy if the comment was changed to describe what the shield actually does.

ebering (changes requested):
Yes, correcting the comment is preferred to nerfing the item.

ebering (approved):

PR #1843, opened 2021-03-22
Make autoexplore/travel wait for barbs to time out.

Aliscans:
As things stood, trying to travel with barbs stuck in you would see the game ask for confirmation, move a single square (if you accept) and then stop because of the HP loss.

This change makes it so that the game waits for the spikes to time out before moving.

You can't remove barbs if you are confused or asleep, but you can't travel then either, and player::is_sufficiently_rested() isn't used for anything else.


ebering:
Thanks for the patch!

Aliscans:
No problem.

PR #1846, opened 2021-03-22
Use https rather than git protocol for submodules

bstrie:
Fixes unexplainable breakage with `git submodule update`.
As discussed with ebering on IRC.


ebering:
While I wouldn't call the breakage unexplainable, using https is best practice. Thanks!

PR #1847, opened 2021-03-23
Give away all of Bilbo's toys

RojjaCebolla:
Fortunately, they only had two.

Put the slinging onto another small slinger (Sp, or maybe Ko), and put the mut resist
onto a species that is only nominally humanoid (VS, or maybe Gr).

### DIFF
--- a/crawl-ref/source/dat/descript/species.txt
+++ b/crawl-ref/source/dat/descript/species.txt
@@ -112,8 +112,8 @@ Scaled mammals the size of horses, Palentongas launch powerful rolling charges.
 %%%%
 Spriggan
 
-Very small, very frail and very fast, Spriggans make highly competent stabbers
-and spellcasters.
+Very small, very frail and very fast, Spriggans make highly competent stabbers,
+slingers, and spellcasters.
 %%%%
 Troll
 
@@ -129,4 +129,5 @@ Vine Stalker
 
 Frail symbionts, Vine Stalkers cannot regain health from potions.
 Their magical reserves absorb damage, and they pack a magic-restoring bite.
+They have an increased resistance to mutagenic effects.
 %%%%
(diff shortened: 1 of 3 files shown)


ebering:
Why? Sp and VS both have a lot going on and are well-developed distinctive species. There's no conservation of species properties design rule we like to follow.

I'm definitely a hard no on giving Sp +4 slings, Sp already has very tilted apts. Slings are cheaper than bows and a fustibalus with higher base damage than a short bow is available, so an Sp considering a ranged weapon is currently strongly steered to slings; no need to make it stronger. We've been trying to move away from having races defined by a single extreme apt; combine that with the general low skill cost of slings and a +4 sling apt doesn't look good on Ko either.

Also not enthusiastic about preserving innate rMut generally, but definitely VS and Gr aren't good choices; they both have a lot going on already (Gr less so).

Goratrix:
My suggestion would be to give +2 Slings to Ko. They currently have 0, and +2 would give them same apt as they have for crossbow, creating a nice decision space for ranged.

riverfiend:
> I'm definitely a hard no on giving Sp +4 slings, Sp already has very tilted apts. Slings are cheaper than bows and a fustibalus with higher base damage than a short bow is available, so an Sp considering a ranged weapon is currently strongly steered to slings;
Take it from me, spriggans do NOT need any new incentive to use slings--They almost need a nerf in that regard, what with the "they're faster" and the "already +2 apt" and the "for some reason a fustibalus is 1-handed even for small species".

WhiteZhark:
Why not give rMut to humans? It fits with their theme of being incredibly boring

sdynet:
> Why not give rMut to humans? It fits with their theme of being incredibly boring

I think it would be better to give it to Dg. They can't get rid of bad mutations through God, and the ability seems to match their background.

RojjaCebolla:
I was thinking that, *because* VS is already quite complicated, tacking on rMut ain't a big deal, cuz the player has already opted-in to a weird experience!

rMut for Dg makes sense too - "your glorious godlike form resists outside meddling!" - but if there's no devteam interest in preserving that mechanic, i can drop it.

PR #1849, opened 2021-03-24
Add some missiles to default autopickup

Implojin:
Curare, atropa, dispersal, and throwing nets are all highly useful items
that can save characters. This commit adds them to autopickup by default.

(9 years ago, curare was removed from autopickup in e11df56, for reasons
that have since been obviated. Everything old is new again!)


ebering:
Squashed and merged, thanks!

PR #1852, opened 2021-03-26
Tier 3demonspawn mutation rework

hellmonk3:
Removed:
- Passive freeze facet. Added the passive freeze mutation to the general mut pool, where it might be interesting.

Reworked:
- Black mark facet. Now gives hex enhancer, black mark, 1 tile radius silence donut
- Hurl damnation facet. Now gives will+ with retaliation on passed will checks, 50% deterministic rtorm, hurl damnation. Hurl damnation is rescaled slightly.

New:
- Corruption/chaos facet. Gives chance to corrode enemies when you damage them, chance to malmutate enemies when you damage them, active ability to forcibly blink away and apply negative statuses to enemies in line of sight (costs exhaustion and skill drain)

Unchanged:
- Robust facet
- Augmentation facet

I think this makes the tier 3 facets more dynamic and interesting.


crawl-ref/source/beam.cc
```diff
@@ -3158,6 +3158,21 @@ void bolt::affect_player_enchantment(bool resistible)
...
             }
         }
+        
+        // punish the monster if we're a willful demon
+        if (you.get_mutation_level(MUT_DEMONIC_WILL))
+        {
+            monster* mon = monster_by_mid(source_id);
+            if (mon && mon->alive())
+            {
+                int damage_roll = 3 + random2(11);
```
nikheizen, line 3224: Maybe this should scale by beam power? Damaging more powerful hexers more heavily would make it more relevant. As it is, it feels like a very pale shadow of Ru aura and I feel like players won't notice it unless it deals the killing blow.

crawl-ref/source/spl-transloc.cc
```diff
@@ -1606,3 +1607,46 @@ void attract_monsters()
...
 }
+
+bool word_of_chaos(int pow)
+{
+    vector<monster *> targets;
+    for (monster_near_iterator mi(you.pos(), LOS_NO_TRANS); mi; ++mi) 
+    {
+        if (!mons_is_tentacle_or_tentacle_segment(mi->type))
+            targets.push_back(*mi);
+    }
```
ebering, line 1618: Is it intentional that this targets friendlies? (Dg interaction danger zone enter here)

crawl-ref/source/beam.cc
hellmonk3: Implemented. Kept the power level fairly low since this is the rank 1 and it also provides will+ though.

crawl-ref/source/spl-transloc.cc
hellmonk3: Fixed, thanks.

crawl-ref/source/spl-transloc.cc
```diff
@@ -1606,3 +1607,47 @@ void attract_monsters()
...
+        if (x_chance_in_y(pow, 500))
+            do_slow_monster(*mons, &you, 20 + random2(pow));
+        if (x_chance_in_y(pow, 500))
+        {
+            mons->add_ench(mon_enchant(ENCH_FEAR, 0, &you));
+            behaviour_event(mons, ME_SCARE, &you);
+        }
+    }
+    
+    you.increase_duration(DUR_EXHAUSTED, 15 + random2(10));
```
semitonal, line 1650: Word of Chaos should ideally use its own cooldown, similar to how berserk and DDoor etc are now implemented - these were split out since there wasn't really any reason for them to interfere with each other (and Word of Chaos should be allowed to work independently from Ru abilities, for example).

PR #1853, opened 2021-03-27
Disable spellcasting skill gain for new berserkers.

Aliscans:
If automatic training was set, and the skills the player actually used all reached 27, the game would start to train the other skills, including the magic skills. This means that someone could offend Trog by training magic skills without casting a spell or visiting the skill management screen.

This change turns magic skills off at the start of the game for a berserker (as already happened when you joined Trog at an altar), and avoids that possibility.

It's pretty easy to experience this problem if you play Meatsprint as a ghoul berserker.


ebering:
Thanks for the patch!

Aliscans:
No problem.

PR #1859, opened 2021-03-30
Resequence early and ecumenical altars

ebering:
Several old overflow altars were incorrectly tagged; this gave certain gods (Ashenzari and Nemelex are stand outs) significantly higher chances of getting a D:1 altar, and don't seem intended.

Many overflow temples to single gods could also place as mini vaults, but the weight for each depended on the god, resulting in an uneven appearance of earlier altars for certain gods. This also prevented non-minivault overflows from placing as frequently, which is sad because they're nice vaults.

Ecumenical altars suffer from the fact that they don't always show up before identified altars; the choice becomes more about a conduct than the chance to gamble for a serious advantage.

This commit sequence addresses all three of the above issues by implementing the following changes:

- All temple gods get exactly 1 guaranteed altar in the overflow range or the ecumenical temple.
- The overflow range is moved to D:3-10.
- There is a 33% chance of exactly 1 random altar to a temple god on D:2.
- There is an 87.5% chance of an ecumenical altar on D:1-3.

There is some concern of startscumming with a D:1 ecumenical altar, but people didn't start scum for D:1 Nemelex, and a player that would quit and re-roll a D:1 ecumenical roll would probably also quit and re-roll a D:2, which we allow regularly. In this case I think the benefit of the interesting choice (roll a random god and maybe get 1* Makhleb on D:1 or maybe get xom), plus the more interesting experience for a new player (try more gods! early! even if you can't get to Temple regularly!) outweighs the start scum concern.

### DIFF
--- a/crawl-ref/source/dat/des/altar/ecumenical.des
+++ b/crawl-ref/source/dat/des/altar/ecumenical.des
@@ -21,7 +21,7 @@ end
 
 NAME:   basic_ecumenical_altar
 TAGS:   transparent
-WEIGHT: 100
+WEIGHT: 200
 : ecumenical_altar_setup(_G)
 MAP
 _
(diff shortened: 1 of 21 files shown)


crawl-ref/source/dat/des/altar/ecumenical.des
```diff
@@ -0,0 +1,259 @@
...
+NAME:   basic_ecumenical_altar
+TAGS:   transparent
+WEIGHT: 200
+: ecumenical_altar_setup(_G)
+MAP
+_
+ENDMAP
+
+NAME:    chequers_ecumenical_altar_simple
+: ecumenical_altar_setup(_G)
```
gammafunk, line 31: It's customary to have common setup lua (shared by a set of vaults) as final line(s) above `MAP` unless statement order is relevant. Looks like that's not the case here, so would recommend moving this there. If you feel that it's better here, that's fine of course, but it's an established style we have.

crawl-ref/source/dat/des/altar/ecumenical.des
```diff
@@ -0,0 +1,259 @@
...
+sSwWw-_-wwwSs
+sSwWw---wwwSs
+@.WwwwwwwwwSs
+sSwwwwwwwwwSs
+sSwwwwwwwwwSs
+sSSSSSSSSSSSs
+sssssssssssss
+ENDMAP
+
+NAME: ebering_ecumenical_altar_webs
```
gammafunk, line 155: Minor formatting detail: align all statement contents so they begin at the same column:
```
TAGS:   transparent
NSUBST: ^ = 4=^. / 4=^..
```
makes vault more readable.

gammafunk (commented):
Vaults look good, really no significant issues, just minor style ones. I only looked at your new ones and fairly quickly at that, but seems ok.

crawl-ref/source/dat/des/altar/ecumenical.des
ebering: Moved in the final merge

crawl-ref/source/dat/des/altar/ecumenical.des
ebering: Thought I'd got all of these. Thanks!

PR #1860, opened 2021-03-31
Remove the Dance card

sdynet:
This card has no unique personality. Remove this card because there are many friends who can play similar roles.


ebering:
As we had discussed on discord, Dance can be a card with personality.

Dancing weapons (if we fix the base types to pick from a list with all the same base damage and only give plain weapons) have different power scaling from some of the other summons in the deck. The big base damage gives them a much stronger starting point, but the power cap on dancing weapons as summons makes the scaling flatter, for a card thats good early when pentagram would give you a weak demon but not as good as a Pan lord.

So I'm closing this, but a PR to simplify what dance summons would be most welcome!

PR #1866, opened 2021-04-01
RST manual: escape the vertical bar key for layer toggle

stahara:
The RST version of the manual needs the | character escaped by a backslash for it to be displayed properly.


ebering:
Thanks for the tweak!

stahara:
You're welcome!  I wrote that improved layer display system years ago, though I don't think I did the docs (the keybind description is general enough that it wouldn't have needed a change).  Amusing that I'd notice and fix a bug in the doc for the same functionality 6 years later.  >8)

PR #1867, opened 2021-04-03
Remove reference to bugzilla

RojjaCebolla:
...which is no longer a thing.


crawl-ref/docs/develop/IRC.txt
```diff
@@ -3,18 +3,7 @@ Developers on IRC
 
 Some of the Crawl developers hang out on the ##crawl-dev IRC channel on
 freenode (irc://freenode/##crawl-dev). You can ask them development related
-questions. Also, users with +v have commit access, so if you have a patch
-which is ready to commit you can poke them about it.
```
ebering, line 7: It is still true that on IRC developers have +v, don't remove this bit of information.

PR #1870, opened 2021-04-04
Consumables panel

nlavsky:
The main goal of adding this panel is to give the player a quick and
easy way to see all the items that can help them succeed in a fight.

It would be ideal to have the full inventory panel, similar to Local
Tiles, but there are a couple of problems. The first is a lack of screen
space: unless you have a big monitor, the choice is either the monster
list (Web Tiles) or the inventory panel (Local Tiles). The second
problem is that only about a half of inventory items are useful in
fights. A filtered view of the inventory will take less space and
present only combat-relevant items.

This commit adds a new panel, which can be configured via seven options:
  * `consumables_panel`: types of items to show on the panel;
  * `consumables_panel_filter`: a list of regular expressions for filtering
    items out;
  * `show_unidentified_consumables`: whether to show unidentified
    scrolls and potions;
  * `consumables_panel_scale`: allows to change the size of icons on
    the panel.
  * `consumables_panel_orientation`: whether to place the panel horizontally
    or vertically.
  * `consumables_panel_font_family` and `consumables_panel_font_size`: font used for displaying the quantities of items.

By default, the panel is displayed in the top left corner, partially covering
the top row of tiles on the map. Non-Barachim chars should always have
at least one out-of-sight row at the top. For Barachim, the game can show
18 rows of tiles with one out-of-sight row at the top as long as you have
736 pixels vertically.

It will look like this:
Showcasing the panel

Later in the game, characters should have around 20-25 different items on the panel when using the default settings:
2 rune char

The panel can be placed vertically too:
Vertical panel
There will be less space for items, but the panel won't cover the LOS square even on small screens.


crawl-ref/docs/options_guide.txt
```diff
@@ -1619,6 +1620,29 @@ always_show_zot = false
         Whether to show the "Zot" status light at all times, even when you've
         got plenty of time left.
 
+consumables = ?!}
```
rawlins, line 1623: I might prefer longer names for these that make clearer what they connect to, e.g. `consumables_panel`, `consumables_panel_filter`, etc.

crawl-ref/source/webserver/game_data/static/cell_renderer.js
```diff
@@ -1198,6 +1198,30 @@ function ($, view_data, main, tileinfo_player, icons, dngn, enums,
                 img_scale);
         },
 
+        draw_quantity: function(qty, x, y, img_scale)
+        {
+            // Copied from tilereg-grid.cc:GridRegion::draw_number()
+            var num = 0;
+
+            if (qty > 9999)
```
rawlins, line 1202: could use `qty = Math.max(0, Math.min(9999, qty))` idiom here

crawl-ref/source/webserver/game_data/static/consumables.js
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        var new_filters = options.get("consumables_filter");
+        if (filters_text !== new_filters.toString())
+        {
+            filters_text = new_filters.toString();
+
+            filters = [];
+            new_filters.forEach(function(pattern) {
+                filters.push(new RegExp(pattern));
```
rawlins, line 111: This would be a big change, so I wouldn't necessarily want to hold up the PR with it, but it might be preferable to implement the filtering on the crawl binary side. Benefits: (i) use the same regex parser that the crawl binary uses (ii) possibly faster (though this would be tbd), and (iii) makes this more easily portable to other build targets. Is this something you have looked into already and could estimate feasibility?

rawlins (commented):
Looks very promising! This is just a quick first-pass review because I was curious how you approached it, so I haven't tried to exhaustively look at the code or test it myself. However, I think this is not too far off from merging *except* in case we think the logic for constructing the consumable bar should be at least partially in the crawl binary. But as I said in the specific comment on this I don't know that that is a prerequisite.

Re the dilemma about UI space, I think a long-term solution where everything is implemented in both UIs is just to allow users to choose what they want in both interfaces. The right sidebar is oddly polarizing where we consistently get requests to have the local tiles stuff instead of the monster list in webtiles, and also to get the inverse request on local tiles. One thing that happens is many local tiles players mainly activate things with the mouse and aren't very comfortable with the key shortcuts, so webtiles is a huge complication for them, so these players would definitely choose the inventory. However, I think that is all independent of this PR, which looks like an interesting UI change on its own terms.

crawl-ref/docs/options_guide.txt
nlavsky: `show_unidentified_consumables` feels descriptive enough, but if it should be renamed to `consumables_panel_show_unidentified_items`, please  let me know.

crawl-ref/source/webserver/game_data/static/consumables.js
nlavsky: Initially I wanted to use a vector of `text_pattern`s for the filter, so I could verify regexps via `text_pattern:valid()`. But it turns out, the game does no such checking for other similar options.
(i) The same regex parser would guarantee the same syntax, but there doesn't seem to be a big difference between crawl and JavaScript regexps. The same basic patterns work fine on both sides:

```
consumables_panel_filter +=  scroll.*(enchant|brand) weapon
autopickup_exceptions    ^= >scroll.*(enchant|brand) weapon
```

(ii) I played a couple of games with the panel on a laptop, which needs half an hour to compile the game, and noticed no slowdown. The worst case would be matching 52 item names against a few dozen regexps each time the player uses or picks up an item. Compared with matching each new game message against hundreds of `force_more_message`s, this shouldn't be too bad. Browsers are slow but not that slow!

(iii) There is a benefit of having filters on the client side. Several existing Web Tiles options can be changed via `window.set_option()` without leaving the game: fonts, `tile_display_mode`, or even minimap colours. Four new options can be changed via the browser console too, which was useful during development.

To summarize, the current state of inventory filtering is good enough, but I agree that filtering should be moved to the crawl binary side. I'll look into that.

Implojin:
I'd like for this panel to be repositionable, and for its default location to not overlap the dungeon viewport.

With default scaling settings the player often has reason to look around with X or whatever; if this panel is positioned at top-left, it will block some of that (already limited) viewport space, which seems like a not-great UI choice.

Maybe you could position it at the bottom-middle, or attach it as a layer toggle to the chat log somehow? Blocking a small right-aligned portion of the combat log seems less intrusive than obscuring the dungeon viewport -- having this panel overlap webtiles chat (displaying the new messages indicator as part of this new panel, maybe?) seems less intrusive still.

nlavsky:
> With default scaling settings ...

This is why there's an option for that :). The player can rescale the panel so it's still readable and doesn't cover the map too much.

Suppose we want to display a panel with 20-25 item icons somewhere on the screen. What would be the best spot for it, suitable for as many different screen sizes as possible?

* The right side has the player info, minimap, monster list, and the chat. The monster list overflows when it has too many monsters. When the chat has several messages, they overlap the monster list and a part of the minimap. There is absolutely no free space for anything else.
* If the panel is between the dungeon viewport and the message area OR above the viewport, then we'll have to use smaller tiles for the viewport or show less lines of messages, or use a smaller font for them. Neither option looks like a good trade-off.
* If the panel is between the message area and the monster list, it'll block a part of the combat log. There are a lot of long, multiline game messages, so it doesn't look less intrusive either.
* The dungeon viewport has out-of-sight areas around the LOS square. If we cover them, the covered parts will still be visible on the minimap and in the X-mode. Now the choice is to either place the panel horizontally or vertically. There's probably more wide screens, so the panel is horizontal. If it's placed at the top, it'll be less likely to cover the LOS square (see the 1st commit message for details).

Later I plan to add an option for placing the panel vertically, along the left side of the viewport. This would make the panel more usable for smaller/vertical screens.

crawl-ref/source/webserver/game_data/static/consumables.js
nlavsky: Done. As a side benefit, there's no need to have (and keep in sync!) `base_type` and `misc_item_type` enums on the JavaScript side.

nlavsky:
@rawlins,
I moved the filtering to the crawl binary side and added an option for repositioning the panel. Is there something else I should be working on?
Also, have you tried the panel in action? I'm looking forward for your feedback!

nojaa3:
@nlavsky 
Could you make the numbers scale separately from the consumable panel's icons? I have relatively small screen and glasses and it is hard to read how many potions I have from full-screen example pictures. Good size for the font in my view is as big as the font of the message log.

nlavsky:
@nojaa3 ,

>as big as the font of the message log

The default font size should definitely be bigger.

Local tiles use this small font to display four-digit quantities, up to 9999. But it's an overkill for the consumables panel. Dropping one digit allows using a bigger and more readable font by default.

Upscaling the 8x14 pixel font from local tiles doesn't give good results, so it's better to have an option for specifying the font.

The same font as in local tiles:

1 0x font

New default font:

16px monospace

nojaa3:
Thank you, new default font is easy to see.

nlavsky:
The WebTiles UI doesn't change very often and it is familiar to a lot of players. That's why an introduction of a new UI element should be as non-disruptive as possible. The panel comes with a bunch of options, but it takes some work to discover them (who likes to read the docs?) and to configure these options.

This new menu gives the player more control over the UI and allows experimenting with the settings without leaving the game. The menu looks like this:
settings_menu

If there is anything else that should be done before the panel is release-ready, please let me know.

rawlins:
I pushed a commit to a branch that is a patch on top of (a rebased version of) this PR, and adds a bunch of features: https://github.com/crawl/crawl/commit/848c20388c4b

``` feat: consumables panel enhancements (clicking, information)

(This is not super well-tested, so I'm not really sure if it is safe to
merge just yet.)

This commit adds the following features to nlavsky's original
consumables panel patch:

* click to show/hide; the leftmost icon when visible is always a hide
  icon.
* right click on the hide icon (only) to get the settings popup
* right click closes the settings popup also. (This should maybe close
  if you click out of this ui, but I wasn't quite sure how to do that.)
* tooltip on consumables gives the item name and inventory letter
* right click on consumables (when in mouse command mode) opens the
  describe popup
* left click (in mouse command mode) just directly triggers the item.
  This is implemented via the actions interface, so doesn't need to add
  much code (as it rests on 6e4aca1). Does this make misclicks too
  easy?

I'm undecided whether this should be minimized or shown by default. It
would be really nice if minimization state at least were stored either
as a cookie or in the save between sessions.

Motivation: The patch as it stands is feature-complete relative to the
original goals (including ones I encouraged), but after testing I
unfortunately don't think it will feel very complete to end users. The
UI has the form of something that should be clickable, so players will
expect it to be clickable. (The tiles even look kind of like local tiles
menus.) Also, players not familiar with the icons will have a lot of
trouble with this: even as a long-time console player I don't know the
icons well. Hence the tooltip.```

Unfortunately I did not make the feature freeze deadline with this work, so I'm not sure if it will make it for the upcoming release, very sorry about that. If it doesn't, it'll be merged right after the tournament.

nlavsky:
@rawlins ,
huge thanks for working on the PR and adding these new features. The clicking and item tooltips felt quite useful during testing. I agree that it is better to delay the merge until after the tournament, though.

I noticed a few problems during testing:

* The hide icon is drawn on top of the first consumable, hiding it underneath.
* It's not possible to use unidentified potions/scrolls by clicking them. Right clicking and quaffing/reading from the item description menu works fine, though.
* The cursor's outline doesn't scale if you resize the panel. This is because `scale` should be the 5th parameter of `renderer.draw_icon()` (consumables.js:287 and 300)
* The tooltips work in Firefox, but not in Chrome. There are a bunch of "Uncaught ReferenceError: tooltip is not defined" errors because of a missing `var` (consumables.js:73).

There's one more thing, which is not a bug, but something that might confuse players. On local tiles, clicking on a wand in the inventory activates its targeter. The targeter appears if you click the rod of lightning on the panel too. But clicking on a wand on the panel immediately fires it instead. This is similar to using wands from the quiver, so I'm not sure what is the correct behavior here.

nlavsky:
I rebased this PR to get the commits required for the click-to-use feature and cherry-picked advil's enhancements from the nlavsky-consumables-panel branch. Also, I added a commit to fix all the issues I found during my testing.

As always, if there is anything left to add to the panel or if there is any feedback, please let me know.

rawlins:
Thanks again for your work on this @nlavsky, I have committed these changes as: https://github.com/crawl/crawl/compare/4b78bf2b85e9...57bb0be0ea2a! Happy to take any further fixes or adjustments you might want to make to this, I think now that the initial merge is over it'll become easier to work on incrementally. Some ideas that have come up:

* I renamed it because I want to eventually allow spells/abilities in this menu somehow. No concrete ideas yet.
* PleasingFungus wants to add a main menu button here, which seems good to me.
* Maybe other commands?
* Maybe an optional setting that automaps the first 10 things in this menu to F1-F10?

Anyways, that's all for the future..I'll go ahead and close up this PR.

PR #1874, opened 2021-04-05
Just one ecumenical altar

dillytin:
Just one ecumenical altar I made.

Please correct it if there's something wrong. It's an experimental first step for me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@
 
 *.sublime-workspace
 *.sublime-project
+crawl-ref/settings/init.txt
+*.txt
+crawl-ref/settings/init.txt
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,3 @@
 
 *.sublime-workspace
 *.sublime-project
-crawl-ref/settings/init.txt
-*.txt
-crawl-ref/settings/init.txt
--- a/crawl-ref/settings/init.txt
+++ b/crawl-ref/settings/init.txt
@@ -85,3 +85,6 @@
 # include                 = 0.12_monster_glyphs.txt
 # include                 = 0.13_monster_glyphs.txt
 # include                 = 0.14_monster_glyphs.txt
+
+
+tile_full_screen = false
--- a/crawl-ref/settings/init.txt
+++ b/crawl-ref/settings/init.txt
@@ -85,6 +85,3 @@
 # include                 = 0.12_monster_glyphs.txt
 # include                 = 0.13_monster_glyphs.txt
 # include                 = 0.14_monster_glyphs.txt
-
-
-tile_full_screen = false
--- a/crawl-ref/source/dat/des/altar/ecumenical.des
+++ b/crawl-ref/source/dat/des/altar/ecumenical.des
@@ -425,8 +425,7 @@ TAGS:   no_monster_gen transparent
 MONS:   plant, gnoll
 SUBST: z = t111
 KITEM: E = never_decay human skeleton / gnoll skeleton / elf skeleton / dwarf skeleton
-: dgn.delayed_decay(_G, "e","hoblin skeleton / hobgoblin skeleton / kobold skeleton")
-FTILE: 1et'_  = floor_grass
+FTILE: 12t'_  = floor_grass
 : ecumenical_altar_setup(_G)
 MAP
       .....
--- a/crawl-ref/source/dat/des/altar/ecumenical.des
+++ b/crawl-ref/source/dat/des/altar/ecumenical.des
@@ -438,10 +438,9 @@ MAP
     ttz'zztt     
    ttzzz'zztt    
   ttz'''''''tt   
-  tt'xxx+xxx'tt  
-  tt'x.....x'tt  
-  ttex..G..xeet  
- tteex..E..xett  
-  ttex.x.x.xtt   
-   ttxxx+xxxtt   
+   t'xxx+xxx't  
+   ttx.....xtt  
+   tEm..G..mEt  
+   ttx.....xtt  
+     xxt+txx   
 ENDMAP
--- a/crawl-ref/source/dat/des/altar/ecumenical.des
+++ b/crawl-ref/source/dat/des/altar/ecumenical.des
@@ -419,3 +419,29 @@ x-----x
 x-----x
 xxb+bxx
 ENDMAP
+
+NAME:   dilly_ecumenical_altar_deadguardian
+TAGS:   no_monster_gen transparent
+MONS:   plant
+SUBST: z = t11
+KITEM: E = never_decay human skeleton / gnoll skeleton / elf skeleton / dwarf skeleton
+: dgn.delayed_decay(_G, "e","hoblin skeleton / hobgoblin skeleton / kobold skeleton")
+FTILE: 1et'_  = floor_grass
+: ecumenical_altar_setup(_G)
+MAP
+         t      
+        ttt     
+       tt_t     
+      tt'tt     
+     tt'ztt     
+     ttz'ttt    
+    ttz'zztt     
+   ttzzz'zztt    
+  ttz'''''''tt   
+  tt'xxx+xxx'tt  
+  tt'x.....x'tt  
+  ttex..G..xeet  
+ tteex..E..xett  
+  ttex.x.x.xtt   
+   ttxxx+xxxtt   
+ENDMAP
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -4,3 +4,6 @@
 
 *.sublime-workspace
 *.sublime-project
+crawl-ref/settings/init.txt
+*.txt
+crawl-ref/settings/init.txt
```
ebering, line 9: You shouldn't be editing the .gitignore file

crawl-ref/settings/init.txt
```diff
@@ -85,3 +85,6 @@
 # include                 = 0.12_monster_glyphs.txt
 # include                 = 0.13_monster_glyphs.txt
 # include                 = 0.14_monster_glyphs.txt
+
+
+tile_full_screen = false
```
ebering, line 90: To specify local options instead of editing this init.txt (which is the default one and shouldn't have changes committed), either edit the usual one for a local install or run crawl with:

`./crawl -extra-opt-last "tile_full_screen = false"`

when you start it from the command line.

crawl-ref/source/dat/des/altar/ecumenical.des
```diff
@@ -419,3 +419,29 @@ x-----x
...
+NAME:   dilly_ecumenical_altar_deadguardian
+TAGS:   no_monster_gen transparent
+MONS:   plant
+SUBST: z = t11
+KITEM: E = never_decay human skeleton / gnoll skeleton / elf skeleton / dwarf skeleton
+: dgn.delayed_decay(_G, "e","hoblin skeleton / hobgoblin skeleton / kobold skeleton")
+FTILE: 1et'_  = floor_grass
+: ecumenical_altar_setup(_G)
+MAP
+         t
```
ebering, line 432:
```suggestion
```

crawl-ref/settings/init.txt
dillytin: what... I thought I manually unchecked those files. thx!

crawl-ref/source/dat/des/altar/ecumenical.des
```diff
@@ -419,3 +419,29 @@ x-----x
 x-----x
 xxb+bxx
 ENDMAP
+
+NAME:   dilly_ecumenical_altar_deadguardian
+TAGS:   no_monster_gen transparent
+MONS:   plant
+SUBST: z = t11
+KITEM: E = never_decay human skeleton / gnoll skeleton / elf skeleton / dwarf skeleton
```
ebering, line 427: `never_decay` should be used very sparingly and with some kind of cue to the player that the corpse is special. I've suggested one take on this but you might have different ideas

crawl-ref/source/dat/des/altar/ecumenical.des
```diff
@@ -419,3 +419,29 @@ x-----x
...
+: dgn.delayed_decay(_G, "e","hoblin skeleton / hobgoblin skeleton / kobold skeleton")
```
ebering, line 428: Not sure what purpose these corpses have in the vault's storyline?

crawl-ref/source/dat/des/altar/ecumenical.des
```diff
@@ -419,3 +419,29 @@ x-----x
...
+     ttz'ttt    
+    ttz'zztt     
+   ttzzz'zztt    
+  ttz'''''''tt   
+  tt'xxx+xxx'tt  
+  tt'x.....x'tt  
+  ttex..G..xeet  
+ tteex..E..xett  
+  ttex.x.x.xtt   
+   ttxxx+xxxtt
```
ebering, line 446: Here the never_decay corpses are entombed, more like a shrine. One way to do it might be this:

```suggestion
   t'xxx+xxx't  
   ttx.....xtt  
   tEm..G..mEt  
   ttx.....xtt  
     xxt+txx   
```

crawl-ref/source/dat/des/altar/ecumenical.des
```diff
@@ -419,3 +419,29 @@ x-----x
...
+: dgn.delayed_decay(_G, "e","hoblin skeleton / hobgoblin skeleton / kobold skeleton")
+FTILE: 1et'_  = floor_grass
+: ecumenical_altar_setup(_G)
+MAP
+         t      
+        ttt     
+       tt_t     
+      tt'tt     
+     tt'ztt     
+     ttz'ttt
```
ebering, line 437: I feel like the path to the altar could be a bit shorter, since the player already passes through the tomb section on the way.

ebering (changes requested):
A good first offering! I pointed out the git housekeeping issues as separate comments.

For the vault itself: `never_decay` is used very rarely and there's always a visual cue about why the corpse isn't decaying. I've suggested one way to do that in this vault but I'm open to other ideas you might have.

Two general changes:
- Trim the outsides of the vault, they'll often not be seen. If you want them to be seen, add border floor.
- Remove all of the spaces at the end of each line.

crawl-ref/source/dat/des/altar/ecumenical.des
dillytin: The intended storyline of those corpses were about the guardian, whose its corpses never decays somehow, of faded altar and its victims killed by the guardian due to trespassing the holy(?) place and abandoned, and decays. I think I should think of how to convey the story or just discard this idea.

crawl-ref/source/dat/des/altar/ecumenical.des
dillytin: Okay!

crawl-ref/source/dat/des/altar/ecumenical.des
```diff
@@ -419,3 +419,26 @@ x-----x
 x-----x
 xxb+bxx
 ENDMAP
+
+NAME:   dilly_ecumenical_altar_fadedgnrine
+TAGS:   no_monster_gen transparent
+MONS:   plant, gnoll; nothing
+SUBST: z = t111
+KITEM: E = never_decay gnoll shaman skeleton / never_decay gnoll skeleton
```
ebering, line 427:
```suggestion
KITEM: E = never_decay gnoll shaman skeleton / never_decay gnoll skeleton
KPROP: E = no_tele_into
```

This is a case where `no_tele_into` is vital, since it's a teleport closet.

crawl-ref/source/dat/des/altar/ecumenical.des
```diff
@@ -419,3 +419,26 @@ x-----x
 x-----x
 xxb+bxx
 ENDMAP
+
+NAME:   dilly_ecumenical_altar_fadedgnrine
+TAGS:   no_monster_gen transparent
+MONS:   plant, gnoll; nothing
```
ebering, line 425:
```suggestion
MONS:   plant
```

This vault is rather recognizable and placing a guaranteed gnoll early will lead to weird spoilers problems.

crawl-ref/source/dat/des/altar/ecumenical.des
```diff
@@ -419,3 +419,26 @@ x-----x
...
+SUBST: z = t111
+KITEM: E = never_decay gnoll shaman skeleton / never_decay gnoll skeleton
+FTILE: 12Et'_  = floor_grass
```
ebering, line 428:
```suggestion
FTILE: 1Et'_  = floor_grass
COLOUR: 1E'_ = green
```

Use the `COLOUR` directive to make sure console players also get a good experience.

crawl-ref/source/dat/des/altar/ecumenical.des
```diff
@@ -419,3 +419,26 @@ x-----x
...
+: ecumenical_altar_setup(_G)
+MAP
+      .....
+     ..ttt..
+    ..tt_tt..
+   ..ttt2ttt..
+  ..ttzz'zztt..
+  .ttzzz'zzztt. 
+  .ttzzz'zzztt.
```
ebering, line 437: This has a chance to create teleport closets with random rolls. To guarantee it doesn't cause problems, make sure every `z` is reachable, like so:

```suggestion
   ..ttt'ttt..
  ..ttzz'zztt..
  .ttz11'11ztt. 
  .ttz11'11ztt. 
```

ebering (approved):
I've pointed out a few things that I'm tweaking when I merge this so you can keep them in mind for future vaults!

Thanks for the submission!

dillytin:
Thanks for all your advice!

PR #1878, opened 2021-04-06
Reduce Makhleb's heal-on-kill

RojjaCebolla:
The recent introduction of Djinni draws further attention to an ongoing DCSS problem, where Makh is a good choice for any character, and is a better "healing" god than Elyvilon. Part of the problem is getting a bunch of free healing regardless of piety earned or invocations investment.

Instead of healing for (roughly) the monster's HD * 1.5, now heal for (roughly) invocation skill + monster's HD * 0.5. 
(`(random2avg(foo * 2) ~= foo`_, right?)_

This is intended to help with: 

1. the highly visible runaway Dj/DD games, 
2. and also with the less-talked-about problem where Makh is a really strong choice any character, not just Dj/DD though they are the poster children.
3. It also incentivizes training Invocations, which could help makh's active abilities become less of a sideshow.

### DIFF
--- a/crawl-ref/source/mon-death.cc
+++ b/crawl-ref/source/mon-death.cc
@@ -1870,7 +1870,7 @@ item_def* monster_die(monster& mons, killer_type killer,
 
                 if (have_passive(passive_t::restore_hp))
                 {
-                    hp_heal = random2avg(you.skill_rdiv(SK_INVOCATIONS, 2), 2)
+                    hp_heal = random2avg(you.skill_rdiv(SK_INVOCATIONS, 3, 2), 2)
                         + random2(mons.get_experience_level());
                 }
                 if (you.species == SP_GHOUL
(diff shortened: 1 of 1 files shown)


RojjaCebolla:
Looks like monster HD ranges from 1 to 15, with 16 for draconians and from 16-30 it's pretty much draconian packs, uniques, big liches, and orbs of fire.  So the player will heal less from those, unless they make a pretty big investment into invo skill.

Could conservatively swap it so that the flat heal comes from HD, and the more random part of the heal is from invo.  Or aggressively make both halves random and able to roll 0!

RojjaCebolla:
I did a bit more wizmode games, and this doesn't actually feel like a nerf - it only affects people who train little or no invo, which is probably not the most common use case?  

Trying a change to the skill multiplier.

PleasingFungus:
(I endorse the general thrust of this change, with some kind of accompanying buff to Major Destruction and/or swapping it out for damnation. I haven't the time to investigate the details right now, though, sorry.)

PR #1879, opened 2021-04-06
Improve Makhleb's Major Destruction

RojjaCebolla:
Throw out the fireball and orb, they cause problems and shouldn't
be using the same targeter as a bunch of bolts. Very Random spells
can be fun, but not when you're paying HP, piety, and a turn.

Add an acid bolt instead.

Not fixed: iron shot and sticky flame are also not bolts.

But at least those don't blow up in your face!

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -2557,12 +2557,11 @@ static spret _do_ability(const ability_def& abil, bool fail, dist *target)
             beam.origin_spell = SPELL_NO_SPELL; // let zapping reset this
             zap_type ztype =
                 random_choose(ZAP_BOLT_OF_FIRE,
-                              ZAP_FIREBALL,
                               ZAP_LIGHTNING_BOLT,
                               ZAP_STICKY_FLAME,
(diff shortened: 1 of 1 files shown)


RojjaCebolla:
```
int power = you.skill(SK_INVOCATIONS, 1)
                    + random2(1 + you.skill(SK_INVOCATIONS, 1))
                    + random2(1 + you.skill(SK_INVOCATIONS, 1));
```
This same power calculation seems to be used for both Minor and Major destruction.  It feels unlikely they should be using the same numbers.  In wizmode with an XL8 character that had 12 invo, a barrage of major destruction casts kills the player - and also fails to wipe out monsters in Lair!  Perhaps another random2 of invo skill should be added?

Another direction could be to keep the fireball and orb, and add in other "splash" spells and drop the bolts.  But I don't think splash spells are good with summons.

sdynet:
How about giving him Damnation?

RojjaCebolla:
That's a very thematically appropriate idea, and it's great that all your summons would be immune to the blast.  But that's one of Demonspawn's coolest gimmicks! 🤔

PR #1880, opened 2021-04-07
Triple the number of level 2 spell gifts that Veh can choose from

RojjaCebolla:
When comparing the [array of gifts](https://gist.github.com/RojjaCebolla/86d9d532445426194319f912e5109000) chosen upon levelup for Veh and Djinn characters, I noticed two things: Dj has a tighter bound, guaranteeing at least one spell of each level. Veh has "clumpier" preferences which may have made sense in the past, but right now there are too few level 2 and level 7 options to support that table.

This PR aims to help with the situation by increasing the giftable options at level 2 from one spell (Searing Ray), to three spells (adding Poisonous Vapours and Dazzling PBAOE).  

Not addressed:
There just aren't many level 2 spells that deal damage, probably because many boring ones have been removed.  This is a bit of a hole that some creative person could fill!

### DIFF
--- a/crawl-ref/source/book-data.h
+++ b/crawl-ref/source/book-data.h
@@ -163,9 +163,9 @@ static const vector<spell_type> spellbook_templates[] =
 {   // Book of Maledictions
     SPELL_CORONA,
     SPELL_HIBERNATION,
+    SPELL_DAZZLING_FLASH,
     SPELL_CONFUSING_TOUCH,
     SPELL_TUKIMAS_DANCE,
-    SPELL_DAZZLING_FLASH,
 },
 
 {   // Book of Air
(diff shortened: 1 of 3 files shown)


semitonal:
I agree that Dazzling Flash might be worth trying at level 2, but do I wonder if it should be excluded from Vehumet's spell list anyway since it doesn't deal any damage.

ebering:
Not as convinced Flash would be ok at level 2; it's a pretty strong disable and bypasses Willpower. Perhaps in the Conjurer spell book it seems too high, but for an Enchanter it's part of an "easy to retry stabs" toolkit. For other characters being level 2 would make it quite cheap as an escape (cheaper than it already is), though maybe a lack of spellpower makes that ok.

Merging the other part.

Issue #1800, opened 2021-03-07
Scarf of shadows reduces length of lighting bolts/shocks reflected at walls


AlexanderPosch:
**PLATFORM and OS or BROWSER:**
webtiles

**DCSS VERSION AND GAME INFORMATION:**
0.27-a0-498-g7ecbb0a

**DESCRIPTION:**
Scarf of shadows reduces the length of lighting bolts/shocks reflected at walls. This does not necessarily seem intentional.

**EXTRA FILES:**
Image of shock without scarf of shadows:
![image](https://user-images.githubusercontent.com/49498206/110225507-665ed180-7ee6-11eb-8ea7-9fe56fa2ab18.png)

Image of shock with scarf of shadows:
![image](https://user-images.githubusercontent.com/49498206/110225521-855d6380-7ee6-11eb-8aec-f81a1faeffa0.png)

AlexanderPosch:
In related news Kobold's reduced los also results in shorter reflected beems (this might be intentional but idk)

PR #1885, opened 2021-04-08
Issue #1800: Make Scarf of shadows not affect the lenght of bounced spells. Also slight buff to player Kobolds, their spells bounce back one space further.

AlexanderPosch:
This pull request addresses Issue #1800:

Los restrictions (like the Scarf of shadows) reduce the length of lighting bolts/shocks reflected at walls.
image

The code checks how long a bounced spell would have been if los had not been restricted and then increases the range of bounced the spell.  (Kobolds nightstalker mutation does not count as a los restriction)

This change does not allow player to affect tiles outside of the los-radius (unless you wear a scarf of shadows and the "Robe of Night" together then it is possible).

Additionally I also gave a slight buff to player kobolds, their spells bounce back one further. Since for kobolds all reflected spells(Shock,Lighting_Bolt,Fire_Ball,Starburst) have range 4 and this results in this:
image
rather than this:
image
Also this change does not allow players to affect tiles outside of the los-radius (unless you wear a scarf of shadows and the "Robe of Night" together then it is possible).

I think it looks nicer and more symmetrical this way and I don't think it is very unbalancing.

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -843,7 +843,7 @@ void bolt::bounce()
     bolt boltcopy = *this;
     if (bounces == 1)
         {
-        extra_range_used -= spell_range(boltcopy.origin_spell, boltcopy.ench_power,true,true)-range;
+        extra_range_used -= spell_range(boltcopy.origin_spell, boltcopy.ench_power,true,(int)you.normal_vision-you.get_mutation_level(MUT_NIGHTSTALKER))-range;
 
         if  (you.get_mutation_level(MUT_NIGHTSTALKER)==3)
         {
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -842,19 +842,8 @@ void bolt::bounce()
     //Also small buff to kobolds making it easier for them to multizap opponents
     bolt boltcopy = *this;
     if (bounces == 1)
-        {
         extra_range_used -= spell_range(boltcopy.origin_spell, boltcopy.ench_power,true,(int)you.normal_vision-you.get_mutation_level(MUT_NIGHTSTALKER))-range;
-
-        if  (you.get_mutation_level(MUT_NIGHTSTALKER)==3)
-        {
-                extra_range_used -= 1;//Makes it easier for kobolds to multizap opponents, (first zap is one cheaper),
-                                      //Spells with range>=los now get reflected back an even number of tiles this also looks cleaner
-        }
-
-        //Prevent bounced bolts from going outside of a radius of you.current_vision
-        extra_range_used += max(0.0,(ray.pos().x+ray.r.dir.x*(range-extra_range_used)-you.pos().x)-you.current_vision);
-        extra_range_used += max(0.0,(ray.pos().y+ray.r.dir.y*(range-extra_range_used)-you.pos().y)-you.current_vision);
-        }
+    
     ASSERT(!cell_is_solid(ray.pos()));
 }
 
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -850,8 +850,11 @@ void bolt::bounce()
                 extra_range_used -= 1;//Makes it easier for kobolds to multizap opponents, (first zap is one cheaper),
                                       //Spells with range>=los now get reflected back an even number of tiles this also looks cleaner
         }
-        }
 
+        //Prevent bounced bolts from going outside of a radius of you.current_vision
+        extra_range_used += max(0.0,(ray.pos().x+ray.r.dir.x*(range-extra_range_used)-you.pos().x)-you.current_vision);
+        extra_range_used += max(0.0,(ray.pos().y+ray.r.dir.y*(range-extra_range_used)-you.pos().y)-you.current_vision);
+        }
     ASSERT(!cell_is_solid(ray.pos()));
 }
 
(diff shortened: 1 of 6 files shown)


crawl-ref/source/beam.cc
```diff
@@ -837,7 +837,19 @@ void bolt::bounce()
...
-
+ 
+    //Keep lenght of bounced Spells constant despite reduced los (scarf of shadows,Robe of Night)
+    //Also small buff to Kolbolds making it easier for them to multizap opponents
+    bolt boltcopy = *this;
+    if (bounces == 1)
+    	{
+    	extra_range_used -= spell_range(boltcopy.origin_spell, boltcopy.ench_power,true,true)-range;
+    	
+    	if  (you.species==SP_KOBOLD)
```
ebering, line 848: this should check for the nightstalker mutation instead of the species directly.

crawl-ref/source/beam.cc
```diff
@@ -837,7 +837,19 @@ void bolt::bounce()
         rg(*ai - ray.pos()) = cell_is_solid(*ai);
     ray.bounce(rg);
     extra_range_used += 2;
-
+ 
+    //Keep lenght of bounced Spells constant despite reduced los (scarf of shadows,Robe of Night)
```
ebering, line 841: typo

crawl-ref/source/spl-util.cc
```diff
@@ -985,10 +985,19 @@ int spell_power_cap(spell_type spell)
...
 {
     int minrange = _seekspell(spell)->min_range;
     int maxrange = _seekspell(spell)->max_range;
+    int los_cap  = (int)you.current_vision;
+    if (ignore_los)
+    	{ 
+    	if (you.species!=SP_KOBOLD)
+    		los_cap  = (int)you.normal_vision;
+    	else
+    		los_cap  = (int)you.normal_vision-3;
```
ebering, line 998: This should check the nightstalker mutation directly and use its calculation.

ebering (changes requested):

Thank you for this. In addition to the in-line changes, this needs to prevent out-of-los attacks when using shadows and Night together. Rare though it may be, we have spent quite a lot of effort moving away from out of los attacks in most cases.

AlexanderPosch:
Thanks for your quick reply and comments.  I'll try to fix the remaining issues.

AlexanderPosch:
> ...this needs to prevent out-of-los attacks when using shadows and Night together. Rare though it may be, we have spent quite a lot of effort moving away from out of los attacks in most cases.

Shock and Lighting Bolt can already attack creatures outside of los by bouncing around a corner:
image
Wearing the Robe and scarf would allow attacking like this:
image


Other spells that can attack creatures outside los are:
- Kills outside los_radius: Gell's Gravitas, Fulminant Prism (in this case it makes sense), Fire Storm (needs luck or los-1), Fire Ball (needs los-2)
- Kills enemy around a corner: Spark, Lightning Bolt, LRD, Fire Storm, Fire Ball

Should I aim to only remove the new possibility for out-of-los attacks or all out-of-los attacks using bouncing? 
I think it might be easier if removing out-of-los spell-attacks were done independently of the bouncing mechanic for all spells by checking if affected targets are within los (or los_radius) of the player.

AlexanderPosch:
The newest change to the code prevents the out-of-los attacks that wearing the "Robe of Night" and a "Scarf of Shadows" allowed. Bolts going around corners can still hit enemies but this was already possible before this pull request.  
If you want _(and I figure out how multiple vectors with scalars)_ I should not be to hard to change the code to also prevent these "around corners" shots.

AlexanderPosch:
What should the range in the spell description show? The "normal range" or the "range restricted to los" or a mixture (like this @--->~~~...)?

AlexanderPosch:
### Summary of all changes:

- Exchanged kobold species checks for nightstalker mutation checks. Fixed typos
- Prevented bounced bolts from going outside of a radius of you.current_vision. This prevents the the out-of-los attacks that wearing the "Robe of Night" and a "Scarf of Shadows" allowed.
- Changed the spell_range_string, range_string so that the actual range of a spell can be seen the in spell description

ebering:
> I think it might be easier if removing out-of-los spell-attacks were done independently of the bouncing mechanic for all spells by checking if affected targets are within los (or los_radius) of the player.

This would be the much better solution. It's far enough out of scope for this PR that I will hunt it down separately when I do the merge. For now you can remove the special case you added (Sorry about sending you in a circle on this!).

I've had a rough weekend, and less crawl time than I had hoped so this isn't a complete review, just dropping a note to let you know where things are. I hope to get to one soon.

@semitonal mentioned that the Kobold special case for bounce range wasn't something she was interested in and I agree. While the geometric symmetry is pleasing, the preferred crawl style is mechanical symmetry.

AlexanderPosch:
Thanks for the update :). It makes sense that mechanical symmetry should come first.
An idea for keeping mechanical symmetry while also implementing some geometric symmetry might be to just completely ignore the night-stalker mutation in the bouncing mechanic. But I can see why this might also be undesirable.

ebering:
Thanks for the patch! I ended up tweaking things quite a bit, you can see the final result in the commit referencing this PR.

AlexanderPosch:
Thank you, It does seem a lot cleaner after your tweaking :) .
What ever is typical (Alexander Posch is fine). But since my contribution was rather little I would also be ok with not appearing.

PR #1890, opened 2021-04-09
Make the Wrath of Trog worthwhile

mainiacjoe:
Wrath of Trog is very thematic with ANGRY: 50, but suicidally dangerous if you cannot manage the *Rage.  This is 5.55x more frequent than Berserkitis 3.  Other unrands with *Rage, besides have ANGRY of only 5, also have beneficial properties such as super-vampiricism, Slay+6, and berserking allies.  Wrath of Trog, though, is simply a +8 antimagic battleaxe.  This hardly compensates for berserking the wielder 10x as often.

This PR brings Wrath of Trog's ANGRY down to 5 and adds beneficial properties to it, to match the other *Rage unrands.  In keeping of the theme with this being Trog's favorite weapon, instead of giving it Troggish properties such as Regen+, Will++, or +Rage--things Trog wouldn't need himself from a weapon--just make it very powerful weapon with no properties other than the *Rage.  Trog's favorite weapon ought, after all, to be as good or better as the best gifts he used to give his followers!

This is accomplished by adding BASE_DAM: +3 and BASE_DELAY: -4 to its entry in art-data.txt, giving it the base damage of an executioner's axe with the base delay of a dire flail.  The enchantment is not changed.  Attached to the commit itself are five graphs comparing the new axe at skills 14, 20 and 26 to a +9 battleaxe of speed and to a +9 anti-magic executioner's axe, plus the fsim.csv.  The higher the target AC and EV, the better Wrath of Trog's relative performance.

The fsim settings were fsim_scale = fighting and fsim_rounds = 10000.  Wrath of Trog was changed in atr-data.txt to comment out the ANGRY line for fsim testing and to add BASE_DELAY = -4 and BASE_DAM = +3, then recompiling just in case.  Player stats were Strength 30 and Dex 20, wearing plate mail with Armour skill 20.  Target monsters were fungus (AC/EV of 0/0), tentacled monstrosity (5/5), wraith (10/10), orb guardian (13/13) and orb of fire (20/20).

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -204,9 +204,8 @@ COLOUR:  ETC_BLOOD
 TILE:    spwpn_wrath_of_trog
 TILE_EQ: axe_trog
 VALUE:   800
-ANGRY:   5
-BASE_DELAY: -4
-BASE_DAM: +3
+ANGRY:   50
+BOOL:    rampage
 BRAND:   SPWPN_ANTIMAGIC
 
 NAME:    mace of Variability
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -204,7 +204,9 @@ COLOUR:  ETC_BLOOD
 TILE:    spwpn_wrath_of_trog
 TILE_EQ: axe_trog
 VALUE:   800
-ANGRY:   50
+ANGRY:   5
+BASE_DELAY: -4
+BASE_DAM: +3
 BRAND:   SPWPN_ANTIMAGIC
 
 NAME:    mace of Variability
--- a/crawl-ref/source/dat/descript/unrand.txt
+++ b/crawl-ref/source/dat/descript/unrand.txt
@@ -16,10 +16,9 @@ Screaming Sword
 %%%%
 Wrath of Trog
 
-A bloodstained axe, forged by Trog himself to have exquisite sharpness and
-balance.  It was his favourite weapon, but was lost many centuries ago.
-It can induce a bloodthirsty berserker rage in warriors who use it in
-battle.
+A bloodstained axe that was the favourite weapon of the old god Trog, before it
+was lost. It induces a bloodthirsty berserker rage in warriors who use it in
+battle and enables them to dash towards their enemies.
 %%%%
 mace of Variability
 
--- a/crawl-ref/source/dat/descript/unrand.txt
+++ b/crawl-ref/source/dat/descript/unrand.txt
@@ -16,8 +16,9 @@ Screaming Sword
 %%%%
 Wrath of Trog
 
-A bloodstained axe that was the favourite weapon of the old god Trog, before it
-was lost. It induces a bloodthirsty berserker rage in warriors who use it in
+A bloodstained axe, forged by Trog himself to have exquisite sharpness and
+balance.  It was his favourite weapon, but was lost many centuries ago.
+It can induce a bloodthirsty berserker rage in warriors who use it in
 battle.
 %%%%
 mace of Variability
--- a/crawl-ref/source/mon-death.cc
+++ b/crawl-ref/source/mon-death.cc
@@ -1619,6 +1619,15 @@ item_def* monster_die(monster& mons, killer_type killer,
             you.increase_duration(DUR_BERSERK, bonus);
             mpr("The necklace of Bloodlust glows a violent red.");
         }
+        else if (player_equip_unrand(UNRAND_TROG))
+        {
+            const int bonus = (2 + random2(4)) / 2;
+            you.increase_duration(DUR_BERSERK, bonus);
+            if (!random2(4))
+            {
+                mpr("Blood oozes out of your axe's blade.");
+            }
+        }
     }
 
     if (you.prev_targ == monster_killed)
--- a/crawl-ref/source/util/art-data.pl
+++ b/crawl-ref/source/util/art-data.pl
@@ -57,6 +57,7 @@ my %field_type = (
     NOTELEP  => "bool",
     NO_UPGRADE => "bool",
     POISON   => "bool",
+    RAMPAGE  => "bool",
     RANDAPP  => "bool",
     RCORR    => "bool",
     REGEN    => "num",
@@ -537,7 +538,7 @@ my @art_order = (
     "BASE_DELAY", "HP", "CLARITY", "BASE_ACC", "BASE_DAM", "\n",
     "RMSL", "unused", "REGEN", "unused", "NO_UPGRADE", "RCORR", "\n",
     "RMUT", "unused", "CORRODE", "DRAIN", "SLOW", "FRAGILE", "\n",
-    "SH", "HARM", "\n",
+    "SH", "HARM", "RAMPAGE", "\n",
     "}",
 # end TAG_MAJOR_VERSION
 # start TAG_MAJOR_VERSION == 35


semitonal:
I feel like the fundamental idea behind the Wrath of Trog is that you're expecting to berserk in every fight when you use it - if anything I'd say that's a better use of *Rage than the normal 5% chance for it to trigger (since even at a 5% chance, you still have to assume that you might get berserked at any inopportune moment). I'd personally much rather see the design go in a direction that keeps the high chance to berserk and compensates for it in some other way (by giving some kind of protection from the post-berserk downsides, and/or making it easier to maintain the berserk state without it timing out, etc?)

hellmonk3:
Something else to note is that the current version is highly abusable as a swap weapon when you *do* want to go berserk. So it should probably have a swap penalty or restriction if it's going to lean into that flavor, in addition to whatever new upsides.

sdynet:
Personally, I think it would be better if we could put Rampage effect and Death' door(Only in berserk stat) on this weapon. :)

mainiacjoe:
I'm fine with leaning in on the *Rage instead of mitigating it.  Certainly my changes were treating the weapon as if it were usually not berserked.  I had only seen semitonal's comment when I went to bed, and what come from that was similar to sdynet's suggestion: Trog's Hand that operates only when Berserk--perhaps boosted to ambrosia-speed regen--and ^Trog bonuses to duration and aftereffects of berserking.

mainiacjoe:
In retrospect, my purpose for the PR was that WoT become something you would want as your main weapon.  Hellmonk's observation is excellent, but a un-wield penalty that isn't accompanied by a boost to the weapon itself would likely turn into never-wield.

mainiacjoe:
As I thought about the unwield penalty, I realized that it needs to be an **_on-wield_** penalty that has a tactical cost but not a strategic cost--something you can rest off, but what you wouldn't want to get in the middle of a fight.  I misunderstood what Hellmonk meant,

mainiacjoe:
Coding related questions #1: This PR failed a linting test from bad whitespace.  Is there a conventions document I could get a link to?  I typically intend to un-Tabify, but I might have forgotten this time.

mainiacjoe:
Coding-related question #2: I know about art-func.h.  Which files handle berserking?

mainiacjoe:
Would it be enough to simply trigger end-of-berserk effects on both wielding and unwielding?  Slow, -Berserk, and a chance of passing out.  You'd not want to switch to it during a battle, you'd need to decide before combat begins and really before exploring.  But if you wield a different weapon while resting in a safe place, no harm done.

rawlins:
> Coding related questions #1: This PR failed a linting test from bad whitespace. Is there a conventions document I could get a link to? I typically intend to un-Tabify, but I might have forgotten this time.

There are scripts in util that you can run to fix lint issues (`util/checkwhite` and `util/unbrace`), should work in any unix environment. Sometimes you need `util/checkwhite -a` to fix non-code files.

> Coding-related question #2: I know about art-func.h. Which files handle berserking?

It's in `player::go_berserk` and `player::can_go_berserk`, as well as code that checks for `ENCH_BERSERK` among various other places. I'd recommend learning to use command line `git grep` for queries like this, i.e. I didn't actually know the answer either but looked at the output of `git grep berserk *.cc` (and `git grep BERSERK *.cc`) to answer this question.

ebering:
> Would it be enough to simply trigger end-of-berserk effects on both wielding and unwielding? Slow, -Berserk, and a chance of passing out.

Sounds both a good tactical disincentive and very flavorful.

In terms of giving the weapon something to incentivize and lean into the *Rage, I suggest adding Rampage! Rampaging one-tile for an attack doesn't incur the berserk movement penalty, so this will help the player keep attacking once Berserked. If this isn't enough, don't apply the berserk penalty to any rampaged movement (even if both moves are moves, and not move followed by attack). Adding Regen and Will++ (all the time, since the weapon has swap penalties) could also help with the incentives. 

@sdynet went as far as to suggest Rampage∞: allowing the player to rampage all the way to a target in-line. This might be a technical headache to implement; I would hold-off implementing that unless the other things have been tried and still come up wanting.

AlexanderPosch:
I know this is somewhat silly but I think it might be fun if players who wield the axe incur the real God Wrath of Trog. Thematically it makes sense: Trog lost it's favourite axe and is now sending it's minions to take it back.

ebering:
> God Wrath of Trog

It does have strong theme, but god wrath is meant to be potentially lethal (even with a single application), which would fail the "good or interesting" goal for unrands (for elaboration [see the long-running unrand design wiki page](https://github.com/crawl/crawl/wiki/Unrand-Review) ).

mainiacjoe:
Adding in post-berserk effects on wield and unwield is something I think I can learn to do. I'm on mobile at the moment but I'm guessing +Rampage is a BOOL in art-data.txt and that sounds great.  As for the Will++ and Regen+, I wonder whether it's even worth making it active only while berserk, given the wield/unwield penalty disincentive to using it as a resist stick.

ebering:
> As for the Will++ and Regen+, I wonder whether it's even worth making it active only while berserk

I think it is not, and also that these props should only be added if the no-movement penalty rampage doesn't seem strong enough.

mainiacjoe:
Thematically, I've wondered whether re-theming it as something whereby the wielder is psychically connected to the actual Trog might explain things better. ^Trog gets the berserk extending bonus and end-of-berserk mitigation, and non-worshippers don't get those bonuses. As for the wield/unwield penalty, something similar to, "Your mind reels as you feel Trog's power surge into you." And "you struggle to regain full control of your limbs as Trog releases them."

mainiacjoe:
I think that Rampage for an arbitrary number of tiles is a poor fit here, but there are few unrand boots and it could work there. Not to mention that I don't know C yet and something that would be a headache to implement would make a lousy first project.

mainiacjoe:
The header of art-data.txt doesn't list rampage as a bool. Was the header not updated, or is rampage not enabled for unrands yet?

semitonal:
> The header of art-data.txt doesn't list rampage as a bool. Was the header not updated, or is rampage not enabled for unrands yet?

There's no handling for rampage on unrands yet, so it'd need to be added as a new bool field. The handling for that is done by util/art-data.pl (which parses art-data.txt to generate all the unrands) - if you look at how eg HARM is handled there, it'd need to be added in the same way to the list of fields at the top of the file, and to the list of ARTPs under @art_order.

mainiacjoe:
I've successfully added rampage to Wrath of Trog and tested it locally after recompile, but I'm having git trouble again trying to push it.  This time, not only is mainiacjoe/wrathtrog behind mainiacjoe/master and I'm not able to update it (I'm trying to avoid -force), but local git on the wrathtrog branch says there's no changes to commit.

ebering:
If on your local branch "wrathtrog" you used the command `git rebase` then the only way to update the remote "maniacjoe/wrathtrog" on GitHub is with a force push command.

mainiacjoe:
Yes, I had used "git rebase master" while on local branch "wrathtrog"  I did this after pulling from crawl/crawl:master to local master, to try to bring branch wrathtrog up to date with trunk.  What command ought I have used instead?  Not knowing this about "git rebase" I see now has been a cause of many of my troubles.

ebering:
That's the correct command sequence. `git rebase master` re-writes the history of your branch to be as if you started working from the current commit referenced by `master`. This destroys information about the commit history. By design, `git push` won't propagate these destructive changes as a default, so you must specify force.

ebering:
This looks like a correct implementation of adding Rampage; but does not yet have the swap penalties, and still has issues with the description.

mainiacjoe:
Yes, I'm still figuring out the swap penalties.  Part of pushing the rampage commit was giving myself another opportunity to try git.

mainiacjoe:
I spent some time trying to figure out the on-wield / unwield penalties.  I figured out how the berserk duration is reduced and the warning messages, but haven't found yet where that is checked to decide whether berserk ends.

mainiacjoe:
In the midst of that I found the code for extending berserk, and added WoT alongside the Necklace of Bloodlust.  I'm concerned that if the on-wield and un-wield penalties are too severe, it will not only discourage swapping but also discourage use altogether.  I thought that improving the benefits of accepting the *Rage while not making it better for those who can circumvent it are worth doing.  As I write this I wonder whether heal-on-kill while berserk might work for this purpose as well.

PR #1893, opened 2021-04-10
Shadow Form description update

MinticePIU:
Shadow Form doesn't drain skills anymore!


semitonal:
Thanks for the PR, have updated the description with some further tweaks as well (it still referenced hostile enchantments instead of willpower).

PR #1894, opened 2021-04-10
add item.description in Lua

mgdelmonte:
Similar to the recently added `spells.describe` function, this adds the string `description` to the Lua `item` object.  The description is the same text one sees in the UI.  Helpful for players using scripts to evaluate relative item values for damage, noise, etc.

### DIFF
--- a/crawl-ref/source/l-item.cc
+++ b/crawl-ref/source/l-item.cc
@@ -13,6 +13,7 @@
 #include "cluautil.h"
 #include "colour.h"
 #include "coord.h"
+#include "describe.h"
 #include "env.h"
 #include "invent.h"
 #include "item-prop.h"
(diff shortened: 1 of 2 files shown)


ebering (changes requested):
Could you re-base this to current master and force-push?

I'm guessing based on your merge-messages that your git repository has crawl master configured as the remote `upstream'. If that's the case, you'll want

```
git rebase upstream/master
git push --force
```

Pushing will automatically update this PR

mgdelmonte:
rebased as requested

rawlins:
Thanks for the commit! Can you use branches for future PRs rather than mgdelmonte:master? This being a PR against your master branch is part of what led to some minor weirdness in trying to merge it and evaluating it in the github interface, rebasing was only a partial fix. (This is why github showed a bunch of extra commits that aren't part of the PR, even after the rebase.)

PR #1896, opened 2021-04-12
Transmutation form adjustments

hellmonk3:
Adjusts some of the transmutation form spells. Blade hands returns to level 5 and the TM start book but loses a small amount of base damage. Dragon form also loses a small amount of base damage. Ice form receives better AC scaling. Hydra form is removed and replaced with air walk, which is an extremely funny reference and hopefully a better spell (but may still need some numerical adjustment and tweaking).


semitonal:
I don't have especially strong opinions on Air Walk (other than it should probably have a different name, the reference is cute but something that's more descriptive would be better), but I wouldn't want to see Blade Hands back in the Tm starter book. The aim of the change in the first place was to move away from giving out starting spells that will end up being used all the way through the endgame, and the slight reduction in base damage doesn't address this. Other starting books are certainly guilty of the same thing, but it's a good direction to keep moving away from.

semitonal:
Closing, since this has been merged separately now.

PR #1897, opened 2021-04-12
Monster flags in Lua

mgdelmonte:
Adds a `flags` function to the Lua `monster_info` object, which returns the list of flags set for the monster.

This is considerably easier and less buggy to use than the `is` function, which requires one to know in advance the names of all the possible flags that can be used.  The flag list for monsters changes constantly as the game evolves.  Because the `is` function raises an error if it is used to check a flag that no longer exists, it is extremely hard to keep Lua helper scripts functional, if they use monster flags via the `is` function.

It's also nice to simply get all the active flags at once, rather than having to query each possible flag against the `is` function to see if it's set.


rawlins:
Following my comment in the other branch, would it be possible for you to pull the new commits out of this PR into a branch (e.g. using `git cherry-pick` with the commit hashes) and resubmit? That would make it easier for someone on our end to see what is going on here; github is showing a bunch of commits from previous PRs of yours that have also involved mgdelmonte:master. The github PR workflow is really intended to have a one branch per PR structure, so it is getting confused because of this.

(This would also allow keeping your master branch exactly aligned with the one in crawl, so would avoid getting tripped up by submodules etc.)

mgdelmonte:
closed and superseded by #1921

PR #1898, opened 2021-04-12
Make mid-lategame more compact.

hellmonk3:
This PR removes 2 floors of lair, 1 floor of sbranch, 1 floor of vaults, 1 floor of slime, and 2 floors of depths with no compensation for xp or items. It does not change extended because that involves a larger overhaul, but somebody should get on that :^)

Think of this as a conversation starter pull request; I don't really expect it to get merged as-is. It may also be worth chopping late dungeon somewhat, bcrawl style, but I think late D tends to work better than 6 floor lair. I support merging the manman xp curve pr as a follow up to this one if approved.

### DIFF
--- a/crawl-ref/source/branch-data.h
+++ b/crawl-ref/source/branch-data.h
@@ -57,7 +57,7 @@ const Branch branches[NUM_BRANCHES] =
       'K', {}, branch_noise::normal, DEFAULT_MON_DIE_SIZE },
 #endif
 
-    { BRANCH_LAIR, BRANCH_DUNGEON, 8, 11, 6, 10,
+    { BRANCH_LAIR, BRANCH_DUNGEON, 8, 11, 4, 10,
       brflag::none,
       DNGN_ENTER_LAIR, DNGN_EXIT_LAIR, NUM_FEATURES,
       "Lair", "the Lair of Beasts", "Lair",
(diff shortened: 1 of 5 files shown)


semitonal:
This would also probably need some adjustments to the mon-pick tables so that the bottom floor of each branch keeps a similar monster set to now, instead of cutting off the spawn tables early. It also looks like some vaults will need updating if they place any monsters by location (eg linesprint is causing compilation to fail due to trying to place Lair:5 monsters).

For a slightly more conservative approach without requiring such big follow-up changes, I'd been thinking of cutting one floor each from Lair, Slime, Vaults and Depths. I wouldn't necessarily be opposed to making more aggressive cuts and changing the XP curve at high levels too, but I'd definitely want to look at some objstat comparisons and get a better understanding of the expected results first.

ebering:
We went with a more conservative 1-floor depths cut with some other adjustments to the XP curve and item generation so far in 0.28.

In the current state I could see cutting a floor from Lair (maybe), but I would favor reducing to 1 S branch per game over two shorter S branches.

I might make a more radical conversation PR at some point in 0.29 on the topic.

semitonal:
I was definitely still interested in cutting a floor from Lair for 0.28 if I get around to it, but then yeah, would be interested in more substantial changes for future versions.

ebering:
Since reaching 0.28 the flattening/shortening conversation that this PR was a part of has petered out, so I'll put it in the history file.

PR #1901, opened 2021-04-13
Create artefact_creation.md

RojjaCebolla:
Basic documentation for artefact creation.  Can be expounded on later if needed.


ebering (changes requested):
Looks like a nice overview, but please wrap to 80 columns!

ebering (approved):

PR #1902, opened 2021-04-14
Simplify GDR

PleasingFungus:
Guaranteed damage reduction is intended to smooth out some of the
nastiness in Crawl combat. It guarantees that your AC will never
roll below some % of its max value when defending against melee attacks
and mundane projectiles (not against spells, etc).

This is a perfectly fine idea, but it's implemented bafflingly: it uses
your base AC OR a hodge-podge of special cases based on your species
and/or form, meaning that it's both spoilery and weird. Some AC is
effectively worth more than others. Not great!

Instead, simplify it to be based purely on the player's current AC.
(Specifically, AC^(1/4) * 18.)

Previous example values:

```
Robe:                 0%
Scale mail (ie Fi):  28%
Plate:               39%
```

New values:

```
AC 6 (ie Fi):        28%
AC 20:		     38%
AC 50:               47%
```

This is a buff to some things and a nerf to other things. It's all over
the place. Chaos reigns.

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -5937,36 +5937,13 @@ int player::armour_class_with_specific_items(vector<const item_def *> items) con
   * useful when the AC roll is inferior to it. Therefore a higher GDR means
   * more damage reduced, but also more often.
   *
-  * \f[ GDR = 14 \times (base\_AC - 2)^\frac{1}{2} \f]
+  * \f[ GDR = 16 \times (AC)^\frac{1}{2} \f]
   *
   * \return GDR as a percentage.
   **/
 int player::gdr_perc() const
 {
-    switch (form)
-    {
-    case transformation::dragon:
-        return 34; // base AC 8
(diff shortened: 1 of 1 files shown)


crawl-ref/source/player.cc
```diff
@@ -6009,36 +6009,13 @@ int player::armour_class_with_specific_items(vector<const item_def *> items) con
   * useful when the AC roll is inferior to it. Therefore a higher GDR means
   * more damage reduced, but also more often.
   *
-  * \f[ GDR = 14 \times (base\_AC - 2)^\frac{1}{2} \f]
+  * \f[ GDR = 18 \times (AC)^\frac{1}{2} \f]
```
nlavsky, line 6012: Should it be ``frac{1}{4}`` instead of ``frac{1}{2}``?

PleasingFungus: Yep, you're quite right, thanks!

PR #1903, opened 2021-04-14
WIP: add Mjölnir, a throwable returning melee weapon

RojjaCebolla:
However you choose spell "Mjolnurr", it's the so-called "hammer of thor".

The main idea is that it's a melee weapon that also can be thrown, and
it has the returning property.  Advice greatly welcome as i tinker with this!

Due to the existence of Eos, an electric+holy mace might not be the most
exciting version of Mjlnr.  Perhaps it'd be better as a throwing dagger, or a
sharpened boomerang.  But what I'd like to try to code is a weapon that
can both be thrown and used in melee.  Potentially interesting is any
codepath in [throw.cc](https://github.com/crawl/crawl/blob/master/crawl-ref/source/throw.cc) that yields "toss away" or "but it does no damage".

With what I've coded so far, Crawl simply crashes when you throw Moljnir,
but i haven't given up yet!

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -1455,6 +1455,16 @@ COLOUR:   ETC_ENCHANT
 INSCRIP:  Rampage+∞
 VALUE:    400
 
+# Mojo is what most Nethack players call it; easier than guessing at spelling.
+ENUM:     MOJO
+NAME:     Mjölnir
+OBJ:      OBJ_WEAPONS/WPN_MORNINGSTAR
+PLUS:     +9
+BRAND:    SPWPN_HOLY_WRATH
+COLOUR:   ETC_ELECTRICITY
+INSCRIP:  Thrown
+BOOL:     elec
+
 # This entry must always be last.
 ENUM:   DUMMY2
 NAME:   DUMMY UNRANDART 2
--- a/crawl-ref/source/art-func.h
+++ b/crawl-ref/source/art-func.h
@@ -1374,6 +1374,31 @@ static void _THERMIC_ENGINE_unequip(item_def *item, bool *show_msgs)
     item->plus = 2;
 }
 
+static void _MOJO_melee_effects(item_def* /*weapon*/, actor* attacker,
+                                actor* defender, bool mondied, int /*dam*/)
+{
+    /* in nethack, mojo does tremendous damage infrequently; we nod to that
+       here with a double-damage electrocution brand that happens half as
+       often as standard elec brand. */
+    if (defender->res_elec() >= 3 || mondied)
+        return;
+    if (one_chance_in(6))
+        {
+            int special_damage = 16 + random2(26);
+            const string punctuation =
+                    attack_strength_punctuation(special_damage);
+            mpr((defender->is_player())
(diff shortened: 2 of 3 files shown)


ebering:
Long ago there were weapons that were both throwing and wielded weapons.

See cfbf888d4b3, 9312efc0, fabd5e008e9f for the life and death of Lehudib's Crystal Spear (and df89f4c26ec4 for the commit making weapons non-throwable which might help you chase your project). I went digging in the archives, the commit messages are a bit terse but [there was a discussion on the crawl-ref-discuss mailing list](https://sourceforge.net/p/crawl-ref/mailman/crawl-ref-discuss/thread/alpine.DEB.2.02.1309220042260.27430%40nawab.imp.fu-berlin.de/#msg31431366).

Something that didn't arise in those old discussions, but (technical and UI hurdles aside) that I think needs to be addressed is the following gameplay question:

```
When would a player chose to throw the weapon instead of swing with it or vise-versa?
```

If it's simply a question of "is the monster in melee range or not?" then this isn't mechanically exciting—why not just make it an unrand boomerang (sadly don't have my Australian mythology handbook with me so I can't suggest some relevant flavour)?

Underlying this is a much deeper mechanical problem that crawl has: there's no meaningful difference between melee-range throwing and swinging with a melee weapon other than the trained skill!

RojjaCebolla:
> If it's simply a question of "is the monster in melee range or not?" then this isn't mechanically exciting—why not just make it an unrand boomerang (sadly don't have my Australian mythology handbook with me so I can't suggest some relevant flavour)?
> 
> Underlying this is a much deeper mechanical problem that crawl has: there's no meaningful difference between melee-range throwing and swinging with a melee weapon other than the trained skill!

Hrmm. Simply making an artifact boomerang does sound like a cleaner concept.

I guess I've personally played a lot of characters that had both a melee weapon skill **and** throwing skill - perhaps that was a mistake on my part.  It seemed like a melee weapon at mindelay was faster, or more damaging than a thrown weapon at mindelay, but after getting it to mindelay, I'd train throwing so I had something to do at range.  So, if it's sensible to play a character who primarily wields a melee weapon but dabbles in throwing, that's the use case.  Though of course such a character is unlikely to run out of ammo, so there's no need for a throwable weapon except the cool factor and a bit of extra damage.

Hrrrrrmmmm.

Aliscans:
Having a melee/thrown split lets you make a weapon more powerful at melee range, where I wouldn't expect there to be much difference with a boomerang.

The Staff of Olgreb can cause damage at melee range or be evoked to cause damage at a distance, so there is a precedent for this sort of idea. It uses MP and damages everything in sight, so it's not an exact match.

The other thing melee weapons do is give you resistances, stat boosts and the like. An artefact boomerang couldn't do that, but instead it would free up the weapon slot for anything you like. If Mjölnir adopts the Nethack restriction of "you can only throw it if you're wielding it", it avoids this.

jv84:
To balance this out, monsters should be able to pick up items and use them. That would be truly wicked.

RojjaCebolla:
pausing this as i'm in the middle of starting a day-job. (non-coding 😞)

PR #1905, opened 2021-04-15
add Lua monster.target_desc function

mgdelmonte:
Adds the `target_desc` function to the Lua `monster` object.  This function returns the targeting description returned if you target a monster with a spell, your weapon, or a thrown item.

Examples:

- when targeting with a spell, the function shows the same description string in the UI ("22% to evade").
- when targeting with a thrown item, the same string as in the UI ("60% to hit")
- when targeting with your weapon, the same string as in the UI ("12% to evade your staff of earth")

This is all helpful information that you can get in the UI but is currently not exposed to Lua.


crawl-ref/source/spl-cast.cc
```diff
@@ -1702,6 +1702,17 @@ static desc_filter _targeter_addl_desc(spell_type spell, int powc, spell_flags f
...
 
+
+string target_desc(const monster_info& mi, spell_type spell)
+{
+    int powc = calc_spell_power(spell, true);
+    const int range = calc_spell_range(spell, powc, false);
+    unique_ptr<targeter> hitfunc = find_spell_targeter(spell, powc, range);
+    vector<string> d = _desc_hit_chance(mi, hitfunc.get());
+    return d.empty() ? "" : d[0];
+}
```
mgdelmonte, line 1713: This helper function required in spl-cast.cc because _desc_hit_chance is inaccessible outside this file.

mgdelmonte:
closed and superseded by #1922

PR #1908, opened 2021-04-16
Fully rename Wizard to Hedge Wizard

Goratrix:
This completes the renaming of Wizard to Hedge Wizard in all relevant code
and also changes the abbreviation to HW, as it is currently the only two-word
job name not abbreviated consistently. Also correctly sort the jobs in
job-data.h (Delver was not placed alphabetically).


ebering (changes requested):
Changing abbreviations does horrible things for our Sequell infrastructure, keeping the abbreviation 'Wz' was intentional. A revised PR that re-names the enum and includes an historical note about the abbreviation would be very appreciated.

Sorry for the delay in this review, as you might have noticed I haven't had much crawl time until recently.

Goratrix:
Thanks for the update! Yes, advil mentioned that the abbrev change causes problems in sequell (it was done for As -> Br), no problem. I actually feel like if the abbrev stays Wz, there's not much point in changing the enum either, maybe it's even better this way when the 'Hedge' part is only in one place and everything else is consistent... so I'd rather cancel this.

PR #1910, opened 2021-04-17
Add variety to boots tiles

Goratrix:
This adds back the boots tiles which were replaced in b2ee965 and
enables variation between the tiles, same as already exists for robes
and helmets.


Goratrix:
Tested in wizmode:

Screenshot 2021-04-17 160321

PR #1912, opened 2021-04-18
Fix the error message on creating a malformed dancing/spectral weapon.

Aliscans:
Previously, if you using the wizard mode create monster command with
"dancing weapon; stone" on a build with asserts compiled in, the game
crashed with "ASSERT(wpn) in 'dungeon.cc' at line 5159 failed.". The
same happens for any non-weapon, or even "nothing", and the same if a
vault was created which included such a monster.

With this change, it prints "Monster 'dancing weapon' requires a
weapon." instead in wizard mode, and the game cannot start if any vault
contains a monster like that.

### DIFF
--- a/crawl-ref/source/mapdef.cc
+++ b/crawl-ref/source/mapdef.cc
@@ -29,6 +29,7 @@
 #include "english.h"
 #include "files.h"
 #include "initfile.h"
+#include "item-prop.h"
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Nice fix, thanks! Apologies for the long delay on merging it.

PR #1913, opened 2021-04-18
Add a help option for the wizard mode create monster command.

Aliscans:
Entering ? as the monster name causes some help text to appear. This is
derived from the MONS section of docs/develop/levels/syntax.txt, but
with a few changes.

1. Use colour - yellow for headings, white for keywords which can be
entered as part of the command, and cyan for other things such as text
the game produces in response to a command, or file references.

2. Remove quotation marks from around keywords. I haven't done this with
cyan text, as this contains longer phrases which may be harder to
separate.

3. Don't give advice on how the command should be used (such as "use
this sparingly").

4. Rearrange the paragraphs a bit, and add a new opening paragraph.

5. Add lists of valid characteristics for mutant beasts and ugly things.

This change means that most of the information you need to create a monster is easily accessible from the create monster command, rather than being hidden away in a file you're told nothing about.

It does mean that sections of two files have much the same content in different formats. I don't know what the best way to handle this would be.

### DIFF
--- a/crawl-ref/source/dat/database/help.txt
+++ b/crawl-ref/source/dat/database/help.txt
@@ -312,3 +312,298 @@ set with <w>Ctrl-F</w>. If <w>Enter</w> is pressed at the prompt without enterin
 that will reset the filter to none.
 You can filter spells by name and by spell school.
 %%%%
+wiz-monster
+<h>Create specified monster</h>
+
+To create a monster with the default equipment for its species, simply enter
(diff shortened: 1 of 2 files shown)


ebering:
Thanks for the patch! I think the documentation duplication is survivable, the monster syntax is very stable these days and when it does change it is usually changed by someone grepping so they'll hopefully find both.

PR #1915, opened 2021-04-18
Further fix webtiles display of defense boosts

nlavsky:
After fc3203cb, defense boosts provided by the "acrobat and divine
shield are no longer coloured blue.

This is because unlike other boosts, these two are handled separately
in `tileweb.cc:_update_statuses()` and don't use `short_text` strings
from duration-data.h.

Fix this by updating `short_text`s in `_update_statuses()`. Also, fix
colouring of corroded AC and weapons on the HUD.

I checked all other defense boosts too and they are all coloured
correctly on WebTiles.


ebering:
Thanks for the patch!

PR #1919, opened 2021-04-19
Allow zoom in via ctrl-'+'

robertxgray:
While testing the zoom for the next Android version, I found it's impossible to zoom in with my keyboard. Some European keyboards like Spanish or German require pressing shift-0 for the equals character, but ctrl-shift-0 doesn't work for zooming it and you get the reassign function instead.

This patch allows ctrl-'+' to be used for zooming in with the affected keyboards.


robertxgray:
According to the Wikipedia, many other keyboards should also be affected: Italian, Danish, Norwegian, Swedish, Dutch, Estonian, Icelandic...

ebering:
Thanks for the patch! And welcome to contributing to crawl!

As you may have noticed, developer time waxes and wanes, so sorry about the delay on merging this.

PR #1920, opened 2021-04-20
Display seven-league ∞ rampage on %

Implojin:
It would probably be better to refactor _itosym() to be capable
of handling ∞, but this commit takes the hack approach already
used by rPois∞.


semitonal:
Thanks for the PR, I went ahead and did the general refactoring instead since I noticed that infinite willpower (in shadow form) should also be displayed.

PR #1922, opened 2021-04-21
added `monster.target_XXXX` Lua functions

mgdelmonte:
**Revised.**  Adds `target_desc`, `target_weapon`, `target_spell`, and `target_throw` functions to the Lua `monster_info` object. `target_desc` returns the targeting description of the monster.  The other function returns the descriptions shown if you target a monster with your weapon, spell, or a thrown item.

Examples:

- `target_desc()` on Sigmund might return "Sigmund, wielding a +0 scythe and wearing a +0 robe"
- `target_weapon()` on Sigmund might return "about 18% to evade your dagger"
- `target_spell("Sandblast")` on Sigmund might return "74% to hit"
- `target_throw(stone)` on Sigmund might return "about 45% to hit"

This is all helpful information that you can get in the UI but is currently not exposed to Lua.  Note that the targeting description of a monster includes helpful information such as its weapon and armour, but this information is not available anywhere else in Lua.  Not even the "long description" of a monster will show it.

This PR supersedes closed PR https://github.com/crawl/crawl/pull/1905, which was messy (included changes from other branches).

### DIFF
--- a/crawl-ref/source/directn.cc
+++ b/crawl-ref/source/directn.cc
@@ -1573,7 +1573,7 @@ string direction_chooser::target_description() const
 void direction_chooser::print_target_monster_description(bool &did_cloud) const
 {
     string text = target_description();
-    if( text > "" ) {
+    if ( text > "" ) {
         mprf(MSGCH_PROMPT, "%s: <lightgrey>%s</lightgrey>",
             target_prefix ? target_prefix : !behaviour->targeted() ? "Look" : "Aim",
             text.c_str());
--- a/crawl-ref/source/directn.cc
+++ b/crawl-ref/source/directn.cc
@@ -1533,17 +1533,17 @@ static void _push_back_if_nonempty(const string& str, vector<string>* vec)
         vec->push_back(str);
 }
 
-void direction_chooser::print_target_monster_description(bool &did_cloud) const
+string direction_chooser::target_description() const
 {
     // Do we see anything?
     const monster* mon = monster_at(target());
     if (!mon)
-        return;
+        return "";
 
     const bool visible = you.can_see(*mon);
     const bool exposed = _mon_exposed(mon);
     if (!visible && !exposed)
-        return;
+        return "";
 
     // OK, now we know that we have something to describe.
     vector<string> suffixes;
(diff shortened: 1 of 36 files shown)


crawl-ref/source/l-moninf.cc
```diff
@@ -321,6 +325,39 @@ static int moninf_get_ev(lua_State *ls)
...
+        ostringstream result;
+        describe_to_hit(*mi, result, false);
+        lua_pushstring(ls, result.str().c_str());
+        return 1;
+    }
+    if (lua_isuserdata(ls, 2))
+    {
+        // to-hit with a thrown item
+        item_def *item = *(item_def **) luaL_checkudata(ls, 2, ITEM_METATABLE);
+        ranged_attack attk(&you, nullptr, item, is_pproj_active());
```
ebering, line 349: This is a confusing api, I would expect if I put a melee weapon in I would get my melee to-hit and not my chance to hit throwing a fumble.

ebering (changes requested):
Thanks for the submission. Sorry we've been a bit behind on our PR work, this one is not ready to go as-is.

This should be split into two functions (or possibly three): one for weapons and one for spells. The one for weapons should detect (or if this turns out to be a pain use two functions) whether the attack is ranged or melee and instantiate the appropriate attack class for `to_hit_pct`.

Also these t functions should  be called `weapon_to_hit`, and `spell_to_hit` since they don't provide the full target desc found in the `x` view.

crawl-ref/source/l-moninf.cc
mgdelmonte: @ebering I've separated this into _four_ functions -- one for the monster itself (there is useful information in the targeter description that is shown nowhere else), and three for the success descriptions given by targeting with a weapon, spell, or thrown item.

crawl-ref/source/directn.cc
```diff
@@ -1533,17 +1533,17 @@ static void _push_back_if_nonempty(const string& str, vector<string>* vec)
         vec->push_back(str);
 }
 
-void direction_chooser::print_target_monster_description(bool &did_cloud) const
```
mgdelmonte, line 1536: No other good way to get the targeter description because it's hidden inside a `print_...` private function.  So I pulled the description creation code out into a public `const` function and then call it from the `print_...` function.

crawl-ref/source/l-moninf.cc
ebering: Four functions! Excellent.

Sorry for the merge delay, your updated version arrived close to the run up to release.

PR #1930, opened 2021-04-27
Modify Grammar Sceptre of Torment

l33t-d00d:
Minor grammatical modification to the message when wielding the Sceptre of Torment.  "Terribly searing pain" (adverb modifying adjective describing noun) becomes "Terrible, searing pain" (compound adjective describing noun).  Both are correct, however, the former is a little awkward because it partially shifts the focus of the sentence from pain to searing.  The latter flows  better because the compound adjective shifts the focus of the introductory descriptions to pain.


PleasingFungus:
Thanks for contributing the fix! Agree with the change.

ebering:
You've been entered into the CREDITS as l33t-d00d for now, let us know if you'd like to be listed another way!

PR #1932, opened 2021-04-28
Split _gift_sif_kiku_gift() into a function for each god.

Aliscans:
Split _gift_sif_kiku_gift() into _give_sif_gift() and _give_kiku_gift(), and rearrange each a bit.

In addition:

The "get a god gift" wizard command now works for Kikubaaqudgha.

The game doesn't crash if Kikubaaqudgha tries to create a spellbook when there are too many items on the level.


PleasingFungus:
LGTM. Nice cleanup!

PR #1934, opened 2021-04-30
Mark XP evokers as being useless if you have one in your inventory.

Aliscans:
1 tin of tremorstones has as many charges as 52 of them do, so there's no benefit to picking up the other 51 of them.


ebering:
`is_useless_item` is used for item coloring, indicating inertness in the quiver, stash search, and a bunch of places where it maybe shouldn't be. However, its assumed to not condition on the player's current inventory and this is important for the following use case: in a portal vault with a full inventory I might want to know that I can drop my (inert) tremorstone tin and pick up a different one on the outside after I finish the portal.

The default autopickup exceptions are set to prevent picking up the other 51 tins of tremorstones.

PR #1939, opened 2021-05-04
Fix what happens when you try to read a spellbook.

Aliscans:
Make the game give a "You aren't carrying any scrolls." message if you try to read() when you have no scrolls in your inventory or on the floor, but there is a spellbook on the floor.

The game had given an inventory prompt with no items on it instead.

The first bit of item_is_selected() dealt with reading scrolls, so the case I've deleted was only used for spellbooks. A restriction in use_an_item() means that you couldn't select a spellbok at the "read which item" prompt it gave you, even if you selected "show all".


PleasingFungus:
Nice find and nice fix! Thanks :)

PR #1941, opened 2021-05-05
Highlight Curl's AC boost on webtiles HUD (12567)

nlavsky:
Highlight boosted AC when palentongas curl up.


PleasingFungus:
Nice fix, thanks!

PR #1942, opened 2021-05-07
Add hard variants of easy ice caves.

hellmonk3:
There were four ice cave layouts that only placed early. Provide these layouts
with an alternate monster set so that they can place on "hard" ice cave levels.
Two maps receive further adjustments; the nsubst in ice_cave_small_giant is
changed so that some 4s actually place, and ice_cave_statue_garden_transporter
is rearranged so that the player can see the landing spot.

There are some very special bosses for players to discover in these vaults, but I
don't think they're too unreasonable compared to double ice fiend.

### DIFF
--- a/crawl-ref/source/dat/des/portals/icecave.des
+++ b/crawl-ref/source/dat/des/portals/icecave.des
@@ -511,10 +511,10 @@ default-depth: IceCv
 # can try to clear the whole cave. The cavey parts should have no chokepoints.
 
 # Foes: animals + frost giant (or rare draconian) & an ice statue.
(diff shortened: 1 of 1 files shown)


gammafunk:
I'll be using some of the work in this PR as part of a rework of monster placement that will further unify the sets of beasts/natural monsters and demons/undead as well as loot items for maps that are already using some common definitions as well as some other maps using very similar ones to the common definitions we have. This will also allow both easy and hard variants by both beasts/natural as well as demon/undead, so we can have greater variety at both difficulties.

I'll leave this PR open and will close with a commit once my work is merged. Thanks for this initial work, as it will become part of what I finally put together.

PR #1948, opened 2021-05-12
Update unident.txt

zhangjoke:
Add translation.
增添翻译。


ebering:
Thanks for the various translations! It looks like you based your work off of an older version of crawl, so I did not include the ones that seemed to be referencing older mechanics. (This is in regards to the entire set of translation pull requests).

PR #1952, opened 2021-05-13
Adding a Zin orc sect vault to branches/orc.des

DreamDustDCSS:
Inspired by Mennas' lore. More details in the commit notes.
I'm still figuring our github, btw. Apologies if I've done anything wrong.
Is it helpful to do pull requests for each vault I make, or should I wait until I have a good number of them before submitting them all at once?

### DIFF
--- a/crawl-ref/source/dat/des/branches/orc.des
+++ b/crawl-ref/source/dat/des/branches/orc.des
@@ -1398,6 +1398,49 @@ x1...1...1x
 xxx..@..xxx
 ENDMAP
 
+NAME:   dreamdust_zin_orc_sect
+TAGS:   no_monster_gen no_item_gen
+DEPTH:  Orc
+ORIENT: float
+WEIGHT: 3
+KMONS:  ? = nothing w:5 / \
+            orc name:pious name_adjective god:zin ; nothing
+KMONS:  1 = patrolling orc warrior \
+            name:devout name_adjective god:zin
+KMONS:  2 = orc knight \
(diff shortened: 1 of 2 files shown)


semitonal:
Thanks for the PR! Submitting vaults in batches is probably the most common approach, but I don't think there's an issue with submitting them individually either if you just have one finished and ready to go.

It's generally best to minimise the use of vault-renamed monsters - the only one here where it really seems necessary for the vault to function is the high priest, since that does explicitly mention Beogh in its regular description and the monster is being made functionally different with a changed spell set. That lets the vault still get its theme across clearly without having as many cosmetic-only renames. The threats in the vault also don't merit giving out so much gold - small piles for flavour would be fine.

Iron grates are usually reserved for cases where it's neccessary for a vault's functionality for a monster to be kept behind a barrier but able to cast spells through it (summoning, smiting, etc) with the player being able to dig the monster out, so the grates aren't really appropriate here, so regular rock or stone as the vault border would be better.

DreamDustDCSS:
Thank you for the feedback!

- Made it so that only the priest has been renamed.
- Changed the gold to regular gold piles.
- Replaced most of the fence with rock walls. I kept the front gate to show that these orcs have separated themselves from the rest of orc society. It also signals to the player that they're entering a vault (even if it's not particularly dangerous). Hope that's ok.

DreamDustDCSS:
Please ignore the two new commits above - meant to do them on a different branch, and I sadly can't figure out how to revert them on Github's web interface. Sigh.

PR #1955, opened 2021-05-15
Metric apts

rawlins:
Main commit message:

```
At the moment, this commit is intended to be up for discussion.

On the one hand, the positive/negative values for aptitudes provide an
easy and convenient shorthand for "good"/"bad", and also make it
relatively easy to do the math for what a particular apt will mean. On
the other hand, we have seen that this scheme often leads to players
overemphasizing the sign vs the actual numeric value; no amount of
documentation etc has led to players (especially newer ones) conceiving
of -1 is more "normal" as opposed to "terrible". For reference (sorry,
did not recalculate this for 0.27, but it hasn't changed that much): in
0.25, the overall mean aptitude was -0.35, with the worst species being
troll at -2.6 and the best species being barachi at +0.77.

This commit is a way I've found to solve some aspects of this problem
without majorly changing the underlying math: where previously
everything except Gn fell on a -5 to +5 range, it simply linearly
rescales aptitudes onto a 0 to 10 range for presentational purposes.
(Internally, nothing changes.)

I do think this makes the skill menu a bit harder to process quickly,
and I've accordingly done some recoloring to try to make good and bad
apts stand out more quickly. But, this still may be a net negative in UI
terms.
```

### DIFF
--- a/crawl-ref/docs/crawl_manual.rst
+++ b/crawl-ref/docs/crawl_manual.rst
@@ -569,7 +569,7 @@ Dodging or Stealth. See Appendix `3. List of Skills`_ for a detailed
 description of all skills present in Crawl. The ease with which a character
 learns a skill depends solely on species. These aptitudes are displayed when
 viewing your skills, and a full table can be viewed in aptitudes.txt (also from
-the help screen during play via '?%').
+the help screen during play via '?%'). Aptitudes generally range from 0 to 10.
 
 You can see your character's skills by pressing the 'm' key; the higher the
 level of a skill, the better you are at it. All characters start with a few
@@ -1470,7 +1470,7 @@ the species.
           which allow them to make extra attacks.
 
 .. note:: Humans are a useful reference point when considering other species:
-          they have 0 for almost all aptitudes; have no special abilities,
+          they have 5 for almost all aptitudes; have no special abilities,
           weakness, or constraints against using certain types of equipment;
           move normally; and gain experience and willpower at a "typical"
           rate. However, you will see that they are categorized as an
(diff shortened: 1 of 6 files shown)


1player:
I did not want to just leave a thumbs down to this proposal, it would have been quite rude since you're very open about this being up for discussion.

Personally I'm not a fan of the new [0,10] scale. I understand the rationale that people give too much importance to the sign of the aptitude, but making it a [0,10] scale makes it even more useless in my opinion, as the sign is a very effective indicator of "worse than average" or "better than average". In my eyes `-1` is much easier to understand and rationalise than a `4`, as in "very slightly below average", whereas 4 has absolutely no more meaning than 11 in an hypothetical [0,27] scale.

randomMesh:
> the sign is a very effective indicator of "worse than average" or "better than average".

This. Please leave it as it is.

jv84:
Weapon accuracies would be unsigned too?

rawlins:
I'm not going to merge this myself, and no other devs have seemed very interested, so I'll go ahead and close this up. The problem remains.

In addition:

> the sign is a very effective indicator of "worse than average" or "better than average".

This is not really accurate, and that people believe and act on some version of this is one of the main motivations of this PR. I haven't updated my numbers for what the exact average is, but as of 0.25, the global apt mean was -0.35, and for combat/magic in particular was around -0.5. In fact, much more important than comparison to the average is that there's a substantial range in the middle that probably should be considered "good" that includes negative numbers, and because of this, the sign on e.g. -1 is quite misleading, especially to newer players.

To reiterate the text in the commit msg: "we have seen that this scheme often leads to players
overemphasizing the sign vs the actual numeric value; no amount of
documentation etc has led to players (especially newer ones) conceiving
of -1 is more "normal" as opposed to "terrible"."

PR #1959, opened 2021-05-20
Lower Absolute Zero's range to 3

sdynet:
This spell has a penalty of randomly selecting the target, but because of its long range, the player was able to offset the penalty by simply using the spell several times. Thus, the Pandemonium lord turned into an ice pillar upon encountering the player. So, to curb brainless play, I'm trying to reduce the range of this spell to 3.

### DIFF
--- a/crawl-ref/source/spl-data.h
+++ b/crawl-ref/source/spl-data.h
@@ -3324,7 +3324,7 @@ static const struct spell_desc spelldata[] =
     spflag::no_ghost,
     9,
     200,
-    5, 5,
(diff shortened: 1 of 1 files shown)


semitonal:
Closing, since Absolute Zero has been removed/replaced for now. It's possible a spell with a similar effect might return in a different magic school in the future, in which case we could look at ways to adjust it to have a better feel (such as giving it low range as per this PR, other suggestions include having a delay on-cast before the instakill takes effect).

PR #1965, opened 2021-05-28
Second Antaeus quote

4Hooves2Appendages:
Add second Antaeus quote.


ebering:
Thanks for the quote!

PR #1968, opened 2021-05-31
Tweak Ozocubu's Refrigeration usefulness check

nlavsky:
Currently, the UI incorrectly tells the player that the spell has
valid targets when there are only plants, butterflies, or player
allies nearby.

This commit changes Refrigeration's usefulness check, so the spell
is grayed out on the quiver if there are no hostile monsters in LOS.


PleasingFungus:
Thanks for finding and fixing this!

PR #1969, opened 2021-06-01
(WIP) Rework and Reflavour: Gnoll Shaman -> Bouda

nikheizen:
This commit reworks Gnoll Shamans and re-flavours them into Bouda.
It is a work in progress and still has one notable bug I am aware of.

### TODO before merge:

- Make the duration expire on allied monsters. _(I need help here, not sure what is missing)_
- Adjust vaults which placed gnoll shamans so they don't have too many bouda or appear with bouda too early.
- Add a couple vault ideas utilizing the strengths and flexibility of the new monster compared to the old one.
- Potentially refactor the 3 monster aura effects (zyme sickness, torpor slow, bouda harm) into a unified function since they share a lot of logic and check a lot of the same properties.

### Changes to consider:

- Group Swiftness (Hunting Call) could also just work as haste/haste other, but I like the flavour with hyenafolk packs.
- Damage %ge probably needs to be adjusted up a bit.
- The weight is currently a bit high in the spawn table for testing purposes, this may need to go down.
- The names for the UI elements I would like some input on. I wanted to use the term "vulnerable", but am concerned about confusion with the scroll of vulnerability.

### Why not just use battlecry/aura of brilliance?

When talking with gammafunk on the discord, they brought up a couple good points about my local branch's ideas. First off the original effect was literally just harm, so it amplified player damage too. This has been simplified as it's a pretty abusable effect. Secondly they pointed out that if the creature is a support character for gnolls, we could just use battlecry/might other as a damage buff. I think that this effect is distinct in a few important ways from battlecry/aura of brilliance:

- The debuff ends immediately when the bouda dies and starts as soon as the bouda ends a move with the player in LOS. The idea here is that it creates a sense of immediacy to kill the bouda which creatures like Orc Knights et al. don't create as much (Orc knights you are more trying to keep out of LOS of their buddies so they can't might as many folks).
- The debuff applies to all damage, which extends the use case of the bouda beyond what gnoll shamans can do. They have a slightly different behaviour with packs of gnolls, but can amplify the power of other spell-casters too for some interesting mid-D encounters and vaults.
- The effect might be interesting on other later-game monsters. I'm looking at the demonic crawler in particular. This effect amplifies melee and poison damage in spider, but also has fun, interesting, and lethal implications when you pair them with hellions/tormenters in pandemonium and hell.

### Inspiration (from the commit message):

> Bouda are characters from Northeast African folklore who are shapeshifters that turn into jackals. They have the power of the Evil Eye which curses others with spectacular misfortune. The term is historically one of exclusion- othering someone by calling them Bouda is analogous to the European tradition of naming folks as witches. In crawl they have all the malevolent energy of their mythological basis, but here it is directed towards the player- the dungeon's one true outgroup.


PleasingFungus:
Don't have time/energy right this second to do a full review, but one small thought:

> Make the duration expire on allied monsters. (I need help here, not sure what is missing)

Do you actually want this to be an ench at all? I wonder if the code that handles halo/umbra might be a simpler way of managing the bookkeeping here. (It's possible that it wouldn't be! I haven't looked into it.)

ebering:
I've picked this up. The flavour is amazing, and I strongly agree about the shaman not managing to be a good support caster The damage amplification aura seems very complicated on the implementation side. I've gone with something simpler in a similar spirit: taking the hunting call in place of haste other and adding weakening gaze (like salamander tyrants) as an instant ability to fit the Evil Eye curse in place of Corona. Hopefully these things will help them "get down to business" faster, though admittedly they don't help bouda shine in non-gnoll groups.

I do think there's something interesting in the monster harm aura idea, though I agree with PF that it is probably better implemented as an area (like halo, silence, and umbra). If you're interested in taking another stab at it do send us a fresh PR!

PR #1978, opened 2021-06-08
Don't put Refrigeration users under penance for tracer checks

nlavsky:
After 73eefd5d, Ozocubu's Refrigeration users can get under penance for
opening the z? or Q menus or just for having the spell on the quiver.
It happens if they worship a god who cares about allies and if there
are allies nearby.

Fix this by skipping conducts when doing tracer checks.


semitonal:
Merged, thanks for the patch!

PR #1979, opened 2021-06-09
Remove -Wiz from blade hand form

dillytin:
Blade hand is now level 6 spell and has become much harder to find. You can only find it from a book of Transfiguration which contains statue form and dragon form. Blade hand is now not decent as before when it was level 5 and in a book of Changes, and not attractive compared to statue form which gives you a good deal of AC and resistance and makes you free from ER from body armour so that you can cast spells easily(and still you can wear shields!). Therefore, I suggest Blade hand form should have no -wiz penalty. It already requires a lot of exp and stats to stabilize. I want to make blade hand attractive as the same level spell: statue form.


dillytin:
Bhand has become 5 level spell, and the reasoning above has also been pointless. Sad.

Issue #1981, opened 2021-06-11
Inaccessible altar vault


ardl:
**PLATFORM and OS or BROWSER:**
webtiles trunk 0.27-a0-1256-gced786f65e

**DESCRIPTION:**
Inaccessible altar vault generated on elf 3. Game is still ongoing so I will post the morgue when its over

![image](https://user-images.githubusercontent.com/70080435/121625167-f1e58280-ca40-11eb-89c1-8bb9d5618692.png)

ardl:
Morgue
https://cbro.berotato.org/morgue/ardl/morgue-ardl-20210612-211307.txt

DreamDustDCSS:
> NAME:   lemuel_upstairs_altar
TAGS:   extra
DEPTH:  Elf, Crypt
ORIENT: float
SUBST:  T:T., )=)}], c:ccx
KMASK:  T = no_monster_gen
MAP
ccccccc
cT...Tc
c.....ccccccc
c..C..+....)c
c.....ccccccc
cT...Tc
ccccccc
ENDMAP

This vault is supposed to have a downstair - } ) ] - in that top corridor part. The idea being that you can only access this altar by taking an upstair from the floor below. 
But since the vault generated on Elf:3, there's obviously no lower floor, which should be the cause of the issue.
I think changing the DEPTH line to "DEPTH: Elf, !Elf:$, Crypt, !Crypt:$" should fix it.

If a dev doesn't fix it, I'll open a PR sometime later. I'm just a little hesitant since something in the documentation caught my eye:

> }{ - Stone stairs - You must be able to reach these from each other. The { upstair is also the stair on which the player will enter the dungeon for arrival vaults. 
)( - Stone stairs, set 2. 
][ - Stone stairs, set 3 - minivaults often fail to get this pair if the level isn't very interconnected or if any larger vaults are using      any stairs.

It sounds to me that this vault using ] is potentially a bad idea. Not sure though.

Issue #1973, opened 2021-06-03
Hatch up into an inescapable room


PhazeonPhoenix:
**PLATFORM and OS or BROWSER:**
webtiles Chrome Windows

**DCSS VERSION AND GAME INFORMATION:**
0.27-a0-1202-ge72e49191c

**DESCRIPTION:**
Took a hatch up and landed in this inescapable room in a vault. Sadly no teleports.

**EXTRA FILES:**

![chrome_20210603-010911](https://user-images.githubusercontent.com/18453/120612718-56359000-c40a-11eb-92da-c1825c5f59a8.png)

https://cbro.berotato.org/morgue/cobaltphoenix/morgue-cobaltphoenix-20210603-081628.txt

Attempt to retrieve backup from CBRO2 failed with the following message:

```
/bin/savegame-backup.sh: line 124: SAVE_MATCHES[0]: unbound variable
```

DreamDustDCSS:
```
NAME:   heliophobic_arrival_ceremonial_halls
TAGS:   arrival no_monster_gen no_trap_gen
ORIENT: float
TILE:   c = wall_marble
FTILE:  '+{cG = floor_marble
COLOUR: 'cG = white
SUBST:  ' = .
MAP
.....ccc+ccc.....
.....cG'''Gc.....
..c..cc'c'cc..c..
.ccc.cG'c'Gc.ccc.
.ccc.cc'c'cc.ccc.
..c..cG'c'Gc..c..
.....cc'c'cc.....
ccccccG'''Gcccccc
cGcGcGc'''cGcGcGc
c'''''''''''''''c
+'ccc'''{'''ccc'+
c'''''''''''''''c
cGcGcGc'''cGcGcGc
ccccccG'''Gcccccc
.....cc'c'cc.....
..c..cG'c'Gc..c..
.ccc.cc'c'cc.ccc.
.ccc.cG'c'Gc.ccc.
..c..cc'c'cc..c..
.....cG'''Gc.....
.....ccc+ccc.....
ENDMAP
```
There are probably multiple ways (and better ones) to fix this problem, but one quick fix would be to tweak the map so there's always a path between the little quadrants and the doors. Something like this:
```
MAP
    .... ....
.....ccc+ccc.....
.....cG'''Gc.....
..c..cc'c'cc..c..
.ccc.cG'c'Gc.ccc.
.ccc.cc'c'cc.ccc.
..c..cG'c'Gc..c..
.....cc'c'cc.....
[...]

PR #1984, opened 2021-06-14
Fixing issues with a couple vaults

DreamDustDCSS:
Fix: lemuel_upstairs_altar

- Should fix the problem originally reported at: https://github.com/crawl/crawl/issues/1981

Fix: heliophobic_arrival_ceremonial_halls

- Should fix the problem originally reported at: https://github.com/crawl/crawl/issues/1973


rawlins:
Thanks for the fixes! I will take the lemuel one as-is. For the heliophobic vault, I'll push an extension; the biggest thing it needs is the `transparent` tag, which will allow levelgen to check its internal pathing. Adding the outer paths will help it place even with that, though.

Issue #1944, opened 2021-05-07
Ghost walking free 


Sobieck:
**PLATFORM and OS or BROWSER:**
webtiles - firefox

**DCSS VERSION AND GAME INFORMATION:**
Version history for your current game:
Game started: 0.27-a0-1137-g3ef0f4ec91
0.27-a0-1139-g28e009f8a6

**DESCRIPTION:**
I believe there might be a problem with map generation. I had a wizlab generated in what appears to be a ghost room. The ghost was just out walking around like the old days. 

![image](https://user-images.githubusercontent.com/6256765/117468678-8b89b380-af1a-11eb-894d-e12dc3afc7a5.png)

DreamDustDCSS:
These are the two vaults:
```
NAME:       mu_enter_wizlab_4
KFEAT:      ! = floor
COLOUR:     ! = magenta
FTILE:      !O = floor_rough_magenta
:           wizlab_portal(_G)
MAP
    xxxxx
    w!!!x
   ww!O!x
  www!!!x
 www..www
www..www
ww..www
w..www
@.www
@www
ENDMAP
```
```
NAME:   ebering_vaults_ghost_xom
TAGS:   vaults_orient_s
: lone_ghost_guarded_loot(_G, "O")
: lone_ghost_extra_loot(_G, ".")
: setup_xom_dancing_weapon(_G)
KMONS:  O = player_ghost
KFEAT:  _ = altar_xom
: vaults_ghost_setup(_G)
MAP
...1...
..d_e..
...O...
xxn=nxx
.......
.......
ENDMAP
```
So it turns out that there's nothing wrong with the wizlab or the level generator. The problem is with ghost vault, which is completely open for some reason.
I can try fixing it in a PR later. Thanks for reporting!

gammafunk:
This bug is in fact due to an interaction of the level generator and the wizlab portal vault. For Vaults branch, vaults exist as special "room vaults" that the layout loads on its own directly into the map. While the layout will never allow these rooms to overwrite each other, information about the original room vault is lost for use by any subsequent non-room vaults. In most cases such non-room vaults don't get placed at all, since the layout doesn't use them. But certain things like portal vaults can place as minivaults, since they use CHANCE.

[...]

PR #1985, opened 2021-06-14
Fix: very strange & serious issues with ghost vaults in the Vaults

DreamDustDCSS:
Many of the ghost vaults in the Vaults are seriously broken. They're just wide open, allowing the ghosts to roam free. I assume this wasn't a simple mistake/oversight when these vaults were created and that some kind of functionality is broken instead, perhaps by a recent-ish change?

My changes just make simple enclosures so the ghosts won't roam. But it's probably better if the original problem (whatever it is) is identified and fixed instead.

Originally reported at: https://github.com/crawl/crawl/issues/1944

Affected vaults:
gammafunk_vaults_ghost_grave
gammafunk_vaults_ghost_cemetery
gammafunk_vaults_ghost_fury_of_okawaru
ebering_vaults_ghost_inner_flame
ebering_vaults_ghost_reflecting_pool
ebering_vaults_ghost_disaster_area
ebering_vaults_ghost_xom
biasface_vaults_ghost_orc_armoury
ploomutoo_gammafunk_vaults_ghost_potion_laboratory
gammafunk_vaults_ghost_tricky_traps
gammafunk_vaults_ghost_crypt

@gammafunk @ebering

### DIFF
--- a/crawl-ref/source/dat/des/branches/vaults_rooms_ghost.des
+++ b/crawl-ref/source/dat/des/branches/vaults_rooms_ghost.des
@@ -71,13 +71,13 @@ COLOUR: t = brown
 : set_feature_name("tree", "dead tree")
 : vaults_ghost_setup(_G)
 MAP
-..x.....
-..x.....
-..n.dt..
-..=.OGt.
-..n.et..
-..x.....
-..x.....
+..cccccc
+..c....c
+..n.dt.c
+..=.OGtc
+..n.et.c
+..c....c
+..cccccc
 ENDMAP
 
 NAME:   gammafunk_vaults_ghost_split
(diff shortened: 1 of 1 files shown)


DreamDustDCSS:
example_open_ghost_vault
Example of what one of these open vaults looks like ^^

gammafunk:
I'm closing this because the ghosts vaults for Vaults are in fact working correctly, although see some issues I describe in comments I'll leave in #1944. I'm closing this PR because your proposed changes aren't valid and the issue isn't specific to ghost vaults. If you want to work on the issue in general or just the ghost vault issue, feel free to open a new PR.

In your screenshot, you attempted to explicitly place these vaults in wizard mode with e.g. `&P`, which will not work. These vaults must be placed via the Vaults layout, which uses a special mechanism to load them inside of rooms that the layout generator creates, making modifications to the terrain inside the vaults after doing so. To test vault rooms, the only easy way is to increase their weight maximum (i.e. set `WEIGHT: 10000`) and reload the level using e.g. `&R`. For ghost vaults rooms you have the added difficulty of triggerikng the chance of generating a ghost vault, although you can modify `_GHOST_CHANCE_PERCENT` at the top of [dat/dlua/ghost.lua](https://github.com/crawl/crawl/blob/master/crawl-ref/source/dat/dlua/ghost.lua).

These ghost vaults are not typical minivault/float vaults, but are special vault "rooms" that work with the Vaults layout. See comments and examples in [dat/des/branches/vaults_rooms.des](https://github.com/crawl/crawl/blob/master/crawl-ref/source/dat/des/branches/vaults_rooms_standard.des). These room vaults seem "wide open" when looking at the vault definition, but properly get walls when used by the Vaults layout. What can break these vaults are minivaults that place via CHANCE, like the wizlab portal in #1944. See comments in that issue for further details.

PR #1999, opened 2021-06-26
More Ecumenical alter Ruins Batch 2

Shapermc:
This commit is for the 7 additional ecumenical alter I created and finally figured out how to merge! My apologies for the extra commit, I was informed this will be deleted or taken care of.


ebering:
I took all of these except tikhal, we really don't want to be adding a lot of monsters to these vaults and I wasn't sure if the concept would get across without them.

You've got your git repository into a strange state, I would on your master branch run the command `git reset --hard origin/master` (where `origin` is the name of the remote corresponding to the crawl official repo) This will destroy the tikhal vault so save that elsewhere before hand.

PR #2009, opened 2021-07-01
Added explicit effect information to regeneration mutation description

dicedlemming:
Numbers sometimes good. This is also how the regeneration is represented on the character sheet


dicedlemming:
Possible to call it HPRegen for exact parity but that would make the tooltip EVEN LONGER and the + kind of implies it

PR #2016, opened 2021-07-06
Fix some issues with "stat zero" warnings.

Aliscans:
Make it so that:

1. A warning is generated if equipping an item reduces a stat below 1.
2. A warning is generated if removing at item (possibly as part of the same action as equipping a different one) reduces a stat below 1 at any point.
3. A warning is not generated if no stat will be reduced below 1 when equipping or removing items.
4. If a warning is generated and the player decides not to proceed, no in-game action takes place. There was a TODO for this.

The existing code didn't do any of these things reliably.


ebering:
Nice, thanks for the patch!

PR #2019, opened 2021-07-08
Update mutation-data.h

dicedlemming:
The spell school it improves is Hexes. Book of Maledictions is the starter Hex book, but Hex is the spell school that gains power.

"Power" is also the terminology used on items that do similar things like Staff of Air, i.e. "increases the power of air spells"

"Enhanced" is more ambiguous when all it does is increase power.


ebering:
Thanks for the patch! How would you like to be listed in the CREDITS?

ebering:
I've credited you as diced lemming for now, let us know if you'd like that changed!

PR #2034, opened 2021-07-16
Strengthening the power of the xom's chessboard

sdynet:
I think this item needs to be slightly buffed considering the risk of not being able to select a useful item or target and having a beneficial effect on the enemy with a low probability.


ebering:
The chess pieces are still in an awkward place, but I don't think a quick power bump is the solution to them. @semitonal favors removal but maybe we can find a good enough compelling design to stay her hand.

PR #2038, opened 2021-07-19
Buff Ko's apt and adjust recommended jobs

dillytin:
Xbow is very difficult to utilise in comparson to bow and it's even worse for Kobold with reduced los. I suggested giving better xbow apt to several other species and nerfing Kobold's but PF doesn't think it is the right way. Instead, Buff Ko's xbow apt by 2(+2 to +4). Kobold with reduced los has difficulty with using ranged weapons, especally when it has 1.0 mindelay, so I think it's still balanced.

Also, AM is enough for Ko's recommended job about ranged weapons. Therefore, remove Hu from it and add Br in it.

### DIFF
--- a/crawl-ref/source/dat/species/kobold.yaml
+++ b/crawl-ref/source/dat/species/kobold.yaml
@@ -15,7 +15,7 @@ aptitudes:
   maces_and_flails: -1
   polearms: -2
   staves: -1
-  crossbows: 2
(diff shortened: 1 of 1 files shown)


dillytin:
It's obvious that you won't like it even if I make it +3, not +4. Then, please consider removing Hunter bg from recommended roles first.

PR #2040, opened 2021-07-20
Fix typo in the manual's Missiles section

githuberto:
This commit fixes a typo in the manual's Missiles section. The verb
should be "throw" since it's being used as an infinitive.


ebering:
Thanks for the fix, how would you like to appear in the CREDITS?

githuberto:
Oh, if this is credit-worthy, Robert Faulk, but this was just a tiny grammar nit :P.

ebering:
Every contribution is credit worthy! It's how we guilt you into contributing more ;)

PR #2041, opened 2021-07-22
Check for monsters, objects & features as soon as you move.

Aliscans:
Check for things after the player moves, rather than waiting for the start of the next turn.
This now uses check_for_interesting_features() in the same way _prep_input() already did, so explore stops if you find an item which autopickup wouldn't pick up.

Change a comment in check_for_interesting_features() so that it now says which the function does.

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -527,6 +527,10 @@ void moveto_location_effects(dungeon_feature_type old_feat,
 
     if (stepped)
         _moveto_maybe_repel_stairs();
+
+    update_monsters_in_view();
+    if (check_for_interesting_features() && you.running.is_explore())
+        stop_running();
 }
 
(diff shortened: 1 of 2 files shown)


ebering:
Gonna hold off on this until after the tournament. As you've found this sort of thing can touch very rickety cross-loaded travel features and I don't have time to test it thoroughly before then.

Aliscans:
That's fine; there are only a few circumstances where the game doesn't announce a monster at all, and this only fixes the one where the monster dies first.

And sorry about not realising what the game did with the return code. I don't know how I missed it.

wheals:
Currently, adding check_for_interesting_features here is causing issues with exploration (it's stopping on seeing most items, even autopickupable ones). Checking for monsters fixes the lack of comes into view messages, but do we gain anything from checking for items here?

PR #2045, opened 2021-07-28
move scroll of noise from bad item to dangerous item

elliottbernstein:
Scroll of noise is not a completely bad item. It's a dangerous item because it will wake sleeping monsters and attract them to you, but assuming you want monsters coming to you, that can be a positive effect. That's a similar usage to a potion of attraction, which is already classified as a dangerous item.

### DIFF
--- a/crawl-ref/source/item-name.cc
+++ b/crawl-ref/source/item-name.cc
@@ -2610,8 +2610,6 @@ bool is_bad_item(const item_def &item)
         case SCR_CURSE_JEWELLERY:
             return !have_passive(passive_t::want_curses);
 #endif
-        case SCR_NOISE:
-            return true;
         default:
             return false;
         }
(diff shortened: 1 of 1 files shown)


jv84:
What's the point? The same can be accomplished by shouting.

elliottbernstein:
> What's the point? The same can be accomplished by shouting. It's a useless item.

According to the comments in the code, it's not a useless item:

> // useless items can hardly be dangerous.

This scroll can put the player in danger. But is it "dangerous" according to the code? This is the question asked to determine if so:

> //Is an item dangerous but potentially worthwhile?

As you pointed out, yelling can have the same effect. So maybe it's not worthwhile either. So is it a bad item?

> // A bad item is always useless.

Well, it can't be useless because it's dangerous ¯\_(ツ)_/¯

elliottbernstein:
Bad is probably the best classification for this scroll! Closing this one out.

PR #2046, opened 2021-07-29
prevent confusing empty corrosion source message

elliottbernstein:
Currently, the empty string results in messages like " corrodes the python!" This change fixes that to specify the source of the corrosion is the player's corrupting presence mutation.


ebering:
Thanks for the patch, how would you like to appear in the CREDITS?

ebering:
Since we are very close to release I'm going to list you under your GitHub author name and if you'd like a change let us know!

elliottbernstein:
Hey @ebering you can list me as Elliott Bernstein. Thank you!

Issue #2017, opened 2021-07-07
Potion of lignification is regarded as useless for Vampire


dillytin:
Versions: 0.27-a0-1442-g87b6fcb595

![캡처](https://user-images.githubusercontent.com/65288602/124761888-3a735b80-df6d-11eb-8abe-e04875979ecb.PNG)

Vamp can use this potion while alive but it's marked as useless

rawlins:
This seems to be specific to the recognized items menu, it works correctly in other cases that I've tested.

robertxgray:
I have been looking into this issue and I think you should consider not returning true for is_lifeless_undead(false) when playing as a vampire. These are the side effects I found:

- Rage and lignification potions are tagged as useless.
- Wanderers can't receive such potions.
- Poison ring tagged as useless while bloddless and other poison related side effects.
- Kikubaaqudgha does not grant Sublimation of Blood.

PR #2055, opened 2021-08-08
Vampires are not (pure) lifeless undead

robertxgray:
Closes https://github.com/crawl/crawl/issues/2017

I have been looking into that issue and I think you should consider not returning true for is_lifeless_undead(false) when playing as a vampire. These are the (probably undesired) side effects I found:

- Rage and lignification potions are tagged as useless.
- Wanderers can't receive such potions.
- Poison ring tagged as useless while bloddless and other poison related side effects.
- Kikubaaqudgha does not grant Sublimation of Blood.

Also, it may render unnecessary a hack in player_res_poison.

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -1468,7 +1468,7 @@ bool player_kiku_res_torment()
 int player_res_poison(bool calc_unid, bool temp, bool items)
 {
     if (you.is_nonliving(temp)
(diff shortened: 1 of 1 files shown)


robertxgray:
If this PR ever gets accepted, you'll probably want to revert this commit: https://github.com/crawl/crawl/commit/e31878e1fe37

PR #2058, opened 2021-08-12
Add Mlioglotl, an Abyss themed unique

amcnicky:
This is a rebuilt, somewhat reworked (balance-wise, mainly) re-submission of PR #1546 
It's been a while since Milo was first submitted, so let's take it from the top:

Overall vision: a new unique with a strong flavour themed around Lugonu and the Abyss.
Flavour vision: a hulking mass (unique thrashing horror) charging after the player whilst rending open the fabric of the dungeon, causing denizens of the Abyss to spew forth.
Gameplay vision: a fast pursuer, combining plain melee damage with the ongoing threat of Abyss summons. Challenging the player to either avoid a fast creature, or to consider a strategic approach to kill Mlioglotl before too many summons are created. Emergent complexity to an engagement with Mlioglotl is added through temporary changes to basic terrain (adding and occasionally removing basic walls, though non-visual changes revert after a short amount of time). A visual novelty is also provided to the player by cosmetically altering the dungeon to Abyss tilesets in areas around the locations where Mlioglotl casts his corrupt_locale.

In-game monster description:
"A demon prince once betrayed his realm through a dark ritual to Lugonu.
Becoming the twisted mass of corruption known as Mlioglotl was his
reward.

Seeing such a powerful acolyte of Lugonu within the mortal plane is a
rare sight indeed, and one should fear the denizens and horrors of the
Abyss that are never far behind."

A few more gameplay details:
- Mlioglotl currently spawns in Swamp, Depths:1-3, Vaults:1-4, Abyss:2- in line with a suggestion from @gammafunk, and a look through the helpful workbook of unique spawn coverage
- The monster profile is 
Speed: 14
Attacks:  {AT_HIT, AF_PLAIN, 12}, {AT_HIT, AF_PLAIN, 12},
       {AT_TRAMPLE, AF_TRAMPLE, 25}, AT_NO_ATK},
Resistances: Poison, elec, cold (similar to thrashing horror)
XP modifier/willpower: 13/120
hit dice/hit points: 15/1350
AC/evasion: 6/2

In terms of feel, this probably puts him as a squishier Saint Rokah with a touch higher and more consistent damage output, who is fast, but without meaningful threat at range. Occasional summons and conservative temporary terrain tweaking are then also thrown into the mix. In play-testing, he generally can be handled well by a buffed melee specialist able to close the gap before meaningful summons have appeared, though spellcasters find him more challenging.

To be explicit, here are a few key points that I have changed based on feedback from #1546:
+ Mlioglotl's corruption has far less capacity for impacting terrain than the player version, and is indeed now a separate function, rather than a 'is monster' flag weaved throughout the player corruption ability.
+ Followers of Lugonu will no longer find Mlioglotl's summons neutral to their cause.
+ Milo will consider corrupting to be a bad idea if he's already in the abyss
+ Milo no longer blinks
+ Milo now deals only plain melee damage, to tone down away from the original chaos branding
+ Milo is now a midgame unique, rather than being tuned as extended/engame.
+ Milo no longer makes use of a tall tile
+ Milo's summons are now normal (not durable)

As always please let me know thoughts and indeed suggested changes to make so that one day you too can enjoy the thrill of Milo and his multi-coloured party zone :)


amcnicky:
mlioglotl

amcnicky:
Following further local playtesting, a bug was identified in which corrupted wall tiles that had been changed to floor tiles allowed the player to get stuck 'inside the wall' once the wall tiles completed their temporary transformation and returned to being walls. To prevent this, and I'm sure similar bugs along the same lines, Milo's corruption is now purely additive in nature (i.e. walls will only ever be added, not removed). This probably makes the unique less problematic in general from a perspective of vault-breaking and other such terrain-manipulating tedium.

ebering (changes requested):
Added Milo to my play testing of PF's god, faced Milo with Okawaru in Swamp 1 after abandoning Ignis.

The corruption spells work well, but its melee damage felt a bit weak for a monster that comes with might (so we expect it to hit a lot). I would suggest a single AT_HIT for 20 over the two dam 12 attacks.

The weakness might also be a result of its casting frequencies (the number with a spell) which are pretty low for the summons. These are weights in a roll out of 200 once the monster decides to cast at all. I would increase all of the casting freqs at least a little, except for might which could stand to go to 20 30, and remove the emergency flag from cause fear.

Very spooky and atmospheric to fight, and happy about how the temp (and non-temp) terrain works.

ebering:
Got very excited about this idea after play testing it, so decided to finish the requested changes and merge it. Thanks for the submission and picking this back up!

For the future, please take a look at our coding conventions and try to follow them! I had to fix quite a few whitespace and bracing style things that were not caught by our automatic code linters.

amcnicky:
Fantastic! Thanks so much. Expecting to get some more crawl time early next week so will check in on the style guide then and may well swing by the dev chat too in case anyone would like to live-discuss elements of the other PRs.

PR #2063, opened 2021-08-14
Update branches.txt

zhangjoke:
Add translation.
增添翻译。


crawl-ref/source/dat/descript/zh/branches.txt
```diff
@@ -1,27 +1,237 @@
+# 汉化项目网站：https://paratranz.cn/projects/808
```
ebering, line 1: This is not the appropriate place for credits. I previously added you as an individual to CREDITS.txt. If there are other members of your project that need to be credited please let us know.

Alternatively we can add "Paratranz.cn Project 808" to the CREDITS file.

PR #2065, opened 2021-08-14
Show turn counts and real game times in lobby

megabug:
This lets spectators find potentially interesting games to view, i.e. from a speedrunning perspective.

`webserver/process_handler.py`:
* Treat the turn count and duration (real game time) as "interesting info" in the wherefile.

`webserver/static/scripts/client.js`:
* Show the new turn count and duration fields.
* While at it, rationalise the units shown for idle times ("min" -> "m", and drop the spaces).

`webserver/static/style.css`, `webserver/templates/client.html`:
* Add new fields.

Note that previous versions of the commit made the real time display increase automatically on the lobby's client side, similar to the idle time. However, this ignored the effect of `IDLE_TIME_CLAMP`; it's not valid to assume that real time is always ticking (instead, it stops after the game is idle for `IDLE_TIME_CLAMP` seconds). In practice only updating the time displays on new wherefile keeps it update to date enough.

### DIFF
--- a/crawl-ref/source/webserver/static/scripts/client.js
+++ b/crawl-ref/source/webserver/static/scripts/client.js
@@ -1099,6 +1099,13 @@ function (exports, $, key_conversion, chat, comm) {
         set("xl", data.xl);
         set("char", data.char);
         set("place", data.place);
+        if (data.turn && data.dur)
+        {
+            set("turn", data.turn);
+            set("dur", format_duration(parseInt(data.dur)));
+
+            new_list.removeClass("no_game_times");
+        }
         set("god", data.god || "");
         set("title", data.title);
         set("idle_time", format_idle_time(data.idle_time));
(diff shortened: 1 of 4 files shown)


crawl-ref/source/webserver/static/scripts/client.js
```diff
@@ -1099,6 +1099,8 @@ function (exports, $, key_conversion, chat, comm) {
         set("xl", data.xl);
         set("char", data.char);
         set("place", data.place);
+        set("turn", data.turn);
```
rawlins, line 1102: This PR overall looks pretty good to me, thanks for the patch! The one caveat here is that for the initial merge it would be best to add a conditional here around these two lines: the python code will require a webtiles server restart, and the js code and the html template will update potentially right away (though `client.js` is dependent on the browser cache so that could take a while too, it's a bit unpredictable). Until a server restart happens (some servers restart very infrequently) this js code will produce an error. It's probably also best to use jquery to hide the columns if `data` doesn't have `turn` and `dur`, or maybe even start them hidden. Eventually this migration code could be removed once all servers have these changes.

Or another way to put this: I recommend testing this code with the python side of things commented out to make sure the js UI doesn't break during migration, which I think it is likely to right now.

megabug: Good point, I hadn't considered that.

megabug:
New version properly handles games not having the new fields (and hides the columns if no games do).

ebering:
Thanks for the patch! I've merged this to master. How would you like to appear in the CREDITS?

megabug:
Thanks. "Matthew Daley" is fine.

PR #2068, opened 2021-08-15
New spell: Explosive Dismissal (L6 Conj)

PleasingFungus:
A combination of the old Aura of Abjuration and Forceful Dismissal spells,
Explosive Dismissal causes all hostile summons in sight to explode, dealing
damage to themselves and everything around them. It's intended to be a spell
that makes players feel smart for turning enemies' strengths against them,
in the same way that Yara's Unravelling does for enemy buffs.

It's specifically not a Summoning spell because big explosions everywhere
play badly with allies, and the Ignition trick of not damaging allies felt
weird. I considered making it Cj/Hex or Cj/Tmut, but just Cj seemed fine.

A variation on this idea would be a spell that deals damage to summons &
their summoners, rather than messing with explosions to *probably* hit the
summoner. That seemed less fun, though.

It's possible that this ends up playing too much like Ignition in practice.
I'm not sure! It seemed interesting enough to implement.


ebering:
Kate just made Yara's explode summons in trunk, riffing on a similar idea. With that done do you think this spell has a place?

PleasingFungus:
Great question! There are two ideas I was playing with here:
1. An abjuration-like effect to counter summons.
2. A way to make the player feel clever and "punish" summoners, in the same way that Yara's already let you "punish" enemies for buffing themselves.

I think the Yara's changes cover the former fine, but I do still think there's room to play with the latter idea. I don't know exactly what that would look like, though, and I don't think it'd look too much like this implementation, so I'll close this for now.

PR #2069, opened 2021-08-15
New god: Ignis, the Dying Flame

PleasingFungus:
DCSS once had many zealot starts, but we've removed most over the years. Choosing a god is an exciting strategic moment in Dungeon Crawl games, and most gods aren't so distinctive on d:1 that it's worthwhile to trade away that moment of choice.

So, what if we could have both?

Ignis is a dying god. Its worshippers have died or abandoned it for the services of other gods; its flame burns low. If you choose to worship Ignis - either by selecting the new Acolyte of Flame start or by finding an ecumenical altar dedicated to it - all the power remaining to the god will be dedicated to you straight from the start, putting you immediately at 5* of piety (150 - more for monks and/or ecu altar followers). But... there's no way to ever get more piety. Sooner or later, you'll have used up everything Ignis has left, and you'll probably want to risk Ignis' (weaker-than-usual) wrath and swap gods.

## Powers

Ignis offers three active powers. Each consume a huge amount of piety, so you get a fairly limited number of uses. They're intended to not really scale beyond the early game, and accordingly don't scale with Invocations at all.

### Foxfire Swarm

Intended to be Ignis's lowest-level ability, Foxfire Swarm fills every open space within radius 2 of the player with friendly foxfires. It costs 8-12 piety, meaning it can be used on average about a dozen times before the player drops below 1* piety and can no longer use it.

I suspect this one is currently extremely overtuned and will need to be nerfed considerably before it's merged into trunk.

### Sea of Fire

Sea of Fire fills every open space in *sight* of the player with smoldering coals, which ignite into flame clouds before the start of the player's next turn. (Monsters that move before the player's next turn will stamp out the coals, just like with Conjure Flame, but unlike Conjure Flame, the flames will burn before the player gets another turn.) It costs 16-24 piety, meaning it can be used about half a dozen times before the player drops below 1* piety.

### Rising Flame

Rising Flame applies a 'rising' duration to the player, which, after 2-4 turns, sends them blasting through the ceiling, 'reverse shafting' to the floor above. It costs 32-48 piety, meaning it can be used about 3 times before the player drops below 2* piety and can no longer use it.

I want this to be the most expensive ability since i think it's the least 'exciting' for the player to use. Players generally prefer to spend their time attacking monsters, so even though escapes are very strong, they're something we should expect them to use rarely but memorably. If it's too weak compared to the other abilities, we can try removing the delay or otherwise messing with it.

### Passive: Inner Warmth

Even at 0*, Ignis still provides passive rF+ to the player. This is mainly intended to make polytheist players (who want to take Ignis all the way to a win) feel better, but it also synergizes with Sea of Fire to some extent.

### Goals

Since Ignis is a god which players are expected to abandon, the wrath needs to be tuned more tightly than with most gods. It should be:
- **Very scary at xl2**, so that it's not right to just burn through Ignis immediately and switch to the first altar you find.
- **A reasonable risk to take around ~xl10-12**, so that people will actually swap, but still noticeable (not a complete pushover even by then).
- **Entertaining** - since players will be experiencing this quite a lot, it should be a fun and memorable experience.

### Effects

The current draft of the wrath is quite rough. It throws the following effects at the player:
- Passive: as long as the player is under Ignis wrath, their rF is capped at 0. (That is, they can get rF- from fire vuln, but they cannot get rF+ or higher in any way.)
- Summon fire elementals.
- Shaft the player.
- TODO: apply 'fire champion' ench to the toughest monster in sight (AF_FIRE and... increased move speed..?)

This would benefit from more brainstorming and thought, though.

### Aftermath

When Ignis is 'mollified', that's the end. Like Jiyva, with no worshippers left and the last of its flame spent in a suicidal attempt at revenge, Ignis perishes. Its altars (if any exist) vanish, and it cannot be found or worshipped again.

(This also prevents trivial polytheist runs, not that it really matters. It's mostly flavour.)

## Background

Ignis is not found in the Temple, so there are only two ways to follow it: a lucky ecumenical altar, or starting as an Acolyte of Flame.

Acolytes of Flame are zealots that start with a basic weapon choice (which, in xom-like fashion, get a SPWPN_FIRE and a negative enchant applied) and a robe.

Experimentally, Acolytes of Flame currently also have a 6/6/0 strength/int/dex spread and get knowledge of Foxfire and some Fire Magic skill. I didn't want to add yet another 'melee background', so I wanted to give the option to choose to either go melee or mage, depending on what you find. That might not actually work, though!

## Miscellany

- If you abandon Ignis and re-worship *before* mollifying (i.e. by going back to the ecu altar you got Ignis from), your piety will reset back to what it was when you abandoned, not back to 5*. No freebies!
- Ignis has no piety decay, since it'd interact a bit awkwardly with people who wanted to play it to the end of the game for polytheist.
- Amulets of faith don't affect Ignis worshippers. Maybe they should provide a discount? I don't feel strongly about it.
- The console colours overlap a bit with other fiery gods. So much fire!
- Rising Flame tries to create a shaft underneath the player when they arrive purely for flavour reasons - you blew through the floor! It won't land you back where you started if you take it, though, more's the pity.

## Remaining TODOs
- **Balance!** (abilities, wrath, and zealot start)
- Don’t choose shaft wrath on unshaftable levels
- Job recommendations (to)
- Job playerdoll tiles
- Tiles (altar, god, abilities, background...)

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -633,13 +633,13 @@ static vector<ability_def> &_get_ability_list()
             0, 0, 0, {}, abflag::berserk_ok },
 
         // Ignis
-        { ABIL_IGNIS_FOXFIRE, "Foxfire Swarm",
-            0, 0, 8 /*avg 12 uses from 150 to <30 piety*/,
-            {fail_basis::invo}, abflag::quiet_fail },
         { ABIL_IGNIS_SEA_OF_FIRE, "Sea of Fire",
-            0, 0, 16, {fail_basis::invo}, abflag::quiet_fail },
+            0, 0, 6, /*avg 15 uses from 150 to <30 piety*/
+            {fail_basis::invo}, abflag::quiet_fail },
+        { ABIL_IGNIS_FOXFIRE, "Foxfire Swarm",
+            0, 0, 9, {fail_basis::invo}, abflag::quiet_fail },
         { ABIL_IGNIS_RISING_FLAME, "Rising Flame",
-            0, 0, 32, {fail_basis::invo}, abflag::none },
+            0, 0, 24, {fail_basis::invo}, abflag::none },
 
         { ABIL_STOP_RECALL, "Stop Recall", 0, 0, 0, {fail_basis::invo}, abflag::none },
         { ABIL_RENOUNCE_RELIGION, "Renounce Religion",
--- a/crawl-ref/source/ability-type.h
+++ b/crawl-ref/source/ability-type.h
@@ -277,8 +277,8 @@ enum ability_type
     ABIL_WU_JIAN_WALLJUMP,
 
     // Ignis
+    ABIL_IGNIS_SEA_OF_FIRE = 1300,
     ABIL_IGNIS_FOXFIRE,
-    ABIL_IGNIS_SEA_OF_FIRE,
     ABIL_IGNIS_RISING_FLAME,
 
     // For both Yred and Beogh
--- a/crawl-ref/source/dat/descript/gods.txt
+++ b/crawl-ref/source/dat/descript/gods.txt
@@ -303,12 +303,11 @@ their life's talents.
 %%%%
 Ignis powers
 
-Followers can draw upon the guttering flames of Ignis' power just a few times
-before only flickering embers remain. They can call a swarm of foxfires against
-their foes perhaps a dozen times, fill their surroundings with a sea of fire
-perhaps half a dozen times, or let Ignis' heat lift them to another level of
-the dungeon perhaps three times. No matter how depleted, Ignis will always
-grant followers resistance to fire.
+Followers can fill their surroundings with a sea of fire, can call a swarm of
+foxfires against their foes, or let Ignis' heat lift them to another level of
+the dungeon. However, the guttering flames of Ignis' power will only sustain a
+limited number of invocations before only flickering embers remain. No matter
+how depleted, Ignis will always grant followers resistance to fire.
 %%%%
 Jiyva powers
 
--- a/crawl-ref/source/religion.cc
+++ b/crawl-ref/source/religion.cc
@@ -391,9 +391,9 @@ const vector<god_power> god_powers[NUM_GODS] =
     },
 
     // Ignis
-    { // TODO
-        { 1, ABIL_IGNIS_FOXFIRE, "call a swarm of foxfires against your foes" },
+    {
         { 1, ABIL_IGNIS_SEA_OF_FIRE, "fill your surroundings with clouds of flame" },
+        { 1, ABIL_IGNIS_FOXFIRE, "call a swarm of foxfires against your foes" },
         { 2, ABIL_IGNIS_RISING_FLAME, "rocket upward and away" },
     },
 };
--- a/crawl-ref/source/spl-summoning.cc
+++ b/crawl-ref/source/spl-summoning.cc
@@ -3050,7 +3050,7 @@ spret foxfire_swarm()
     bool unknown_unseen = false;
     for (radius_iterator ri(you.pos(), 2, C_SQUARE, LOS_NO_TRANS); ri; ++ri)
     {
-        if (_create_foxfire(you, *ri, GOD_NO_GOD, 15))
+        if (_create_foxfire(you, *ri, GOD_NO_GOD, 10))
         {
             created = true;
             continue;


perryprog:
I gotta admit, this sounds super cool to me from a lore perspective (I don't know from a gameplay or mechanical perspective how most of this would work out, I'm very bad and newish at DCSS). I do have a couple of thoughts for some potential areas for flare, pun intended.
* Instead of destroying Ignis's altars when he's mollified, maybe replace them with some sort of innate ruin, to serve as a reminder to what was previously there? If Ignis is already "weak", it'd make sense for the altars to already be a reflection of that—maybe the altar could visually have a flame that is directly proportional to his presence in the realm.
* Considering how his abilities are guaranteed to be limited use, it might make sense to have a significantly lower failure rate—this might also make sense in terms of how much "attention" Ignis would be giving the player, as you're basically his last hope at survival.
* Amulet of Faith could probably reduce the piety usage of his abilities by some constant factor. Maybe could make it an option for people who want to actively use him past where he's "designed" to be used?
* Rising Flame could probably have some sort of "explosion" fiery thing done to any monsters nearby when it's used. Doesn't need to be too impactful, but that just makes a cool mental image for me.
* I feel like Ignis's wrath should almost be made "personal"—my imagining of it is Ignis is viewing you as his last hope, and you forsaking him would be something that could hurt him emotionally. In game this could maybe come through as some sort of internal dialog you hear of Ignis berating you, or some sort of hallucination/vision of Ignis himself before he finally poofs. Again, not too sure how this would work out gameplay wise, as I'm not sure if any of that is actually feasible.

PleasingFungus:
Hi! Thanks for the ideas! I unfortunately don't have time to go into an in-depth response right now, but a few small notes:

> Considering how his abilities are guaranteed to be limited use, it might make sense to have a significantly lower failure rate...

Yep! The failure rate is 0.

> Amulet of Faith could probably reduce the piety usage of his abilities by some constant factor...

Yeah, I had some thoughts along those lines. There's a note to that effect somewhere in the PR (either in a TODO comment or a commit message, I forget). I haven't decided about it, but it's a reasonable idea.

> maybe the altar could visually have a flame that is directly proportional to his presence in the realm.

This is cool! Sort of hard to justify the extra work when (a) you rarely re-visit altars anyway and (b) most worshippers won't even have a single altar in the game (since altars will only exist if you roll one through a faded altar). But it's a very thematic idea!

perryprog:
> This is cool! Sort of hard to justify the extra work when (a) you rarely re-visit altars anyway and (b) most worshippers won't even have a single altar in the game (since altars will only exist if you roll one through a faded altar). But it's a very thematic idea!

In that case, maybe it could be re-worked into a method of some kind of limited piety building—maybe an ability that, at the cost of some finite resource (or maybe gold), would place a small shrine to Ignis. This would give you a (very) small amount of piety, and this would be the object that'd have the indication of how much of his "divine essence" is left. If he dies, these would turn into the sort of ruined monuments I suggested earlier. This would also give a bit of background on how all the other shrines could have originally appeared—they were, at one point, constructed by other dungeon delvers on the same quest you are.

The one thing I don't want this mechanic to ruin is the "limited use" aspect of this god that's so unique. The best way I can think to keep that is to either not do this at all (and it is a lot of extra work for not too much output) or to have a severe amount of diminishing returns on how much piety this altar-construction ability would grant you. Hopefully that makes sense.

ebering:
> I feel like Ignis's wrath should almost be made "personal"—my imagining of it is Ignis is viewing you as his last hope, and you forsaking him would be something that could hurt him emotionally. In game this could maybe come through as some sort of internal dialog you hear of Ignis berating you, or some sort of hallucination/vision of Ignis himself before he finally poofs. Again, not too sure how this would work out gameplay wise, as I'm not sure if any of that is actually feasible.

@perryprog as far as "a dialog" goes, crawl supports pure flavour god speech in the form of [a god speech database](https://github.com/crawl/crawl/blob/master/crawl-ref/source/dat/database/godspeak.txt). If you've got ideas for a set of Ignis wrath lines elaborating on your imagined dialog we'd be happy to include them! The text generator supports certain variables so the lines can be filled in like mad-libs, [for details see the editing guide](https://github.com/crawl/crawl/blob/master/crawl-ref/docs/develop/monster_speech.txt). The guide is for monster speech but god speech uses the same system. Feel free to stop by #crawl-dev on libera.chat if you have questions.

A visible avatar would be more difficult, since players expect monsters on the grid to have some interactivity.

crawl-ref/source/stairs.cc
```diff
@@ -636,6 +638,42 @@ static bool _level_transition_moves_player(level_id dest,
...
+        mpr("In a burst of heat and light, you rocket briefly upward... but you can't rise from here.");
+        return;
+    }
+
+    mpr("With a burst of heat and light, you rocket upward!");
+    untag_followers(); // XXX: is this needed?
+    floor_transition(DNGN_ALTAR_IGNIS /*hack*/, DNGN_ALTAR_IGNIS,
+                     whither, true, true, false, false);
+
+    // flavour! blow a hole through the floor
```
ebering, line 672: for more flavour: place an Ignis altar at the departure point? Riffing on @perryprog here.

crawl-ref/source/ability.cc
```diff
@@ -646,6 +648,15 @@ static vector<ability_def> &_get_ability_list()
...
             0, 0, 0, {}, abflag::berserk_ok },
 
+        // Ignis
+        { ABIL_IGNIS_SEA_OF_FIRE, "Sea of Fire",
+            0, 0, 8, /*avg 12 uses from 150 to <30 piety*/
+            {fail_basis::invo}, abflag::quiet_fail },
+        { ABIL_IGNIS_FOXFIRE, "Foxfire Swarm",
+            0, 0, 16, {fail_basis::invo}, abflag::quiet_fail },
+        { ABIL_IGNIS_RISING_FLAME, "Rising Flame",
+            0, 0, 32, {fail_basis::invo}, abflag::none },
```
ebering, line 658: These costs feel about right; however in the `a`bility menu they both are listed as `extremely large`. A better indication of their relative cost in that menu would be nice, so that it's not only stuck in the second pane of `^`.

ebering (commented):
I played a HuAF and a MiAF to the point of mollifying Ignis (or death in the first case). Both were fun, I could probably stand to think a bit more about sea of fire use cases. I think it has some good ones for its price but I didn't totally puzzle out the secret tech.

The HuAF died to wrath; I was checking what had targeters and burned most of my piety on an unnecessary rising flame, and switched very early. Fire elementals were pretty scary.

The MiAF mollified Ignis successfully through lair. The wrath felt about right, scary enough to require some doing but not so bad it felt like a mistake. The stat split was not so noticeable on a Mi, this one had plate just outside the entry vault and promptly forgot about foxfire.

In both games I kept looking at the action table in my dump to try to puzzle out where I stood piety-wise. A bit more UI info would be helpful here, either in the form of Ru-like "after you use this ability your piety will be: stars" (with randomization I understand that can't exactly be the solution), or maybe some indicator like "If you used all of Ignis' remaining power on this ability you could use it n more times".

Some quick interface notes from playtesting:

- rising flame shows as available in the quiver even on D:1.
- there are no targeters for foxfire swarm and sea of flame.
- status light or better indication that gaining rf is impossible during wrath (rip 2 my potion of resistance the first time I got fire elemental wrath)
- `^O` layout looks a bit weird in console? Not sure if anything can be done about that

crawl-ref/source/god-wrath.cc
```diff
@@ -2012,6 +2014,62 @@ static bool _wu_jian_retribution()
...
+{
+    return random_choose(
+        " says: my hands, my teeth, my burning tears: these will never betray me!",
+        " says: attack, my last stalwart servants!",
+        " wails with rage, and rains down burning fury!",
+        " calls forth spirits of fire to unmake you!",
+        " says: burn! Burn until nothing is left!",
+        " says: these will know how to inflict the punishment due to you!",
+        " says: with a thousand fires I burned cities, but these will suffice for you!",
+        " cries: Betrayer! Apostate! Perish!");
```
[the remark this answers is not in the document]
PleasingFungus: Yeah, that was planned but I forgot to add a TODO for it. Will do. (Forgot we already had a god speak specific DB; I was afraid I'd have to make a new one.)

crawl-ref/source/stairs.cc
PleasingFungus: Interesting ideas, but it feels odd to do things at the departure point (since the player won't see it until later), and not sure how much sense an altar makes flavour wise.

PleasingFungus:
> ^O layout looks a bit weird in console? Not sure if anything can be done about that

I looked into this. Looks like Abyssal Knight works the same way. I think it's fixable - I'd merge `dgn-overview.cc`'s `_get_altars()` calls to temple and non-temple gods into one call so that they wouldn't have to print on a new row & out of alphabetical order - but probably out of scope for this PR.

Shapermc:
Ignis Early/Quick Impressions: 
Initially Ignis sounded like a cool concept I was very interested in messing around with. After a couple hours of messing around with Ignis both as a melee support and mage support, I don’t like Ignis. I might come to like Ignis, but as it stands Ignis feels like a time sink w/o payoff. Hopefully I can explain that through some feedback points: 

•	Spells feel like they’re in “the wrong order” since they don’t go from least to most expensive (Sea/Swarm/Rising would make more sense for order). 
•	Starting “book” is pretty light, forcing casters to branch much sooner, and are more at the whim of the floor god placing books. Adding conj flame seems to fit with the overall theme, and should be helpful, though maybe too on the nose. I also have an idea below for extending the spell portion of Ignis too. 
•	Currently there’s no “invocation synergy” for Ignis. This feels like a missed opportunity, and I think that training invo could lower the cost of the abilities. This would make two things happen: extend the life of Ignis (needed), and helps you build something to take forward (currently missing). 
•	Sea of flame isn’t fun to use. It has all the “problems” cj flame has (too much setup) and it punishes players who screw it up (traps you in place, can’t even blink w/o taking damage). I see possible options to solve this being a) change to clouds, or b) give temp flame immunity during the sea’s duration (also giving the player a second use for sea of flames as a short term fire immunity buff). 
•	Wrath if very harsh if you abandon Ignis early. I literally laughed at the screen when I got 4 durably summoned fire elementals on like d7 or something. It just seemed impossible. I also heard about players getting double shafted within like 20 or so turns of itself. 
•	You use up your abilities too fast. Depending on the order you cast things at you can get as few as 4 uses of the god’s abilities overall before they’re exhausted. At max you can maybe get 7-8 uses of abilities. 

So, those are the issues I’ve seen with some preliminary poking around. The “distinctive on D1” goal of the god is met, but at what cost? Ignis feels like starting with a *debt* you have to repay to the game. You get a negative enchanted weapon, and essentially a pile of panic buttons you can press, but not very often. And you’re saddled with the burden to dump this god at the right time and not earlier. Likely this will put many characters in bad spots when they dump, so that also will need to be strategic, and likely a hard decision. It will be interesting to get a player who’s suddenly weak in lair 4 or something. 

Which leads to the biggest problem: So Ignis got you through d1-d10 and a little bit of lair… now what? Abandon god and get wrathed? Keep god and get nothing, and build towards nothing (i.e. piety)? These aren’t exciting positions to be in. The cost of the abilities are also so high it’s easy to use those panic buttons early and be left with no piety well before you’re able to survive abandoning Ignis. 

I played a few different games leaning melee and a few leaning caster. I did play the casters a bit more hybrid, but honestly it was both times due to lack of variety in spells and finding a good weapon on the ground. Ignis feels like they support both roles well, and starts you out with what I’d consider great starting stats for a hybrid. It didn’t feel like either were too weak, though I’m sure Melee dudes will feel like the skill points in spellcasting/fire magiks are “a waste” but it’s fine. Possible consideration to drive home the “hybrid feel” could be to start the player in animal skins, don’t know why but robe screams caster to me :woman_shrugging:  

I’m no game dev, but I like thinking about Crawl, so here’s some ideas to help ease some of the burden of Ignis’ debt: 
•	Make invocations decrease the piety cost of the spells. Power creep is a real thing and scaling damage or something on invocations would break it probably, but if a player can build up invo and hold off on using the panic button a bit, this would likely extend out the life of the god. 
•	Give players something that is more permanent from the god. As stated the invocation could be something that carry’s over. Possibly a one-time free branding of a weapon with flame, but only when below 2* piety, which would make carrying around a “useless god” more rewarding (and it doesn’t make wrath any easier). Maybe just Ignis grants a fire spell every other level ala Djinni. 
•	Doing something to give the players more abilities to use. Either scaling the power so that the cost can be cheaper and more frequently used, or possibly one more super cheap spell that is a bit more frequently usable. 

I wanted to give a full writeup on my thoughts as thanks for allowing me to participate in the open testing. As it stands I realized with the last character I was playing that I only had one use of an ability left and I wasn’t even to lair yet. Made me feel like I was about to have the burden of figuring out how to dump this hard to deal with god and then re-building everything from the start. That’s why I said that Ignis feels like a Debt you have to pay off. As it stands I likely won’t be playing much Ignis after a couple hours of testing. 

If you have any questions, just let me know, happy to expand on anything.

ciroH:
It would be very cool if an Ignis altar was generated in Ice caves, the best zone for a weakened fire god.

Jackiecrazy:
a thought: uniques are a non-renewable resource that are intended to be a challenge, perhaps some piety can be regained by killing those?

PleasingFungus:
Ciro: Ice Caves generally generate around when Ignis is supposed to fall off and be replaced, sadly.

Jackie: I do like the idea of tracking unique kills for some purpose! I had the idea a while back of a "Nametaker" spell that levelled up by killing uniques - maybe someday I'll come back to it.

Zap-zapper:
https://www.youtube.com/watch?v=vI0FETSHAv4

_Alone with this vision
Alone with this sound
Alone in my dreams
I carry around
I will not take from you and you will not owe
I will protect you from the fire below
It's not in my mind
It's here by my side
Go tell the world that I'm still alive
Alone with this vision
Alone and blind
Go tell the world that i'm still alive..._

PR #2072, opened 2021-08-16
Six new wizlabs + assorted wizlab updates

hellmonk3:
Adds new wizlabs for Maxwell, Yara, Borgnjnor, and Alistair, plus revives modified versions of Eringya and Tukima, which were removed in 0.17. These are still a bit wip at the moment and could use a lot of new tiles, but they're in a finished enough state that I'm opening this pr for balance and design feedback. As part of the new wizlabs, I added several monsters, mostly courtesy of PleasingFungus, and a new cloud type. This pr also removes the golem set from Tomb of Doroklohe (because those monsters are used in Maxwell) and adds a few jobbed draconians to the end of Iskenderun (to make that wizlab a bit more dynamic).


ebering:
Ran these with a wizmoded DgSu based on my current trunk game, which had just finished 2 runes.

All felt like good challenges, required consumable uses and the occasional reset. This particular char was really poorly suited to face Maxwell (an lbolter trying to airstrike constructs).

Tukima's felt a bit light on monster density, but that is possibly because I did it first and was much more careful with noise versus the dancing weapons.

I didn't look closely at the loot but nothing struck me as deeply disappointing. I missed several of the Alistair storylines because I was casting lbolt and mcc outside the ~~frat house~~ mansion but oh well.

Seems like these are in a good place to clean up the commit history and windmill smash the merge button.

Gonna give putrid mouths both their miasma breath and stat drain breath when you do.

PR #2074, opened 2021-08-18
Add a few more actions for Okawaru Duel spectators

RojjaCebolla:
Turns out that `_fruit_name_` was removed, so i had to expand the scope of this PR to bring those back!


ebering:
Great flavours, thanks for the patch!

PR #2076, opened 2021-08-19
Simplify the ranger card

sdynet:
There are too many monsters with similar roles on this card. I want to cut out too many branches and leave only those who match the name of this card.

Modified Effect: Now it summon only these three(centaur warrior, naga sharpshooter, deep elf master archer).


PleasingFungus:
Seems reasonable to me. Thanks for this!

PR #2082, opened 2021-08-21
prompt for auto-travel to purchasable shopping list item

elliottbernstein:
When a shopping list item becomes purchasable, a non-blocking and easy-to-miss message is currently displayed. 

This change asks the player for a Y/n response for whether or not they want to auto-travel back to the shop where each newly purchasable item is located.

### DIFF
--- a/crawl-ref/source/shopping.cc
+++ b/crawl-ref/source/shopping.cc
@@ -2021,7 +2021,6 @@ void ShoppingList::gold_changed(int old_amount, int new_amount)
     {
         ASSERT(min_unbuyable_idx < list->size());
 
-        vector<string> descs;
         for (unsigned int i = min_unbuyable_idx; i < list->size(); i++)
         {
             const CrawlHashTable &thing = (*list)[i];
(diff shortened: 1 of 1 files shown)


ebering:
Commit c0b50483a added the gold message to the default explore stop. Is a prompt to travel immediately really necessary?

Prompts are modal and interrupt the player's flow (they can't resume auto explore immediately with o) and the shopping list already has travel functionality.

elliottbernstein:
Glad to see someone else addressing this as well. Based on what you described, that might be enough. If so, I'll come back and close out this PR.

elliottbernstein:
@ebering I'm still blowing past this message. It's probably just my button mashing style of play, might be harder to miss while playing a non-fighter character. Is there a good way to get other players to chime in on what they'd prefer between the two behaviors?

ebering:
Have you added a personal force more for this particular message in your rc? 

Crawl is configurable, but also turn-based. We don't set the defaults for rapid button mashing, which generally wants more modal interruptions (which change input handling and demand specific inputs) and less nonmodal gentle measures (like the default explore stops).

elliottbernstein:
I have not edited my RC in that way yet. But your comment reminded me, I could put  the behavior in my PR behind a feature flag.

rawlins:
I don't think this PR would work for general use as-is. I'll take the two parts, interruption/alerting and autotravel in turn:

Interruption/alerting: `yesno` should only be used for relatively extreme things, exactly because it is quite intrusive and can't be very easily overridden (I believe it can with custom lua). You can get a sense for what it is currently used for by checking out `git grep yesno` but some examples include wasting an item, abandoning a religion, targeting yourself, etc. Basically things that can't be undone. I don't think missing a shopping list item fits.

If shopping list messages were to trigger an alert by default beyond stopping autoexplore, the first thing to try would be adding it to the force_more list, which as ebering is pointing out is easy enough to do in your own rc. The syntax is: `force_more_message += You now have enough gold to`. In principle this could also be added as a default force_more (in `dat/defaults/messages.txt`.) However, I'm not even sure this is a great idea, I suspect most players would prefer to not get a `more` for this case. But I haven't done anything resembling a poll, either, so maybe I'm wrong.

Autotravel: improving autotravel to the most recent indicated shopping list seems interesting but I think it would need a different UI, because of the issues with using `yesno`. Here I would want to improve the `$` menu or perhaps add something to the autotravel interface that makes it easy to get to the specific item. Some possibilities (there may be more, this is just brainstorming):
* Indicate more clearly which item was just mentioned in `$`
* Set the next default travel target for `G` if there isn't ongoing travel to the location of the shop
* Add something new to `G`, where e.g. `G$` goes to the most recent shopping list target. (Though this maybe gets a bit confusing with the autotravel features built into `$`.)

Any of these ideas involves about 1 more keystroke than the current PR, but also doesn't assume the player should immediately deal with going to the shop, and so is less intrusive (even with a force_more for this case I think).

I won't close this yet in case you are interested in reworking this PR towards an idea like one of those in the previous paragraph.

elliottbernstein:
Thanks for the suggestions. I actually found a solution that works well for getting my attention; I added the message to `flash_screen_message`. I'll close out this PR!

PR #2090, opened 2021-08-24
Fix `target_desc` crash bug

mgdelmonte:
This helper function was created to assist with the clua function `target_spell`, but it crashes if `targeter_addl_desc` doesn't return a function for getting more info.  Easy to reproduce by testing `target_spell("Foxfire")` on any monster.


ebering:
Hi, I'm not really sure what your git process is but your pull requests always seem to have very convoluted histories.

I recommend starting a fresh branch from crawl/master for each pull request if at all possible. Additionally, when maintaining your local copy you can apply local changes after upstream changes by using `git pull --rebase`.

PR #2094, opened 2021-08-26
Update package name for DejaVu font on Ubuntu

RojjaCebolla:
It's been unavailable under the old name for awhile now


ebering:
We indeed only use -core. Thanks for pointing this out!

PR #2098, opened 2021-08-29
New spell: Scorch (L2 Fire)

PleasingFungus:
Scorch is a damage spell that strikes a random enemy in sight, dealing fire
damage to them and (if it beats AC & resists) applying Cerebov's rF- debuff
to them for a few turns. Currently, it does 2d5 damage at 0 power, scaling to
2d10 at 50 power.

Scorch is intended to provide additional variety to the Fire Elementalist
early game (reducing reliance on Foxfire) and to make Cinder Acolyte feel
more distinctive from FE by replacing its Foxfire entirely.

Open questions:
- Is the 'random targeting' gimmick actually fun, or should the player have
  a bit more control over what they bap?
- Would the spell feel more fun with a limit on range (& perhaps some
  compensatory buff)?
- Are the numbers reasonable?

TODOs:
- Add the spell to books properly.
- Add a message when spells become castable due to level up, if there isn't already such a message.

### DIFF
--- a/crawl-ref/source/book-data.h
+++ b/crawl-ref/source/book-data.h
@@ -484,7 +484,7 @@ static const vector<spell_type> spellbook_templates[] =
 },
 
 { // Book of Scorching
-    SPELL_FOXFIRE,
+    SPELL_SCORCH,
     SPELL_FIREBALL,
     SPELL_SUMMON_CACTUS,
 },
--- a/crawl-ref/source/dat/descript/spells.txt
+++ b/crawl-ref/source/dat/descript/spells.txt
@@ -1184,6 +1184,11 @@ Scattershot spell
 Blasts a cone-shaped area in front of the caster with metallic fragments. Its
 damage is strongly reduced by armour.
 %%%%
+Scorch spell
+
+Scorches a random foe. Those injured by this have their resistance to fire
+burned away for a short time.
+%%%%
 Seal Doors spell
 
 Imparts seals on nearby doors and stairs, preventing those not allied with
--- a/crawl-ref/source/god-abil.cc
+++ b/crawl-ref/source/god-abil.cc
@@ -1401,7 +1401,8 @@ bool vehumet_supports_spell(spell_type spell)
         || spell == SPELL_FROZEN_RAMPARTS
         || spell == SPELL_MAXWELLS_COUPLING
         || spell == SPELL_NOXIOUS_BOG
-        || spell == SPELL_POISONOUS_VAPOURS)
+        || spell == SPELL_POISONOUS_VAPOURS
+        || spell == SPELL_SCORCH)
     {
         return true;
     }
--- a/crawl-ref/source/job-data.h
+++ b/crawl-ref/source/job-data.h
@@ -130,13 +130,13 @@ static const map<job_type, job_def> job_data =
 
 { JOB_CINDER_ACOLYTE, {
     "CA", "Cinder Acolyte",
-    7, 5, 0,
+    6, 6, 0,
     { SP_HILL_ORC, SP_BASE_DRACONIAN, SP_OGRE, SP_DJINNI, SP_GNOLL },
-    { SPELL_FOXFIRE },
+    { SPELL_SCORCH },
     { "robe" },
     WCHOICE_PLAIN,
     { { SK_FIGHTING, 3 }, { SK_WEAPON, 3 },
-      { SK_FIRE_MAGIC, 1 }, {SK_SPELLCASTING, 1} },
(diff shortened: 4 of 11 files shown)


ebering:
Here's a version of a grid I used in thinking about positional magic and ensuring the diversity of the elemental starting backgrounds (which I should maybe put in docs/develop/spells but it'll just drift out of date). It compares the 4 elemental backgrounds and Cj. Explanation of the attributes:

- *Ranged* does damage at range. for emphasis range is >= 4
- *Instant* does damage immediately
- *Corridor* works well in corridors
- *Accurate* Guaranteed hit, levels annotated with (r) for "ranged accurate"
- *DOT* does continuous damage over time from one cast


Bg  | Ranged | Instant | Accurate | DOT | Corridor | Defense
--- | ------ | ------- | -------- | --- | -------- | -------
AE | 1,4  | 1,2,4 | 2,4(r)   |     | 1      | 3
FE | 1,5    | 5       | 1(\*),4,5(r) | 3, 4 | 4,5 | 3
IE |        | 1       | 1,3,4    | 3,4 | 1,3 | 3,4
EE | 3    | 1,3   |         |     | 1,3   | 2,4
Cj | 1,2,4  | 1,4     | 1,2,4    | 2   | 1,2,4    | 3

To make the very early game a bit clearer, I've repeated the chart with just levels 1-3:

Bg  | Ranged | Instant | Accurate | DOT | Corridor | Defense
--- | ------ | ------- | -------- | --- | -------- | -------
AE | 1      | 1,2     | 2        |     | 1,2      | 3
FE | 1      |         | 1(\*)    | 3   | 3(\*\*)  | 3
IE |        | 1       | 1,3      | 3   | 1,3      | 3
EE | 3      | 1,3     |          |     | 1,3      | 2
Cj | 1,2    | 1       | 1        | 2   | 1,2      | 3

While foxfires do sometimes wander off or select an inappropriate target, I'm counting them as accurate as they are signifcantly more accurate than sandblast, stone arrow, shock or searing ray. I put two \* on the corridor for conjure flame, because it's a corridor spell but not like the other ones in that column; one could easily make the case that FE still has a hole there. Every bg except Cj has "holes" in the first 3 spell levels, which are only partially filled by level 4. 

When I made my "hang on" irc remark I was concerned that Scorch would fill in too many of the holes. It does fill in the Instant and Corridor spots somewhat. So, with this framework in mind here's my take on answers to your questions after trying early DrFE, HuFE, and FeFE:

- Is the 'random targeting' gimmick actually fun, or should the player have a bit more control over what they bap?

I don't think that it should be targetable (even with dodge) as that would make it even more "Corridor". Additionally, the random targeting fits with fire's vague vibe from the positional magic redesign. It helps more than foxfire in the emergency of "surrounded in a corridor", but without range and targeting it doesn't fully get an early FE into corridor fighting mode.

- Would the spell feel more fun with a limit on range (& perhaps some
compensatory buff)?

Tested it at range 3, felt good with the constrained range and damage values.

- Are the numbers reasonable?

With its current damage values, at the early stage of the game for an FE it seems to be comparable to foxfire unless you lopsidedly train fire magic over conjurations. Even then, it looks like it won't out-damage foxfire until after foxfire reaches max power---a good spot IMO. This property helps it not push FE too much into corridor mode early: foxfire is half the price for more damage sides! Because of how AC interacts with foxfire Sear is better for punching through armour, but enough early monsters have low AC that I don't think using Sear in a corridor always beats using Foxfire from an effective position.

So, I waive my initial reservations. This is a cool spell.

Gotta see how it plays with the CA start but don't wait to hear back from me there if you want to merge since CA is already very different from the mage backgrounds.

PleasingFungus:
Extremely cool chart!

> It helps more than foxfire in the emergency of "surrounded in a corridor", but without range and targeting it doesn't fully get an early FE into corridor fighting mode.

Right - I was thinking of 'corridor' being "the ability to fight in corridors", but if it's "being *incentivized* to fight in corridors* then I def agree Scorch doesn't really push that far corridorwise.

> I don't think that it should be targetable (even with dodge)...

Right, yes - I'd imagined the alternative being some other targeting gimmick, e.g. "targets the strongest enemy [for some definition of strongest]" or somesuch. Though I'm biased, I agree it feels good at range 3 - I actually think the range limit makes it feel better, by allowing a little more control over what you're going to hit.

> Gotta see how it plays with the CA start but don't wait to hear back from me there if you want to merge since CA is already very different from the mage backgrounds.

I playtested a little and it seemed quite respectable. XL 1 is a little rough (esp for DECA) but you've got Ignis so who cares.

Thanks again for the in-depth thoughts!

PR #2100, opened 2021-08-30
Improve wanderers

hellmonk3:
Wn is a really cool "random start" option, but has a couple annoyances. You can get some truly awful starts due to the way skill points are assigned, and are basically guaranteed to have awful stats due to the way stat points are assigned. This pr tries to rectify those problems and adds some new things that wn can get at game start. 

The new model assigns the equipment first, based entirely on species apts, then assigns stats and skills afterward depending on what we got. It tends to lead to more lopsided (ie useful) stat distributions and higher levels of skill in things that are useful, though these are by no means guaranteed.

Some new cool options I added:
- melee "good item" can get an upgraded base type or vorpal brand instead of enchant
- "good armour" can roll chain and "good shield" can roll enchanted buckler or prot buckler instead of kite shield
- "good magic school" can roll a 4 spell randbook rather than a background starting book
- wn that start with the sandblast spell get some stones
- it is possible to start with a hunting sling and sling skill
- evocation items can get some additional wand types
- throwing can start with boomerangs

I tested this primarily by rolling sets of 10 huwn, so it'd be wise to look at some more esoteric species before smashing that mf merge button.

### DIFF
--- a/crawl-ref/source/ng-wanderer.cc
+++ b/crawl-ref/source/ng-wanderer.cc
@@ -3,8 +3,10 @@
 #include "ng-wanderer.h"
 
 #include "item-prop.h"
+#include "item-use.h"
 #include "items.h"
 #include "jobs.h"
+#include "newgame.h"
 #include "ng-setup.h"
 #include "potion-type.h"
 #include "randbook.h"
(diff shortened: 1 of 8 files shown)


crawl-ref/source/ng-wanderer.cc
```diff
@@ -536,6 +636,10 @@ static vector<spell_type> _wanderer_good_equipment(skill_type & skill)
     case SK_EARTH_MAGIC:
     case SK_POISON_MAGIC:
     case SK_HEXES:
+        if (one_chance_in(3))
+            return _give_wanderer_major_spells(skill);
+    // deliberate fallthrough
```
PleasingFungus, line 641: Let's label this `// deliberate fallthrough to SK_SPELLCASTING`. Unlabeled fallthroughs are [dangerous](https://nethackwiki.com/wiki/Yeenoghu#.22A_ludicrous_bug.22) :)

crawl-ref/source/ng-wanderer.cc
```diff
@@ -704,6 +789,11 @@ static void _maybe_memorise_start_spell(const vector<spell_type> &spells)
             ++lvl_1s;
             to_memorise = s;
         }
+        // give stones for use with sandblast
+        if (s == SPELL_SANDBLAST)
+        {
+            newgame_make_item(OBJ_MISSILES, MI_STONE, 10 + random2avg(41, 5));
```
PleasingFungus, line 795: Ooh, fans will be excited! (Also, don't forget to run `util/unbrace` and `util/checkwhite`.)

RojjaCebolla:
what happens if ogres or other species that can't wear chain roll "good armor"?

ebering:
Made some tweaks, fixes, and adjustments:

- tilted the skill point distribution to be a bit more lopsided to the items granted (a few guaranteed points)
- added some incompatibilities to not overflow the player's hands (no shield quarterstaffing unless fo)
- guaranteed a defense skill as well as an offense (some of the triple offense rolls were too jank imo)
- guaranteed spell full wanderers have spell casting skill even if they don't roll it (see the commit comment for more

I have some further suggestions but ran out of implementation time today:

- give a chance for stealth's good item to give atropa or datura (not sure how much of either)
- figure out some interesting `decent_item` picks for at least some of dodging, stealth, fighting, and unarmed, the ones I rolled felt like they got a consumable too often? Not sure exactly what to do here though, maybe give a plus to an already picked item? Aux armour or Jewellery is probably too good?
- give big species a dragon scales or tla if they drew good armour (instead of a fallback robe of sadness)
- don't give out kite shields for good shields because of the delay changes

ebering:
Continued prodding this over the past day:

- Hand out aux armour in some places (but not too much) for dodging and stealth and oddly-shaped species
- Per @semitonal remove the job books from the possibility: they steer the character a bit too much into that job but strange, instead give a 3-spell randbook for good spells (4 but level 3 capped for spellcasting).
- The lower spell numbers were suggested by @gammafunk . With this scheme a wanderer can roll between 2 and 6 spells, more in line with the book changes we've made to other backgrounds.
- I added a bunch of new consumables to the mix
- I also gave a chance for decent fighting to gift an extra weapon plus.

PR #2102, opened 2021-08-31
Improve ratskin cloak behavior and unmark it as evil

perryprog:
- ~~Add plain rats to its spawn list~~
  - Undid this on advice from PleasingFungus, as plain rats are effectively useless
- If you follow a good god, don't spawn hell rats
- Since it now can't spawn evil allies if you follow a good god, unmark it as evil

### DIFF
--- a/crawl-ref/source/ouch.cc
+++ b/crawl-ref/source/ouch.cc
@@ -538,7 +538,10 @@ static void _maybe_spawn_rats(int dam, kill_method_type death_type)
     if (!x_chance_in_y(capped_dam, you.hp_max))
         return;
 
-    monster_type mon = coinflip() ? MONS_HELL_RAT : MONS_RIVER_RAT;
+    monster_type mon;
+    mon = is_good_god(you.religion)
+          ? MONS_RIVER_RAT
+          : coinflip() ? MONS_HELL_RAT : MONS_RIVER_RAT;
 
     mgen_data mg(mon, BEH_FRIENDLY, you.pos(), MHITYOU);
     mg.flags |= MG_FORCE_BEH; // don't mention how much it hates you before it appears
--- a/crawl-ref/source/ouch.cc
+++ b/crawl-ref/source/ouch.cc
@@ -546,12 +546,9 @@ static void _maybe_spawn_rats(int dam, kill_method_type death_type)
         if (!god_hates_monster(rat) && one_chance_in(++seen))
             mon = rat;
 
+    // If there's no valid creatures to pull from (e.g., follower of Oka), bail out.
     if (!seen)
-    {
-        // No valid creatures to pull from (e.g., follower of Oka), bail out.
-        mpr("Your cloak shudders, but is subdued by a divine force.");
         return;
-    }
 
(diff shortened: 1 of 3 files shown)


crawl-ref/source/ouch.cc
```diff
@@ -538,7 +538,10 @@ static void _maybe_spawn_rats(int dam, kill_method_type death_type)
     if (!x_chance_in_y(capped_dam, you.hp_max))
         return;
 
-    monster_type mon = coinflip() ? MONS_HELL_RAT : MONS_RIVER_RAT;
+    monster_type mon;
+    mon = is_good_god(you.religion)
+          ? MONS_RIVER_RAT
+          : coinflip() ? MONS_HELL_RAT : MONS_RIVER_RAT;
```
PleasingFungus, line 544: What do you think about something like this? (untested)
```
    set<monster_type> rats = { MONS_RIVER_RAT, MONS_HELL_RAT };
    int seen = 0;
    monster_type mon = MONS_NO_MONSTER;
    for (monster_type rat : rats)
        if (!god_hates_monster(rat) && one_chance_in(++seen))
            mon = rat;
    if (!seen) // e.g. Oka hates all
        return;
```

perryprog: Ah, yeah, I do like that better. One thing I'm unsure of is how Oka should respond to the cloak—I feel like Oka should disapprove of wearing this cloak, regardless of if it functions, as that seems more in line with what he expects. I think this could be either done as it is now, which is indirect in that any rats the cloak spawns while following Oka causes them to be aggro on you (which is hilarious), or it could be done similar to other egos like Running and Cheibriados, and cause a piety loss.

PleasingFungus: You shouldn't be able to end up with allies under Oka regardless (so I don't think piety penalties are relevant).

I think that making the rats hostile is a bit of a rude trap for players, personally. If you die to that, would that be fun? Or would it feel like the game played a 'gotcha' on you?

perryprog: Well, it is marked as a dangerous item in that case, so I feel less bad about it. It's also pretty transparent in that it spawns allies for you. https://github.com/crawl/crawl/blob/df715e3894bdcf4ccc677eb781822a17384bfcd5/crawl-ref/source/item-name.cc#L2716-L2722

perryprog: On second thought, it could also be an opportunity for some Oka-specific flavor. Maybe if a rat would be summoned but you follow Oka, there's a custom line along the lines of "Your ratskin cloak rustles, but does nothing" or "Your ratskin cloak vibrates, but a divine force holds it back". If you want to keep the hostile behavior (what currently happens), maybe a line of "A river rat scurries out from under your cloak, and is turned against you by divine influence".

PleasingFungus: The behavior you mentioned above dates to Ru's hated by all. Oka shares code with it, but is flavored a bit differently. The "divine force" approach you mention SGTM.

PleasingFungus:
Looks good! Thanks for working on this.

PR #2106, opened 2021-09-01
New Spell: Flame Wave (L4 Fire/Conj)

PleasingFungus:
Flame Wave is a channeled spell (like Searing Ray) that blasts your
surroundings with ever-wider waves of flame. It's intended to be a
fun (and spectacular!) capper to the Fire Elementalist start.

Fact sheet:
- Fireball damage.
- Channel costs 2 MP per turn.
- Can be channeled twice normally, for a max radius of 3.
  - Veh range extension allows a third channel at radius 4.
  - Noise starts at 10 (shout) and goes up by 2 per turn.
- Burns trees (very efficiently!)

Open questions:
- How do I get autocombat to work?
- What books should this actually be in? (Need a new one?)
- Should sleep cancel channeling? (For this & for other spells?)
- Are the numbers reasonable? (Damage, noise?)

Points for improvement:
- Would be nice to explain how Veh works in the spell description.
- It would be nice (both for this and Searing Ray) to allow
  cancelling channeling in response to a warning prompt without
  wasting the turn.


RojjaCebolla:
When you first described this on IRC, i thought this was going in an anti-ramparts direction: when you cast it, poof, aoe1 of fire, and then you can move after that, for POOF aoe2 of fire in the new spot! That kinda hot-feet, fly and fleet spell might be fun too.

PleasingFungus:
I get what you mean!

One of the main reasons that fireball wasn't dropped from the FE start when other L5 spells were removed is that it'd leave Sticky Flame as the top spell. That would encourage a very "kiting"-heavy playstyle - light a guy on fire, run away, repeat - that seemed quite tiring to play for large stretches of the game.

I don't think we're looking for other "fly and fleet" spells, but it's fun to think about!

crawl-ref/source/status.cc
```diff
@@ -555,6 +555,14 @@ bool fill_status_info(int status, status_info& inf)
         }
         break;
 
+    case STATUS_FLAME_WAVE:
+        if (you.props.exists(FLAME_WAVE_KEY))
+        {
+            inf.light_colour = WHITE;
+            inf.light_text   = "Wave";
```
ebering, line 562: Would like some cool plusses like searing ray!

PleasingFungus: Done. :)

PR #2112, opened 2021-09-02
Allow / to be used instead of shift in UIDirectionChooserView.

Aliscans:
This change means that /+direction is an alternative for shift+direction, which matches the as it already is when triggering the "long walk" command.

Put ' and ; before * and / in targeting_help_1, as the above change means the latter keys are used as modifiers with use_modifier_prefix_keys set.

In unmangle_direction_keys(), only let * modify characters which CONTROL() maps to \x00-\x1F. In particular, this means that ** is *.


ebering:
Thanks for the patch!

PR #2116, opened 2021-09-04
Prevent stabs when berserk

mainiacjoe:
Berserk prevents most actions, it ought also prevent attacking in "just the right spot."

### DIFF
--- a/crawl-ref/source/attack.cc
+++ b/crawl-ref/source/attack.cc
@@ -1735,7 +1735,7 @@ int attack::player_stab(int damage)
 void attack::player_stab_check()
 {
     // XXX: move into find_stab_type?
-    if (you.duration[DUR_CLUMSY] || you.confused())
+    if (you.duration[DUR_CLUMSY] || you.confused() || you.berserk())
     {
         stab_attempt = false;
         stab_bonus = 0;


mainiacjoe:
This is an edge case, it requires a stabbable status effect to land and then berserk to begin afterwards.  A player that can generate hexes from a wand or spell or greenDr can do this, also hexing allies such as Hepl Hexer, Yred Vampire Knight.  Berserkitis makes the going berserk afterwards after the status effect lands more likely.

hellmonk3:
oppose changing this since berserk does not turn off other special weapon properties like cleave and reach. Admittedly rather hard for players to actually take advantage of this one, unlike the others.

mainiacjoe:
I hadn't considered that, and it makes sense.  Confusion does cancel reach, but not cleave.  So stabbing and reach are something you can do while berserk, but not while confused, whereas cleave you can do whenever.

mainiacjoe:
I realized a difference though between stabs and reach/cleave.  Only axes cleave, and only polearms reach, but any weapon can stab because stabs depend on a status effect in the target.  Short Blades get a bonus to stabs, but stabbing is not intrinsic to the weapon class like cleaving and reach are.

Paperbell:
I would miss using BiA to chunk enemies with lots of fast might-boosted distraction stabs if this was merged.

PleasingFungus:
I think you're arguably right from a flavour point of view. However, being able to stab while berserk isn't so egregiously weird that it'll surprise or upset many players, and it's nice to keep things simple.

PR #2118, opened 2021-09-05
Added CMake/vcpkg MSVC build

rds1983:
Hello,
I've noticed that official way of building at Windows/MSVC is somewhat dated.
Hence I've created new one from scratch that uses CMake and vcpkg.
It replaces all 3rd party libs from "contrib" with ones at vcpkg repository.
The only exception is "lua", since which older required version is unavailable in the vckpg repo.
Unsure if this PR has any need, but I decided to create it regardless. :)


gammafunk:
MSVC compilation is now working via msbuild and is even integrated into CI, so it gets regular testing. Hence this approach isn't need, so I'm closing this PR. Thanks for the submission, and further improvements to MSVC builds are welcome.

PR #2121, opened 2021-09-07
Fix: don't list walljump twice while silenced

### LINKED ISSUE #2120: Reading a scroll of silence causes "Wall jump" button to appear twice
Local tiles, macOS 11.5.2

0.28-a0-419-g3dc288482e

**DESCRIPTION:**
Steps to reproduce:
1. Create a character, enter Wiz mode (for convince—this same issue occurred naturally in a game for me)
2. Give yourself a scroll of silence
3. Join WJC and set your piety to anything above about 50
4. Observe the set of buttons in the "Abilities" section of the sidebar
5. Read the scroll of silence
6. Observe the set of buttons in the "Abilities" section of the sidebar
   - Expected results: the buttons remained unchanged
   - Actual results: a second wall jump button appears, functionally identically to the first. It does not appear in the <kdb>a</kdb> menu, nor anywhere else that I can see.

Notes:
- I tested this with one other god and did not have the same issue

**EXTRA FILES:**
Before silenced:
image

After silenced:
image

Example save with the issue already-present (same turn as the above screenshot):
[PerryWiz.cs.zip](https://github.com/crawl/crawl/files/7122945/PerryWiz.cs.zip)

perryprog:
This specifically affected the "Abilities" section of the sidebar, but I'm not sure where else. It seems likely that this is legacy code from when walljump was more of a pseudo-ability, but since whether something can be used while silenced is determined by the failure basis (e.g., if it's an invocation), there's no need for this check to exist in get_god_abilities anymore.

Closes #2120.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -3943,7 +3943,11 @@ vector<ability_type> get_god_abilities(bool ignore_silence, bool ignore_piety,
     }
 
     if (!ignore_silence && silenced(you.pos()))
+    {
+        if (you_worship(GOD_WU_JIAN) && piety_rank() >= 2)
+            abilities.push_back(ABIL_WU_JIAN_WALLJUMP);
         return abilities;
+    }
 
     // Remaining abilities are unusable if silenced.
     if (you_worship(GOD_NEMELEX_XOBEH))
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -36,6 +36,7 @@
 #include "god-abil.h"
 #include "god-companions.h"
 #include "god-conduct.h"
+#include "god-passive.h"
 #include "hints.h"
 #include "invent.h"
 #include "item-prop.h"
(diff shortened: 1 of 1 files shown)


crawl-ref/source/ability.cc
```diff
@@ -3941,11 +3941,14 @@ vector<ability_type> get_god_abilities(bool ignore_silence, bool ignore_piety,
             abilities.push_back(static_cast<ability_type>(anc_type));
         }
     }
-    if (silenced(you.pos()) && you_worship(GOD_WU_JIAN) && piety_rank() >= 2)
-        abilities.push_back(ABIL_WU_JIAN_WALLJUMP);
 
     if (!ignore_silence && silenced(you.pos()))
+    {
+        if (you_worship(GOD_WU_JIAN) && piety_rank() >= 2)
```
gammafunk, line 3947: You would probably want to do this like
```suggestion
        if (have_passive(passive_t::wu_jian_wall_jump))
```

perryprog: Nice. DIdn't know about that.

PR #2122, opened 2021-09-07
Fixes two unrelated bugs concerning Phantom Mirror and the monster

Nasst:
version of Phantom mirror.

Bug 1 is simple: When a monster casts phantom mirror on another monster,
the casting monster is unentangled/unconstricted. This behaviour is unintended,
as clearly, the monster being copied should be freed instead.

Bug 2 is more subtle: When a monster is copied, its constriction status is
not copied (as intended). However, the enchantment applied by BVC to indicate
direct constriction IS copied. This results (in tiles) in a graphical bug, where
the monster has the BVC icon, but can move around and act as normal.

### DIFF
--- a/crawl-ref/source/mon-cast.cc
+++ b/crawl-ref/source/mon-cast.cc
@@ -3263,10 +3263,10 @@ monster* cast_phantom_mirror(monster* mons, monster* targ, int hp_perc, int summ
         return nullptr;
 
     // Unentangle the real monster.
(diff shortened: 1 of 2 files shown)


Nasst:
Screenshot from 2021-09-28 11-04-29
Screenshot from 2021-09-28 11-04-32

This is the second bug in action. Note how the stone giant was able to move even though it appears to be under BVC.

PR #2124, opened 2021-09-08
Add nine large Temple maps with 18-21 gods

mainiacjoe:
There are two with 18 gods, two with 20 gods, and five with 21 gods.


nicolae-carpathia:
archimedes_tesselation is missing an upstair. It looks like one of the B glyphs is supposed to be turned into a {, but the NSUBST is missing. (The map is listed under 21 gods but there are 22 B glyphs.)

mainiacjoe:
Yes, that was the intention, that one B become the upstairs.  Do I fix this here, or at the branch on my fork that this is pulled from?

crawl-ref/source/dat/des/branches/temple.des
```diff
@@ -2985,6 +2985,85 @@ cc.....cccc+ccccccccccc...B...cc.....cccc+ccc
                                   ccccc              
 ENDMAP
 
+NAME:    mainiacjoe_temple_starburst
+PLACE:   Temple
+ORIENT:  encompass
+WEIGHT:  10
+SUBST:   c : c
```
nicolae-carpathia, line 2992: This SUBST doesn't appear to do anything. I am not opposed to just sitting around doing nothing _in principle_, mind you, but this is the wrong place for a SUBST to do that.

nicolae-carpathia:
> Yes, that was the intention, that one B become the upstairs. Do I fix this here, or at the branch on my fork that this is pulled from?

I think the usual procedure is to make the changes on your local branch, commit them, and then push them to this remote branch.

crawl-ref/source/dat/des/branches/temple.des
```diff
@@ -2794,6 +2794,64 @@ llllllClllllllllllllllllllEllllllllDlllllllllllllll
 llllllClllllllllllllllllllEllllllllDlllllllllllllll
 ENDMAP
 
+NAME:    mainiacjoe_temple_archimedes_tessellation
+PLACE:   Temple
+ORIENT:  encompass
+WEIGHT:  10
+SUBST:   c : cvb
+SHUFFLE: lw
+MAP
```
nicolae-carpathia, line 2803: A question of aesthetics: the big areas of deep water can have shallow water added, so instead of big blocks of red and blue, you get big blocks of red with blocks of mixed blues, which looks off. If this isn't the intent, I would add TAGS: no_pool_fixup to the vault definition.

nicolae-carpathia:
I like the layouts. Big fan of geometrical patterns and tesselations. One small thing: starburst has a lot of empty space, which might lead some wasted time with autoexplore, but I'm not sure how much of a concern that actually is since it's just Temple. I'll tag in gammafunk. @gammafunk 
Other than that, and the issues mentioned elsewhere, I think these are fine to merge.

PleasingFungus:
Merged, thanks! :)

PR #2128, opened 2021-09-11
Prune functions

Aliscans:
Delete some unused functions.
Avoid compiling some functions into a console binary where they are only used with tiles.

I haven't changed what any functions do here. There are functions which don't seem to be doing anything useful when called in console mode, but I have left them as they are.


rawlins:
I'm a bit late on this, but what exactly is the gain in being so specific with the `#ifdef`s? I would generally prefer `#ifdef USE_TILE` to be on code that is actually tiles-specific, not just code that happens to be only used on tiles. (And in general, would rather reduce than increase the amount of conditional defines.)

Aliscans:
> I'm a bit late on this, but what exactly is the gain in being so specific with the `#ifdef`s?

None of the functions I changed were used anywhere, so I was mainly thinking in terms of shorter compile times and smaller binaries. It only saves 12kB on my system, so the effect per user is only small.

PR #2132, opened 2021-09-15
Use the standard names for monsters in "comes into view" messages.

Aliscans:
Change the "comes into view" messages to use the names the game normally uses for each monster by preference, rather than a name based on the monster type.

This is particularly noticeable with groups of dancing weapons or ugly things

Different names continue to be used when a lot of different monsters come into view at once.

This should fix bug #0012649.


ebering:
Thanks for the patch!

Issue #1947, opened 2021-05-11
Issue with creating specific granite statues in vaults


DreamDustDCSS:
Hello! I recently started making vaults, and I ran into a problem with placing _specific_ granite statue tiles in my vaults.

"G" or "dngn_granite_statue" places a random statue from the granite statue pool. If I want to place a particular statue from the pool, I can do so (for about half of them) by using KFEAT and/or TILE.
For example, if I want to make a holy-themed vault, I can place the angel statue with:
TILE: G = dngn_statue_angel
or 
KFEAT: ? = granite_statue
TILE:    ? = dngn_statue_angel

However, this strangely only works for some granite statues. I'm not a programmer or anything, but I think the problem might be with [dc-feat.txt](https://github.com/crawl/crawl/blob/master/crawl-ref/source/rltiles/dc-feat.txt)?
Short excerpt as an example:

> statue_imp
%syn DNGN_STATUE_IMP
statue_orb
statue_cerebov
%syn DNGN_STATUE_CEREBOV
statue_demonic_bust
%syn DNGN_STATUE_DEMONIC_BUST
statue_twins
statue_orb_guardian
%syn DNGN_STATUE_ORB_GUARDIAN

Some of the statues have synonyms(?), and I'm able to place these specific statues by using it. But for the ones that don't (like statue_orb or statue_twins in the excerpt), I'm unable to. Trying to use "statue_twins" or "dngn_statue_twins" doesn't work for some reason I don't understand. Maybe they need a synonym added to this file or they just aren't defined properly somehow?

List of the statues with this problem: 
statue_dwarf, statue_archer, statue_tengu, statue_princess, statue_hydra, statue_orb, statue_twins, statue_sword, statue_sigmund, statue_triangle

rawlins:
Sorry for the slow reply on this. You have identified the solution! Feel free to add a `%syn` line to `dc-feat.txt` as part of a vault patch. For an example of a patch that does this (I somewhat arbitrarily pulled it from git blame): a17c66b2ac69ae2487

(I couldn't swear that there isn't any other secret way to get at these tiles without a `%syn` line, but setting a readable `%syn` is good practice even if there is.)

PR #2135, opened 2021-09-15
Adding 13 new vaults

DreamDustDCSS:
These vaults are based on different granite statues that were previously inaccessible for vault-making. (See: https://github.com/crawl/crawl/issues/1947)
(Although I should note that I still can't get dngn_statue_dwarf to work, sadly...)

Descriptions for each vault are in the commits. Feedback very welcome!


hellmonk3 (commented):
Sorry it's taken so long for anyone to get to this. I like the theme of a lot of these vaults, especially the new entry vaults. I think that most of them are good with some minor tweaks but merry_men might need some more involved change because it tends to produce isolated areas.

PR #2136, opened 2021-09-15
Turn Dissolution into a human when Jiyva dies.

aaronegolden:
It doesn't make sense for Dissolution to remain the "high priest of
Jiyva" if Jiyva died for having lost all its followers.

Fixes https://crawl.develz.org/mantis/view.php?id=11741

### DIFF
--- a/crawl-ref/source/daction-type.h
+++ b/crawl-ref/source/daction-type.h
@@ -52,6 +52,7 @@ enum daction_type
     DACT_ALLY_HEPLIAKLQANA,
     DACT_UPGRADE_ANCESTOR,
     DACT_REMOVE_IGNIS_ALTARS,
+    DACT_JIYVA_WORSHIPERS,
     NUM_DACTIONS,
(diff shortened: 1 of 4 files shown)


semitonal:
I feel like the existing flavour positions Dissolution as pretty devout - lore-wise, perhaps Dissolution is faithful but a single follower just isn't enough? And TRJ is big and full of jellies and so counts as lots of followers, and the player has special protagonist powers so also counts as enough to keep Jiyva alive. I do agree that it'd be cute to do something special to Dissolution if they do end up as the only remaining Jiyva worshipper, whether that's removing their spells or transforming them back but flavouring it as Jiyva's power having left them? Labelling them as a false priest just doesn't quite seem right to me!

ebering:
Implemented Kate's suggestion in the linked commit. If Dissolution winds up alive with Jiyva dead their spells are replaced with a Cantrip, plus custom casting messages suggesting mourning, loss, and/or unanswered prayers.

Issue #2039, opened 2021-07-19
Shop Vault breaks connectivity


ardl:
**PLATFORM and OS or BROWSER:**
webtiles trunk

**DCSS VERSION AND GAME INFORMATION:**
0.27-a0-1591-gc56fcc985a
**DESCRIPTION:**
Shop vault breaks connectivity, all other stairs on the level have been used.

**EXTRA FILES:**
![image](https://user-images.githubusercontent.com/70080435/126229952-aff55499-14ad-4872-ba19-d0f771a73800.png)

*To upload a save file directly to github, compress it to a .zip archive, and
then github will accept the upload. (It will not take a .cs file directly.)*

*If this is an online game, see [these instructions](https://crawl.develz.org/wiki/doku.php?id=mantis:savebackup)
for more information on how to make a save backup; you can provide the
resulting link here.*

rawlins:
Can you post a seed when you have one?

ardl:
Yes, I'll post the seed when the games over. I went back thru with passwall to check if there was a stair in the area, and there was only a down hatch.

![image](https://user-images.githubusercontent.com/70080435/126718828-9cb9b9f4-e7d1-48a8-975a-30948088ff44.png)

ardl:
Seed is 5164812597112276993

PR #2137, opened 2021-09-15
Edit a vault to prevent breaking level connectivity

DreamDustDCSS:
Issue was reported here: https://github.com/crawl/crawl/issues/2039
Based on the seed provided, the vault was nicolae_take_the_stairs
My fix is to just add some extra padding (floor tiles) around the outside.


ebering:
Thanks for the fix!

Issue #2126, opened 2021-09-10
Escape hatch can trap player in grunt_nemelex_the_gamble


catern:
**PLATFORM and OS or BROWSER:**
local tiles, Linux, 0.27.0

**DESCRIPTION:**
Took an escape hatch and ended up in one of the rooms of grunt_nemelex_the_gamble, which have no exit.

**EXTRA FILES:**
Save: [Watoaf.cs.gz](https://github.com/crawl/crawl/files/7140409/Watoaf.cs.gz)
(in wizmode because I used it to escape before realizing I should report this)

PR #2138, opened 2021-09-15
Prevent players getting stuck in a vault's teleport closets

DreamDustDCSS:
Issue was reported here: https://github.com/crawl/crawl/issues/2126
The vault is grunt_nemelex_the_gamble
It's a relatively small vault that heavily relies on triggers and has two closets for players to potentially get stuck in.
My proposed solution is just to add "no_tele_into" as a Tag. This will prevent any problems with unexpected teleports/shafts/hatches.


ebering:
Thanks! fixed this a different way.

PR #2140, opened 2021-09-15
Add a zoo-themed Lair entry vault

mainiacjoe:
Places a building with five cages, each containing one of two tough monsters from each S-branch and from Lair itself.  Has a "zookeeper" named monster.


crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -440,6 +440,63 @@ MAP
     ..@...
 ENDMAP
 
+NAME:    mainiacjoe_lair_zoo
+TAGS:    lair_entry transparent
+TAGS:    no_monster_gen no_item_gen
+ORIENT:  float
+SHUFFLE: NALPS
```
hellmonk3, line 447: The cages need to be masked with no_tele_into, since there's no way out if the player gets teleported in and doesn't have another teleport/dig/blink (admittedly rather unlikely, but possible).

crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -476,7 +476,7 @@ FTILE:   Ll~ = floor_dirt
 SUBST:   T : TG
 KFEAT:   # = iron_grate
 KFEAT:   O = enter_lair
-MONS:    gnoll ; whip
+KMONS:   z = gnoll name:zookeeper n_rpl n_des n_noc ; whip
```
hellmonk3, line 479: The zookeeper can probably be a real monster rather than popcorn, since you aren't making the player fight anything else. Any midlevel humanoid with a whip is probably fine, maybe an orc knight? I'm not sure the rename is really necessary, either.

crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -440,6 +440,63 @@ MAP
...
 
+NAME:    mainiacjoe_lair_zoo
+TAGS:    lair_entry transparent
+TAGS:    no_monster_gen no_item_gen
+ORIENT:  float
+SHUFFLE: NALPS
+NSUBST:  P = 1:P / *:`
+SUBST:   P = Pp
+KMONS:   P = anaconda
+KMONS:   p = shock serpent
```
hellmonk3, line 451: The second subst on these isn't necessary, you can just do KMONS: P = anaconda / shock serpent. Likewise, if you're going to always make three cages use floor_dirt you can just make L,  P, and N all nsubst to ' and define the ftile at the end. No need to do it in three places.

mainiacjoe:
I think it's ready

crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -465,6 +465,7 @@ SUBST:   T : TG
 KFEAT:   # = iron_grate
 KFEAT:   O = enter_lair
 KMONS:   z = orc knight ; whip
+KMAXK:   \WANPLS = no_tele_into
```
nicolae-carpathia, line 468: no_tele_into is a KPROP, not a KMASK. Also, the \ should be a `.

crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -453,12 +453,11 @@ KMONS:   N = emperor scorpion / orb spider
...
-KFEAT:   Aa = shallow_water
+KFEAT:   A = shallow_water
 NSUBST:  S = 1:S / *:W
 KMONS:   S = bunyip / goliath frog
-KFEAT:   Ss = shallow_water
+KFEAT:   S = shallow_water
 NSUBST:  L = 1:L / *:`
-SUBST:   L = Ll
 KMONS:   L = catoblepas / eight-headed hydra
 KFEAT:   ` = floor
```
nicolae-carpathia, line 462: There's a lot of redundancy here. 
One thing to keep in mind is that KFEAT/KMONS/KMASK/KITEM/ITEM/MONS etc. apply only after all the various SUBST and SHUFFLE and NSUBST have happened, even ones that are defined after them in the vault definition. (KPROP and the various tile/colour changing commands are the odd ones out here; they only take into account the substitutions that are defined before them.) You don't need to have three separate lines saying KFEAT: ` = floor. (In fact, because ` doesn't have any meaning on its own, the feature there will be set to floor by default. So you don't even really need those KFEAT lines at all.) Because of this, it's also easier to simply have all your K-statements in one part, then all the substitution statements in another part, rather than mixing them around like you've done here.

You can also combine NSUBST lines into one:
NSUBST: P = 1:P / *:`, N = 1:N / *:`, A = 1:A / *:W, S = 1:S / *:W, L = 1:L / *:` 
which can be shortened even further:
NSUBST: P = P / `, N = N / `, A = A / W, S = S / W, L = L / `
since if no numbers are specified for the replacement glyphs, the last one will be interpreted to have *= in front of it and the rest will be interpreted to have 1=.

Likewise, you can combine the KFEAT lines:
KFEAT: AS = shallow_water

hellmonk3:
Did some syntax cleanup, squashed and merged from branch. Thanks!

PR #2144, opened 2021-09-19
Add tile_number_ratio_pixels option

Antern:
It's hard to read icon numbers for some people on retina displays,
so the option was added to customize font size for them.

tile_number_ratio_pixels = 28, for example:
Screen Shot 2021-09-19 at 07 05 07
.


Antern:
Redundant due to https://github.com/crawl/crawl/pull/2147

PR #2145, opened 2021-09-19
Add a new vault: Nikola's Laboratory

DreamDustDCSS:
Inspired by Nikola the Mad Inventor, Nikola's Laboratory is a Depths-tier vault with a theme of lightning, constructs, and traps.

image
Breakdown:
- At the entrance are two lightning spires (hint for relec) as well as a guardian golem and a spellforged servitor. This is to introduce players to the vault and set the laboratory theme.
- From there, you can take two paths (with two rooms each) to the central chamber. The hallways between these rooms are lined with stone pillars that have a chance of being lightning spires. There's also a pressure plate trap that will randomly spawn in one of the hallways - stepping on it will surround you with guardian golems.
- One room has a strange machine in it.
- One room has guardian golems and spellforged servitors behind a grate (I think of them as robots in storage). The servitors can hit you with airstrikes from behind it. Trying to walk past into the next hallway will trigger the grate to fall - releasing the enemies.
- One room has dispersal traps, a raiju band, and a couple lightning spires in it. Very appropriate for a Nikola vault, IMO. Haha
- One room has a couple crystal guardians and guardian golems in it.
- The central chamber is accessed via transporter, where you'll face Nikola and two guardian golems. (If Nikola already appeared in your game, he'll be replaced by a wizard with a tweaked spellset)
- There are also two lightning golems held here in containment, and once you kill Nikola (or the wizard), they'll be set free by a dead man's switch. Not sure how balanced this is, but it's probably fine? Let me know.
- The loot in the central chamber is a manual, an acq scroll, an air/transloc spellbook, and a lightning rod.

A few more notes:
- I don't know much about programming, so the triggers for this vault are probably sloppily written. I mainly just copy/pasted from other vaults and did some tweaks.... Let me know if there's a better way to implement them. I can at least confirm that the triggers worked properly during playtesting.
- Including a strange machine might be controversial. I don't want to encroach on the revamped Tukima wizlab, but I thought adding _one_ strange machine to this vault should be fine. It's too thematic to pass up!
- Including guardian golems and spellforged servitors will definitely be controversial since they aren't standard enemies. However, I think they work quite well here based on the theme (a mad scientist making robots / magical constructs in his lab) as well as the gameplay. Especially the guardian golems - although they can't attack, their injury bond will make other enemies (like the crystal guardians and Nikola himself) more tanky, and their inner flame explosions are cool & potentially dangerous. I'm biased of course, but I had fun with the constructs during my playtesting.

Any feedback is very welcome!

### DIFF
--- a/crawl-ref/source/dat/des/variable/nikola_laboratory.des
+++ b/crawl-ref/source/dat/des/variable/nikola_laboratory.des
@@ -1,0 +1,163 @@
+##############################################################################
+# Nikola's Laboratory (by DreamDust)
+
+{{
+function lab_drop_grates(data, triggerable, triggerer, marker, ev)
+    if data.trig == true then
+        return
+    else
+        data.trig = true
+    end
+
+    local did_change = false
+    for slave in iter.slave_iterator("wall_phase", 1) do
+        if dgn.feature_name(dgn.grid(slave.x, slave.y)) == "iron_grate" then
+          if you.see_cell(slave.x, slave.y) then
+            did_change = true
+          end
+          dgn.terrain_changed(slave.x, slave.y, "floor", false)
+        end
+    end
+    if did_change then
+      crawl.mpr("There is a click. Suddenly, the grates are dropped!", "warning")
+    else
+      crawl.mpr("There is a click. Nothing appears to happen.", "warning")
+    end
+end
+
+function callback.drop_trap_stepped(data, triggerable, triggerer, marker, ev)
+  if data.triggered == true then
+    return
+  end
+  local x, y = marker:pos()
+  if not you.see_cell(x, y) then
+    return
+  end
+  data.triggered = true
+  dgn.grid(x, y, "floor")
+  local spawns = 0
+  for p in iter.rect_iterator(dgn.point(x-1, y-1), dgn.point(x+1, y+1)) do
+    if (dgn.mons_at(p.x, p.y) == nil
+        and p ~= dgn.point(you.pos)
+        and dgn.is_passable(p.x, p.y)
+        and spawns < data.max) then
+      if (dgn.create_monster(p.x, p.y, data.mons)) then
+        spawns = spawns + 1
+      end
+    end
+  end
+ if (spawns > 0) then
+   crawl.mpr(data.msg)
+ end
+end
+
+function callback.lab_deadmans_switch(data, triggerable, triggerer, marker, ev)
+  for slave in iter.slave_iterator("floor", 1) do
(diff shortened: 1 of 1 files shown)


DreamDustDCSS:
Another note: Nikola's Lab is in a new, separate file (https://github.com/crawl/crawl/pull/2145/files) since it's trigger-heavy.

DreamDustDCSS:
Also not sure why the checks failed... :( 
It works fine for me.

> Running test #32: 'youdie.lua'.
test error: .../work/crawl/crawl/crawl-ref/source/dat/dlua/test.lua:23: No enter_pandemonium created at Depths:3 (map dumped to level-fail.map)
test error: ...er/work/crawl/crawl/crawl-ref/source/test/unique.lua:14: Invalid depth for Depths in spec "Depths:5"
test error: ...work/crawl/crawl/crawl-ref/source/test/uniquegen.lua:28: Invalid depth for Depths in spec "Depths:5"

Very weird.

randomMesh:
Very nice!

> The loot in the central chamber is a manual, an acq scroll, an air/transloc spellbook, and a lightning rod.

 "Laboratory" sounds like there should be some nice potions too for looting.

semitonal:
High-tier Zot-only monsters like electric golems (and orbs of fire, killer klowns etc) are supposed to remain sealed in Zot as a general rule. There are some very rare exceptions (like a single electric golem as a wizlab boss), but for the most part they shouldn't be placed outside Zot, and definitely not in multiples as a trap. In general Lua traps should also be used very sparingly - one as the focus of a vault is maybe reasonable, but they don't seem like they add much here other than making the vault more spoilery. Just having the vault make you fight monsters normally and then use a transporter to fight the boss seems like it might work better - transporting into an enclosed space is already a more straightforward way to increase threat level and doesn't have the same kinds of spoiler issues (as long as you can see the area you're transporting into).

The checks failing look like they're probably caused by the recent changes to Depths and are unrelated to the vault, so those shouldn't be an issue.

DreamDustDCSS:
> Very nice!
> 
> > The loot in the central chamber is a manual, an acq scroll, an air/transloc spellbook, and a lightning rod.
> 
> "Laboratory" sounds like there should be some nice potions too for looting.

Thanks!
The loot is already pretty generous, so I feel like adding potions would put it over the top. Nikola also seems more like an "electricity & robots" kind of scientist rather than a "mysterious bubbling potions" one, at least to me. 
However, your comment has inspired me to make another vault with a mad wizard scientist conducting human experimentation. Can have a bunch of potions as loot! I'll submit it when I have another full batch of vaults ready.

PR #2147, opened 2021-09-20
Add tile_grid_pixels option

Antern:
The option allows to resize inventory and spells icons.
It works in the same way as tile_cell_pixels for the dungeon tiles

Screenshot was made with the next options:
tile_cell_pixels = 40
tile_grid_pixels = 40
Screen Shot 2021-09-20 at 02 58 46
.


ebering:
Thanks for the patch, sorry for the delay in merging!

Antern:
Thank you a lot, no need for me to reapply the patch every new version anymore.

> How would you like to appear in CREDITS.txt?
Antern

PR #2156, opened 2021-09-23
Adjusting the power cost of Fedhas

sdynet:
Fedhas' plants are very powerful and the piety cost of them seems reasonable. However, in actual battles, due to restrictions on plants being unable to move, players had to summon plants again or bring their enemies to the place where they were. Also, Fedhas didn't give the player useful passive even if there were no plants, so the player had to  pay a lot of fee for every battle to get God's help.

So I'm going to lower the cost of the piety a little bit. However, the abuse of power is also not desired, so MP costs are increased in return. Instead of being able to summon plants more often, players will now have difficulty using power several times in a single battle(Lack of MP).

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -507,13 +507,13 @@ static vector<ability_def> &_get_ability_list()
 
         // Fedhas
         { ABIL_FEDHAS_WALL_OF_BRIARS, "Wall of Briars",
-            3, 0, 2, {fail_basis::invo, 30, 6, 20}, abflag::none},
+            4, 0, 1, {fail_basis::invo, 30, 6, 20}, abflag::none},
         { ABIL_FEDHAS_GROW_BALLISTOMYCETE, "Grow Ballistomycete",
-            4, 0, 4, {fail_basis::invo, 60, 4, 25}, abflag::none },
+            6, 0, 3, {fail_basis::invo, 60, 4, 25}, abflag::none },
         { ABIL_FEDHAS_OVERGROW, "Overgrow",
-            8, 0, 12, {fail_basis::invo, 70, 5, 20}, abflag::none},
+            12, 0, 10, {fail_basis::invo, 70, 5, 20}, abflag::none},
         { ABIL_FEDHAS_GROW_OKLOB, "Grow Oklob",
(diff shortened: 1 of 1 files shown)


gammafunk:
I don't agree that piety costs of Fedhas' abilities are too high. It's been my experience that the summoned allies are indeed very strong, and that the current costs work well at preventing their abuse. All Fedhas allies scale their duration with Invocations, so the solution to preventing having to summon allies multiple times in a fight is to train more Invocations. This has the added benefit of increases the allies' strength, so its an especially useful skill to train for Fedhas. But it's not been my experience that Fedhas' allies' duration are too low even with more modest levels of invocations one is likely to have by, say, end of Lair. Ballistomycetes in particular are quite strong when first available, and able to win fights versus powerful single or grouped monsters, especially when more than one is made. 

Uses of Overgrow and Fedhas Oklobs are limited by their piety costs, but these limitations feel reasonable. You'll likely only get to make a 8-12 Oklobs before dropping  from 6* to below 5*, which is a good limit for max size of an oklob farm. And making Overgrow cheaper will let players break into a significantly higher number of vaults, which I don't think is desirable.

It's true that Fedhas doesn't have a passive, but from the standpoint of being strong and useful, it certainly doesn't need one. It provides a strong active all the way down to 2*. The 2* briars active will also be getting a modest buff this version where briars will block incoming projectiles. Fedhas is an "all-invocations" god, which is an acceptable design. It's certainly is indeed a useless god if you burn through all of your piety, but in practice, eventually you'll not dip below 3* with reasonable play. I've found it quite doable to have enough piety to break into the Zot:5 vault and use a sizeable Oklob farm at that entry point. I think Fedhas' current power level is good, and switching to MP cost isn't going to be a very effective counter-balance lowered piety costs.

However I will say that Fedhas' design is a bit lackluster in that it's "only" an ally god, and provides little in the way of strategic depth beyond managing piety to have higher level abilities available. I originally wanted the god to feature some kind of passive, but there were no good ideas. Some other compelling gameplay idea that worked in tandem with the "stationary ally" theme that Fedhas currently has would be welcome, and that could certainly come in the form of passives. However balance-wise, I think Fedhas' is in a good spot and piety cost reductions aren't necessary at the moment.

PR #2157, opened 2021-09-24
Don't say the map boundary can be burnt or dug through.

Aliscans:
Prevent get_feature_desc() from stating that terrain on the map boundary can burn or be dug through, as it can't. This reflects a restriction in destroy_wall().

Fix a typo in a comment.

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -2477,8 +2477,8 @@ void get_feature_desc(const coord_def &pos, describe_info &inf, bool include_ext
     }
 
     // mention that permanent trees are usually flammable
-    // (expect for autumnal trees in Wucad Mu's Monastery)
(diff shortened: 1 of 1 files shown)


Aliscans:
> I sort of wonder if the map border should just be visibly always permarock, instead of hiding this information in the description? 

I don't know. Using a special tile for the level boundary would sometimes make the "which way should I explore" subgame very easy for someone on a level with a portal and no magic mapping.

With this change, the player needs to check to discover that this is part of the boundary. Most won't do this as finding the boundary isn't all that helpful, and the chance of a wall being part of the boundary is pretty low.

For implementation, you'd want new tiles for indestructible trees. The game already has the autumnal ones, but they'd look odd in the Swamp.

PleasingFungus:
> Most won't do this as finding the boundary isn't all that helpful, and the chance of a wall being part of the boundary is pretty low.

That just makes it more tedious, though...

PR #2158, opened 2021-09-25
Bailey revisions

hellmonk3:
This pr cuts bailey spawn depth to D:7-10 and increases chance per floor to compensate; this solves the current problem where many baileys are trivial at the post-lair end of the depth range. Many baileys are quite hard at the early end of the depth range, so some of them may need their difficulty adjusted downward. I've also done a pass over most of the maps to introduce randomization in monster placements and clean up some of the more egregiously bad map design. To this end, I cut bailey_axe_1 and bailey_polearm_4, substantially reworked bailey_axe_3, and added 4 new maps (elven captive, vampire castle, froggy fort, enchanted keep) that place a few more atypical monsters but still fit the map theme.

Final bailey count is 7 axe maps and 8 polearm maps. I think this is fine, but there's probably room for another axe map if someone has ideas.


crawl-ref/source/dat/des/portals/bailey.des
```diff
@@ -1195,3 +1196,72 @@ tt------....--t-
...
+ccccc.;;;;..333.c34c....ct                            
+cdf4c.;;;;;;;...c43c+cc+ctt                           
+cfffc.;;;;;;;;..c33..c....tt                          
+cff4c..;;;;;;;;.c....c.---.t                          
+ccc+cc.;;;;;;;;.cccc+c.---.t                          
+  c43c.;;;;;;;;...333c.---.t                          
+  c33+..;;;;;;;;;...3c---.tt                          
+  cccc...;;;;;;;;...cc---.t                           
+   c........;;;;;...c.--..t                           
+   c.........;;;....c.--.tt
```
ebering: The ';' forest will create teleport closets

crawl-ref/source/dat/des/portals/bailey.des
```diff
@@ -1195,3 +1196,72 @@ tt------....--t-
...
+   c........;;;;;...c.--..t                           
+   c.........;;;....c.--.tt                           
+   c...x..x........cc.--.t                            
+   c333x..xc+ccccc+c.---.tt                           
+   c...x..xc..333.4c.----.tt                          
+   c...x33xc4.333..c.----.ww                          
+ ttc...x..xccccccc+c.---.www                          
+ ttccccx..x.............www                           
+ wwwwwwx..xwwwwwwwwwwwwwww                            
+ wwwwwwx..xwwwwwwwwwwwww
```
ebering: Might look better to have all of the border the than the entrance be `t` ?

crawl-ref/source/dat/des/portals/bailey.des
```diff
@@ -509,32 +508,34 @@ xxxxxxxxxxx..<..xxxxxxxxxxx
...
+xxxc.cccWW.......WWccc.cxxx
+xxxc1.1.WW.......WW.1.1cxxx
+xxxc.cccWW.......WWccc.cxxx
+xxxc1.1.WW.......WW.1.1cxxx
+xxxc.cccWW.......WWccc.cxxx
+xxxc1.1.WW.......WW.1.1cxxx
+xxxc.cccWW.......WWccc.cxxx
+xxxc1.1.WW.......WW.1.1cxxx
+xxxcccccWW.......WWcccccxxx
+xxxxxxxccW.......Wccxxxxxxx
```
ebering, line 460: A bit worried about water cheesing the bridge now that the water is shallow, but I guess there are enough kobolds and the bridge is wide enough and monsters "step aside for ranged friends" some now so it might be ok. 

An alternate suggestion would be to add hallways leaving from the ambush spots to the room south of the bridge, so that once the "back up and re path" part of the AI is triggered the kobolds open the door (releasing the back up) and come down the bridge.

ebering (commented):
Made some commentary here and there, mostly small stuff. The tweaks to existing maps generally look good, though I had some concerns about the kobold deep water one that survives (trying to keep the classic flavor without the badness in a different way).

The new maps look generally good but I haven't given them a play through yet.

crawl-ref/source/dat/des/portals/bailey.des
```diff
@@ -1138,3 +1136,55 @@ tttttwwwwww.........wwwwwwttttt
...
+# 6 gold piles, 4 good potions/scrolls, 4 good armours / superb items
+NAME:   bailey_polearm_elven_captive_hellmonk
+ORIENT: encompass
+TAGS:   bailey_polearm no_item_gen no_monster_gen no_pool_fixup
+: item(dgn.loot_potions .. " / " .. dgn.loot_scrolls)
+ITEM:   any armour good_item no_pickup
+: orc_with_polearm(_G)
+: orc_warrior_with_polearm(_G)
+: orc_knight_with_polearm(_G)
+MONS: deep elf pyromancer / deep elf zephyrmancer
```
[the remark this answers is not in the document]
ebering: Playthrough feedback: With my just-arrived at D:7 DrWn when I loaded this map in wizmode I got 3 orc warriors for the bridge and door guards, which was hard to manage without flight+deep water cheesing. Unsure if 3 or 4 is too many and making the bridge guards plain orcs is better, or if it's "just right". Also flew over the moat to manage the knight, and was able to vuln mind blast the zephyrmancer. So disregard my concerns about the zephyrmancer.

crawl-ref/source/dat/des/portals/bailey.des
```diff
@@ -880,11 +882,11 @@ MONS:   orc warrior ; spear w:5 | trident w:5 | halberd | hand axe | war axe /\
...
 cccccccccccccccccccccccccccccccccccccccccccccccccccccc
 cc.<.......1cccccccc....1ccccccccc....1ccc..........cc
-c..........12cccccc.....13c.....c.....14+....5......5c
-c..........12c.$$.c.....13+.!!!.+.....14+.......???..c
-cA.........12+.$$.+.....13+.!!!.+.....14+.......???.<c
-c..........12c.$$.c.....13+.!!!.+.....14+.......???..c
-c..........12cccccc.....13c.....c.....14+....5......5c
+c......c...12cccccc.....13c.....c.....14+....5......5c
+c.....ccc..12c.$$.c..cc.13+.!!!.+..c..14+.......???..c
+cA.....c...13+.$$.+..c..13+.!!!.+.....14+.......???.<c
```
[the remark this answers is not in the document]
ebering: Played this one, I think the central `c` in both of the first two rooms should be a `G`.

crawl-ref/source/dat/des/portals/bailey.des
```diff
@@ -1195,3 +1196,72 @@ tt------....--t-
...
+            scale mail | chain mail
+KFEAT:  H = granite_statue
+NSUBST: d = 2:d / 2:e / 1:| / * = dde|, f = 1:f / *=f..
+NSUBST: ; = 20=t. / *:., - = 4:H / 12 = 1. / *:.
+NSUBST: 2 = 5:2 / *=2., 3 = 6:3 / *= 3..., 4 = 3:4 / *=4., G = 1:5 / *:G
+NSUBST: ! = 1:6 / 2=6H / *:H
+SHUFFLE: A<
+TILE: t = dngn_tree_dead, H = dngn_gravestone
+: dgn.delayed_decay(_G, 'f', 'orc skeleton')
+: set_feature_name("tree", "dead tree")
```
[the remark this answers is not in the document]
ebering: Also set console `COLOUR`s  for your dead trees

crawl-ref/source/dat/des/portals/bailey.des
ebering: After playing the map I'm still concerned about water cheese but it's just luring the orc warriors into shallow water, the initial attackers close faster than I could get flame wands. The hail of brigand darts makes that pretty unattractive though, so maybe its ok?

crawl-ref/source/dat/des/portals/bailey.des
```diff
@@ -1201,3 +1201,59 @@ ccc+cc.;;;;;;;;.cccc+c.---.t
...
+: item(dgn.loot_potions .. " / " .. dgn.loot_scrolls)
+: easy_axe_fighter(_G)
+MONS:   human ; hand axe ego:vorpal | war axe ego:vorpal . scale mail
+MONS:   vault guard ; broad axe . chain mail | plate armour
+KMONS:  4 = bullfrog
+KMONS:  5 = prince ribbit, blink frog / cane toad w:1
+KFEAT:  4 = shallow_water
+SHUFFLE: AB
+SUBST:  ; = W:3 ., B = <
+NSUBST: ! = 24 = 1... / 24 = 2... / *:.
```
ebering, line 1185: After playing, these good axe humans hit pretty hard for the early end of the depth; maybe only give them an ego axe half the time.

crawl-ref/source/dat/des/portals/bailey.des
```diff
@@ -450,16 +450,17 @@ ENDMAP
 #
 # Apart from weapons and armours left by the monsters, there are about 10
 # potions useful in battle. If those monsters drink them first, of course...
-# Plus, there is always one stack of very good potions.
 NAME:   bailey_axe_2
 ORIENT: encompass
 TAGS:   bailey_axe no_rotate
 : axe_returning(_G)
```
ebering, line 389: the swarm of boomers doesn't really work here, and also baileys give out too many boomers for modern crawl standards.

crawl-ref/source/dat/des/portals/bailey.des
```diff
@@ -450,16 +450,17 @@ ENDMAP
...
-MONS:   gnoll ; mundane halberd | halberd ego:electrocution ident:type
-MONS:   orc warrior / orc knight
-NSUBST: e = 1:e / *:d
+MONS:   gnoll ; halberd ego:draining | halberd ego:electrocution ident:type
+MONS:   orc warrior ; halberd ego:electrocution ident:type . plate armour /\
+        orc knight
```
ebering, line 392: why doesn't mr orc knight get to be in on the pole action?

crawl-ref/source/dat/des/portals/bailey.des
ebering: A further follow up on this. These good axe humans are spookier than the frogs outside the throne room, even with the D:10 version of the same char. The pacing feels a bit weird on the map.

crawl-ref/source/dat/des/portals/bailey.des
```diff
@@ -1257,3 +1257,57 @@ ccWWc+c+c!!!!!!cc
    cc...cc
     ccccc
 ENDMAP
+
+# A short but intense spriggan/ogre bailey
+# Nine items, with a chance for some delicious dragon scales
+NAME:   bailey_polearm_enchanted_keep_hellmonk
```
ebering, line 1227: The first time I played through this I first encountered ogres then the ogre spriggan mix, then the inner baddies. That felt like a good progression, but with the way the map rolls the second play through I rolled spriggans first, which made the following ogres feel kind of silly. Maybe de-randomize the placement a little bit to get better pacing.

ebering:
Hmm, my playtesting commentary got kind of long, maybe should have wrapped it in a single review. Anyway, the general thrust of these looks good. A few more things to look out for that you might want to check before (or right after) merge:

- throwable quantities given to monsters, currently giving out 4 stacks of boomers means a player can leave a bailey with 70 boomers (lol)
- make sure the boss monsters have the appropriate weapon types (I noted one orc knight missing a pole arm but didn't check em all with care) as appropriate for the bailey type

PR #2162, opened 2021-09-28
Remove archer monsters hidden bonus damage

Nasst:
This commit removes the HD * 4/3 bonus damage from monsters with the archer tag. This bonus, besides being hidden, lacks an analogous bonus in melee monsters (those with the fighter tag). The end result of this is rather unintuitive from the player's perspective, since when the player uses ranged weapons, they usually hit for a little bit less than their melee counterparts, and this is dramatically different from monsters (a merfolk javelineer, for example, hits for a little bit more than an impaler with a +9 demon trident).

Monsters more heavily impacted by this were buffed recently with the change to attack speed.


Nasst:
Closed, since the consensus seems to be that making the damage transparent is the way to go, instead of removing it.

PR #2165, opened 2021-09-29
Implements showing % to-hit for breath abilities.

Nasst:
There's a lot of repetition, but I could't find a way to avoid it
cleanly. There's a lot of subtle differences between many breath
abilities.


ebering:
Since this PR @semitonal did some work implementing a general targeter framework for abilities and integrating them into both the modern quiver and the targeter classes. Building on her work I've implemented this feature a different way that has much less duplication. Thanks for the suggestion, the patch also pointed out some helpful starting points.

PR #2166, opened 2021-09-30
Add 7 new vaults

DreamDustDCSS:
7 new vaults - descriptions in the commit notes. Any feedback welcome!

### DIFF
--- a/crawl-ref/source/dat/des/branches/shoals.des
+++ b/crawl-ref/source/dat/des/branches/shoals.des
@@ -2110,6 +2110,53 @@ cww......L
 ccccJJKKLL
 ENDMAP
 
+NAME:   dreamdust_from_the_deep
+TAGS:   no_monster_gen no_item_gen no_tide
+DEPTH:  Shoals, !Shoals:$
+ORIENT: northwest
+WEIGHT: 3
+KFEAT:  X = open_sea
+SUBST:  # = xxx~
+NSUBST: ~ = 8:z / 2:Z / *:w
+KFEAT:  wzZ = deep_water
+NSUBST: - = 6:1 / 2:2 / *:-
+KFEAT:  -12 = shallow_water
+TILE:   -12 = dngn_shallow_water_murky
+: set_feature_name("shallow_water", "some foul and slimy water")
+FTILE:  .3G?x = floor_slime
+TILE:   x = wall_oozing
+TILE:   G = dngn_gravestone
+: set_feature_name("granite_statue", "a stone stele inscribed with bizarre and terrifying symbols")
+KMONS:  z = octopode zombie
+KMONS:  Z = kraken zombie
+KMONS:  1 = octopode ; trident ego:chaos | dagger ego:chaos
+KMONS:  2 = apocalypse crab / great orb of eyes
+KMONS:  3 = tentacled monstrosity
+KITEM:  ? = ring of the octopus king, \
+            randbook title:Mad_Ravings spells:malign_gateway&summon_horrible_things numspells:2
+MAP
+XXXXXXXXXXXXXXXXXXXXX
+X~~~~~~~~~~~~~~~~~
+X~~~~~~~~~~~~~~~~~~
+X~~~~~~~####~~~~~~~~
+X~~~~~~-xxxxx~~~~~~~
+X~~~~~xx----xx~~~~~~
+X~~~~xx------xx--~~~
+X~~~xx--xxxx------~~
+X~~#x--xx..xx--x#~~~
+X~~#x---..3Gx--x#~~~
+X~~#x-----.?x--x#~~~
+X~~#x--xx--xx--x#~~~
+X~~~xx--xxxx------~~
+X~~~~xx------xx--~~~
+X~~~--------xx~-~~~~
+X~~~~--xxxxxx~~~~~~~
+X~~~~~~~####~~~~~~~~
+X~~~~~~~~~~~~~~~~~~
+X~~~~~~~~~~~~~~~~~
+X
+ENDMAP
+
 ###############################################################################
 #
 # <<4>> Shoals ending vaults
--- a/crawl-ref/source/dat/des/variable/float.des
+++ b/crawl-ref/source/dat/des/variable/float.des
@@ -4943,6 +4943,35 @@ MAP
  .........
 ENDMAP
 
+NAME:    dreamdust_cat_pyramid
+TAGS:    no_monster_gen transparent no_rotate
+DEPTH:   Crypt, !Crypt:$, Depths, !Depths:$
+ORIENT:  float
+WEIGHT:  3
+TILE:    c = wall_tomb
+COLOUR:  c = yellow
+FTILE:   -S1+ = floor_sandstone
+KFEAT:   S = granite_statue
+TILE:    S = dngn_sarcophagus_pedestal_left
+TILE:    G = dngn_statue_cat
+KITEM:   $ = gold q:100
+KMONS:   1 = sphinx
+KMONS:   . = nothing w:100 / felid zombie
+MAP
+       ...
+      ..c..
+     ..ccc..
+    ..cc$cc..
+   ..cc$S$cc..
+  ..cc1---1cc..
+ ..cc-------cc..
+..ccccc+++ccccc..
+.ccccccG.Gcccccc.
+.................
+ ......G.G......
+     .......
+ENDMAP
+
 
 ###############################################################################
 #
--- a/crawl-ref/source/dat/des/variable/float.des
+++ b/crawl-ref/source/dat/des/variable/float.des
@@ -10313,3 +10313,61 @@ x..x..x11xx...........xx11x..x..x
 x1.x.1x11*xxx...{...xxx*11x1.x.1x
 xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 ENDMAP
+
+NAME:    dreamdust_experimentation
+TAGS:    no_monster_gen no_item_gen
+ORIENT:  float
+DEPTH:   D:13-, Depths
+WEIGHT:  5
+KPROP:   ? = no_tele_into
+COLOUR:  c = white
+TILE:    c = wall_brick_white
+FTILE:   -+G?d = floor_rough_white
+TILE:    G = dngn_statue_demonic_bust
+KFEAT:   t = demonic_tree
+ITEM:    potion of ambrosia / potion of berserk rage / \
+         potion of degeneration / potion of haste / \
+         potion of invisibility / potion of lignification / \
+         potion of might / potion of mutation w:30
+KMONS:   1 = wizard
+: if you.in_branch("D") then
+NSUBST:  - = 2:1 / *:-
+: else
+NSUBST:  - = 6:1 / *:-
+: end
+SUBST:   ? = 2344
+: if crawl.one_chance_in(2) then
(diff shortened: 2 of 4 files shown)


crawl-ref/source/dat/des/branches/shoals.des
```diff
@@ -2110,6 +2110,53 @@ cww......L
 ccccJJKKLL
 ENDMAP
 
+NAME:   dreamdust_from_the_deep
+TAGS:   no_monster_gen no_item_gen no_tide
+DEPTH:  Shoals, !Shoals:$
+ORIENT: northwest
+WEIGHT: 3
+KFEAT:  X = open_sea
```
gammafunk, line 2118: Avoid using glyphs that already have a default definition as described in the DES syntax doc. You will see older vaults use `X` quite a bit in this way, but this is something we discourage. It works much better for syntax highlighting and vault reading in general if you can see a standard glyph and know that it has its usual definition. I recommend using a simple scheme for custom glyph usage that you try to follow consistently. For example, try using the lower-case glyphs `apqrsyz` for custom solid terrain elementals (skipping the item list glyphs on `defghijk`). You can use `DEFHJKLMNOPQRSZ` either alternately if you prefer, or specifically for things that will be eventually some kind of open terrain. For custom loot items, it's recommended to use `defghijk` even if definiting via `KITEM` and likewise it's recommended to use `1234567` with either `MONS` or `KMONS` whenever possible. If you run out of numeric glyphs, you can fall back to e.g. upper-case glyphs mentioned above. 

Generally avoid using crawl display glyphs like `z` or `Z` or `#` just because those are crawl console glyphs when numeric or other appropriate glyphs are available. The glyphs ``-'"`;!`` are great to use for large areas when placing monsters or to designate inside/outside areas for vault logic.

crawl-ref/source/dat/des/branches/shoals.des
```diff
@@ -2110,6 +2110,53 @@ cww......L
...
+KFEAT:  -12 = shallow_water
+TILE:   -12 = dngn_shallow_water_murky
+: set_feature_name("shallow_water", "some foul and slimy water")
+FTILE:  .3G?x = floor_slime
+TILE:   x = wall_oozing
+TILE:   G = dngn_gravestone
+: set_feature_name("granite_statue", "a stone stele inscribed with bizarre and terrifying symbols")
+KMONS:  z = octopode zombie
+KMONS:  Z = kraken zombie
+KMONS:  1 = octopode ; trident ego:chaos | dagger ego:chaos
```
gammafunk, line 2132: We avoid using player species monsters that are not in normal circulation, mostly because they're defined for non-gameplay cosmetic reasons and aren't designed as interesting monsters in combat. In this case, I don't think there's a good connection to the Octopus King unrands, since this is an abyss-themed vault. I'd recommend simply using a Shoals-appropriate monster like a merfolk impaler and remove the placement of that unrand. The malign gateway spellbook and existing monster set more than establish the theme. 

I don't think the kraken zombie adds much to this vault either, other than being a somewhat weak tentacle connection. But there is no need for an additional undead theme. In general for thematic vaults, to get the point across I recommend choosing one strong thematic idea and developing that one idea well rather than using multiple semi-related themes. If you really like the Kraken, a live one would actually work perfectly fine. A Kraken zombie is also mostly strictly weaker than a normal Kraken. If you really like the "undead" connection, a spectral kraken would be more effective.

crawl-ref/source/dat/des/branches/shoals.des
```diff
@@ -2110,6 +2110,53 @@ cww......L
...
+ORIENT: northwest
+WEIGHT: 3
+KFEAT:  X = open_sea
+SUBST:  # = xxx~
+NSUBST: ~ = 8:z / 2:Z / *:w
+KFEAT:  wzZ = deep_water
+NSUBST: - = 6:1 / 2:2 / *:-
+KFEAT:  -12 = shallow_water
+TILE:   -12 = dngn_shallow_water_murky
+: set_feature_name("shallow_water", "some foul and slimy water")
```
gammafunk, line 2125: I would avoid redefining feature names here. This is something we try to reserve for portal maps where a change in theme is more self-contained and all encompassing (the entire level is different). In general, for these types of vaults that need to establish a theme but also need to integrate into the existing level, go for a "less is more" approach. Carefully chosen custom monsters and loot as well as possibly some use of custom tiles for walls can establish the theme well without using multiple custom descriptions that can confuse the player as to whether they have some in-game effect.

crawl-ref/source/dat/des/variable/float.des
```diff
@@ -4901,6 +4901,77 @@ MAP
...
+WEIGHT: 3
+COLOUR: bm = red
+TILE:   m = dngn_transparent_wall_red
+FTILE:  -mA? = floor_volcanic
+KFEAT:  ! = .
+MARKER: ! = lua:fog_machine { cloud_type='flame', pow_min = 10, pow_max = 10, \
+                              delay = 10, size = 1, excl_rad = 0 }
+: if you.in_branch("D") then
+KMONS:  ? = sun demon
+KMONS:  - = nothing w:50 / demonspawn
```
gammafunk, line 4917: Comments above re: species monsters also apply to unclassed demonspawn, since those are no longer in circulation. I would recommend moving this vault a bit deeper in Dungeon and simply doing some reversal of roles: have a black sun boss with sun demons and possibly some other fire or necromancy themed threats for the later dungeon. Placement would probably start at `D:12` at the earliest, where the black sun would make a great `8` level spawn. If you wish to place the vault earlier at its existing `D:6-10` range, you could consider simply having some lower-level spectral monsters (fast monsters or just those with decent melee damage that spawn in the `D:6-10` range make good spectral candidates). For combining the themes of Fire and Necromancy, there are potentially a lot of viable choices for mid and later Dungeon, but it will need to be something other than unclassed demonspawn.

crawl-ref/source/dat/des/variable/float.des
```diff
@@ -4901,6 +4901,77 @@ MAP
...
+            randbook disc:fire disc2:necromancy spells:scorch \
+            slevels:8 numspells:4 title:The_Black_Sun, \
+            ring of protection from fire / ring of fire / \
+            staff of fire / scroll of immolation q:3
+: else
+KMONS:  ? = orb of fire
+KMONS:  - = nothing w:50 / black sun / \
+            draconian monk w:5 / draconian scorcher
+KITEM:  ? = manual of fire magic w:5 / \
+            randbook disc:fire disc2:necromancy spells:irradiate \
```
gammafunk, line 4928: Irradiate isn't consistent with this theme of "Necromancy and Fire" you've established; it's not even using the two schools of your custom spellbook. The name does sort of suggest Sun, but I'd recommend to sticking with spells that use one of the two established themes. Starburst is the most obvious spell to use here. It's also worth pointing out that you've actually gone against the theme of the game's existing notion of "Black Sun". The Black Sun of the classed demonspawn monster isn't one with a connection to Fire; it's a more figurative "Sun" that just radiates evil. The Black Mark ability that defines this demonspawn class does the following:

> Empowers the caster and any nearby allies to debilitate their foes on a
successful melee attack, allowing them to inflict draining, cause weakness, or
deplete magical power.

Although the more literal Fire theme you've established is consistent and well integrated into the vault, you've not really used draining, weakening, or anti-magic effects in loot or monsters that are central to the special black sun demonspawn outside of the one use of the black sun itself that only places in some locations. It's possible to use something like orange demon to apply weak to the player or have draining or anti-magic weapons at monster equipment or loot, for example.

crawl-ref/source/dat/des/variable/float.des
```diff
@@ -4901,6 +4901,77 @@ MAP
...
+DEPTH:   Crypt, !Crypt:$, Depths, !Depths:$
+ORIENT:  float
+WEIGHT:  3
+TILE:    c = wall_tomb
+COLOUR:  c = yellow
+FTILE:   -S1+ = floor_sandstone
+KFEAT:   S = granite_statue
+TILE:    S = dngn_sarcophagus_pedestal_left
+TILE:    G = dngn_statue_cat
+KITEM:   $ = gold q:100
```
gammafunk, line 4957: It's better to have randomization by using normal gold piles. If you need more gold than three piles would typically general, simply put more gold in additional places. In general gold only loot is less interesting as vault loot for a player this late, so I'd actually recommend introducing some good general or consumable loot in the form of `%` (normal quality random), `*` (higher quality random), or `|` (higher quality mostly equipment). Placing these in front of your cat statue will theme as being "offerings", since that seems to be what you're establishing. Alternately you can use the tables `dgn.loot_scrolls` and `dgn.loot_potions` to make higher quality consumable potion and scroll loot, if you like that more thematically. Gold only is fine if you strongly prefer this, but simply use more `$` instead of defining custom gold counts here, if you go that route.

crawl-ref/source/dat/des/variable/large_themed.des
```diff
@@ -2316,3 +2316,100 @@ xxxx-xx..xx....xx-xxx-x
...
+NSUBST: b = 1:' / *:b
+# Everything else
+KFEAT:  abcx = stone_wall
+KFEAT:  deX = metal_wall
+KFEAT:  f = crystal_wall
+TILE:   zabcdx = wall_lab_stone
+TILE:   deX = wall_lab_metal
+FTILE:  -.`"'T<G = floor_gauntlet
+KFEAT:  < = escape_hatch_up
+KMONS:  < = patrolling minotaur ; any armour . \
```
gammafunk, line 2382: Use of `any armour` and `any weapon` won't work well for monster equipment. For `any armour`, the minotaur may get equipment it can't actually use, and the item will never get created. For `any weapon`, it may get a launcher with no ammunition. You might look into using the (dgn.monster_weapon)[https://github.com/crawl/crawl/blob/master/crawl-ref/source/dat/dlua/dungeon.lua#L915] function with a class like "knight" or "warrior" and then just defining some good potential armours:
```lua
kmons("< = patrolling minotaur ; " .. dgn.monster_weapon("knight", nil, "good_item") .. 
    " . plate armour | chain mail | scale mail")
``` 
You can get fancier and add in a shield (which won't get used if the minotaur gets a 2h weapon), give it some javelins, random wands etc. It might work best if you just pick a few alternate items for each "slot", which can be weighted, and you can have an entry for each slot of "nothing" as well.

DreamDustDCSS: I tried looking into this briefly, but I think I may be too dumb. Haha
I ended up just making it a regular patrolling minotaur after doing other tweaks to the vault (the mino is now always in the center, so having javelins shouldn't matter as much as when it could previously generate on the outer perimeter of the vault). Hope that's ok.

DreamDustDCSS:
Thank you for the very detailed and valuable feedback! I'll definitely be referring back to these comments for my future vault-making.

I've tried to incorporate almost all of your suggestions. My only concern is that the ice island vault has some sloppy-looking monster-placement code... it works, but there's probably a much better way of doing it than using a bunch of if-else statements? Not sure. Also, it seems I failed a check somehow with the black sun cult vault, but I don't understand what I did wrong. :(

Btw, I reworked the labyrinth vault to make it smaller and simpler. I also made an additional gauntlet exit vault based on it.

Thank you for patience! If you have any other improvements to suggest, please let me know. And feel free to reject my vaults if they're bad... the previous vaults I've submitted probably have even more issues than this latest batch, lol.

PR #2169, opened 2021-10-01
Add a usefulness check for Elyvilon's self-healing abilities

nlavsky:
Without such check, it's easy to over-heal and waste piety.


semitonal:
Merged, thanks for the patch!

PR #2173, opened 2021-10-06
Gate of hell

Aliscans:
The game used the same feature for gates leading into the Vestibule of Hell from the outer dungeon, and ones leading there from other Hell branches. They're used like separate features, behave like separate features, and look like separate features if you use tiles. On the other hand, they have the same description, and there's no easy way to make them look different in a console display.

As I use different glyphs locally for "up only" and "down only" stairs & arches, the last of these annoys me.

This change replace DNGN_ENTER_HELL with DNGN_EXIT_COCYTUS, DNGN_EXIT_DIS, DNGN_EXIT_GEHENNA or DNGN_EXIT_TARTARUS for gates leading out of those branches. It replaces the name and description for each, so that the colour and glyph can easily be changed.

The second and third commits remove the references to Hell features in feat_is_branch_entrance() and feat_is_branch_exit(). This should make the functions easier to follow, and make the code a little shorter. I've modified the functions which call them to accommodate this change.

These change "hints" mode so that it treats the gate into Hell as a branch entrance rather than a set of stairs. They also make the dlua is_branch_entrance() function count the gate into Hell, but only because it's easier that way. Everything else should work the same as before.


crawl-ref/source/dat/descript/features.txt
```diff
@@ -426,14 +431,29 @@ A gateway to the Iron City of Dis
...
+%%%%
 A gateway to the freezing wastes of Cocytus
 
 A gateway leading to Cocytus.
 %%%%
+A gateway leading out of Cocytus
+
+A gateway leading back to the Vestibule of Hell, and whatever horrors you left
+behind there.
+%%%%
```
nlavsky, line 456: Instead of adding the same description four times, you can add it once and then use the `<feature>` syntax for reusing descriptions. For example,

```
A gateway leading out of Dis

<A gateway leading out of Gehenna>
```

Aliscans: Thanks; I've changed that now.

PR #2175, opened 2021-10-08
Feat: make Inner Flame smite targeted

perryprog:
After some discussion on #crawl-dev, this seems like it could be a good idea. It's often desirable to want to choose the inner flame target as someone not necessarily on a band's front line, or just over allies (commonly fire fox).

This is my first time doing "real" code-basey stuff, so this should be reviewed with care—it's very likely I didn't get this right on the first go (though in testing it seems fine). I'm also not sure if killing away all the beam code was the right move, but it seems okay... I'm also not sure if that "player_to_hit" value was important that was in the zap-data entry I killed.


perryprog:
Oh—this also needs checking/testing for behavior when inner flaming allies. I did not do either yet, but I have to go eat. (I'm also not sure how that "should" be refactored when going from beam to non beam spell stuff.)

semitonal:
A lot of the enchantment handling was done using beam/zap code originally yeah, so some of that (including the willpower check) would need to be re-implemented for a smite-targeted version. The player_to_hit value in zap-data.h is used as a power modifier on enchantments (see zap_ench_power in beam.cc - in inner flame's case, power is multiplied by 3), so that would also need some handling.

PR #2184, opened 2021-10-12
Add 7 more vaults

DreamDustDCSS:
Details in the commit notes - feedback welcome!
(First 5 vaults are pretty simple, last 2 a little more complicated btw)

### DIFF
--- a/crawl-ref/source/dat/des/branches/zot.des
+++ b/crawl-ref/source/dat/des/branches/zot.des
@@ -2797,9 +2797,9 @@ KFEAT:  _ = altar_zin
 TILE:   G = dngn_statue_angel
 MONS:   angel god:zin dbname:zin_angel, \
         killer klown-shaped pillar of salt, \
-        killer klown, \ 
+        killer klown, \
         butterfly, \
-        chaos spawn, \ 
+        chaos spawn, \
         glowing shapeshifter
 KFEAT:  O = altar_xom
 KFEAT:  ^ = dispersal trap
--- a/crawl-ref/source/dat/des/branches/zot.des
+++ b/crawl-ref/source/dat/des/branches/zot.des
@@ -2786,6 +2786,41 @@ MAP
 ............?............
 ENDMAP
 
+NAME:   dreamdust_zin_vs_xom
+TAGS:   no_monster_gen
+ORIENT: float
+DEPTH:  Zot, !Zot:$
+COLOUR: v = white
+TILE:   v = dngn_metal_wall_white
+FTILE:  _`1 = floor_w_marble
+KFEAT:  _ = altar_zin
+TILE:   G = dngn_statue_angel
+MONS:   angel god:zin dbname:zin_angel, \
+        killer klown-shaped pillar of salt, \
+        killer klown, \ 
+        butterfly, \
+        chaos spawn, \ 
+        glowing shapeshifter
+KFEAT:  O = altar_xom
+KFEAT:  ^ = dispersal trap
+NSUBST: - = 4=2 / 4=2- / -
+NSUBST: ? = O / 2=^ / 2=3 / 2=3? / 2=4 / ?
+SUBST:  ? = ":200 W:100 n:60 x:40 O:40 v b ^ 4 5 6 T U V Y
+COLOUR: " = random
+MAP
+xxxxxxxxx@xxxxxxxxx
+x.........????????x
+x.......---???????x
+xvvvvnG.---???????x
+xv`1`vv.---???????x
+xv_`1`1.---???????x
+xv`1`vv.---???????x
+xvvvvnG.---???????x
+x.......---???????x
+x.........????????x
+xxxxxxxxx@xxxxxxxxx
+ENDMAP
+
 ##############################################################################
 ##############################################################################
 #
--- a/crawl-ref/source/dat/des/variable/float.des
+++ b/crawl-ref/source/dat/des/variable/float.des
@@ -4901,6 +4901,33 @@ MAP
 .cccccccccc
 ENDMAP
 
+NAME:   dreamdust_wasp_nest
+TAGS:   no_monster_gen
+ORIENT: float
+DEPTH:  D:12-, Lair, Spider, Swamp, Depths
+WEIGHT: 5
+NSUBST: . = 6=1 / 6=1. / .
+: if you.in_branch("Depths") then
+SUBST:  x = b
+COLOUR: b = cyan
+MONS:   spark wasp
+: else
+TILE:   x = wall_desolation
+COLOUR: x = lightgray
+MONS:   hornet
+: end
+MAP
+     xxxx
+   xxx..xxx
+ xxx..xx..xxx
+xx..xx..xx..xx
+@.xx..xx..xx.@
+xx..xx..xx..xx
+ xxx..xx..xxx
+   xxx..xxx
+     xxxx
+ENDMAP
+
 
 ###############################################################################
 #
--- a/crawl-ref/source/dat/des/variable/large_themed.des
+++ b/crawl-ref/source/dat/des/variable/large_themed.des
@@ -2316,3 +2316,25 @@ xxxx-xx..xx....xx-xxx-x
 xxxxxxxxx-x.....xxxxxxx
 xxxxxxxxxxx.@@@..xxxxxx
 ENDMAP
+
+NAME:   dreamdust_cave_troll
+TAGS:   no_monster_gen
+ORIENT: northwest
+DEPTH:  D:10-, Lair
+NSUBST: - = 1 / 2=? / .
+KFEAT:  ^ = web
+MONS:   deep troll
+: dgn.delayed_decay(_G, "?", "yak corpse")
+MARKER: ! = lua:fog_machine { cloud_type='flame', pow_min = 10, pow_max = 10, \
+                              delay = 10, size = 1, excl_rad = 0 }
+MAP
+xxxxxxxxxx
+x--xxxx
+x-!-xxxx
+xx-.xxxxx
+xxxx.^xxx
+xxxxx.xx
+xxx^.xx
+xxx@xx
+x
+ENDMAP
--- a/crawl-ref/source/dat/des/variable/large_themed.des
+++ b/crawl-ref/source/dat/des/variable/large_themed.des
@@ -2338,3 +2338,63 @@ xxx^.xx
 xxx@xx
 x
 ENDMAP
+
+NAME:   dreamdust_under_siege
+TAGS:   transparent no_monster_gen no_pool_fixup
+ORIENT: south
+DEPTH:  D:12-, Orc:1, Depths, !Depths:1
+WEIGHT: 5
+TILE:   c = wall_stone_smooth
+FTILE:  +'{([>-O = floor_cobble_blood
+# Monster Placement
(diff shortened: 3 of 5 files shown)


crawl-ref/source/dat/des/branches/zot.des
```diff
@@ -2786,6 +2786,41 @@ MAP
...
+TAGS:   no_monster_gen
+ORIENT: float
+DEPTH:  Zot, !Zot:$
+COLOUR: v = white
+TILE:   v = dngn_metal_wall_white
+FTILE:  _`1 = floor_w_marble
+KFEAT:  _ = altar_zin
+TILE:   G = dngn_statue_angel
+MONS:   angel god:zin dbname:zin_angel, \
+        killer klown-shaped pillar of salt, \
```
gammafunk, line 2799: Recommend defining "terrain" monsters like this with KMONS on an appropriate glyph like `a` or `p` so it doesn't get confused with the list of monsters the player actually fights on `1-7`.

crawl-ref/source/dat/des/branches/zot.des
```diff
@@ -2786,6 +2786,41 @@ MAP
...
+        killer klown, \
+        butterfly, \
+        chaos spawn, \
+        glowing shapeshifter
+KFEAT:  O = altar_xom
+KFEAT:  ^ = dispersal trap
+NSUBST: - = 4=2 / 4=2- / -
+NSUBST: ? = O / 2=^ / 2=3 / 2=3? / 2=4 / ?
+SUBST:  ? = ":200 W:100 n:60 x:40 O:40 v b ^ 4 5 6 T U V Y
```
gammafunk, line 2808: Although the chance for it is pretty low in this particular case, you want to avoid a substitution that can create teleport closets. The best way to do this is to use a checkerboard pattern for your glyph:
```
---"?"?"?"x
---?"?"?"?x
---"?"?"?"x
---?"?"?"?x
---"?"?"?"x
```
You may want to modify you substitution accordingly. Since much of the `?` will still become floor, it will still look randomized, just not quite as much as it would with the uniform block of `?`. You can get a bit fancier and try to add some "seeds" of specific terrain or a second interspersed substitution pattern, but you just have to ensure that your final result can never make teleport closets. Masking the substitution area with e.g. `no_tele_into` and other tags isn't an alternative to this, since we want to avoid hidden disconnected zones. If the disconnected zones are visible (i.e. through glass) and masked appropriately to not allow teleport and not generate items/monsters, then it's fine.

crawl-ref/source/dat/des/variable/float.des
```diff
@@ -4901,6 +4901,33 @@ MAP
...
 
+NAME:   dreamdust_wasp_nest
+TAGS:   no_monster_gen
+ORIENT: float
+DEPTH:  D:12-, Lair, Spider, Swamp, Depths
+WEIGHT: 5
+NSUBST: . = 6=1 / 6=1. / .
+: if you.in_branch("Depths") then
+SUBST:  x = b
+COLOUR: b = cyan
```
gammafunk, line 4912: two colours you want to avoid using for normal rock walls are cyan and lightcyan. In console, the former is used by metal walls and the latter is used by clear (transparent) rock and stone walls. Special wall colors aren't terribly necessary here to communicate the theme given the monster list and internal layout, and to a large extend the branch-default wall tiles will look better. If you're strongly attached to colored walls, you might consider what wizlab_cloud does: use `ETC_SILVER` crystal walls, which are cool since they change color with each UI update:
```
COLOUR:     b = silver
```
Probably as class as you can get to communicating "air-themed wall". This works for pretty much all locations. What I recommend doing is make the outer border be rock of the default coloring/tiling for the level, then have the internal structure be your `b` glyph, implying this is the area "built" by the wasps. Possible to also embed some `b` in the internal walls surrounding the central area, but ideally from the outside the vault looks relatively normal.

crawl-ref/source/dat/des/variable/mini_monsters.des
```diff
@@ -2317,6 +2317,66 @@ cccc....cccc
...
+.pppppp.c...c.
+.pWWWWp.cc+cc.
+.pppppp.......
+..............
+ENDMAP
+
+NAME:   dreamdust_gnoll_shepherd
+TAGS:   no_monster_gen no_item_gen transparent
+DEPTH:  D:7-11, Lair
+KMONS:  1 = gnoll shaman name:gnoll_shepherd \
```
gammafunk, line 2344: Don't think this is notable enough to warrant a custom monster name/entry. It's just gnoll shaman with yaks or dream sheep. Certainly notable to see dream sheep before Lair, but not to the point where we'd define a custom monster name. It's sort of especially awkward because the shaman isn't a band leader, and it's entirely possible for it to get separated from its "flock". 

If you really like the theme of a gnoll shaman that's a shepherd, I recommend you make some kind of more elaborate "hill" vault with grass etc that somehow contains the flock. Ideally some slightly more elaborate structure to go alongside (a hut or cottage?) and some nice thematic loot. Otherwise I'd just place a gnoll shaman with its custom quarterstaff and its flock and that's fine as a fight for the player to encounter.

crawl-ref/source/dat/des/variable/large_themed.des
```diff
@@ -2316,3 +2316,85 @@ xxxx-xx..xx....xx-xxx-x
 xxxxxxxxx-x.....xxxxxxx
 xxxxxxxxxxx.@@@..xxxxxx
 ENDMAP
+
+NAME:   dreamdust_cave_troll
+TAGS:   no_monster_gen
+ORIENT: northwest
+DEPTH:  D:10-, Lair
+NSUBST: - = 1 / 2=? / .
+KFEAT:  ^ = web
```
gammafunk, line 2325: web trap doesn't feel quite right here, since that's something we always use with spider-type monsters. I recommend use of net traps, and try to place the net trap such that the troll might step on it as they chase the player. Note that it's also possible to only have a chance to place the trap on each glyph or on certain glyphs, if you're concerned it's too nasty.

crawl-ref/source/dat/des/variable/large_themed.des
```diff
@@ -2316,3 +2316,85 @@ xxxx-xx..xx....xx-xxx-x
...
+xx-.xxxxx
+xxxx.^xxx
+xxxxx.xx
+xxx^.xx
+xxx@xx
+x
+ENDMAP
+
+NAME:   dreamdust_under_siege
+TAGS:   transparent no_monster_gen no_pool_fixup
```
gammafunk, line 2343: This is probably a vault that wants `patrolling`, since it's trying to establish a theme of defenders under siege. If the "attacking" monsters wander off after level generation, the player will just encounter a fortress, potentially even an empty one if defenders also wander off.

crawl-ref/source/dat/des/variable/large_themed.des
```diff
@@ -2316,3 +2316,85 @@ xxxx-xx..xx....xx-xxx-x
...
+KFEAT:  O = altar_yredelemnul / altar_kikubaaqudgha
+KPROP:  -O = bloody
+MONS:   ancient champion / vampire knight / lich, \
+        angel / nothing, vault guard, daeva / angel
+:    end
+: end
+MAP
+  ...........................
+ .............................
+...............................
```
gammafunk, line 2386: I'd recommend trimming the surrounding floor on `.` to only two squares deep around the entire perimeter of the vault. You only need a maximum of 45 squares in the case when your `NSUBST` for `.` places everything, which obviously wouldn't happen in practice, and and two squares of `.` surrounding the vault gives you significantly more than this. The smaller the vault is, the more likely it is to successfully place.

crawl-ref/source/dat/des/variable/large_themed.des
```diff
@@ -2316,3 +2316,85 @@ xxxx-xx..xx....xx-xxx-x
...
+MONS:   orc / orc warrior, orc zombie / ogre zombie, \
+        orc skeleton / ogre skeleton, necromancer
+: else
+# Depths
+:    if crawl.coinflip() then
+# Holy defenders
+TILE:   G : dngn_statue_angel
+KFEAT:  O = altar_zin / altar_elyvilon / altar_the_shining_one
+FTILE:  -O : floor_limestone / floor_w_marble
+MONS:   angel / vault guard, zombie, simulacrum, lich / ancient lich w:3
```
gammafunk, line 2373: I'd recommend simply using a mix of cherubs, angels, and daevas as defenders instead of having vault guards. It's fine to use any holy monsters you feel are appropriate, including the possibility of a pearl dragon "boss". Those are all going to be something a player in Depths can handle.

For the "attacking" monsters, I recommend mixing in some spectres with the simulacra. It actually works best if you look at the depths monster set and hand pick interesting zombies, simulacra, spectres. For zombies, fast monsters work nicely, or at least monsters with heavy melee attacks. Many random zombies you'll get will be functionally useless, e.g. spriggan air mage zombies. For simulacra and spectres, there's a similar problem of some being useless or nearly so. Fast and heavy or multiple attacks work well. 

If you want a feel that's more a mix of relative popcorn, it's fine to just use a simple `zombie` or `simulacra` and let it be random, but I would recommend at least mixing in some of `spectre`, since those don't lose monster speed as derived undead.

crawl-ref/source/dat/des/variable/mini_monsters.des
```diff
@@ -2317,6 +2317,66 @@ cccc....cccc
...
+KMONS:  2 : yak / dream sheep
+NSUBST: . = 5=2 / .
+MAP
+.....
+..1..
+.....
+ENDMAP
+
+NAME:   dreamdust_looks_like_rain
+TAGS:   transparent no_monster_gen
```
gammafunk, line 2355: This one might want `patrolling` since there's a rare cloud type placement that will seem out of context if the player encounters the first time when the vault monster has wondered off.

crawl-ref/source/dat/des/variable/mini_monsters.des
```diff
@@ -2317,6 +2317,66 @@ cccc....cccc
...
+MONS:   sky beast
+NSUBST: . = 2=1 / 3=1. / .
+KFEAT:  A = stone_arch
+: if you.in_branch("D") then
+KMONS:  A = air elemental / wind drake
+: else
+KMONS:  A = titan / storm dragon
+: end
+: if crawl.coinflip() then
+MARKER: A = lua:fog_machine { cloud_type='rain', pow_min = 10, \
```
gammafunk, line 2366: You can use thunder clouds in Depths, where titans and storm dragons are immune to elec. It might get some collateral kills, but monsters in Depths are stronger, there's only one of this cloud, and we don't care as much if we kill a few stragglers. Seems an appropriate thematic upgrade for Depths, at least.

crawl-ref/source/dat/des/branches/zot.des
```diff
@@ -2786,6 +2786,41 @@ MAP
 ............?............
 ENDMAP
 
+NAME:   dreamdust_zin_vs_xom
+TAGS:   no_monster_gen
+ORIENT: float
+DEPTH:  Zot, !Zot:$
```
gammafunk, line 2792: Not actually necessary to exclude Zot:$, as we allow random vaults on that level unless they somehow couldn't function properly. It even places killer klowns, which a monster the often only first encounters on Zot:5 itself.

PR #2185, opened 2021-10-13
Rework rework Ely

sdynet:
Too much time has passed since Ely was released, and many of the possibilities they was able to go have already been taken by other gods. I've been thinking a lot, but I don't think it's a good choice to use complex methods to make Ely's mechanism interesting. Therefore, I condensed only the core part of god and refined it simply and clearly.

The powers of Ely is now as follows:
*1 - Guardian Spirits
*2 - Heal Other(Strict neutral)
*4 - Divine Protection(Purifying yourself, recovering a small amount of HP/MP as much as part of the damage and invalidating the damage leading to death. Also, it allows the killing of enemies.)

That's all. If you want true healing, take a potion.


crawl-ref/source/actor.cc
```diff
@@ -330,6 +330,7 @@ int actor::spirit_shield(bool calc_unid, bool items) const
 
     if (is_player())
         ss += you.get_mutation_level(MUT_MANA_SHIELD);
+        ss += have_passive(passive_t::ely_guardian_spirit);
```
KABoissonneault, line 333: Seems like misleading indentation.
Did you mean to put this in the `is_player()` branch? If so, add `{ ... }`
If not, it should be indented with the code outside the branch

sdynet: Um, I'm sorry, but can you tell me in more detail? I didn't understand well due to lack of skills.

KABoissonneault: The way you wrote the code, it looks like as if you meant to write
```
    if (is_player())
    {
        ss += you.get_mutation_level(MUT_MANA_SHIELD);
        ss += have_passive(passive_t::ely_guardian_spirit);
    }
```
That is, the `ss` value would only be added if the actor is the player.

But what you actually wrote is equivalent to 
```
if(is_player())
{
  ss += you.get_mutation_level(MUT_MANA_SHIELD);
}
ss += have_passive(passive_t::ely_guardian_spirit);
```
That is, the check is made for all actors. Notice the line with Ely's guardian spirit is not indented.

The fact that it looks like one thing but is actually something else is why I called it "misleading". I didn't call it wrong, because I don't think it would have an effect on any actor other than the player. It's still an issue for future maintainers, who might be confused about the intent.

sdynet: oh, ok. I'm I understood. Thank you. :)

semitonal:
Closing this - Elyvilon's had some related changes in b5831b96, inspired by the idea here of simplifying and reducing the number of abilities.

PR #2190, opened 2021-10-15
feat: allow checking if there is enough space for Malign Gateway

nlavsky:
This commit disables the spell in the quiver and z/Z menu when there
is not enough space to open a gateway.


ebering:
Thanks for the patch!

PR #2195, opened 2021-10-19
Clarify setting up python on msys2 windows

ArmiesAndCastles:
I had problems setting up python, using msys2 on windows. This clarification may help those, unfamiliar with python environment.


ArmiesAndCastles:
> Was this from within the msys2 shell?

Yes.

> we'd rather provide documentation on what leads to this sort of path issue

If the issue is with msys2 and/or windows, then a workaround may be the best solution.
In any case, I don't insist on adding this to the documentation. I'll look at this issue some more, but you can close it.

PR #2198, opened 2021-10-20
Forbid temporary form lava/deep water exploration

### LINKED ISSUE #1975: Dragon form auto-explores over lava and deep water
**DCSS VERSION 
TRUNK, 1202

**DESCRIPTION:**

Dragon form considers itself as flying permanently, which causes auto-explore over deep water and lava. It may lead player into the unexpected consequences, such as draining or maybe instant death(?).

ArmiesAndCastles:
player::permanent_flight should not return true for temporary forms.
All forms appear to be temporary.
Resolves #1975.


ebering:
Thanks for the patch!

PR #2202, opened 2021-10-23
fix: disable miscast-created allies when allies are forbidden

nlavsky:
As a followup to cf4520f7, this commit blocks nameless horrors from
Spellbinder under Okawaru or Sac Love.


semitonal:
Merged, thanks!

PR #2209, opened 2021-10-30
Increase the piety cost of Rising Flame

AdamPG:
Cinder Acolyte has proven to be an incredibly strong start. In
particular, the Rising Flame power changes the game in ways that
make the entire early and mid-game into a foregone conclusion for
any experienced reasonably-high-win-rate player. While strong
starts such as Berserker already exist, these are all still
susceptible to losing if one makes the various mistakes that
players make in an average game of Crawl. Rising Flame, however,
removes all consequences for mistakes.

At starting Ignis Piety, the player gets four uses of Rising
Flame. When they activate it, they are shafted upwards after a
delay of 2+random2(3) turns (an entire turn faster than a
teleport scroll!). This ability is probably the most powerful
escape ability in the entire game and is simply too strong to be
available from the start of the game in such a large quantity. A
"get out of jail free card" for a player's worst mistakes is a
cool idea, but it should only be available one time, similar to
Tso or Kiku's capstone abilities.

As such, this PR increases the piety cost of Rising Flame by four
times.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -647,7 +647,7 @@ static vector<ability_def> &_get_ability_list()
         { ABIL_IGNIS_FOXFIRE, "Foxfire Swarm",
             0, 0, 9, {fail_basis::invo}, abflag::quiet_fail },
         { ABIL_IGNIS_RISING_FLAME, "Rising Flame",
-            0, 0, 96, {fail_basis::invo}, abflag::none },
+            0, 0, 0, {fail_basis::invo}, abflag::none },
 
         { ABIL_STOP_RECALL, "Stop Recall", 0, 0, 0, {fail_basis::invo}, abflag::none },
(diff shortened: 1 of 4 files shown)


PleasingFungus:
Hi, thanks for the PR! I agree with the conclusion here, but I'm concerned about the UI of the implementation (how to communicate this very high cost to players, since it doesn't show up in `a` in an explicit way) and the interaction with other abilities.

Kate suggests that we could instead make Rising Flame an explicitly once-only ability, and adjust costs for the other two active abilities somewhat. What do you think?

AdamPG:
I agree with Kate, that sounds like a better implementation.

AdamPG:
I went ahead and implemented + tested Kate's recommendations. An increase of 1/3 to the abilities more-or-less accounts for the free use of Rising Flame (see the commit message), which seems appropriate.

PleasingFungus:
Thanks! Nice change :)

PR #2211, opened 2021-11-01
fix: correct levdes.vim comment syntax highlightingSeems to have been introduced as a typo in 763bd81.

perryprog:


ebering:
Thanks for the patch!

PR #2213, opened 2021-11-03
text: fix, reword, and extend a few mutation descriptions

nlavsky:
* To make style more consistent, this commit capitalizes stat names,
  adds a missing serial comma, and replaces +Stealth with Stealth+.
* "Stab" has a special meaning for melee attacks, so replace it with
  the spiked tentacle's attack verb, "pierce".
* Clarity blocking Rnd*Rage is a well-hidden mechanic, which is not
  mentioned anywhere in the game, so let's fix this too.


semitonal:
Merged, thanks for the patch!

PleasingFungus:
+1 on thanks. Good fixes!

PR #2214, opened 2021-11-03
Fix: remove sleepwalking, refactor duplicate methods

perryprog:
Being asleep wasn't considered a condition in being unable to move, which led to issues like not having your evasion reduced while asleep.

I investigated this after causative mentioned it in #crawl.

### DIFF
--- a/crawl-ref/source/actor.h
+++ b/crawl-ref/source/actor.h
@@ -336,7 +336,6 @@ public:
     virtual int  dragon_level() const;
 
     virtual bool paralysed() const = 0;
-    virtual bool cannot_move() const = 0;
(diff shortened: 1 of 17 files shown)


semitonal:
I'm pretty sure this is intentional, and cannot_move() and asleep() are assumed to be distinct checks with differing behaviour in a number of different places - player::cannot_act() exists specifically for cases that should check both. I don't have a strong opinion either way in terms of unifying them (either in general or just specifically to give reduced evasion while asleep), but it'd require follow-up changes to the other checks if so.

perryprog:
Force-pushed to correct staging for 6376081.

PR #2215, opened 2021-11-04
Improve the monsters in some overflow vaults

hellmonk3:
Overflow vaults with a uniq_altar tag can place in Depths. Some of these vaults
place very easy monsters, like goblins with spears, which are suitable for the
overflow altar range but not suitable later. This commit either adds an
alternative monster set for Depths placement or limits the range of the vault
to dungeon only, depending on which method seemed best.


PleasingFungus:
Merged with some slight tweaks, thanks!

PR #2217, opened 2021-11-08
Assorted vaults

hellmonk3:
here's another vault pr, thank you for your time


nicolae-carpathia:
> here's another vault pr, thank you for your time

you're welcome, i didn't try them out in practice yet but they look pretty good to me

PR #2219, opened 2021-11-09
Ignis scorched temple vault & two small fixes

pdpol:
This PR comprises 3 commits. The first is a rare temple vault for Ignis in the vein of those for Lucy/Jiyva that there have been impassioned arguments made in discord for and against. The other two are small cleanups to things I was looking at while working on this that could be totally wrong and would be more than happy for someone to let me know I have no idea what I'm doing.
 
The Ignis vault is based predominantly on volcano features. I am hoping to somehow imply the "dying" aspect of Ignis with the size/shape of the lava pool and definitely looking for suggestions on how to make that read better.

As for the two fixes, I think the comment documenting usage of fog machines just has the wrong default value specified for `pow_rolls` based on https://github.com/crawl/crawl/blob/3d4abc6862629c0cc18f666097ec6b1ddd03e039/crawl-ref/source/dat/dlua/lm_fog.lua#L101.

Aside from that, when I was looking through `temple.des` I happened to be referencing `minmay_temple_bridge_d` a bunch since I was cribbing it's style heavily and all, and I became convinced that there either could never be 18 altars in that vault as written or I can't count/read. Either is possible. Thanks for your time!

### DIFF
--- a/crawl-ref/source/dat/des/branches/temple.des
+++ b/crawl-ref/source/dat/des/branches/temple.des
@@ -4131,3 +4131,52 @@ ccQ..gh.hg..Lcc
   ccP.O.N.Mcc
    ccccccccc
 ENDMAP
+
+# 12 gods
+# plus 7 Ignis altars
+NAME:   pdpol_scorched_temple
+TAGS:   no_cloud_gen temple_variable temple_altars_12
+PLACE:  Temple
+ORIENT: encompass
+WEIGHT: 2
+KFEAT:  I = altar_ignis
+MARKER:  D = lua:fog_machine { cloud_type = "grey smoke", \
+             pow_max = 3, delay_min = 10, delay_max = 30, \
+             start_clouds = 0, size = 1, excl_rad = -1 }
+SUBST: A = AAAAG
+KPROP:  D = no_tele_into
+: set_feature_name("stone_arch", "broken pillar")
+SUBST: ^ = ^^.
+KFEAT: ^ = web
+SUBST: D = l
+SUBST:  " = -..
+SUBST:  ' =  LL--.
+SUBST:  L =  ll-
+COLOUR: - = brown
+TILE:   A = dngn_crumbled_column
+RTILE:  x = wall_volcanic
+FTILE:  GBIA^+. = floor_rough_red
+FTILE:  - = floor_rough_brown
+MAP
+xxxxxxxxxxxxxxxxxxxxx
+xxxxxxxxxxxxxxxBxxxxx
+xxxxxxxxxxxxxB...Bxxx
+xxxxxxxxxxxx.......xx
+xxxxx..^xxxG.......Gx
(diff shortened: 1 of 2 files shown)


ebering:
I'm in favor of an Ignis temple, but this doesn't really read dying to me. The lake of fire is round and intact, and Ignis is mixed in with the other gods. Perhaps you meant to randomize it with the `SUBST: ' = l:12 .` line, to create the idea of a drying lava pool?

I think that's an important first step, but since this is a rare temple you can feel more at liberty to custom make the lava-lake into one that's drying. To really sell it I'd try to find a pattern that evokes the cracked icebergs on a mostly frozen lake.

The other thing to think about is how do the Ignis altars relate to the other gods nearby? In your current design Ignis is most present, but doesn't stand out. There are a couple of ways you could approach making Ignis feel abandoned:
- make a central "inner sanctum" with only Ignis altars, and either manually or automatically ruin it (using the level builder, happy to help you figure this out), with an outer, less elaborate but in-tact part housing the other gods
- don't go crazy Ignis: make a more traditional temple layout that's not a volcano and have a small nook (like a hearth or something) with an Ignis altar and some chained clouds of smoke pulsating (not fire anymore)
- have each altar on its own  "platform" over a lava pool, with Ignis's partially filled in
- same as above but with chained cloud generators of fire flanking the altars, Ignis gets only smoke
- clever use of the broken_pillar statue variant, webs, and/or dead trees (could also make a burnt tree variant tile for this) around the Ignis altar(s)

pdpol:
Thank you for the feedback @ebering! I like these suggestions and will try to rework this with some of those ideas over the weekend.

I did try to "fray" the edges a bit with the `SUBST` line, and originally tried to do something where the outer most layer was very likely to be floor, and the next layer in from that slightly less likely, thinking it would give it a more dried up look but really it just looked checkerboardy. I really like the idea of a cracked iceberg approach.

I'll play around with some of those other ideas too. Definitely interested in anything you could point me towards to use the level builder.

pdpol:
I had some time to take another crack at this trying to incorporate a lot of @ebering's feedback.

I initially wanted the broken pillars to be based off of granite statues and not stone arches, similar to a lot of the ecumenical vaults. Because I also want guaranteed granite statues in the two smaller rooms for the other gods I couldn't really figure out how to properly target just the ones in the big Ignis room to run `set_feature_name` on, so ultimately used another feature. Curious if there's a way to do this?

ebering:
> Curious if there's a way to do this?

Sadly not a straight-forward way to do it. We do have fountains and stone arches so I think the broken pillar granite statues could go in Ignis' room.

Thanks for the submission!

I switched the roles of arches and statues and did some tweaks when merging to make the "smaller but still tended" part and the "originally giant but now walled off" part pop and contrast.

I also implemented a version of the "traditional temple but the ignis altar is neglected" in  08698888 if you want to take a look at that take on the concept.

Issue #1935, opened 2021-04-30
Shapeshifters in ooze form maintain engulfment after shifting forms to a non-engulfing monster


splitpunched:
**PLATFORM and OS or BROWSER:**
CKO, webtiles

**DCSS VERSION AND GAME INFORMATION:**
version 0.27-a0-1113-gb2b6fe9c5e - it may have been slightly earlier, but none of the patches between them would have affected this

**DESCRIPTION:**
Character was engulfed by a quicksilver ooze shapeshifter. Once the shapeshifter changed into another form, I was still treated as being engulfed until I got away from the shapeshifter, even though it was no longer in a form capable of engulfing.

Relevant parts of the log:
```
You cannot cast spells while unable to breathe!
Your lungs strain for air!
...
The quicksilver ooze changes into a storm dragon!
...
You slip free of the ooze engulfing you.
* * * LOW HITPOINT WARNING * * *
Your lungs strain for air!
The storm dragon changes into a catoblepas!
You take 6 damage, and have 98/215 hp.
You slip free of the ooze engulfing you.
You gasp with relief as air once again reaches your lungs.
```

**EXTRA FILES:**
Included is the morgue file, [split.txt](https://github.com/crawl/crawl/files/6406874/split.txt)

PR #2225, opened 2021-11-13
fix: really end engulf upon polymorph

Implojin:
At present, the bug described in #1935 still occurs in trunk.
dd09596130 made an attempt to fix this, but clear_far_engulf()
only applies to non-adjacent monsters by default.

This commit forces the engulf removal upon monster polymorph,
hopefully fixing the bug.


semitonal:
Merged, thanks for the fix!

PR #2227, opened 2021-11-14
Buff Beogh's Smite

nlavsky:
... by letting it set ally targets correctly.

Also, tweak Smite's damage message to include attack strength punctuation.


semitonal:
Thanks for the patch!

PR #2228, opened 2021-11-14
Remove a redundant "your" from a message.

Aliscans:
The item_name variable in _equip_weapon_effect() uses DESC_YOUR, so doesn't need "your" to be added later.


semitonal:
Good catch, thanks!

PR #2229, opened 2021-11-15
Add "flight" to the stash search prefix for armour "of flying".

Aliscans:
Searching for "fly" returned any equipment which gives flight, but not potions, and searching for "flight" returned potions and artefacts, but not boots of flying (as "find" searches artefact descriptions, but not ego item ones).

This change means that a search for "flight" finds everything. Searches for "fly" are unchanged.


PleasingFungus:
Looks good, thanks!

PR #2233, opened 2021-11-18
Simplified Divine Vigour

sdynet:
The effectiveness and duration of Divine Vigour become stronger in proportion to the Invocation skill level. However, the usefulness of this power fluctuates excessively depending on the skill level. It's also annoying to guess how much the effect increases depending on the skill level. Therefore, I would like to simplify this power like Heroism.
I plan to fix the effect of this force to 'MaxHP/MaxMP 30% increase' and lower the cost slightly.


semitonal:
I don't think Divine Vigour is overly complicated personally - it scales from a 5% bonus at 0 invocations to a 50% bonus at 27 invocations. Having it work that way is a good way to encourage training invocations heavily for a strong payoff, which I think is a big part of Elyvilon in particular compared to, say, Okawaru, where training invocations increases duration but doesn't change the strength of the effects, so you tend to just get the minimum possible skill to activate the abilities.

ebering:
I agree with Kate that invocations scaling is pretty central to Ely's design. Closing this.

PR #2235, opened 2021-11-20
A couple small barding cleanups

pdpol:
I was reading some Palentonga code to try to be less bad as them and
came across two things that looked like reasonable cleanups.

* Comment in acquire.cc suggests only nagas have increased chance to
  acquire a barding but I think it applies to palentongas too since it's
based off you.wear_barding().
* wears_barding() checks the SPF_SMALL_TORSO flag instead of SPF_BARDING
which works because they both apply to the same species but I assume is
just coincidental.

Tested out trying to wear bardings and acquire a bunch on
nagas/palentongas and humans as a control and I don't think I broke anything.


PleasingFungus:
Nice fixes, thanks!

PR #2244, opened 2021-11-26
FEAT: Mark identify scrolls as useless once everything is identified.

Aliscans:
Create a "have identified all potions and scrolls" variable. Scrolls of identify are treated as useless if it is set. It's updated on game load, and also whenever set_ident_type() changes something.

### DIFF
--- a/crawl-ref/source/item-name.cc
+++ b/crawl-ref/source/item-name.cc
@@ -1954,9 +1954,7 @@ bool item_type_known(const object_class_type base_type, const int sub_type)
 
 // Has every item a scroll can identify (i.e. all scrolls and potions) been
 // identified by "you"?
-static bool _is_everything_identified;
-
-void check_if_everything_is_identified(void)
+void check_if_everything_is_identified()
 {
     const object_class_type types[] = {OBJ_SCROLLS, OBJ_POTIONS};
 
@@ -1969,20 +1967,20 @@ void check_if_everything_is_identified(void)
         {
             if (!item_type_known(t, s) && unidentified++)
             {
-                _is_everything_identified = false;
+                you.props.erase(IDENTIFIED_ALL_KEY);
                 return;
             }
         }
     }
-    _is_everything_identified = true;
+    you.props[IDENTIFIED_ALL_KEY] = true;
 }
 
-bool set_ident_type(item_def &item, bool identify)
+bool set_ident_type(item_def &item, bool identify, bool check_last)
 {
     if (is_artefact(item) || crawl_state.game_is_arena())
         return false;
 
-    if (!set_ident_type(item.base_type, item.sub_type, identify))
+    if (!set_ident_type(item.base_type, item.sub_type, identify, check_last))
         return false;
 
     if (in_inventory(item))
@@ -2008,7 +2006,8 @@ bool set_ident_type(item_def &item, bool identify)
     return true;
 }
 
-bool set_ident_type(object_class_type basetype, int subtype, bool identify)
+bool set_ident_type(object_class_type basetype, int subtype, bool identify,
+                    bool check_last)
 {
     if (!item_type_has_ids(basetype))
         return false;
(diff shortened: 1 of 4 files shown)


crawl-ref/source/item-name.cc
```diff
@@ -1952,6 +1952,31 @@ bool item_type_known(const object_class_type base_type, const int sub_type)
     return you.type_ids[base_type][sub_type];
 }
 
+// Has every item a scroll can identify (i.e. all scrolls and potions) been
+// identified by "you"?
+static bool _is_everything_identified;
```
rawlins, line 1957: Would prefer not to do something like this with a static, and if with a static, with something that ensured a default initialization (e.g. via https://isocpp.org/wiki/faq/ctors#construct-on-first-use-v2). See below for a completely non-static suggestion. I realize you are doing initialization by running the check at startup but it's still better to ensure default initialization absolutely in case the code is run in an unexpected way -- as-is in this patch `_is_everything_identified` has an undefined value until the `check...` function is run. (E.g. I couldn't swear that that startup code is called during tests or scripts, though it might be. Similar with arena.)

Overall, I think it might be simpler, instead of using a static, to set a prop on the player when they have identified everything, and then condition the id uselessness check on that. You wouldn't even need a meaningful value -- just whether it is set. This would let the value get saved, not need a static, and side-step the issues about default initialization (if the check is whether a prop is set, the default is a per-player no, automatically).

crawl-ref/source/item-name.cc
```diff
@@ -1952,6 +1952,31 @@ bool item_type_known(const object_class_type base_type, const int sub_type)
...
+
+void check_if_everything_is_identified(void)
```
rawlins, line 1959: style note, we don't use `void` argument lists, this is more of a C thing I think.

crawl-ref/source/item-name.cc
```diff
@@ -1952,6 +1952,31 @@ bool item_type_known(const object_class_type base_type, const int sub_type)
...
+{
+    const object_class_type types[] = {OBJ_SCROLLS, OBJ_POTIONS};
```
rawlins, line 1961: This function could return immediately if its condition (however implemented) is already marked as satisfied, as long as it has a default initialization.

crawl-ref/source/startup.cc
```diff
@@ -363,6 +363,7 @@ static void _post_init(bool newc)
     update_vision_range();
     init_exclusion_los();
     ash_check_bondage();
+    check_if_everything_is_identified();
```
rawlins, line 366: as noted above, if this were implemented with a player prop, this call would be unnecessary.

rawlins (changes requested):
Good change, but I have some comments on the implementation. My recommendation is either: (i) ensure that the `static` has a real default initialization, or (ii) implement this using a prop on the player rather than a `static`. My preference is for (ii), though (i) is closer to the current patch.

crawl-ref/source/startup.cc
Aliscans: > Good change, but I have some comments on the implementation. My recommendation is either: (i) ensure that the `static` has a real default initialization, or (ii) implement this using a prop on the player rather than a `static`. My preference is for (ii), though (i) is closer to the current patch.

Regarding (i), I suppose [Zero Initialisation](https://en.cppreference.com/w/cpp/language/zero_initialization) is more a C thing, too. In this case, _is_everything_identified is initially false.

I've changed it to use (ii), though.

> This function could return immediately if its condition (however implemented) is already marked as satisfied, as long as it has a default initialization.

I've changed it to skip this check when called by _identify_last_item(). Other than in wizard mode, I don't think any other function can identify a scroll or potion once scrolls of identify are redundant.

> as noted above, if this were implemented with a player prop, this call would be unnecessary.

As a new version could contain different items which need to be identified, I've changed it to only calculates it for a new game or when loading a game with a different version string.

Aliscans:
Thanks for doing those.

PR #2245, opened 2021-11-27
fix: don't force picking up of newly IDd equipment

perryprog:
This just re-uses the logic that was being used for wands, which also avoids the edge-case where if you floor-id something without using autopickup triggering (like when AP is disabled) then it was possible to have the item ID'd, marked as don't-pickup in `\`, and still picked up due to NEEDS_AUTOPICKUP_KEY set in its props.

The old code that led to this logic seems to imply there was an intentional hack for "previously-unknown items", but in testing, I found no unexpected behavior with the new version.

I also chose _not_ to merge the repeated code, as I unsure that it's too significant of an issue, and I have frankly no clue what the difference between the `if (fully_identified(item)) return false` and `get_ident_type(item) → return false` bits is in each of the branches.

### DIFF
--- a/crawl-ref/source/items.cc
+++ b/crawl-ref/source/items.cc
@@ -953,13 +953,11 @@ static bool _id_floor_item(item_def &item)
         if (fully_identified(item))
             return false;
 
-        // autopickup hack for previously-unknown items
-        if (item_needs_autopickup(item))
-            item.props[NEEDS_AUTOPICKUP_KEY] = true;
-        identify_item(item);
-        // but skip ones that we discover to be useless
-        if (item.props.exists(NEEDS_AUTOPICKUP_KEY) && is_useless_item(item))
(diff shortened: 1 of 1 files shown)


ebering:
Thanks for the patch!

ebering:
See the discussion in this commit: https://github.com/crawl/crawl/commit/87f88a2ef8972ccea23c5b95a7b874a2a91189b8 this patch (even when fixed subsequently) caused problems. Simply leaving out the `NEED_AUTOPICKUP_KEY` logic would implement your desired behavior, however there are users who like the current behavior. The description "auto pickup unidentified x" is ambiguous, particularly in light of floor id, so in this case I think the better way to go is with a new game option.

perryprog:
Nice catch—I’ll see what I can do; I should be able to get that written in a few days when I’m less busy.

PR #2246, opened 2021-11-27
Remove "const" from the parameter to seen_item().

Aliscans:
751e42844d described removing the const from a reference within seen_item() as a "major hack". 11 years later, fixing this involves removing the const from this function and one additional method, with no changes needed for stack_iterator.


PleasingFungus:
Seems reasonable. Thanks for the change!

PR #2247, opened 2021-11-27
Lua: Fix missing args for print_hint()

powerbf:
The Lua binding for print_hint() was missing the optional 2nd and 3rd arguments.


ebering:
Thanks for the patch! I'm merging several of the Lua PRs in a batch, so this will get closed in a bit when that's done. How would you like to appear in CREDITS.txt?

PR #2248, opened 2021-11-28
hint mode: prevent temporary portals being misindentified as downstairs

powerbf:
The call to is_feature('>', gc) will return true for a temporary portal because '>' is the command to use it, and it's not an altar. Not sure if that's the expected behaviour of that function or not, so didn't modify it. Instead, moved test for a portal entrance before that call so that portals are correctly identified in hints mode.


PleasingFungus:
Looks reasonable. Thanks for the fix!

PR #2249, opened 2021-11-28
Don't highlight spells Vehumet offers if you've read a book about them.

Aliscans:
This changes the meaning of "blue spell at the top of the list" from "spell Vehumet is offering you" to "spell which is only available because Vehumet is offering it to you". Vehumet can withdraw the offer, but you cannot lose the book.

This change doesn't affect the religion screen.

### DIFF
--- a/crawl-ref/source/monster.cc
+++ b/crawl-ref/source/monster.cc
@@ -1247,7 +1247,6 @@ bool monster::drop_item(mon_inv_type eslot, bool msg)
             mprf("%s drops %s.", name(DESC_THE).c_str(),
                  pitem.name(DESC_A).c_str());
         }
-        pitem.props[DROPPER_MID_KEY].get_int() = mid;
 
         if (!move_item_to_grid(&item_index, pos(), swimming()))
         {


PleasingFungus:
Thanks for submitting this!

The change in this PR is considerably broader than seems to be intended. This breaks messaging for memorizing Veh spells and updating the Veh gift timer.

46f201e74d8ac3219957b3d53fa18bc193b731c6 is closer to right, but I'm still not sure this is a net positive change. With this change, how do players tell which spells Veh is offering them? Do they have to consult their notes..?

What do you think?

Aliscans:
Sorry, of course it was doing too much. I would also change MemoriseDelay::start() though. The game gives a special message when you learn a spell Vehumet is offering, but Vehumet isn't really doing anything for you if you've seen it in a book anyway.

If you wish to see what the last spell Vehumet offered you was, you can check the religion screen. This doesn't work with the last set of spells, but it doesn't matter then. If you've read a book with one of the spells in it, Vehumet can't take the spell away from you, and the gift timer does nothing once you've been offered your last gift.

What do you see as the drawbacks for this change? I know that it makes the "reset the gift timer" thing less obvious once you have read a spellbook about a spell, but I always seem to wait much longer than 200 turns between spells anyway. The big advantage I see is that the spells Vehumet can take away from you are highlighted with this change.

Aliscans:
Sorry, didn't realise the PR would close. I don't know if there's a way to reopen it properly, so I've created a new version under https://github.com/crawl/crawl/pull/2278.

> The change in this PR is considerably broader than seems to be intended. This breaks messaging for memorizing Veh spells and updating the Veh gift timer.

Sorry, I don't know why I hadn't seen that originally. As you say, changing vehumet_accept_gift() could, at least in theory, affect gameplay.

I've kept the changes to _sort_mem_spells(), SpellLibraryMenu::entry_colour() & MemoriseRegion::update(), as they all show Vehumet's offers in a different way to other spells.

I've also kept the change to MemoriseDelay::start(), as the "Vehumet" message suggests that you're relying on Vehumet in some way. If you've read a book containing a spell, you could abandon your faith and still lear the spell just as easily.

I've left learn_spell(), but I wonder if the "don't call a spell offered by Vehumet 'impossible'" thing should simply be deleted. Treating Vehumet differently made sense when you simply couldn't learn "book" spells with a 100% fail rate, but it doesn't tell the player anything now. It's not directly related to this PR, though.

> [46f201e](https://github.com/crawl/crawl/commit/46f201e74d8ac3219957b3d53fa18bc193b731c6) is closer to right, but I'm still not sure this is a net positive change. With this change, how do players tell which spells Veh is offering them? 
The religion screen shows if Vehumet is offering you a spell, and tells you what it is until you're offered the last set of spells. After that, there is no gameplay difference between a spell you've seen in a book, and one you're being offered by Vehumet and have also seen in a book.

The main benefits I see for this change are:

1. The memorise spell list is in order of difficulty except when there's a spell on it which Vehumet is providing. It's easier to look through the list when the spells are in order of difficulty.

2. The game shows the player which spells are only available because Vehumet is offering them. Until the last set of spells, this gives the player a motive to learn (say) LRD now as it may be useful later on. As things stand, someone using the default options has no easy way to check if a "blue" spell is one of these or not.

PR #2250, opened 2021-11-28
State when a terrain blocks movement, but not reaching or magic.

Aliscans:
State in the description when a terrain blocks movement, but doesn't restrict reaching weapons or LOS_NO_TRANS spells.

At present, the former covers orcish idols, granite statues and malign gateways (the last only shows a description before the tentacle appears), and the latter also covers grates.

This knowledge could be useful for (e.g.) cheibrodos_spider_deadly_dance.


ebering:
Thanks for the patch!

PR #2254, opened 2021-11-28
bugfix: lua moninf_get_flags must init flags

mgdelmonte:
As with `is`, the `flags` function needs to init `mi_flags`.  Currently the function will fail until flags are initialized, which you can workaround by calling `is` first with a string, but that's silly.


PleasingFungus:
Heh, nice catch. Thanks for the fix!

Issue #2081, opened 2021-08-20
Travel trail not drawing correctly


Goratrix:
After commit https://github.com/crawl/crawl/commit/14faa605fdf6afe43d831ae06f5ba015b416325d from PR https://github.com/crawl/crawl/pull/2041, travel trail display with autoexplore doesn't work properly:

1. when autotravel is interrupted, the travel trail is almost always missing the last 'step', see below screenshot.

![Screenshot 2021-08-20 200055](https://user-images.githubusercontent.com/54112839/130275722-2305419b-3932-4b1a-8598-ba183b3bba3d.png)

2. There are also situations where the trail is not drawn at all, I'm not sure when exactly that happens, seems to be sporadic. Trying to pin-point it, it seems to happen when autotravel walks to a loot pile dropped by a killed enemy. Example below, I was standing on the red X, killed an orc from range, then autotraveled to where he dropped the pile, there is no travel trail. Same thing happens with the last 'pass' when autoexploring a cleared-out level right before the 'Done exploring' message, there is no travel trail drawn.

![Screenshot 2021-08-20 201146](https://user-images.githubusercontent.com/54112839/130276100-b516066c-dabb-4e64-b7da-aabc4a6a29b9.png)

Edit: hmm, looks like the specific case of travel trail not drawn when autoexploring to dropped loot piles exists even before this commit, so it's not related. Still not sure if intended or bug? Feels like a bug, imho travel trail should draw with every movement. Maybe it was introduced with some of the previous commits that changed autotravel to loot piles, there were a few of those.

PR #2259, opened 2021-11-29
Travel trail fixes (#2081)

nlavsky:
This PR fixes two travel trail-related bugs reported in #2081. These visual bugs happen because some events stop travel and clear delays *a bit* earlier than before.


ebering:
Thanks for the patch!

PR #2260, opened 2021-11-29
Add the most nicolae demonic rune vault possible

nicolae-carpathia:
I have reached my apotheosis: I have put a rune in a shop.

Since the demonic rune repeats if you don't pick it up, I figured it
would be the best option for a rune shop. (The abyssal rune also
reappears if you don't pick it up, but the theme is easier to fit into
Pandemonium.) If you already have the rune, the shop instead places
randarts of various kinds.

Out of 25 generated shops, the stats on the price of a demonic rune:
Minimum: 3804
Maximum: 8647
Average: 6640.76
St Dev:  1412.1

The vault also places fat stacks of cash, three other shops, and
demon-summoning monsters from other branches as visitors. Enjoy!

### DIFF
--- a/crawl-ref/source/dat/des/branches/pan.des
+++ b/crawl-ref/source/dat/des/branches/pan.des
@@ -2521,6 +2521,106 @@ cc..B..B..cc
    cccccc
 ENDMAP
 
+################################################################################
+# A Pandemonium lord with a mercantile bent has set up a bazaar in Pandemonium.
+# There are three mostly-usual stores, but the center store is where the Pan
+# lord keeps their most special merchandise: the demonic rune of Zot. 
+# If you already have the demonic rune, it's got a zigfig for you.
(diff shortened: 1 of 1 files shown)


Gopall:
This certainly is an interesting idea for a shop and a rune vault.

nicolae-carpathia:
> This certainly is an interesting idea for a shop and a rune vault.

thank you! with your support, and the support of other viewers like you, we can finally convince the other devs to let me merge this vault

semitonal:
FWIW, since this has been waiting for a little while and I know nicolae is very keen on it, my main objection was just to the initial implementation (with a lower threat level and the zigfig as a fallback), rather than to the concept itself. If it's been changed to make it more comparable in threat level to the other guaranteed rune vaults, I think Pan is a perfectly fine place for something like this.

PR #2261, opened 2021-11-29
Updated documentation of use_animation

burnish3d:
Changed '-=' to '+=' to correctly describe default behavior.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -1553,7 +1553,6 @@ default_show_all_skills = false
         even untrained and untrainable ones. As usual, '*' toggles back to
         trainable mode. This option does not affect hints and tutorial games.
 
-
 monster_list_colour = friendly:green, neutral:brown, good_neutral:brown, strict_neutral:brown
 monster_list_colour += trivial:darkgrey, easy:lightgrey, tough:yellow, nasty:lightred
         (List option)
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Hm, sort of unclear to me what the style here is. It might be trying to tell the user "here's how to set this", rather than defining "this is how this is set?" If so, telling the user to += something that's already included wouldn't be very useful.

What do you think?

semitonal:
I'm not sure how closely it's kept to overall but the rule for options_guide is supposed to be that it lists the default setting, so I think technically the most consistent thing here would be to list it in the same way as `explore_stop` does (so `=` for initally setting the option, and then `+=` when it splits onto a second line).

PleasingFungus:
Oh, https://www.reddit.com/r/dcss/comments/r52d2w/any_way_to_disable_boomerang_animation/hmn6upy/?utm_source=reddit&utm_medium=web2x&context=3 had more discussion of this (coming to the same conclusion as kate, I think?)

burnish3d:
So would that be
use_animations = beam, range, hp, monster_in_sight, pickup, monster, player
use_animations += branch_entry
?

burnish3d:
OK, so I took a look at the other three examples, fire_order_ability, use_animations, and monster_list_colour.  fire_order_ability and use_animations are both correct and were pointed out due to a misunderstanding, but I have updated monster_list_colour to be consistent (without touching the prose).

burnish3d:
@semitonal Hey just wanted to ping you on this to get your attention again :)

ebering:
Thanks for this! How would you like to appear in CREDITS.txt?

burnish3d:
Put me down as Zachary Chandler. Neat!

PR #2263, opened 2021-12-01
fix: typo: Correct a typo in the description of the Book of Pain

Hurricos:
As the title states.


PleasingFungus:
Nice catch, thanks!

PR #2267, opened 2021-12-02
Rekindle the Black Torch

ebering:
This is an implementation along the lines of [the design sketch for a new Yred included in the 0.28 plan](https://github.com/crawl/crawl/wiki/Rekindle-the-Black-Torch-%28a-new-Yredelemnul%29).

The executive summary:

- At 0* Yred grants passive reaping of living and demonic creatures, as well as recall. Things with a corpse come up as zombies, things without as spectrals (including those poor souls whose corpse was exploded). These reaped creatures stick around until the player leaves the level.
- Yred Piety is proportional to the number of Yred servants in the player's los.
- At 2* Yred grants the "Dark Bargain" active, which trades some zombies/specters for a long-duration Yred servant summon. Summon quality scales with piety, and the summon list now covers all (non-harmful) n, z, V, and W monsters (except Vampire magi who are kiku nerds).
- At 4* Yred grants "Drain Life". Same as old drain life, but now costs zombies/specters as well as mp.
- At 5* Yred grants "Enslave Soul". Same as old enslaved soul. This is the only interlevel-recallable ally in new Yred.

The design goals of this redesign are:
- Preserve the identity of Yred (zombie momentum plays, reap and harvest the living, bone dragon pals).
- Minimize reliance on interlevel recall.
- Reduce the total number of permanent allies drastically.
- Experiment with alternative piety systems.

Still todo:
- [x] Dark Bargain ability targeter
- [ ] Dark Bargain ability icon
- [x] Figure out the "piety scale"
- [x] Figure out the reaping rate
- [x] Figure out the ability costs
- [x] Implement: monk and ecu gift of zombies (the piety kind)
- [x] Implement: faith increases the reaping rate, removing it removes some Zs
- [x] Fix penance for doing holy
- [x] Fix abandonment messaging / behavior
- [x] Save compat


semitonal:
Extremely minor but maybe the darkvision passive should also be moved to just be available at any piety level, since the purpose of it is to stop you from getting an accuracy penalty around allied profane servitors, but if you spend most/all of your piety on getting one then you'd lose the passive.

ebering:
After some initial testing I moved dark bargain down to 2*.

semitonal:
Just so it doesn't get lost in the scrollback, ShaperMC noted in Discord that Dark Bargain appears to not cost any zombies as a Djinn, possibly because the zombie cost is based on MP cost?

ebering:
Good catch.

ebering:
Tweaks and changes based on the initial round of feedback: Dark Bargain's invo power scaling is improved, Enslave Soul now works on all intelligences and the status lasts longer.

ebering:
With the power of magic this doesn't require any save compat code. Old Yred enslaved souls will just become spectres on load. Old Yred gifts will still follow the player on transfer, count for piety when in LOS and can be paid for New Yred abilities, which gives such a player something "nice" but is probably the gentlest way to do the transition.

PR #2268, opened 2021-12-03
Qaz's power cost adjustment

sdynet:
In early games, the risk is greater than the benefits of worshiping Qaz. His signature characteristic, the unstoppable noise, is a fatal disqualification, but the powers given are too expensive to overcome the ordeal and require long training to be strong. So, I have somewhat eased the cost of power. I hope the players will be able to overcome the hostile environment more actively.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -595,11 +595,11 @@ static vector<ability_def> &_get_ability_list()
 
         // Qazlal
         { ABIL_QAZLAL_UPHEAVAL, "Upheaval",
-            4, 0, 3, {fail_basis::invo, 40, 5, 20}, abflag::none },
+            2, 0, 2, {fail_basis::invo, 40, 5, 20}, abflag::none },
         { ABIL_QAZLAL_ELEMENTAL_FORCE, "Elemental Force",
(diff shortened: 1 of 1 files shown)


semitonal:
I'd tend more towards saying that Qazlal's abilities are appropriately costed, and many other god abilities are much too cheap. I could see reducing the cost of Upheaval slightly, since that is supposed to be a little more spammable, but I wouldn't be keen to reduce the costs of the other two abilities personally.

sdynet:
> I'd tend more towards saying that Qazlal's abilities are appropriately costed, and many other god abilities are much too cheap. I could see reducing the cost of Upheaval slightly, since that is supposed to be a little more spammable, but I wouldn't be keen to reduce the costs of the other two abilities personally.

Well, I understand. But the key is not the cost. Qaz's early games were too harsh, and I intended to give a small advantage accordingly. If you don't want to adjust the cost, what do you think is a good solution?

semitonal:
I did go ahead and slightly reduce the MP and piety cost of Upheaval, and the MP cost of Elemental Force, alongside some other god ability changes to tone down some strong abilities on other gods: 46f1710d2ca8

PR #2271, opened 2021-12-05
Externalise hints text (part 1).

powerbf:
Moved a lot of text from hints.cc to hints.txt.
As part of this, I also moved the Lua function hints_type from dlua into clua, so that it can be used in hints.txt.


ebering:
Thanks for this patch! This sort of db work is nice to have, it was unfortunate for the timing of this that a separate hints edit landed at the same time and created some merge conflicts. I've cleaned it all up in the merge.

PR #2277, opened 2021-12-07
consider Summonings useless after sacrificing love

jktomer:
With Abjuration gone, every spell in the Summonings school is inaccessible to
characters that have sacrificed Love to Ru. The skill is therefore useless to
them; we needn't bother allowing them to train it, and certainly shouldn't
offer manuals of Summonings from acquirement.


ebering:
Thanks for the patch! How would you like to appear in the CREDITS?

jktomer:
`Jonathan Klabunde Tomer` would be great :) Thanks!

PR #2283, opened 2021-12-10
Main menu button for the action panel

nlavsky:
This PR adds a main menu button to the action panel and refactors code for drawing actions and item glyphs.

### DIFF
--- a/crawl-ref/source/initfile.cc
+++ b/crawl-ref/source/initfile.cc
@@ -4982,6 +4982,8 @@ void game_options::write_webtiles_options(const string& name)
 
     tiles.json_write_bool("show_game_time", Options.show_game_time);
 
+    tiles.json_write_bool("action_panel_disabled",
+            Options.action_panel.empty());
     tiles.json_write_bool("action_panel_show",
             Options.action_panel_show);
     tiles.json_write_int("action_panel_scale",
--- a/crawl-ref/source/initfile.cc
+++ b/crawl-ref/source/initfile.cc
@@ -4982,6 +4982,9 @@ void game_options::write_webtiles_options(const string& name)
 
     tiles.json_write_bool("show_game_time", Options.show_game_time);
 
+    // TODO: convert action_panel_show into a yes/no/never option. It would be
+    // better to have a more straightforward way of disabling the panel
+    // completely
     tiles.json_write_bool("action_panel_disabled",
             Options.action_panel.empty());
     tiles.json_write_bool("action_panel_show",
--- a/crawl-ref/source/tileweb.cc
+++ b/crawl-ref/source/tileweb.cc
@@ -527,6 +527,8 @@ wint_t TilesFramework::_handle_control_message(sockaddr_un addr, string data)
     else if (msgtype == "main_menu_action"
              && mouse_control::current_mode() == MOUSE_MODE_COMMAND)
     {
+        // TODO: add a control message type that can send arbitrary commands
+        // (possibly just as a string, like the lua API for this)
         process_command(CMD_GAME_MENU);
     }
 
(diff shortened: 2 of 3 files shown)


crawl-ref/source/initfile.cc
```diff
@@ -4982,6 +4982,8 @@ void game_options::write_webtiles_options(const string& name)
 
     tiles.json_write_bool("show_game_time", Options.show_game_time);
 
+    tiles.json_write_bool("action_panel_disabled",
+            Options.action_panel.empty());
```
rawlins, line 4986: This is fine for now but in the long run it may be worth adding a real option that positively disables the action panel. I think I'd rather do it by converting `action_panel_show` into a yes/no/never thing, but I think that will need some further work on the options API.

crawl-ref/source/webserver/game_data/static/action_panel.js
```diff
@@ -305,51 +301,121 @@ function ($, comm, client, cr, enums, options, player, icons, gui, util,
...
-                                      {slot: filtered_inv[selected - 1].slot});
+                                      {slot: filtered_inv[selected - NUM_RESERVED_BUTTONS].slot});
                 }
             }
         }
     }
 
+    function draw_action(texture, tiles, offset, scale, needs_cursor, text)
+    {
+        tiles = Array.isArray(tiles) ? tiles : [tiles];
```
rawlins, line 343: maybe refactor to call the glyph drawing code from here instead of the main loop? Looks like the quantity/highlight code is duplicated right now unless I'm missing something

crawl-ref/source/webserver/game_data/static/action_panel.js
```diff
@@ -386,13 +452,10 @@ function ($, comm, client, cr, enums, options, player, icons, gui, util,
...
-        // XX This should definitely be a different/custom icon
-        renderer.draw_gui(gui.PROMPT_NO, 0, 0, adjusted_scale / 100);
-        if (selected == 0)
-        {
-            renderer.draw_icon(icons.CURSOR3, 0, 0, undefined, undefined,
-                adjusted_scale / 100);
-        }
+        // XX The "X" should definitely be a different/custom icon
+        draw_action(gui, gui.PROMPT_NO, 0, adjusted_scale, selected == 0);
+        draw_action(gui, gui.STARTUP_STONESOUP, cell_length, adjusted_scale,
```
rawlins, line 457: it's worth adding a TODO note here, something like:
```
// TODO: select tile via something like c++ `tileidx_command`
```

(Currently no way to do this in js that I know of, probably not worth doing in this PR but worth flagging.)

crawl-ref/source/tileweb.cc
```diff
@@ -524,6 +524,11 @@ wint_t TilesFramework::_handle_control_message(sockaddr_un addr, string data)
         // XX only set this flag if a relevant option has actually changed
         Options.prefs_dirty = true;
     }
+    else if (msgtype == "main_menu_action"
```
rawlins, line 527: eventually it would be better to have a control message type that can send arbitrary commands (possibly just as a string, like the lua API for this). This doesn't need to happen in this PR, but it'd be worth adding a TODO note to this effect near this message type

rawlins (changes requested):
Looks good from reading the code, one refactoring suggestion and a bunch of places where I would leave TODO comments in the code as pointers to future changes.

nlavsky:
I added a regression in "feat: don't hide the action panel unless it is explicitly disabled", so I had to redo it. Some of the show/hide panel code was dependent on `filtered_inv.length`, so I added a check to ensure we draw the panel only after receiving inventory data.

crawl-ref/source/tileweb.cc
nlavsky: Done. If there is anything else I should change, please let me know.

PR #2286, opened 2021-12-10
Add `you.known_items` Lua function

mgdelmonte:
This exposes the list of known, identifiable subtypes to Lua.  It's the same list as given by the `\` "display known objects" command, minus the unknowns.  Without this, it's impossible to know (within Lua), that one has already seen a particular kind of potion or scroll.

### DIFF
--- a/crawl-ref/source/l-you.cc
+++ b/crawl-ref/source/l-you.cc
@@ -807,6 +807,36 @@ static int l_you_abil_table(lua_State *ls)
 }
 
 
+/*** Known items.
+ * @treturn table The list of identifiable item subtypes you know.
+ * @function known_items
+ */
+static int you_known_items(lua_State *ls)
+{
+    lua_newtable(ls);
+    int index = 0;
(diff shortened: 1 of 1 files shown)


ebering:
Thanks for this latest batch of lua api changes!

Out of curiosity, what sort of things do you do with the clua api? Do you have a bot? Helpful interface extensions? Things I can only dream of? If you have some neat demos I'd love to take a look, and it might also be interesting to have a write-up as a guest post on our dev blog.

mgdelmonte:
@ebering I've built a pretty complete bot.  It started as a small script to do better targeting for Fireball, but it grew and grew and grew.

ebering:
Cool! Does the fireball targeting script still exist too? Is the bot source publicly viewable? How well does your bot play, we could have a bot showdown between qw and mw! (By a tradition that predates me many crawlbots have the naming convention -w for some letter, which may or may not be an author initial, so I've called yours mw as a placeholder).

If you have thoughts about the bot development process, extending and using crawl's lua api, and think you could write up something neat I'd be happy to put it up on the dev blog as a guest post.

PR #2287, opened 2021-12-10
Remove piety decay

CrawlOdds:
Proposal: remove piety decay from all gods, and reduce piety gain to rebalance this.

This is currently unimplemented; this PR is to explore whether such a change might be accepted.

### Reasoning

- Piety decay mostly acts as a small, soft penalty on time.
  - How strong the penalty is is very variable between characters, but any god with piety decay gives some incentive to minimize time when not at max piety.
- This soft penalty on time incentivizes some play that the player may find tedious; especially manual exploration (even if autoexplore is safe), and conserving MP to minimize rest time.
- Piety decay does not create many interesting decisions. It is generally not enough of an incentive to make the player take any real risk; only to select the quicker of safe options.
- There are not many degenerate behaviors that a soft penalty on time fixes. For the degenerate behaviors that do exist (like shouting for thousands of turns), the Zot clock is a partial solution here.
- Piety decay is unequal between characters; slow characters' piety decays faster since it is measured in time not turns.
- One positive effect of piety decay is to make piety harder to track. I don't know how important this effect is - I think that piety is randomly gained and therefore hard to track anyway, but I could be wrong.

### Suggested implementation

- Remove piety decay on every god except Uskayaw.
- Adjust piety gain rates so that the net piety gain of the average character doesn't change.
  - TODO: I don't know how best to determine the size of these adjustments.

### Alternatives

One alternative is to tone down piety decay for gods for which it matters most (at least, Sif). As a rule of thumb, a player not using abilities and playing in a normal way should gain piety on average.


randomMesh:
I like this.
Piety decay always feels as if you did something wrong. Not fun.

apranvoku:
Agreed, it doesn't add much to the game and discourages slow, thought-out play.

Gopall:
I actually thought that this was already a planned change.

burnish3d:
Is there a way to measure how much piety is gained and how much is lost due to decay over the course of a game? I believe lost due to decay could be estimated using turn count and the average rate of decay, but I don't know how one would get at how much piety was gained. Anyway, if you could do that you could take a sampling of morgues to try and estimate how much the rate would need to change to preserve the old behavior.

Or, could just remove it and tune down in an ad-hoc fashion. I feel like most players spend most of their time piety capped, or dieing before they lose a significant chunk of piety due to decay anyway.

PR #2265, opened 2021-12-01
Storm Form should display weapon name as "Storm paws", "Storm tentacles" or "Storm fists" depending on species

hedk3:
Currently, this form always displays "Storm fists" no matter the active species, despite tiles showing that the player character's anatomy is maintained in this transformation.

To be consistent with Statue Form and Blade Hands, this minor change uses the same logic used in Statue Form to fix this.

PR #2290, opened 2021-12-12
fix: acknowledge Storm Form in player-act.cc

hedk3:
A follow up to #2265. player-act.cc did not take into account the new 0.27 spell Storm Form in _hand_name_singular. This corrects this, and adds a correction to an edge case with felids that was encountered while testing.


ebering:
As written this didn't compile, though it did point out the correct lines. Thanks!

PR #2293, opened 2021-12-12
Ban Fedhas/Jiyva from Thunderdome (arenasprint), replace Teleport Self with Blink Away

Oneirical:
Following the removal of Teleport Self, the dungeon sprint map Thunderdome became broken. Occasionally, a creature with Teleport Self would attempt to spawn, flooding the message box with error messages, freezing the map and making it impossible to complete.

This fix replaces all instances of Teleport Self with Blink Away. It also bans Jiyva (as an exploration god, it is completely useless) and Fedhas (oklobs can be spammed with extreme piety gain, and placed on every arena entrance for easy spawnkills, trivializing the entire challenge).

My previous GitHub account (hedk3) got suspended for using a temporary email address, so this isn't actually my first commit.

### DIFF
--- a/crawl-ref/source/dat/des/sprint/arena_sprint.des
+++ b/crawl-ref/source/dat/des/sprint/arena_sprint.des
@@ -696,7 +696,7 @@ function arena_sprint_init_boss_table()
                    "spells:mephitic_cloud.15.wizard;" ..
                           "summon_hydra.15.wizard;" ..
                           "invisibility.15.wizard;" ..
-                          "teleport_self.15.wizard.emergency",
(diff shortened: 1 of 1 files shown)


semitonal:
Thanks for the patch! I've merged this but have left Fedhas as enabled for now - for sprints, it's not as much of a problem for some gods or combos to be extremely powerful, and gods are typically only disabled if they're mostly/completely non-functional. For example, Djinn Warper has been shown off as a very silly broken strategy in meatsprint, but it's probably fine for that to just function as a different way to balance the challenge of the map and to let some tactics be much more overpowered than we'd allow in the main game.

PR #2294, opened 2021-12-13
Re-Add potions of cure mutation

apranvoku:
There are some edge cases in the game where a player will be heavily mutated or just needs to remove a certain very bad mutation, and can quaff multiple potions of mutations and roll unluckily, still keeping the one mutation they wanted to remove and even adding further undesirable mutations. Re-adding potion of cure mutation as a low weight potion would fix this, and also make extended more desirable without fear of losing from a plethora of debilitating mutations.

### DIFF
--- a/crawl-ref/source/dat/des/variable/float.des
+++ b/crawl-ref/source/dat/des/variable/float.des
@@ -4689,8 +4689,8 @@ TAGS:   no_monster_gen no_trap_gen no_tele_into
 DEPTH:  D:4-, Lair
 ORIENT: float
 KMASK:  -'DE12 = no_item_gen
-# Cancellation placed only for mutagenic clouds.
-ITEM:   potion of resistance ident:type, potion of cancellation ident:type q:2
+# Cure Mutation placed only for mutagenic clouds.
+ITEM:   potion of resistance ident:type, potion of cure mutation ident:type q:2
 # Replaces a gold pile in late Dungeon
 : item(dgn.good_aux_armour)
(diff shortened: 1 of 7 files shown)


ebering:

When the devteam removes something it is with a reason we find rather compelling, and we are not likely to re-add what was removed in an identical form. (Though a reworked version that fixes the problems that led to the original's removal is always possible.)

In this case their removal was part of a larger mutation overhaul in commit b9e5f3fc1c77 . Your proposal needs to engage with this reasoning (you can find further elaboration in the development chat logs from that time). In particular you should explain what interesting gameplay comes from "hoarding !cmut until receiving a malmutation you don't want to play with". 

The situation you sketch in your opening, where a player is stuck with a malmutation and must play around it (consider doing some zig floors, adjust how their character is played, etc.); equally nice are when a powerful positive mutation a malmutation coexist, is it worth the gamble re-rolling? This kind of dilemma is exactly what the mutation system aims to provide. Impactful malmutations that cause a character to stop short of all 15 runes are a good thing, since there are ways to avoid the worst mutation situations with tactics. More recently the probabilities of !mut were adjusted 564068e7f76a25 with the aim of further moving in the direction of requiring a malmutated player to deal with it.

I'm going to close this, since your PR is a basic re-add of the old !cmut potion and not a rework that aims to engage with the reasons that led to its removal. If you have a re-work idea you want to chat about before putting in the implementation effort feel free to drop by the development chat in #crawl-dev on libera.chat!

PR #2298, opened 2021-12-16
Ammo's mulch rate and generation probability adjustment

sdynet:
It's not important, but the mulch rate of ammo is hidden information. Since I don't think there's any real benefit to adjust the figures individually, I'd like to divide the rate ratio into three types.
General(Except for the exception, all)
Susceptible to destruction(special dart)
Very susceptible to destruction(net)

I think the supply and demand of bows and crossbows need to be smoother because they are main weapon. I think it is ideal that the amount of ammo is somewhat insufficient when the player does not have the help of god. In line with the developers' intention to turn throwing weapons into strategic resources, I'm will not touch javelin. However, I think the quantity of darts and boomerangs is somewhat insufficient to achieve the strategic goal. Of course, excessive use of special darts is not allowed, so I will apply two times the probability of destruction to them like curare.

### DIFF
--- a/crawl-ref/source/item-prop.cc
+++ b/crawl-ref/source/item-prop.cc
@@ -662,15 +662,15 @@ struct missile_def
 static int Missile_index[NUM_MISSILES];
 static const missile_def Missile_prop[] =
 {
-    { MI_DART,          "dart",          0, 12, 2,  true  },
+    { MI_DART,          "dart",          0, 20, 2,  true  },
 #if TAG_MAJOR_VERSION == 34
     { MI_NEEDLE,        "needle",        0, 12, 2,  false },
 #endif
-    { MI_STONE,         "stone",         2, 8,  1,  true  },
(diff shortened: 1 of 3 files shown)


crawl-ref/source/item-prop.cc
```diff
@@ -662,15 +662,15 @@ struct missile_def
 static int Missile_index[NUM_MISSILES];
 static const missile_def Missile_prop[] =
 {
-    { MI_DART,          "dart",          0, 12, 2,  true  },
+    { MI_DART,          "dart",          0, 20, 2,  true  },
```
ebering, line 665: This significantly deccreases the mulch rate of curare.

ebering (commented):
Thanks for the suggestion, I think at this time we don't want to take this shift. If we're going to make ranged an "all the time main weapon" thing the better way to go is just to remove the items and give unlimited ammo. Doing this outright has balance implications so it needs to be a part of a broader ranged redesign.

PR #2301, opened 2021-12-19
Update mutation-data.h

apranvoku:
Add weight list at top of documents for generic muts non-specific to god or race. Rebalance weights, keeping percentages but capping weights at 100. This will make future re-weights more intuitive and easier.

### DIFF
--- a/crawl-ref/source/mutation-data.h
+++ b/crawl-ref/source/mutation-data.h
@@ -2,61 +2,7 @@
 
 #include "tag-version.h"
 
-/*Mutation weights used when acquiring mutations. (Not exhaustive)
-Mutations are classified as good or bad. Good/bad mutations should
-total 100 weight. Zero weighted muts can go at the bottom.*/
 
-//Good Mutations.
-short MUT_IRIDESCENT_SCALES_WEIGHT = 6;
-short MUT_STRONG_WEIGHT = 5;
-short MUT_CLEVER_WEIGHT = 5;
-short MUT_AGILE_WEIGHT = 5;
-short MUT_MP_WANDS_WEIGHT = 5;
-short MUT_CLARITY_WEIGHT = 5;
-short MUT_WILD_MAGIC_WEIGHT = 5;
-short MUT_HORNS_WEIGHT = 4;
-short MUT_TALONS_WEIGHT = 4;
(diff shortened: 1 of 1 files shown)


apranvoku:
Just want to stress, the actual percentages of each mutation was kept the same, essentially rounding it to the nearest percent, except for horns which was moved from weight 7 to 4 (5.47% to 4%) to achieve a round total weight of 100. 

Total weights and new weights can be seen here: https://docs.google.com/spreadsheets/d/1_e2a1rrm6cYwzzMjQpFlLhha5MGGP0SeATRht29zYjM/edit?usp=sharing

crawl-ref/source/mutation-data.h
```diff
@@ -2,6 +2,62 @@
...
+short MUT_WEAK_WEIGHT = 8;
+short MUT_DOPEY_WEIGHT = 8;
+short MUT_CLUMSY_WEIGHT = 8;
+short MUT_BERSERK_WEIGHT = 7;
+short MUT_SCREAM_WEIGHT = 6;
+short MUT_SUBDUED_MAGIC_WEIGHT = 6;
+short MUT_DEVOLUTION_WEIGHT = 4;
+short MUT_INHIBITED_REGENERATION_WEIGHT = 3;
+short MUT_TELEPORT_WEIGHT = 3;
+short MUT_NO_POTION_HEAL_WEIGHT = 3;
```
ebering, line 59: Thank you for doing the work of re-calculating and calibrating the weights!

It's somewhat unfortunate that mut_data is an array, so that mutations can't be grouped in a way that makes it easy to see group weights. However, since these weight numbers are not declared except in the mutation data structure, crawl style is not to create separate constants for each one. If you want to leave an easy-to-use summary of the good and bad total weights, a comment containing the table at the start of the array is the way to go.

If you want to be able to move the mutation_defs to group them, you'll need to refactor mut_data from an array to a map. Probably not worth the effort.

ebering (changes requested):
The numbers themselves look good, but see my comment on style and implementation!

ebering:
Per the discussion in #crawl-dev, it doesn't really make sense to try to standardize these weights: there are individual species restrictions that shift the distribution for some species, and comments drift out of date. Sorry I sent you on a goose chase!

PR #2306, opened 2021-12-25
Attempting to update Zot Defence.

Gopall:
Hello. 

Zot defence was last in the game during 0.15. I've heard it was removed because it was broken or out of date. But I'm not certain about that. I wanted to play zot-def, so I updated it to the latest version of crawl (0.28) for fun.  
Given the reasons why it may have originally been removed, I can't say if it should or shouldn't be added back to the game. Anyways, it's here in case the main branch or any of the forks want to use it.

A brief overview of what zot defence was:
> Zot Defence was an inverted version of crawl. It was a bit like a tower defence game mode. It took place on a single level and the player had to stop monsters from getting to the orb of zot. In order to win you had to get 15 runes to pick up the orb and exit the level. When you killed monsters you got zotpoints. You were able to spend zotpoints to use certain special abilities that you got when you leveled up.

 I've only done a little balance testing in this game mode, but so far it seems like this is much easier than the normal mode of crawl. If you do play this, feel free to tell me about any bugs or issues you encounter.


Gopall:
Here are some of my comments and notes about the game mode and it's potential issues (in it's current state):

The monster pathfinding sometimes bugs out. If you are outside the orb chamber, they will sometimes walk up to a wall infront of the orb chamber and stay there. But if you go back to the orb chamber they will follow you there.

The teleportation ability you get later on is rather powerful and I think it's the cheapest ability you have at 2 zotpoints. This may be a balance issue. Some abilities may be unbalanced in terms of their zotpoint cost.

Curse skulls used to be stationary when ZotDef was last in the game. I haven't made a modified version of the curse skull (yet), so they will wander around if you make them.

Monsters only spawn directly on the staircase. Judging from the old code, this is similar to how they were supposed to spawn in the gamemode.  Although it may have been that they were simply supposed to spawn in close proximity to rather than directly on top of the stairs. 
In any case, this means you can prevent them from spawning by blocking the stair cases with stationary plants you can create. But that also prevents you from winning the game since killing the enemy monsters is required for runes to spawn. I had to force them to spawn in a specific spot, because they were previously spawning in blocked off regions of the map that were used as treasure chambers (for when the player unlocked teleportation).

Some of the things that were previously used to balance certain abilities are now gone. Creating a portal to a bazaar used to permanently drain HP.  Using the zot abilities to acquire an item used to drain levels and creating gold also drained levels.  So I added zot point costs to those abilities. One ability, remove curse, has been removed.

Runes seem to spawn at random locations on the level floor.

ebering:
Hi, this seems like you started quite a bit of work but have trailed off. ZotDef was originally removed because it fell into disrepair, there just wasn't enough consistent interest in maintaining the game mode with balance and polish (no "mysterious pathfinding errors" or other exploitable bugs). We want to distribute something that does more than just run.

As you've learned by this exercise, ZotDef used a bunch of mechanics that have since been radically changed, and there's a lot more to do building on what you've done to get it to a really good place. Unfortunately, there isn't really interest among the dev team in doing that work, and there's little enthusiasm for taking on the maintenance burden again.

This is a great trip down crawl memory lane to read. I do hope you (or someone else) with the desire and energy maintains and releases a ZotDef fork which gives the game mode its polished due. If it stays not too far from current trunk it will continue to work with web tiles, and some of the server admins are amenable to hosting forks.

PR #2307, opened 2021-12-25
Display potion/scroll rarity on the item knowledge screen

nlavsky:
This PR adds information about relative generation weights to the `\` screen. It'll look like this:

unided_consumables

Also, it replaces strings like `It is found rarely compared to other types of scroll.`
with a more succinct `It is a rare scroll.` in item descriptions.


ebering:
Nice! Thanks for the patch!

PR #2309, opened 2021-12-26
Small tweaks to LCS tiles

Goratrix:
There is an LCS rendering bug reported a few times by dilly in discord, examples:

unknown

unknown

While I was not able to reproduce it locally, the affected tile (crystal_spear1.png) has issues which may be causing this. It has size 27x27 instead of 32x32 and there is a stray pixel in the exact corner where the redering artifact appears. This commits fixes the tile and also tweaks the other tiles to be symmetrical (currently some of the directions are smaller than others and some are not properly centered).


ebering:
Thanks for the fix!

PR #2310, opened 2021-12-26
feat: add option autopickup_decide_on_sight

perryprog:
This is a second take at #2245, which was reverted in 87f88a2 as it turns out some people [rely](https://xkcd.com/1172/) on the behavior I thought I was fixing (understandably so, funny comics aside).

I would appreciate any feedback on the option naming and it's description in options_guide.txt since I'm not confident I'm explaining it well—although to be fair, I think this is an unintuitive (and possibly niche?) option to begin with.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -25,8 +25,8 @@ The contents of this text are:
 3-  Interface.
 3-a     Dropping and Picking up.
                 autopickup, autopickup_exceptions, default_autopickup,
-                pickup_thrown, assign_item_slot, pickup_menu_limit,
-                drop_filter
+                autopickup_decide_on_sight, pickup_thrown, assign_item_slot,
+                pickup_menu_limit, drop_filter
 3-b     Passive Sightings (detected and remembered entities).
                 detected_monster_colour, detected_item_colour,
                 remembered_monster_colour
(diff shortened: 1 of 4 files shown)


perryprog:
After some discussion with Yermak on IRC, here's an alternate version of this that feels a bit more intuitive and more in-line with what it seems that people who rely on the current behavior would use it for:

* Rename the option to just be "pickup_one_each", where if true, autopickup will pick up the first type of equipment that you identify, regardless of what autopickup_exceptions is set to.
* This is basically (or completely, but I'm not positive if that's true) equivalent to the option implemented in the PR

gammafunk:
After looking over the subtle issues going on here, I've decided to close this PR for now. As perryprog points out from his discussion with Yermak, there's apparently a desire for autopickup behaviour that uses logic where, during floor identification the first unidentified item of an equipment type is always marked for autopickup and without applying exceptions. The current logic of autopick decisions happening before performing floor identification will generally result in the first unidentified item of a type being autopicked even if there's an exception for its type. As far as I can tell, players are happy with this, and there may not be many users of a new option to make autopickup decisions only after floor identification. Perhaps @ebering can weigh in on whether we still need the option implemented in this PR, or if we need Yermak's suggested approach. For now, I'm closing this PR.

PR #2312, opened 2021-12-26
fix: remove extra spaces in irradiate's messages

perryprog:
This also simplifies the verb selection by using `actor::name` and `actor::conj_verb` wholly instead of with a ternary expression.


perryprog:
This PR is now (more or less) redundant with the changes made in 6b9ff3a, so I'm closing this.

PR #2314, opened 2021-12-27
fix: display_bar should clamp negative values

perryprog:
If a bar was displayed of something with a negative value (like with EV) then text such as "EV: none (normally )" could be displayed. This is a bit odd, so let's make cases like that display "EV: none" instead.

I'm unsure if this breaks anything, but a bit of testing didn't show anything unexpected. The function isn't used for much besides showing player accuracy in `_player_spell_stats` (which seems to only be called when accuracy isn't -1) as well as monster EV, will, and AC.

Before:
image

After:
image

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -4501,22 +4501,26 @@ static void _describe_mons_to_hit(const monster_info& mi, ostringstream &result)
 }
 
 /**
- * Print a bar of +s and .s representing a given stat to a provided stream.
+ * Print a bar of +s representing a given stat to a provided stream.
  *
- * @param value[in]         The current value represented by the bar.
- * @param scale[in]         The value that each + and . represents.
+ * @param value[in]         The current value represented by the bar. Clamped
+ *                          to zero.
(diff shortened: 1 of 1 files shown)


crawl-ref/source/describe.cc
```diff
@@ -4501,22 +4501,26 @@ static void _describe_mons_to_hit(const monster_info& mi, ostringstream &result)
...
+ *                          to zero.
+ * @param scale[in]         The value that each + represents.
  * @param name              The name of the bar.
  * @param result[in,out]    The stringstream to append to.
- * @param base_value[in]    The 'base' value represented by the bar. If
- *                          INT_MAX, is ignored.
+ * @param base_value[in]    The 'base' value represented by the bar, clamped to
+ *                          zero. If INT_MAX, is ignored.
  */
 static void _print_bar(int value, int scale, string name,
```
randomMesh, line 4514: While you are at it, you could make the `string name` parameter a const reference.

perryprog: Good call, thanks!

ebering:
Thanks for the patch!

PR #2319, opened 2021-12-29
Implement unrandart Club of the Night

Monkooky:
Implements a +10 unrandart club with the following properties:
Stabs like a short blade
Gives stealth++
On kill, gives receding Darkness effect.

Each such kill gives a ~25% reduction in line of sight, up to 75%. This reduction decays by ~3% per turn.

I think it's cool and fun, but you should merge it for the pun.


semitonal:
We already have quite a few sources of LOS radius adjustment, and it works reasonably on a permanent (Kobold/Barachi) or semi-permanent (equipped scarf of shadows/robe of Night) basis, but having heavily variable, duration-based LOS adjustments doesn't sound like a great way to use the effect.

PR #2322, opened 2021-12-29
make throatcutter a urand

Monkooky:
Throatcutter's image name was inconsistent with
standard artifact naming conventions.
This corrects that.


Goratrix:
Looks like Zealot Sword also has this prefix, if you want to add that to the PR... unrand_zealot_sword.png

ebering:
We typically don't worry about conventions too hard here, since moves of binary files contribute a lot to repo size. Since throat cutter just got a new tile I took the opportunity to move it. Thanks!

PR #2328, opened 2022-01-02
Spelling Standardization

klorpa:


ebering:
Hi, you seem to be working from a very out of date copy of the crawl repository, was this some old commits that we'd merged previously?

In general we're not too worried about spelling cannonization in the comments (it can mess up `git blame`), so I'll merge the Fannar change separately.

ebering:
Ooop looking at my style handbook themself is accepted for the reflexive form of singular they and Fannar uses gender neutral they/them pronouns.

So I'm closing this.

PR #2329, opened 2022-01-02
Square up scroll rarities

RojjaCebolla:
They'll look prettier when they are all the same kind of number.
(3, 4, 6, 10, and 14 to the second power)


ebering:
Per recent #crawl-dev discussion this kind of internal numerical aesthetic is unimportant, the gameplay impact of the weights is what matters most. Closing.

PR #2332, opened 2022-01-02
Better support for mobile browsers

### LINKED ISSUE #2221: Feature request: webtiles: bring up the on-screen keyboard on iPadOS, iOS
- `webtiles`, on https://underhound.eu:8080
- Firefox for iPad 38.1
- Safari for iPad 15.0 (605.1.15)
- Firefox for iOS 39.0
- Safari for iOS 15.0 (604.1)

- 0.27

**DESCRIPTION:**

It could be fun to be able to play DCSS on the iPad (and possibly iPhone).

Modern browsers seem to do a decent job at rendering the webtile game, and you can even poke at tiles to move to locations, which is nice! However, there is no input field, visible or hidden/virtual, that would bring up the on-screen keyboard on these devices. Additionally, there is no operating-system-provided way for the user to force bringing up an on-screen keyboard on these devices. This means that any commands apart from "move to this tile" cannot be made in the browser on iPad or iPhone, without connecting an external physical keyboard. Of course it is possible to e.g. click the chat box at the bottom right, and you will get the on-screen keyboard for typing comments, but it will disappear immediately when you click outside the chat box.

It would be nice if the webtiles web UI had some mechanism of requesting the on-screen keyboard to be present. 

Thanks!

robertxgray:
This is a minimum set of features and fixes to make the WebTiles interface playable on mobile browsers.

- Added an optional auxiliary input field that can be focused to enable  virtual keybards.
- Added a new option tile_web_mobile_input_helper to enable or disable  the input field.
- Fix touch zoom and scroll disabled when a popup is shown.
- Allow popups to be closed with touch events.
- Fix some minor display issues on small screens.

Tested on Android: Chrome and Firefox. Can someone test it on iOS please?

Closes #2221


robertxgray:
https://user-images.githubusercontent.com/9676118/147891000-051b7e26-a4b6-4cfd-83ed-01944ffbb322.mp4

crawl-ref/docs/options_guide.txt
```diff
@@ -2353,6 +2353,13 @@ tile_web_mouse_control = true
         WebTiles. Regardless of the value of the setting, the minimap will
         respond to mouse control.
 
+tile_web_mobile_input_helper = auto
```
rawlins, line 2356: Could you clarify why there needs to be an option? I.e. what's the use case for setting this to either true or false?

crawl-ref/source/webserver/game_data/static/focus-trap.js
```diff
@@ -162,10 +162,11 @@ function focusTrap(element, userOptions) {
       capture: true,
       passive: false
     });
-    doc.addEventListener('touchstart', checkPointerDown, {
+    // Capturing touch events disables zomming and scrolling on mobile browsers
```
rawlins, line 165: this looks to me like it would probably break focus-trap on mobile though?

Also if we were to keep this we'd want to remove the corresponding `removeEventListener` call

crawl-ref/source/webserver/static/scripts/client.js
```diff
@@ -1709,5 +1713,29 @@ function (exports, $, key_conversion, chat, comm) {
         return x > 0 ? x : -x;
     }
 
+    function is_mobile()
```
rawlins, line 1716: Anything in `static` is not versioned, and so updating both at the same time usually leads to problems (stale caches, servers needing to restart). I'm not immediately sure what a clean solution is for this code (i.e. something that doesn't touch `client.js`), which does seem to need access minimally to `handle_keypress`. For making changes that do touch both versioned and non-versions js code it is usually best to write the versioned code so that it doesn't crash if the `client.js` hook is missing (and this can be removed later once servers catch up). The change here would probably be to make the `game.js` event handlers only installed if `is_mobile` is present in `client`. But if you can figure out some way to do this without touching `client.js` that might be better.

rawlins (changes requested):
Looks like something we should merge before 0.28 if possible, just a few comments from reading the code. I can probably test on ios at some point.

crawl-ref/docs/options_guide.txt
robertxgray: I didn't find a way of detecting if the user is using a virtual keyboard, so is_mobile() looks for a touch screen instead. I hope that will be correct in most cases, but it's far from perfect. Unless we enable the input helper for everybody, I thought users would benefit from being able to override the detection. I guess the most common use case will be devices featuring both a touch screen and a keyboard.

Zireael07: Unfortunately there are devices that are touch capable wuthout using a virtual keyboard, e.g. more and more Windows tablets.
And, equally as unfortunately, I don't know of a better way of detecting "mobile", as neither touch capability nor screen size are indicative any longer.

robertxgray: While the current criteria is likely to render many false positives, it is not a big issue IMHO. The input helper is semi-transparent and covers only a small part of the map. We can make it even more discrete if needed. A false negative is a bigger problem because the game becomes unplayable, but I can't imagine a situation where that should happen.

crawl-ref/source/webserver/game_data/static/focus-trap.js
robertxgray: I tested again and didn't find any issues, but I'm probably missing some of the places where the trap is used because I'm unfamiliar with it. Please warn me if you can think of any test in particular.

An example use case where removing that listener (or an alternative workaround) is needed can be seen on the video. At 01:45 I look for an altar and the results are shown outside of the screen. That's a common situation because the browser zooms the search input. If the listener is active, I can't zoom out or scroll to see the results.

Thank you for the warning about the removeEventListener call. Note I commented the code, should I have deleted it instead? I'll have another look at this because it would be better not to modify that file.

crawl-ref/source/webserver/static/scripts/client.js
robertxgray: I'll try to move as much code as possible to game.js. Is that right?

There are also changes in static/style.css. Should I use game_data/static/style.css instead?

crawl-ref/source/webserver/game_data/static/focus-trap.js
robertxgray: I was able to revert the changes in focus-trap.js by excluding touch events in ui.target_outside_game.

crawl-ref/source/webserver/static/scripts/client.js
robertxgray: Done.

crawl-ref/docs/options_guide.txt
rawlins: Sounds good for now. We may want to eventually figure out some custom detection for some of the more popular windows devices if possible.

crawl-ref/source/webserver/static/scripts/client.js
rawlins: (yes, css changes in `game_data` are also preferred unless it is something that is completely version-agnostic. But sometimes it has to happen.)

Walther:
Trying to test this, ran into issues when setting up.

```
 python3 webserver/server.py
2022-01-08 22:02:14,679 INFO: Loaded server configuration from: /home/walther/git/crawl/crawl-ref/source/webserver/config.py
2022-01-08 22:02:14,685 INFO: Loading data from base.yaml
2022-01-08 22:02:14,685 INFO: Updating live games config
2022-01-08 22:02:14,685 INFO: Loaded game config seeded-web-trunk (from base.yaml).
2022-01-08 22:02:14,685 INFO: Loaded game config tut-web-trunk (from base.yaml).
2022-01-08 22:02:14,685 INFO: Loaded game config sprint-web-trunk (from base.yaml).
2022-01-08 22:02:14,706 INFO: Listening on :8080
Traceback (most recent call last):
  File "webserver/server.py", line 24, in <module>
    server_main()
  File "webserver/server.py", line 21, in server_main
    webtiles.server.run()
  File "/home/walther/git/crawl/crawl-ref/source/webserver/webtiles/server.py", line 519, in run
    userdb.ensure_settings_db_exists()
  File "/home/walther/git/crawl/crawl-ref/source/webserver/webtiles/userdb.py", line 45, in ensure_settings_db_exists
    settings_db)
NameError: name 'settings_db' is not defined
```

Did not find any relevant docs or tips with quick greps of the codebase for `settings_db` and such - i'm possibly missing some non-obvious setup step or similar. The relevant codepath seems to be here https://github.com/crawl/crawl/blob/a21510a5d8e6c8dd9d1675faab8011a20b666e62/crawl-ref/source/webserver/webtiles/userdb.py#L41-L46

but i'm not sure what the correct fix or workaround would be.

Another file where `settings_db` is mentioned is here https://github.com/crawl/crawl/blob/a21510a5d8e6c8dd9d1675faab8011a20b666e62/crawl-ref/source/webserver/config.py#L51-L53
but that is commented out by default.

---
Context: WSL2 Ubuntu 20.04, built with `make -j$(nproc) WEBTILES=y`

rawlins:
> Did not find any relevant docs or tips with quick greps of the codebase for `settings_db` and such - i'm possibly missing some non-obvious setup step or similar. The relevant codepath seems to be here
> 
> https://github.com/crawl/crawl/blob/a21510a5d8e6c8dd9d1675faab8011a20b666e62/crawl-ref/source/webserver/webtiles/userdb.py#L41-L46

@Walther that looks like a bug in trunk on a fresh setup, I'll fix it soon but for testing purposes it'd be safe to simply comment out that line.

Walther:
Thank you!

Commented out the lines, built and hosted on my raspberry pi 4 for easy access on the devices on the local network.

iPad Pro, Firefox 40.2 (7167)
IMG_0323

iPad Pro, Safari `AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.1 Safari/605.1.15`
IMG_0324

<details><summary>iPhone screenshots, click to expand</summary>

iPhone, Firefox 40.2 (7167)
IMG_0370

iPhone, Safari `AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.2 Mobile/15E148 Safari/604.1`

IMG_0371
IMG_0372
</details>

This certainly works perfectly fine on the iPad, and as a fun little proof-of-concept (but not necessarily ideally playable) on the iPhone.

Half-jokingly, showing a software `<Tab>` button somewhere on the UI in addition to the input field could be a nice touch, as the default OS software keyboards don't provide one and it is fairly useful in the game. Modifiers like `<Ctrl>` etc are of course missing as well, but perhaps not dealbreakers. These kind of things could quickly end up as rather large scope creep.

On the iPad the screen height seems to work reasonably well, as the virtual keyboard will be shown as a floating pop-out. On the iPhone, however, the game will try to use the full screen height of the phone calculated from when the keyboard was not present, and the viewport ends up being super tall. This makes a lot of scrolling necessary after opening the keyboard, when the viewport height gets cut essentially in half. This, too, might not be important, as playing on the small iPhone screen is probably not going to be a great experience even if the height calculations would be adjusted.

rawlins:
Thanks for the patch @robertxgray and the detailed report @Walther! I'm possibly going to merge this roughly as-is today, though I think it's pretty rough on ios. But it's still better than what happens now.

I've done some quick testing both on an iPhone and a not-very-new iPad pro on iPadOS 14.6.1. This was rebased against master, so does include a large batch of menu changes that happened since this PR was submitted. 

> On the iPad the screen height seems to work reasonably well, as the virtual keyboard will be shown as a floating pop-out.

@Walther this wasn't the default for me, is it possible that this is persisting from other websites or apps? Looks like Safari or iPadOS might save the popout state. (In fact I didn't even know that the popout option existed, so it was pretty useful to see your screenshots for this.)

Notes:

* agreed that this probably isn't playable on iPhone, but it's still a minor incremental step.
* keyboard issues: as noted above I don't think popout is a default or ever happens automatically. The default iPad keyboard shifts the display, rather than zooming, which is pretty painful for crawl. From what I understand this behavior is different on Android, and some quick research has convinced me that it would take me a while to get something better working on iOS, so I'm not going to try to fix this before merging. Using a popout keyboard as @Walther did is the solution for now, but I suspect many ios users (like me) don't even know about this feature. Some links for future reference: https://blog.opendigerati.com/the-eccentric-ways-of-ios-safari-with-the-keyboard-b5aa3f34228d?gi=a9f920adc42 https://medium.com/@im_rahul/safari-and-position-fixed-978122be5f29
* keyboard focus issues: it would be better if tapping on the map view didn't hide an open keyboard, seems tricky to do though. Similarly, tapping out of a popup with the keyboard still open loses it under at least some circumstances (not sure why), and entering a popup submenu (e.g. looking at an item description in inventory) always hides the keyboard. Some of these may be interactions with focus-trap?
* some broken interactions with menus/popups, that I think I might be able to fix some of these. Things I noticed: Tapping out of a submenu activates an item in a menu underneath. It's very easy to lose input focus which renders some menus that don't have good mouse support (e.g. the skill menu) unusable without closing and reopening. Multiselect menus aren't currently working (they've changed since this PR so maybe related to that). These seem to be the most needed things to fix before merging. Are these issues iOS specific?
* input box: if the input box starts with its text selected (which maybe shouldn't be possible in the first place), shift is on for some reason, which is very confusing. For some reason, this happens systematically when I open the input after tapping out of a popup, no idea why. Also, the existence of text in this box triggers ios autocompletes based on where the cursor is.
* Agreed that the keyboard on ios is pretty impoverished for crawl and that tab is a notable lack. A solution to this might just be a different key binding, e.g. bind `CMD_AUTOFIGHT` to space instead of tab. As a default binding I think this would be a non-starter as space is used to clear mores, and people have their fingers hard-wired for this, but we do distribute alternate keyboard layouts. If someone produced an alternate binding in rc form we could distribute it, like the dvorak binding etc.
* tap and hold for right click would be very helpful, probably not something I will do now though
* Another pain point was that tap does not interrupt autotravel. This might be non-touch-specific.

rawlins:
I have merged this! Commit sequence: https://github.com/crawl/crawl/compare/ee0477cf1a57...a885d4b3d6c0

Hopefully this doesn't break anything on Android, but I won't be surprised if it needs some followup in the next few days. It's still only marginally playable I'm afraid, but it's not completely unplayable any more.

PR #2333, opened 2022-01-02
Update typed-ast to 1.4.3

robertxgray:
This is required to run the server on Python 3.10+


rawlins:
This seems fine to merge, though to actually run the server I'm pretty sure this isn't necessary (only the `base` reqs should be -- the readme used to have some confusing wording about this but I fixed it recently.)

PR #2336, opened 2022-01-04
fix: phial should always engulf the center target

### LINKED ISSUE #2331: Phial not working correctly on Convoker
Webtiles, CKO, Chrome,  Dungeon Crawl Stone Soup version 0.28-a0-1389-gc726bef33b (webtiles), TeSu^Sif

Somewhere just before turn 77658 I used Phial of Floods on a convoker who was reciting recall. The order of operation in the messages shows that I hit him with the blast of water, then the recall still went off afterwards, and he was not given the waterlogged status. Unsure how to give more information. I have screen capped and relayed the combat log from the point of recall being used through my next action after the recall went off: 

image

Please let me know if I can help in any other way.
[ShapermcTeSu.txt](https://github.com/crawl/crawl/files/7799837/ShapermcTeSu.txt)

perryprog:
Previously, if the phial of floods knocked a target back onto a non-floor or water feature like a staircase, then no flooding would generate at that location (fine) and consequently no waterlogging/engulfing (not fine).

Let's instead make create_feat_splash return a set of coordinates, and add back the center location before iterating over monsters to waterlog.

Closes #2331.

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -2419,8 +2419,7 @@ void bolt::affect_endpoint()
         set<coord_def> splash_coords = create_feat_splash(pos(), is_player ? 2 : 1, num, dur);
         dprf(DIAG_BEAM, "Creating pool at %d,%d with %d tiles of water for %d auts.", pos().x, pos().y, num, dur);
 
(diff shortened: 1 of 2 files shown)


ebering:
Thanks for the patch!

PR #2340, opened 2022-01-07
refactor and a fix: wearing/taking off armour

perryprog:
This refactors a good bit of the logic for taking off armour, as well as a few other things in the inventory wear/unwear section.

I also removed a few TODOs, clarified the semantics of the return values of the wear/unwear armour functions (they are not used anywhere anyway), and added a few doc comments for some of the less intuitive functions.

As a part of this, it fixes a tiles bug where clicking to equip items that were dangerous to put on would give you a prompt twice: this is seen in the changes in tile_item_use.


ebering:
Thanks for the patch!

PR #2341, opened 2022-01-08
fix: moving staircases shouldn't be considered invisible

perryprog:
When you climbed down a staircase while Xom was sliding them around, you could get the message " slides away from you right [...]". The stair was being described by feature_description_at which only uses the player map knowledge, which means the staircase would be DNGN_UNSEEN. This uses feature_description instead to get around that.


ebering:
Thanks for the fix!

PR #2351, opened 2022-01-11
Modification the captain's cutlass

sdynet:
The existing functions were fun, but not vicious. The added option will certainly warrant the pirate's notoriety.


ebering:
This is a megabuff to an already fun and strong unrand, so I think we'll pass on the effect. Nets are limited ammo for a reason, unlimited netting (even with a melee requirement) is too much.

PR #2353, opened 2022-01-11
Warn_before_sad_lava_incident

burnish3d:
Adds a check before potentially accidental item destruction.


ebering:
Thanks for the patch! Merged with tweaks.

PR #2354, opened 2022-01-12
Don't grant free piety to Jiyva worshippers

nlavsky:
Fully exploring the Slime Pits is a risk-free and tedious way of getting exploration piety under Jiyva. Let's not give players an incentive to do this.


ebering:
Thanks for the patch!

PR #2355, opened 2022-01-12
fix: make Trog tantrum if orb of energy is used

kiedtl:
An orb of energy is primarily used for enhancing spellcasting, so it
makes sense that it should be considered a wizardly item and thus that
Trog should frown upon its use.

### DIFF
--- a/crawl-ref/source/god-item.cc
+++ b/crawl-ref/source/god-item.cc
@@ -307,7 +307,8 @@ bool is_hasty_item(const item_def& item, bool calc_unid)
 bool is_wizardly_item(const item_def& item, bool calc_unid)
 {
     if ((calc_unid || item_brand_known(item))
-        && get_weapon_brand(item) == SPWPN_PAIN)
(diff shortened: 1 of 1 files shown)


ebering:
Thanks for the patch and welcome to contributing to crawl! (Or, if you're a returning long time contributor welcome to the GitHub, GH thinks this is your first contribution).

How would you like to appear in the CREDITS.txt?

kiedtl:
> Thanks for the patch and welcome to contributing to crawl! (Or, if you're a returning long time contributor welcome to the GitHub, GH thinks this is your first contribution).

Thanks, I am indeed a first-time contributor. I had been meaning to make a patch for some time.

> How would you like to appear in the CREDITS.txt?

I don't mind either way. Honestly, I don't think that a single `+2 -1` patch is enough to merit it.

ebering:
Everyone gets to be in the credits if they contribute a patch!

kiedtl:
> Everyone gets to be in the credits if they contribute a patch!

In that case, sure, I have no objections :)

PR #2356, opened 2022-01-12
feat: add flavorful descriptions for each orb ego

kiedtl:
The descriptions should probably be put into items.txt at some point (along with the descriptions for ammo brands).

### DIFF
--- a/crawl-ref/source/dat/descript/items.txt
+++ b/crawl-ref/source/dat/descript/items.txt
@@ -743,8 +743,7 @@ forces they could not hope to control.
 %%%%
 orb
 
-A sphere of crystal, carried in place of a shield. Magic scintilliates within
-its interior.
+A sphere of crystal, carried in place of a shield.
 %%%%
 orb of zot
 
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -1838,38 +1838,28 @@ static string _describe_armour(const item_def &item, bool verbose)
             switch (item.brand)
             {
             case SPARM_LIGHT:
-                description += "\n\nThis pale golden orb is slightly warm to the "
-                        "touch. Within you can see a thin coil of some "
-                        "otherworldly black metal, scintilliating with "
-                        "eldritch energy.";
+                description += "Within is a thin coil of some otherworldly "
+                        "black metal, scintilliating with eldritch energy.";
                 break;
             case SPARM_RAGE:
-                description += "\n\nIntricately engraven on the surface is the "
-                        "image of a raging wolf, with two blood-red gems "
-                        "for eyes. The orb is quite warm, and when you "
-                        "touch it, visions of endless carnage and "
-                        "destruction flood your mind.";
+                description += "Engraven on the surface is the likeness of a "
+                        "raging wolf, with two blood-red rubies for eyes.";
                 break;
             case SPARM_MAYHEM:
-                description += "\n\nIt is made out of a dark blue material, "
-                        "unnaturally heavy in your hands. As you hold it, you "
-                        "feel a supernatural force urging you to lash out at "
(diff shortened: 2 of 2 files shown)


kiedtl:
Descriptions, for convenience:

### orb of light
> This pale golden orb is slightly warm to the touch. Within you can see a thin coil of some otherworldly black metal, scintilliating with eldritch energy.

### orb of wrath
> Intricately engraven on the surface is the image of a raging wolf, with two blood-red gems for eyes. The orb is quite warm, and when you touch it, visions of endless carnage and destruction flood your mind.

### orb of mayhemcase SPARM_MAYHEM:
> It is made out of a dark blue material, unnaturally heavy in your hands. As you hold it, you feel a supernatural force urging you to lash out at everything around you.

### orb of guile
> It is shaped in the form of an unblinking eye, with a single blue gemstone sitting in its catlike pupil.

### orb of energy
> Encased within the transparent material, you can see a spiral of silver wire wrapped around an iron rod. Seven pink gems adorn the orb, carven in the likeness of some small extinct mammal.

### orb of bugginess
> As you stare into the sphere, you are filled with an unnatural longing for horse-drawn carriages.

semitonal:
It sounds reasonable to add a little more flavour to the orb descriptions, but these are a little too flowery for Crawl's usual description style - we avoid directly addressing the player in item descriptions with things like "You feel <x>" or "You can see <x>", and flavour descriptions like this should be kept very brief (usually just a single short sentence) since there's a lot of information to communicate on description screens. Your orb of guile description here I'd say is probably the best example of the sort of style and length to aim for.

This kind of flavour description should definitely also be in the database files as you mention rather than in C++, but I'm not sure at a glance how easy that is to do for things like orbs where it's tied to armour egos rather than subtypes.

kiedtl:
> It sounds reasonable to add a little more flavour to the orb descriptions, but these are a little too flowery for Crawl's usual description style - we avoid directly addressing the player in item descriptions with things like "You feel <x>" or "You can see <x>", and flavour descriptions like this should be kept very brief (usually just a single short sentence) since there's a lot of information to communicate on description screens. Your orb of guile description here I'd say is probably the best example of the sort of style and length to aim for.

Thank you for the feedback, I'll revamp the descriptions come Monday (as my hands are tied until then).

> This kind of flavour description should definitely also be in the database files as you mention rather than in C++, but I'm not sure at a glance how easy that is to do for things like orbs where it's tied to armour egos rather than subtypes.

I did look into doing this, but in the end I gave up on trying to understand certain bits of the code.

Could this be merged first, and then a PR that fixes this issue (and the ammo descriptions) be merged later?

ebering:
I'm not as convinced of per-ego flavor descriptions, neither scarves nor magic staves have such a thing. Item descriptions (like so many of our description menus) are already jam-packed with information, for mundane items (as opposed to artefacts) we don't want to overwhelm the player with flavour text.

The description of orb I wrote was a quick one-off, and changing it to suggest a broader realm of possibilities of orb decoration and style would be welcome.

kiedtl:
> I'm not as convinced of per-ego flavor descriptions, neither scarves nor magic staves have such a thing.

They should, in my humble opinion.

> Item descriptions (like so many of our description menus) are already jam-packed with information, for mundane items (as opposed to artefacts) we don't want to overwhelm the player with flavour text.

It's just one or two sentences at most, though if you like I can shorten it further. If that's not enough, currently the flavor description is in the same paragraph as the "carried in place of a shield" description. If it was made a separate paragraph again, would that make it easier for a reader to skip the flavor text?

> The description of orb I wrote was a quick one-off, and changing it to suggest a broader realm of possibilities of orb decoration and style would be welcome.

"Magic scintilliates in the interior" seemed as bland and general as a flavor text could be to me.

PR #2357, opened 2022-01-14
Add new option to hide the blood

mhcerri:
Add a new option show_blood to control if the blood splatters and
bloody corpses should be shown or not. The new option defaults to
true to maintain the previous behavior.

### DIFF
--- a/crawl-ref/source/tiledgnbuf.cc
+++ b/crawl-ref/source/tiledgnbuf.cc
@@ -11,6 +11,7 @@
 #include "tiledoll.h"
 #include "tilemcache.h"
 #include "tilepick.h"
+#include "options.h"
 
 DungeonCellBuffer::DungeonCellBuffer(const ImageManager *im) :
     m_buf_floor(&im->m_textures[TEX_FLOOR]),
@@ -223,6 +224,9 @@ void DungeonCellBuffer::draw()
 void DungeonCellBuffer::add_blood_overlay(int x, int y, const packed_cell &cell,
                                           bool is_wall)
 {
+    if (!Options.show_blood)
+        return;
+
     if (cell.is_liquefied && !is_wall)
     {
         int offset = cell.flv.special % tile_dngn_count(TILE_LIQUEFACTION);
--- a/crawl-ref/source/tilepick.cc
+++ b/crawl-ref/source/tilepick.cc
@@ -2849,7 +2849,7 @@ tileidx_t tileidx_item(const item_def &item)
 #endif
 
     case OBJ_CORPSES:
-        if (item.sub_type == CORPSE_SKELETON)
+        if (!Options.show_blood || item.sub_type == CORPSE_SKELETON)
             return _tileidx_bone(item);
         else
             return _tileidx_corpse(item);
--- a/crawl-ref/source/tileweb.cc
+++ b/crawl-ref/source/tileweb.cc
@@ -1621,11 +1621,13 @@ void TilesFramework::_send_cell(const coord_def &gc,
             write_tileidx(next_pc.cloud);
         }
 
-        if (next_pc.is_bloody != current_pc.is_bloody)
-            json_write_bool("bloody", next_pc.is_bloody);
+        if (Options.show_blood) {
+            if (next_pc.is_bloody != current_pc.is_bloody)
+                json_write_bool("bloody", next_pc.is_bloody);
 
-        if (next_pc.old_blood != current_pc.old_blood)
-            json_write_bool("old_blood", next_pc.old_blood);
+            if (next_pc.old_blood != current_pc.old_blood)
+                json_write_bool("old_blood", next_pc.old_blood);
+        }
 
         if (next_pc.is_silenced != current_pc.is_silenced)
             json_write_bool("silenced", next_pc.is_silenced);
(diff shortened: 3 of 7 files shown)


mhcerri:
Sorry, I had to make a few adjustments to make sure it was working in tiles, webtiles and console.

no_blood

dillytin:
How do you distinguish between corpse and skeleton?

mhcerri:
Hi, @dillytin. The description and everything is still the same.

mhcerri:
Just to add some background. It would be nice to me to be able to hide the gore when playing around young kids.

crawl-ref/source/tiledgnbuf.cc
```diff
@@ -223,6 +224,9 @@ void DungeonCellBuffer::draw()
 void DungeonCellBuffer::add_blood_overlay(int x, int y, const packed_cell &cell,
                                           bool is_wall)
 {
+    if (!Options.show_blood)
+        return;
+
     if (cell.is_liquefied && !is_wall)
```
rawlins, line 230: I'm pretty sure this block is for Leda's Liquefaction, not blood, so the options check should maybe be in the `else if` condition below

crawl-ref/source/tileweb.cc
```diff
@@ -1621,11 +1621,13 @@ void TilesFramework::_send_cell(const coord_def &gc,
             write_tileidx(next_pc.cloud);
         }
 
-        if (next_pc.is_bloody != current_pc.is_bloody)
-            json_write_bool("bloody", next_pc.is_bloody);
+        if (Options.show_blood) {
```
rawlins, line 1624: maybe better just to send `false`. Also, `old_blood` is not used if `bloody` is false, so can be ignored here.

crawl-ref/source/tilepick.cc
```diff
@@ -2849,7 +2849,7 @@ tileidx_t tileidx_item(const item_def &item)
 #endif
 
     case OBJ_CORPSES:
-        if (item.sub_type == CORPSE_SKELETON)
+        if (!Options.show_blood || item.sub_type == CORPSE_SKELETON)
```
rawlins, line 2852: Maybe could be merged without this, but I do agree with dilly's comment that this distinction matters for gameplay purposes if you're playing a character using certain spells/abilities, so it would be best to have some other indicator of skeleton vs corpse if at all possible. The most obvious thing might be to use a `TILE_FLAG..`, but I don't see any that could be repurposed.

Other brainstorming: I could also imagine recoloring the skeleton tiles, though I guess you are trying to not depict even the color of blood. Or some kind of simple visual tweak to the skele tiles? For reference, console distinguishes them pretty simply, like this: `†÷`, so whatever's here wouldn't need to be super perfect, and doesn't need to implement all the body shapes for the corpse-skele case IMO.

Also, extremely minor, but there's some flavor stuff that `_tileidx_corpse` does that is unrelated to blood.

rawlins (changes requested):
Thanks for the PR! Some quick comments, the important one to fix to merge this is the (I think accidental) Leda's interaction.

crawl-ref/source/tiledgnbuf.cc
mhcerri: I will fix that.

crawl-ref/source/tilepick.cc
mhcerri: Thanks for the feedback, @rawlins. I will experiment with some alternatives on how to represent the corpses.

randomMesh:
Why remove only the blood? If you really want to protect children, replace the weapons with lollipops and candy bars as well...

mhcerri:
@randomMesh, not sure what's the purpose of such comment. Nothing is being replaced, it's just one option for people who want to use it.

randomMesh:
I just wanted to express my struggle with understanding the fact that it seems to be normal for Crawls community to remove harmless blood stains "bEcAuSe oF thE cHiLdrEn".

> It would be nice to me to be able to hide the gore when playing around young kids.

It's a f*cking Roguelike with monsters, axes and dark magic, just don't play it when kids are around!

For me it's a direction i don't want to see Crawl is heading. Male Harpies were the beginning, now removal of blood, what next?

Sorry for the rant.

rawlins:
@randomMesh this is not an appropriate place for such a rant, and attacking contributors (regardless of whether you agree with the direction of their commit) in this fashion constitutes unacceptable harassment and a violation of github community guidlines (https://docs.github.com/en/github/site-policy/github-community-guidelines). This is not your first strike, I'm giving a tempban. If you chose to return to commenting on this project after that, and continue to engage in behavior inconsistent with the github community guidelines, it will become permanent.

RojjaCebolla:
image

rawlins:
For context, a suggestion in dev chat was that you could use a tombstone for what are regularly skeletons, so that's why @RojjaCebolla posted that.

Also, just to make a specific statement on this option as a dev team member: it definitely seems to me like it would be a good thing to have for accessibility purposes. Lots of games have it (https://www.taminggaming.com/en-us/search/accessibility/174-Turn+Off+Blood) so it is very far from a strange or unusual thing, and I don't think that children would even be the only use case for it. So thanks for putting together a patch for it. I don't think we'll get this in before the 0.28 release which is coming up very sharply, but some time soon after the tournament may be possible.

PleasingFungus:
Now that Hellmonk has reworked magic to not rely on corpses, this is workable without further modification. I've merged it with a few fixes that Rawlins suggested. TY for the patch! :)

PR #2360, opened 2022-01-15
fix: click + key modifier detection in tiles

### LINKED ISSUE #898: Shift+Left Click sometimes equips weapon instead of dropping it
When reporting a bug, please include the following:

**PLATFORM:**
Local tiles, tested on Linux

**VERSION:**
0.22.1

**DESCRIPTION:**
When you have a weapon in the inventory that you want to drop, Shift+Left Clicking it sometimes equips it instead of dropping it. I can't find any specific way to reproduce this behavior, except that when the issue starts to occur, it continues to occur until you save the game and load it again. If I do find a specific way to reproduce it, I'll comment to let you know.

dessaya:
Fixes #898

The issue was caused because `SDL_GetModState` returns an OR'd combination of the modifier keys, but `SDLWrapper::get_mod_state` was comparing against the exact values of the flags.

### DIFF
--- a/crawl-ref/source/windowmanager-sdl.cc
+++ b/crawl-ref/source/windowmanager-sdl.cc
@@ -646,28 +646,9 @@ unsigned int SDLWrapper::get_ticks() const
     return SDL_GetTicks();
 }
 
-tiles_key_mod SDLWrapper::get_mod_state() const
+unsigned char SDLWrapper::get_mod_state() const
 {
-    SDL_Keymod mod = SDL_GetModState();
-
-    switch (mod)
-    {
-    case KMOD_LSHIFT:
-    case KMOD_RSHIFT:
-        return TILES_MOD_SHIFT;
-        break;
-    case KMOD_LCTRL:
-    case KMOD_RCTRL:
-        return TILES_MOD_CTRL;
-        break;
-    case KMOD_LALT:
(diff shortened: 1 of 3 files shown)


dessaya:
Actually you are right, this commit does not completely fix #898. I'd say it fixes 90% of it :sweat_smile: 

I think there are two causes for #898:

1. The incorrect handling of the bitmask when some other modifier key is pressed (e.g. shift + alt)
2. If you press some modifier keys and at the same time the SDL window loses or gains focus (e.g. press shift, then click on some other window, then release shift and come back), SDL gets confused and reports a wrong state for the modifier keys.

I think (2) was being exacerbated by (1), especially if you use e.g. alt+tab or ctrl+alt+arrow keys to move around the desktop.

This commit fixes (1). I have no idea how to fix (2), but there is a workaround: if you ever find that shift is not working properly, just press and release it, and SDL will fix its state. Previously this workaround did not work just as well because you would also have to press/release ctrl and alt.

rawlins:
Thanks, after your comment I did a bit of searching around and found: https://github.com/libsdl-org/SDL/issues/671 which looks like it would have caused exactly the #898 behavior given current modifier handling. It's not clear that there was an explicit fix in SDL2 itself, but that workaround in the report looks reasonable to do anyways.

PR #2361, opened 2022-01-15
feat: Replace Teleportitis with Blinkitis

Oneirical:
Teleportitis has been named many times by DCSS fans as the mutation with the least counterplay and the most dangerous effects. Obtaining Teleportitis with no quick way to cure it is a run ender for many characters, effectively locking them out of the extended endgame if they cannot realistically obtain extra mutation potions or convert to Zin/Jiyva. Every step taken in a new level could easily bring the player into a death trap - and as bad mutations tend to stack, -Scroll in conjunction to Teleportitis can make escape impossible. Compared to other bad mutations, Teleportitis is horrible to have even at level 1, and encourages tedious play, such as constantly retreating to rest on previously explored levels instead of simply resting at the risk of ending up in a death trap while the player is healing.

This pull requests replaces the dreaded Teleportitis with a more dubious, potentially useful, Xom-approved counterpart in the same design principle as the "dangerous, but with a perk" bad mutations such as Berserkitis or Subduded Magic. Now that the Abyss can no longer be scummed, bad mutations should have their severity reduced, as their purpose as a way to discourage scumming is now obsolete.

With this mutation, upon taking damage that is greater than 10% of the player's max HP and upon succeeding on a [mutation level]/4 chance roll, the player will randomly blink. This can be useful in some cases to escape trouble, but most of the time, the unpredictable effects can be disastrous, whether it's stopping stair-dancing strategies in their tracks, getting chain-blinked into a dangerous area, or, if one is playing a melee character, be forced to eat a few ranged attacks as one attempts to return in range.

I have additional ideas to convert the most infamous bad mutations into these "dubious" designs, but I wanted to test the waters first by addressing the most dreaded mutation out of the lot with a simple idea.


semitonal:
The premise here is sort of a flawed one - teleportitis is by no means a "run ender", and there are plenty of ways to cure it or to win despite it. In addition, even if mutations that "locked players out of the extended endgame" were a thing that existed, this wouldn't necessarily be a problem. The extended endgame is optional, and if you fail to build your character well, or fail to prepare correctly around mutators in the first place and allow yourself to become heavily malmutated, then it's both expected and intended that it could become extremely difficult to clear these optional areas.

PR #2364, opened 2022-01-16
Menu highlight option

Aliscans:
I've made a couple of changes which I think would improve how the new menu layout works, particularly for console users. I've:

1. Created an option (menu_highlight) to control how a highlighted menu row is shown. I know this affects console, and doesn't affect local tiles, but I don't know how webtiles shows things.

"background" gives the selected row a grey background.
"inverse" switches the foreground and background colours.
"square" adds [ and ] to the selected row.

This does not affect the start menu, or any other menu which doesn't use the Menu class.

My idea here is that you can't control how third party clients display text, but there should be something here which looks reasonable on most systems.

2. Changed the shop menu so that space is "toggle selection for highlighted item" rather than "buy"; a synonym for  '.' rather than Enter.

There's nothing in the instructions to say what space should do, or say how to select an item except by letter, so this is closer to what I expected to happen when I first saw it.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -74,7 +74,7 @@ The contents of this text are:
                 show_travel_trail, skill_focus, default_show_all_skills,
                 monster_list_colour, view_delay, use_animations,
                 darken_beyond_range, force_more_message, flash_screen_message,
-                cloud_status, always_show_zot, action_panel,
+                cloud_status, always_show_zot, menu_highlight, action_panel,
                 action_panel_filter, action_panel_show_unidentified,
                 action_panel_scale, action_panel_orientation,
                 action_panel_font_family, action_panel_font_size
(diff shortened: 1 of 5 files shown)


rawlins:
I'm sort of actively working on this and have some stuff that is not consistent with this, will look at it after that. It might be a good idea to check in re ongoing projects before submitting a PR :-).

rawlins:
I've taken the shop commit, which is straightforward. I'm not sure about the highlight option, but I don't think I want to merge it as-is:

* A lot of the angst about this was not actually people who wanted a better version of it, but just people who wanted it to go away. I've provided an option for them already.
* In my testing I don't see any problem with darkblue (the new 8 color default) in terms of readability, so I'm not sure the other options here are urgently needed -- the putty readability issue was fairly specific to lightgray.
* I'm really wary of anything that changes item text, especially when it requires assumptions about what subclasses do. (This is why I was *extremely* happen to fully remove the old selection code -- granted it was a worse implementation than yours, but still had encapsulation issues.)
* The inversion idea may be valuable and a higher contrast mode would be good, but this implementation I think will lead to some readability issues with some color combos and it would need to be calibrated a bit more. Also, inverting is already implemented at a very low level (in `libunix.cc` for linux/mac) just for a very narrow set of color combos, and so it might be better for the low-level thing to be expanded with an option. (E.g. invert both magenta and lightmagenta on a magenta backgorund, instead of just magenta, and so on.) I wrote the new color display diagnostic popup with an eye to calibrating such an option.

rawlins:
I'm going to go ahead and close this -- I think that console menu readability was in an ok state by the time of release, at least for people who wanted it, and the notes above still stand. I'm not opposed to further increasing readability by e.g. improving the contrast for highlight inversion, but I think the implementation would be somewhat different.

Aliscans:
No problem. The main reason I wanted to do it was to help myself.

I've selected colours for Crawl which mean that I can easily distinguish them from one another, and I can read any of them against a black background. Some have a similar brightness to dark grey, making that parts of the inventory menu were hard to read with a dark grey background.

I know there are other ways I could address this problem, but this one works for me.

Dark grey on red (another combination which was added to Crawl recently) is also a problem, but I rarely go into the skill menu to see it.

PR #2365, opened 2022-01-16
Adjusting Zin in hell

mapreduceflower:
### the main keywords of Zin are as follows : 
* good god
* mutation resistance
* hell effect resistance
* protection & control

### hell has been readjusted as follows : 
* increase the weight class of monsters and increase the number of neutral monsters.
* hell effect have been divided into two.

changes have been made, but there is no correction for Zin.
>* Although the hell effect is divided into two, resistance works at a 50% probability only on one side, effectively functioning at 25% (nerf)
>* The power of recite in hell does not function properly due to increased the number of neutral monsters, increased the weight class of monsters, and changes in hell effect (nerf)
>* Sanctuary is also blocking the half effect by changing the hell effect (nerf)

nerf about god are possible at any time. but Zin has a identity of a good god specialized in hell, this identity speicified in hell effect resistance and sanctuary, and this nerf undermines his identity. Zin has become as difficult to see as an attractive option in hell as in the past.

### my proposal is as follows :
provide Zin with appropriate compensation for hell rework.
>* easy option: Zin's hell resistance now resists the hell effect from 50% to 100%.
>* hard option: Zin's hell resistance now resists the hell effect from 50% to 100%. Sanctuary temporarily invalidates the permanent hell effect.
>* strange option: Reading the book of inferno when recite in hell. If successful, the permanent hell effect is temporarily invalidated


semitonal:
Recite is already a very complex ability - giving it a new set of branch-specific behaviour doesn't sound like a good direction to me. It's also worth noting that while Zin doesn't protect from the passive Hell effects, the timed effects became much more nasty with the Hell effect changes, so the existing protection from those goes up in relative value. Zin remains an extremely strong god in the late portions of the game, so I'm not convinced that any compensation is needed here.

ebering:
Agree with Kate's assessment of this. Also, the hell rework made hell nastier for every character and every god, intentionally, so I don't think Zin needs specific compensation.

PR #2366, opened 2022-01-18
Allow upward stone stairs to be used for LRD

RojjaCebolla:
Tested: 
image

Swamp can be a difficult place in which to find any valid LRD targets.
For those who play in tiles mode, it can be confusing that up-stairs
aren't targets for LRD; it's clearly a stone structure and is described
as such as well. 
image

Down stairs on the other hand, are a hole in the floor. 
image Also, they're more friendly to stair-dancing, because that tends to be safe, cleared territory, so of the two, upstairs are the more interesting as a last-ditch LRD spot. One needs a lot of AC to sit on a stairwell and blow it up!

Floated this idea in IRC but I thought it might get more thoughts and comments if I do it up as a PR. Thanks for lookin!

### DIFF
--- a/crawl-ref/source/spl-damage.cc
+++ b/crawl-ref/source/spl-damage.cc
@@ -1188,6 +1188,9 @@ static const map<dungeon_feature_type, feature_frag> fraggable_terrain = {
     { DNGN_ORCISH_IDOL, { "rock", "stone idol" } },
     { DNGN_GRANITE_STATUE, { "rock", "statue" } },
     { DNGN_PETRIFIED_TREE, { "rock", "petrified wood" } },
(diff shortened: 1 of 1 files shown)


ebering:
This is pretty net-hacky. It also opens the "why can't I LRD floor" can of flavour worms. Currently everything LRD-able is (or can be rendered, ugh open doors) impassable; what's LRD-able is a bit of a rough spot for clarity but that's at least something. Adding stairs would break the pattern, for a dubious gameplay benefit (the lack of LRD sources in Swamp makes it stand out, guaranteeing a few on every floor would diminish this).

Picking out only upstairs (as opposed to all stairs (what about hatches??)) is also a clarity shortcoming from the point of view of mechanical clarity; while the reasons for only upstairs may be "clear" in tiles, it's not clear in console.

Issue #2346, opened 2022-01-10
Shift + Movement + Door creak makes you move one more turn


dillytin:
**DCSS VERSION AND GAME INFORMATION:**
0.28-1435-g8239

**DESCRIPTION:**
Shift movement makes your char move to a certain direction until something changes. If you use shift movement to opening a door, it usually just opens the door and stops. However, if a door creaks loudly and you have no enemy in los, somehow your char decides to move one step further. This is just one of **extremely trivial and unimportant bugs** I've managed to find. 

**EXTRA FILES:**
![image](https://user-images.githubusercontent.com/65288602/148727642-03c33e5f-e998-490e-90cb-e4dc3315ea84.png)
This is when you just open a door and it doesn't creak. Your char doesn't move.

![image](https://user-images.githubusercontent.com/65288602/148727562-243b64ed-5acb-4917-816e-539e622d3e61.png)
You can see your char moves onto the door and takes 2.0 turns.

PR #2368, opened 2022-01-19
Stop running when opening a door with shift + direction

liammcd:
Using shift + direction to open a door would continue running for another turn when the door creaked, the player would remain stationary if the door remained silent. Reported in issue #2346

### DIFF
--- a/crawl-ref/source/movement.cc
+++ b/crawl-ref/source/movement.cc
@@ -1103,8 +1103,6 @@ void move_player_action(coord_def move)
              || !you.running)
         && !attacking && feat_is_closed_door(env.grid(targ)))
     {
-        if (you.running == RMODE_START)
-            stop_running();
         open_door_action(move);
         move.reset();
         return;
--- a/crawl-ref/source/movement.cc
+++ b/crawl-ref/source/movement.cc
@@ -1103,7 +1103,8 @@ void move_player_action(coord_def move)
              || !you.running)
         && !attacking && feat_is_closed_door(env.grid(targ)))
     {
-        stop_running();
+        if (you.running == RMODE_START)
+            stop_running();
         open_door_action(move);
         move.reset();
         return;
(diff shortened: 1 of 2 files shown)


rawlins:
Thanks for the first PR! This is a good try but it runs up against the fact that starting/stopping travel is way more baroque that it initially appears and extremely (some would say "too") configurable. In general, optional run stops shouldn't be hardcoded but instead should rely on the values in `runrest_stop_message`.

I believe what is happening with this bug is that the regular door message (in `player_open_door` in `player.cc`) is just on the default message channel, but the creak message happens on MSGCH_SOUND, which on the default settings is ignored for running purposes (by the line `ignore += sound:`, in [crawl-ref/source/dat/defaults/runrest_messages.txt](https://github.com/crawl/crawl/blob/master/crawl-ref/source/dat/defaults/runrest_messages.txt)). I *think* the reason it stops after that is just the usual feature detection of shift-travel.

A possible solution is that, despite the noise, this `mprf` should just happen on the same channel as regular door messages so it skips this ignore. (I see a few other messages in that function that, while they do involve noise, maybe should be on the regular channel as well for similar reasons.) It may also be possible to add a specific `runrest_stop_message` (maybe something like `runrest_stop_message += .*it creaks loudly.*`, though I haven't carefully checked this wording) that would override the channel ignore. Either of these solutions seems ok to me on first glance.

Feel free to take another stab at the fix if you like (rebasing in PRs is fine so you could keep working in this one).

liammcd:
Thanks for the feedback, definitely more nuanced than expected. I moved the creaking message to the regular message channel as well as the berserk version as this had the same behaviour.

ebering:
Thanks for the patch! I ended up fixing this by just changing the runrest_stop defaults, without making a code change, in the referencing commit. `runrest_stop_message` and `runrest_ignore_message` are two names of a common pseudo-list option where the first match (stop or ignore) is the one that applies, so adding an appropriate stop match before the default `runrest_ignore_message += sound:` was what I did.

PR #2372, opened 2022-01-20
Improve the plural of "<name>'s ghost", and the spelling of wilful.

Aliscans:
1. Pluralise ghosts with the same name as "2 ghosts called <name>". I came across two identical ghosts in a vault, and thought that "2 <name>'s ghosts" wasn't very good English.

Two other possible approaches to this issue are:
a. Only merge the names of identically named ghosts if they are being merged at "genus" level (at which point the names disappear).
b. Only allow each name to be used for a ghost once in a game.

2.  Change "willful" to "wilful", and "skillful" to "skilful", as the longer version is a US spelling.

### DIFF
--- a/crawl-ref/source/dat/descript/monsters.txt
+++ b/crawl-ref/source/dat/descript/monsters.txt
@@ -785,7 +785,7 @@ skilled in archery.
 centaur warrior
 
 A muscular centaur, veteran of numerous battles, armed with well-made weapons
-and able to use them skillfully.
+and able to use them skilful.
 %%%%
 chaos spawn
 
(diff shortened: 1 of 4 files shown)


crawl-ref/source/dat/descript/monsters.txt
```diff
@@ -785,7 +785,7 @@ skilled in archery.
 centaur warrior
 
 A muscular centaur, veteran of numerous battles, armed with well-made weapons
-and able to use them skillfully.
+and able to use them skilful.
```
ebering, line 788:
```suggestion
and able to use them skilfully.
```

Still want the adverb

Aliscans: Sorry; I thought I'd changed that. The "draconian knight" description also has the wrong spelling of "skilfully".

PR #2373, opened 2022-01-20
update acquirement armor gen debugging with new egos

christopherlandry:
debug acquirement generation couldn't track new armor egos.  Also lacked compiler check to ensure they're in sync the way weapons did, and fixed a small bug where it assumes num weapon brands always > armor egos.

### DIFF
--- a/crawl-ref/source/wiz-item.cc
+++ b/crawl-ref/source/wiz-item.cc
@@ -875,11 +875,11 @@ static void _debug_acquirement_stats(FILE *ostat)
     int randbook_spells = 0;
 
     int subtype_quants[256];
-    int ego_quants[NUM_SPECIAL_WEAPONS];
+    int ego_quants[256];
 
     memset(subtype_quants, 0, sizeof(subtype_quants));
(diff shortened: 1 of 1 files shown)


crawl-ref/source/wiz-item.cc
```diff
@@ -1173,9 +1175,17 @@ static void _debug_acquirement_stats(FILE *ostat)
...
             "harm",
+            "shadows",
             "rampaging",
             "infusion",
+            "light",
+            "rage",
+            "mayhem",
+            "guile",
+            "energy",
+            "debug_randart",
```
christopherlandry, line 1186: item-prop-enum.h's list of egos ends with NUM_REAL_SPECIAL_ARMOURS and then NUM_SPECIAL_ARMOURS. Not sure what the real_special is for so put a placeholder here for that one.

ebering: REAL_SPECIAL is meant for the end of the egos that can spawn. I'll tweak this when I merge it.

PR #2375, opened 2022-01-20
The final element of the Hell reform

nlavsky:
"A hideous howling noise" was a lie: the horn has never produced a sound.

This commit gives the horn a proper noise.


ebering:
Thanks for the patch!

PR #2387, opened 2022-01-28
feat: add a new randart name

Implojin:
of @god_name@'s Attic


ebering:
Thanks for the patch!

PR #2392, opened 2022-01-31
Fix experience gain from noxious bog kills

elliottbernstein:
This fixes a bug in which monsters killed by Eringya's Noxious Bog never result in experience gains for the player.


elliottbernstein:
I can see this fix being an issue if monsters cast the spell, although I'm not aware of any monsters who cast it currently.

I was unsure of how to make `oppressor` you or the casting monster, so would appreciate advice on how to do so, if it's an important case to cover.

PleasingFungus:
Fixed in 89e12c0570e3e488ff8f3b7d25d32e2fbd9837b6, thanks! (Didn't see this PR before I merged my own fix, sorry.)

PR #2398, opened 2022-02-02
acquirement can now weight base class types

christopherlandry:
Refactors acquire.cc to allow weighting of base class types.  Previously, scrolls and other random methods relied on a simple array of base types that were randomly shuffled and drawn from like a deck of cards.  With this change, instead the game randomly chooses a weighted item and checks that against the list of previously acquired items to ensure uniqueness (which is reset upon the player picking from their acquirement selection).  
Marking this as a draft because, while it seems to work fine in game when invoked via scroll, my attempt to add a [g] random to debug acquirement in wizard mode is generating odd/bad reports in items.stat.  I can work on fixing that but I wanted some feedback this was even a worthwhile change.

Also, I took a stab at weighting the base types in a skill-appropriate matter but I don't think it's very good. Probably needs to be less all-or-nothing based on stats, early acquirement is very heavily weighted towards specific skills on a character.

bonus - more comments on how acquire works and method signatures therein

### DIFF
--- a/crawl-ref/source/acquire.cc
+++ b/crawl-ref/source/acquire.cc
@@ -85,6 +85,7 @@ static int _skill_rdiv(skill_type skill, int mult = 1)
 static int _acquirement_armour_subtype(bool divine, int & /*quantity*/,
                                        int /*agent*/)
 {
+    //would realistically make sense to move XOM unwanted armor check here
     const equipment_type slot_type = _acquirement_armour_slot(divine);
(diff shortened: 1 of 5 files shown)


Hellmonk:
I believe we want to keep current acquirement behavior, ie skill-independent weighting between categories. Possibly some larger refactor of acquirement would make category weighting by skill sensible, but this pull request is three years old and acq code has changed significantly since its creation.

PR #2401, opened 2022-02-04
Give berserk players a chance to break down the door

liammcd:
Let the player get in on the door breaking action. Gives a 20% chance to break down the door when berserk. Adds Lua marker property door_berserk_verb_break to customize the verbiage.


PleasingFungus:
Thanks for submitting this! The flavour is great, but the mechanical impact seems undesirable - it's just kind of a weird and hard to understand gotcha for players on the rare occasion that they open doors while berserk. I considered this while implementing door breaking and intentionally didn't include it - apologies for not including a comment to clarify!

PR #2407, opened 2022-02-07
Implement CMD_REVEAL_OPTIONS on GNU/Linux

robertxgray:
I have implemented CMD_REVEAL_OPTIONS on GNU/Linux in the same fashion as Mac, but there are two details where I'm not sure if I'm getting the expected behaviour. Can you confirm please?

First, if there's not custom configuration in the user's home, the global configuration is shown instead.

Second, there's no icon in the menu option.
imagen

### DIFF
--- a/crawl-ref/source/command-type.h
+++ b/crawl-ref/source/command-type.h
@@ -156,7 +156,7 @@ enum command_type
 
     CMD_SHOW_CHARACTER_DUMP,
     CMD_GAME_MENU,
-#ifdef TARGET_OS_MACOSX
+#if defined TARGET_OS_MACOSX || defined TARGET_OS_LINUX
     CMD_REVEAL_OPTIONS,
 #endif
     CMD_LUA_CONSOLE,
--- a/crawl-ref/source/main.cc
+++ b/crawl-ref/source/main.cc
@@ -1964,6 +1964,10 @@ public:
         add_entry(new CmdMenuEntry("Show options file in finder",
             MEL_ITEM, 'O', CMD_REVEAL_OPTIONS));
 #endif
+#ifdef TARGET_OS_LINUX
+        add_entry(new CmdMenuEntry("Show options file in a file manager",
+            MEL_ITEM, 'O', CMD_REVEAL_OPTIONS));
+#endif
         add_entry(new CmdMenuEntry("", MEL_SUBTITLE));
         add_entry(new CmdMenuEntry(
                             "Quit and <lightred>abandon character</lightred>",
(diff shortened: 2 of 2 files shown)


rawlins:
can any devteam members evaluate this for linux compat? I guess it hinges on how widespread `xdg-open` is. I'm sort of inclined to click the merge button if no one responds (but I will be doing that without any testing myself).

robertxgray:
Will #2873 make CMD_REVEAL_OPTIONS obsolete?

rawlins:
> Will #2873 make CMD_REVEAL_OPTIONS obsolete?

I think it's going to be a while, if ever, before manual rc editing is gone; even if we were to merge that PR as is, there's a whole set of options handled purely in lua that it doesn't deal with. (I still haven't merged this myself because I can't test it, but maybe I'll actually set up a linux device soon.)

Aliscans:
> Will #2873 make CMD_REVEAL_OPTIONS obsolete?

I don't think so. Crawl uses Lua in configuration files, so they always have the potential to be more than just the set of "option = value" statements #2873 gives.

There may be a few rough edges at the moment, though. When I use the function on my system, I get the following output:

File manager showing contents of a directory from the Crawl distribution

First, it opens a file manager (the one in `/etc/mailcap` in this case) to show the directory which contains the default save file. I think this works as it's supposed to.

If `mailcap` doesn't have a file manager in it, it skips this step.

Crawl in-game menu with an error message spread across the bottom half of the screen.
Then it prints a badly formatted warning message. This may be completely hidden if you use tiles, as that doesn't do much with the original window.

I don't know if printing `stdout`/`stderr` messages from a running file manager could make it harder to use, but I think you should be able to call something like replay_messages() to view them later on, whatever happens at the time.

rawlins:
Thanks for testing, @Aliscans! Given this, I would definitely wonder what happens in tiles on your system, because if it opens a console-based file manager in the same terminal that crawl was started from, I don't know how usable the results would be in general.

The other thing I was wondering is whether there should be a check if `xdg-open` is installed in the first place, somehow. I am assuming it must be pretty widespread, but is it 100% guaranteed? On mac the comparable assumption is safe because `open` is a macos system utility that is installed in `/usr/bin` with every OS version going back maybe to the original OS X release. (And its behavior for this case is safe, it will only ever open a new finder window.)

Also, re the original question about the icon, there's still no icon yet for that menu option, so it just doesn't show anything.

robertxgray:
Oh, I tough this was a tiles-only feature. I think I'm just going to close this PR if you don't mind and consider other options.

rawlins:
FWIW I think it would be completely fine to limit it to tiles linux builds if you had something that works well for that case..

Aliscans:
> Thanks for testing, @Aliscans! Given this, I would definitely wonder what happens in tiles on your system, because if it opens a console-based file manager in the same terminal that crawl was started from, I don't know how usable the results would be in general.

With local tiles, it opens in the window I ran Crawl from, and blocks the window Crawl opens itself until I finish. If I close the window I ran Crawl first, it blocks the game for a second or so, but doesn't produce any visible output.

With webtiles, it prints the warning messages as part of the server's output, with or without DGAMELAUNCH. I don't know if it would make sense to use a file manager at all.

> The other thing I was wondering is whether there should be a check if `xdg-open` is installed in the first place, somehow. I am assuming it must be pretty widespread, but is it 100% guaranteed?

What might be better would be to have an option to control what this command does (maybe with `xdg-open '%s'` as a default). It doesn't stop the command from failing, but it does give the user the tools to fix it if it does.

robertxgray:
Thank you for your time. This needs a big rework, so I'll better close it for now.

PR #2415, opened 2022-02-12
Rework the static discharge formulas

hellmonk3:
Primarily so that damage can be displayed in future. Don't rely on division by
power to recur, just caclulate the number of arcs to do upfront and do them at
full power. Additionally, random round all of the formulas to avoid messy power
breakpoints. New numbers give mostly similar damage at low power, I think.
At high power with a lot of targets things can get a bit wilder. If it's a
problem, can cap power at 50 instead of 100.

I didn't actually make a commit to display damage yet, will do that later probably.

### DIFF
--- a/crawl-ref/source/spl-damage.cc
+++ b/crawl-ref/source/spl-damage.cc
@@ -991,6 +991,8 @@ int airstrike_space_around(coord_def target, bool count_unseen)
     return empty_space;
 }
 
+static const int AIRSTRIKE_POWER_DIV = 7;
+
 spret cast_airstrike(int pow, coord_def target, bool fail)
 {
     if (cell_is_solid(target))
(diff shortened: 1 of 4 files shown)


crawl-ref/source/spl-damage.cc
```diff
@@ -2580,6 +2593,11 @@ spret cast_discharge(int pow, const actor &agent, bool fail, bool prompt)
     return spret::success;
 }
 
+int discharge_max_damage(int pow)
+{
+    return 6 + (pow + 7) / 8;
```
ebering, line 2598: this doesn't use your fancy new constants

ebering (commented):
actual damage numbers look good to me, the display function doesn't look right for discharge damage per arc though.

PR #2424, opened 2022-02-15
Fix warning with OTR up + poison immune summons

### LINKED ISSUE #2419: Pointless warning for OTR + Haunt
Webtiles

.28

**DESCRIPTION:**
Casting Haunt while OTR is active generates a warning 'really summon while emitting a toxic aura?' 
However all Haunt's summons are poison immune, making this warning unneeded and potentially confusing.

CipHuK:
Fixes #2419

Previously we warned about summoning with OTR up, even for summons
with poison immune. Now another method is added to cover summons with
poison immune to remove pointless warning.

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -4245,12 +4245,12 @@ void bolt::handle_stop_attack_prompt(monster* mon)
     {
         string verb = make_stringf("charm %s", mon->name(DESC_THE).c_str());
         monster_info mi = monster_info(mon);
+        bool has_rpois = get_resist(mi.resists(), MR_RES_POISON) > 0;
 
-        bool stop = get_resist(mi.resists(), MR_RES_POISON) <= 0
-                  ? rude_stop_summoning_prompt(verb)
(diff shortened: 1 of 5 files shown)


crawl-ref/source/beam.cc
```diff
@@ -4244,8 +4244,12 @@ void bolt::handle_stop_attack_prompt(monster* mon)
     else if (flavour == BEAM_CHARM)
     {
         string verb = make_stringf("charm %s", mon->name(DESC_THE).c_str());
-        if (rude_stop_summoning_prompt(verb))
-        {
+
+        bool stop = mon->res_poison() <= 0
```
PleasingFungus, line 4248: When dealing with prompts for user-visible info, we should use monster info rather than the raw monster values.

PleasingFungus (commented):
Thanks for writing this! I'm a little concerned about fragility of these checks, but it does seem nicer for UI reasons.

PleasingFungus:
I did some fixups and merged this in. Thanks! :)

PR #2425, opened 2022-02-15
Add annotations for distracted monsters.

Aliscans:
Most of the reasons for a monster_info having MB_DISTRACTED give a description when the monster is selected on the level map, but not all.

This creates two flags to fill in the gap.
- Monsters which you can see, but which can't see you because of invisibility, are described as "can't see you" in the examine surroundings command and monster list summary.
- Monsters which are distracted (possibly because they are attacking a summoned monster rather than you) ar described as "not watching you" or as "distracted".


ebering:
Thanks!

PR #2428, opened 2022-02-18
corpseless

hellmonk3:
Still in early stages, but I'm opening this pr now for discussion and better progress tracking.

Motivation: corpses-as-items have tended to be problematic. Corpses were changed to be less item-like to prevent players from carrying them around to do tedious zombie setup stuff a long time ago, but the current version still has major problems wrt luring, especially after the duration of derived undead was shortened. Secondarily, corpses as items leads to a lot of item stack clutter. Crawl has other, better models for these types of spells, specifically by just applying an effect on monster death instead of interacting with corpses. I will note that pleasingfungus counterproposed just making corpses rot really fast, which does solve most of the luring problem, but is kind of silly. Anyway, here's the current plan:

-Remove corpses as items, but keep the visual effect of dead monster corpses/skeletons and put that in its own file (corpsespatter.cc?)
-Merge pain and animate skeleton into a rift wizard style "death bolt" (needs a different name under crawl naming conventions) that animates monsters it kills
-Rework corpse rot as a higher level necro/poison duration spell
-Turn animate dead into a death channel style effect
-Turn simulacrum into a single target irresistible status (somewhat like infest) that spawns a power-dependent number of simulacra when the monster dies
-Overhaul everything Kiku does except book gifts. Passively boost undead duration and give a chance to spawn extra when you make one, grant a new low level active (tbd - possibly cigotuvi-like scaling defenses) to replace corpse drop, remove the corpse requirement for invoking torment, allow ranged pain brand.
-Do something about beogh revival. No real plan for this yet, sorry. Could just give passive ally lifesaving or something.
-Whatever stuff I forgot about that still uses corpses


RojjaCebolla:
dchan, animate dead, and corpse rot are all things you cast *before* killing stuff now, right? and meanwhile, everything is attacking you and you aren't killing it.

Are you considering having some of these effects work after kill? Like maybe the game can know that X monsters have recently died while within your LoS, and so you can cast some of these sorts of things after-the-fact, and less like pre-combat buffs?

Monkooky:
Gonna suggest 'Necrosis' as a name for painimate skeleton.
Alternately, Painimate Skeleton.

crawl-ref/source/beam.cc
```diff
@@ -5461,6 +5463,29 @@ mon_resist_type bolt::apply_enchantment_to_monster(monster* mon)
...
+            if (you.see_cell(mon->pos()))
+            {
+                mprf("%s writhes in agony%s",
+                     mon->name(DESC_THE).c_str(),
+                     attack_strength_punctuation(dam).c_str());
+                obvious_effect = true;
+                if(mons_can_be_zombified(*mon) 
+                    && !mons_class_flag(mon->type, M_NO_SKELETON))
+                {
+                    mon->add_ench(mon_enchant(ENCH_NECROTIZE, 0, agent(), 1));
```
RojjaCebolla, line 5481: this is cool as heck

PR #2431, opened 2022-02-19
Move, remove, repair: dat/database and dat/descript

RojjaCebolla:
Small fixes based on output from util/db_lint

compiles and runs fine on my system, idk why the build partly failed


RojjaCebolla:
pretty sure there was nothing wrong with this PR? but it didn't have c2a82b186b8f5f3c9aebea940ff380b436c03abc so the build failed for that reason. 

merged that in and forcepushed

RojjaCebolla:
@ebering this would fix some missing descriptions, and would fix 1285b157 which removed one too many xom miscast text entries, because harmless miscasts are still extant

PR #2432, opened 2022-02-20
Onboarding: set hp_colour to higher defaults

RojjaCebolla:
This might subtly hint to newer players that things have
gone awry sooner.
Before: yellow at 50%, red at 25%.
Now: yellow at 70%, red at 40%.

With devteam approval, I'd also like to set other warnings a bit higher, like hp_warning, mp_warning, autofight_stop.


ebering:
This looks good. Adjusting some of the other warnings a bit higher is maybe overkill. autofight_stop in particular can be a UI screw when set too high for players who don't yet appreciate having it high.

PR #2433, opened 2022-02-20
Feat: demigods' holy and unholy words are more powerful

RojjaCebolla:
The demigod shtick is uninspiring for some players: they get higher
stats and otherwise the species is a bit dry. To give them more of a
twist based on their divine nature, increase the power of scrolls of
holy word and scrolls of torment.


ebering:
This one is a hard no. The fact that some players preconceptions about what the name "demigod" implies doesn't align with Dg's design niche has been long outstanding and is just accepted. As a design "no god, slow level, but big stats and MP" is concise and creates interesting gameplay all on its own, and there's no desire from the devteam dilute this from a place of flavour.

Even if a flavour argument for changing Dg were compelling, this particular iteration is rather strange in crawl lore: no god in the pantheon can be called upon for both holy word and torment (indeed, those that could offer one hate the other!), many gods offer neither (these not being their particular domains), and while a Dg's divine ancestry is suggested to be distant, the pantheon is well documented as full of jealous gods—unlikely to turn a blind eye to the poor company their offspring might be keeping.

PR #2435, opened 2022-02-20
Discourage pillar dancing and improve starting kits

PleasingFungus:
Especially early on, players can find themselves in a situation where they're right next to an enemy and will probably die if they fight that enemy. (A hobgoblin, a yak, what have you.) If the player has no consumables available, or isn't in much risk of dying from one hit, the best plan is often to run around in circles until the monster either (a) gets in one hit (and thus loses enough energy to create a 1-space gap) or (b) loses a turn from random energy, and then to flee up stairs, close a door, etc. Or just to run around until the player has regenerated to full hp/mp and restart the fight!

(Random energy was intended to discourage this, but doesn't succeed.) 

This is fiddly and unfun. It's not good gameplay. Removing random energy solves (b), but in many ways it just makes the situation worse - without the ability to create a gap, players are forced to continue "pillar-dancing" until they fully regenerate.

So, in addition to removing random energy, this PR adds a new 'parting attack' mechanic for monsters. Whenever the player moves away from a foe, that foe has a 1/5 chance of launching an attack. This means that running around while next to a threatening enemy is unlikely to be net positive over time. (If you have at least one space between you and your foe, you can still pillar dance to your heart's content, but that's no worse than it was before. And you should usually be able to get to stairs or something in those cases.)

This does make tactical repositioning slightly harder, but only slightly. If you're next to a gnoll and want to move across all of LOS to get into a corridor before the gnoll's pals enter melee, you'll be attacked about once en route (on average), which is still better than spending 2 or more turns in melee with multiple enemies.

To compensate, and to add more early-game options and decision-making in general, this PR adds a variety of new items to different backgrounds. The goal is to give players buttons they can use to deal with one or two rough fights, without overcompensating and trivializing the early game or neutering the ID game. The current list of changes:

- All 'mages' ('pure casters') start with a potion of magic.
- Fighters get an additional potion of might.
- Gladiators get a few throwing weapons, roughly half of what the AM
  throwing start got.
- Monks get two potions of ambrosia. (See, it's divine.)
- Hunters get a throwing net. (Not sure about this one.)
- Brigands get an additional poisoned & curare dart.
- Artificers trade their xom piece for nine charges of iceblast.
- Wanderers get an additional random potion or scroll.
- Delvers get nothing, for now, since I'm already pretty happy with
  how they play. They're a challenge anyway, really. :)
- Berserkers and Cinder Acolytes likewise get nothing. They already
  have perfectly good early game buttons.
- Abyssal Knights start at 60 piety (just over 2*) instead of 38 (just
  over 1*). This should allow them to use Bend Space if needed.
  (I think they're still quite weak.)
- Chaos Knights get Artificer's xom chesspiece.
- Transmuters get a potion of lignification, which should work well
  with their unarmed combat focus. It's also very thematic.
- Warpers get a few more boomerangs of dispersal.
- Arcane Markspersons get a few boomerangs of dispersal too. I'm sort
  of hoping to rework this background more broadly at some point.
- Enchanters get another potion of invisibility.

All of these are subject to change.

Other things tried in earlier versions of this PR:
- "Closing attack": monsters have a 1/6 chance of attacking when moving adjacent to you. Felt too intrusive into 'normal combat'.
- "Following attack": monsters have a chance of attacking when moving into a space that you just left. Encouraged zig-zagging while retreating.

Other things that have been discussed, but not implemented:
- Tighten time pressure? (Difficult with DCSS's map generation, which has a lot of big empty spaces.)
- Make all enemies fast or all players slow? (Pretty big change.)
- Make monsters shout more when following you? (Details unclear)


sdynet:
I understand your intention and see the mechanism as interesting. But I think it's better to give the monster a move speed bonus, not to damage the player. This is because the damage-causing method creates complex problems in special situations.


Ex)
1. Wu jian's users move frequently, especially when using the power of god. How are their actions handled? Who has the priority? If the player uses the heavenly storm and moves, does he have to be beaten continuously? Or will you give him a special passive when using this power?
2. Someone pushed and pulled you next to the monster several times. If you are unlucky, you'll bounce around like an American football and die.
This is too much, so are you going to make a decision before your turn comes? At this time, assume that you are wearing the fenser's glove. Do you judge parting attack only once, and riposte all? And how do you explain this to the player? It's kind of doubtful.


I guess the way you give the enemy a move speed bonus will have a similar effect to your intention and will not create a special situation. Also, rather than being attacked immediately, I presume that the reason "you're getting hit because you're slow" will cause less psychological rejection to the player. This is also the way for Sp. It's a pity that he was beaten by a snail just because he was next to him. Of course, if you insist that it's all Sp's fault... I have nothing to say(sad crying)

PleasingFungus:
> Wu jian's users move frequently, especially when using the power of god. How are their actions handled? Who has the priority? If the player uses the heavenly storm and moves, does he have to be beaten continuously? Or will you give him a special passive when using this power?

Wu Jian grants immunity to opportunity attacks while launching martial attacks (whirlwind/lunge). This is already implemented; I'll document this.

> Someone pushed and pulled you next to the monster several times. If you are unlucky, you'll bounce around like an American football and die.

That has no effect on this. The attacks only trigger when the player manually moves, not when they are moved by some other force.

PleasingFungus:
Thanks for the feedback! :)

ebering:
Commentary on the starting kits. My general feel is that the aop/energy change is lateral in where it places the difficulty and doesn't need heavy comp. So my commentary on starting kits is entirely about how much they're shifting the difficulty curve.

The following felt too generous and can be cut back:

- Monk does not need anything I think. There is (after the early altar work I did) a <1.25% chance of a given game not having a faded or identified altar on D:1-3, so Monks can use their starting bonus quickly if they really need something. It's not a tactical desperation play, maybe one ambrosia is sufficient, but getting ** plus faded altar piety out of the gate is massive.
- The second potion for Figher and Enchanter is too much, one is plenty. (MuEn/MuFi do fine even with the aop change and no potions!)
- The two javs for Gl were fun and flavorful as a fight opener but also very strong. One would be more than enough. Gl's nets are really really good, good enough that Gl is probably ok with maybe one more net.
- Wr and AM's extra dispersal boomers. Wr's got a nice chunk of consumables already. Not sure what to give AM, but a launcher and L1 slow is very generous already. Happy to chat more about a bigger rework of the background as well as brainstorm small trinkets to give an AM.
- CK getting anything, even the chesspiece (which I'd like to take another iteration on this version). Not because CK would be to strong with a bump, but for challenge zealot reasons.

Things that felt good as is:
- A magic potion for magi.
- The Tm !lig is both flavorful and can nicely smooth a pain point (or more with the free identification of it).
- Br ammo bump.
- AK piety bump. Still on the weaker side but having enough piety to get a bend space in feels in similar territory to the Tm !lig.
- Be/CA getting nothing.
- The Wn bonus consumable.

Things I'm not sure about but also lack good answers for:
- See AM
- Hunter's net; feels like it steps on Gl's most distinctive consumable.
- CK/Ar and the chesspiece

Could defer on both of these last few, giving them nothing for now and thinking about it over the cycle.

PleasingFungus:
Thanks for the feedback! I've implemented most of your changes; leaving fi/en/ck/hu unchanged for now.

Definitely agree that it'd be better to have a more distinctive item for hunters - could maybe give them and/or AM a scroll of immolation? Read-IDing ?immo is so funny, though. Undecided.

acrobat3:
I think parting attack will work well in many situations.
Because most of dangerous monsters overwhelm the player's regen in many situations.
But if you have enough regen in early game(vine stalker, troll, amulet of regen, TLA, artifact...)
Parting attack will make even more boring pillar dance than now.

For example, you are playing HuMo, in D:4, exp level 8, maxHP 70, ac 7, ev 12, and have amulet of regen.
You are right next to the monster and you can not easily beat the monster with your current HP.
Then you must calculate the monster's AvEffDam.(yourself or using wiz mode fight simulation)
You can recover a hp of 1.01 (=0.8+0.21) per turn, and a parting attack is triggered with a 20% probability,
so If AvEffDam is lower than 5.05, it is worth trying a pillar dance to save consumable items.

Edmund with +2 flail of holy wrath(AvEffDam=2.9): can pillar dance
yak(AvEffDam=3.4): can pillar dance
wight with +4 great sword(AvEffDam=5.7): can't pillar dance
wight with +3 scimitar(AvEffDam=4.3): not perfectly safe but still can pillar dance
(I calcurate these using wiz mode fight simulation)

It is extremely boring to calculate AvEffDam of monster and compare with player's regen many times.
and also there is no random energy, so you have to pillar dance very very long time(you can't create a gap).
Whenever the player moves away from a foe, give player -regen status for one turn maybe solve the problem...?

And I think probability of parting attack(20%) is too low, so it is too much luck-dependent.
What about whenever the player moves away from a foe, give player irresistible small damge instead of actual monster attack?

acrobat3:
Here's my feedback on good-start test branch play!
I played 6 games. 
Mostly melee background, because I thought melee characters are more affected by parting attack.
And played until it became a character that could easily win or until it died.

1. No pillar dancing was really fun to play.
Restarting the battle at no cost except for boring key input was really bad.
But player still can pillar dance if it has enough regen and defensive stat.(even more boring pillar dance than before)

2. It felt that one attack per three steps limited tactical positioning too much.
To discourage pillar dancing more effectively, it may be better to do something related to player's regen.

3. I think Improving starting kits compensate enough on D:1~D:2.
But on D:3~L:6, the difficulty was quite higher than before.
Especially when chose weak combo or useful items are not generated enough.
If you want to maintain the overall difficulty without making D:1~D:2 trivial, it seems good to consider other ways like nerfing ood(with current starting kit buff).

4. I found something weird.
When I pillar dance ogre or two-headed ogre, they never did parting attack after first parting attack.
(My human fighter pillar danced for hundreds of tiles but only got one parting attack from ogre or two-headed ogre.)
I think this may be related to slow attack speed of them.
Is it a bug or intended?
Not that important though, because normal speed race still can easily get a gap from orge and two-headed ogre.

PleasingFungus:
Thanks for the in-depth feedback! :)

1. Yep, known. My hope is that this won't come up too often, since once you have a bunch of regen and defense, it's less likely that there's a significant incentive to pillar dance. I'll probably revisit this.
2. Noted!
3. Reasonable. Hellmonk suggested tweaking D spawns to be a little gentler. I'll take a look and see what I think.
4. Ogres are sort of weird right now. I think it's somewhere between a bug and a badly designed behaviour. I'm going to take another look and try to make ogres always get parting attacks if they're able to keep moving next to you (but not if you'd be able to get a gap from them when walking away - if that makes sense).

quhsi:
Some random person feels that this makes playing formicids much more difficult. At least that you have to be much more selective with which monsters or monster groups to engage.

PR #2436, opened 2022-02-21
feat: submodularize qw; streamline stress/run qw

perryprog:
This commit soft-depends on:
- crawl/qw#2 for Depths being four levels deep
- crawl/qw#3 for armour switching behavior with the slot rename
- crawl/qw#4 for automatic starting

I say "soft-depends" because it can be merged before or after this without any breakage besides what's already happening.


rawlins:
I think this is unfortunately now stale with respect to major changes in the qw repository, so I'm going to go ahead and close it. (Which doesn't preclude the idea getting resurrected, but it would have to have a bunch of extra steps.)

PR #2437, opened 2022-02-22
Add a demonspawn cloak slot mut: demonic wings

RojjaCebolla:
The demonic wings mutation first grants Rampage, then permanent flight,
blocking the cloak slot. At level 3, the player begins to repel missiles
with their wings, which can hopefully help with some of the situations
that they find themselves in due to Rampage.


RojjaCebolla:
Notes: other ds body muts only block a slot at level 3, is that important to stick to? Should i switch level 2 and 3 of this mut?

Permanent rampage may be seen as annoying by some; having done a 15-rune game with +rampage body armor, i feel pretty confident that it's easy to get a feel for and work around. But, we could maybe have a toggle for the ability. Or, re-use rampage's "am i moving closer to enemy?" check, but instead of doubling movement, just cut the cost of such movement in half.

PleasingFungus:
Don't think we'd want to add a toggle - those are quite fiddly.

Implojin:
Demonic rampage sounds very flavorful and although I'm obviously biased, I like this concept~

RojjaCebolla:
Before setting it to rmsl, I was thinking the t3 of the mut could provide +SH, but then that seemed a bit dilutiony of what's special about "large bone plates". but some kind of defensive benefit from the wings, to offset Rampage risk.

semitonal:
Giving all three of Rampage, flight and rMsl seems like it'd be doing too much for a body slot mutation. rMsl also overlaps with one of the scales mutations, so shouldn't be re-used on a body slot. Something more suitable would probably be to only give Rampage and block the cloak slot at level 3. I'm not sure what the level 1 and 2 effects would be in that case, but they should be fairly minimal (compare to level 1 or 2 of an aux attack mutation, which are pretty low-impact).

PleasingFungus:
To be clear, I believe kate is not saying that this proposed mut is *too powerful*, but that it has *too many ideas* - too many distinct mechanical elements, some of which trample on other existing DS muts.

Think about it as if you have a 'complexity budget' for this idea, where more distinctive and novel elements take more points. Perhaps you have a budget of 10 points, where rampage costs 6, flying costs 3, and rMsl costs 5. That's way over budget! Players will already be excited and interested by getting a rampage mut - no need to over-egg the cake.

RojjaCebolla:
Ah I see. Hmm. What about a a "wing buffet" aux attack that doesn't do much damage? It'd be mostly just a hook you can hang some +Slay on.

or no damage at all, but it knocks back enemies?

Implojin:
What would you all think about the idea of a rampage ds mut that included rampage, dazzle, stab upgrade as a 3 tier thing?

I think the game doesn't do a great job of communicating in-game that rampage can be used on stab characters at present; it's just kind of left for players to infer that it might be possible -- even then, seven-league break the rule and wake things up. Explicitly making this particular combo available as a possibility for Ds might go a long way towards communicating this to players?

This might(?) also be interesting to some megazig characters: Dazzle, rampage, and spriggans knife can be an interesting combo, but wearing them ordinarily takes up 3 gear slots.

RojjaCebolla:
i for one find spriggan knife's power level to be extreme and i wouldn't normally want to hand that out anywhere else. but maybe on one o' them top-drawer muts!  i don't feel qualified yet to think about t3's, i wanna muck around with body slots and t2's first. (Although, those weaker ones are seen in far more games, aren't they?)

PleasingFungus:
Yeah I'm going to get around to nerfing sprigknife at some point.

RojjaCebolla:
rip to an incredibly good one. can't believe i doggedly wielded gyre and gymbal for so long with spriggan's knife as a backup 😆

RojjaCebolla:
idea slush pile:

- a weak aux attack
- some increased defensive stat while flying (like tengu)
- +1 EV per adjacent enemy
- increased speed going up stairs (cuz flying!)
- increased speed when above or adjacent to lava. (thermals? hellishness?)
- 1 aut discount on rampages
- whenever you move/rampage (_or for simplicity: whenever you move/sit, like acrobat_) ...
  - 2 ac, 2 ev, 2 sh
  - deal 2 damage to adj creatures at end of move
  - +2 slaying (applies to next action)
  - +will

ebering:
This conversation kind of petered out. Your final result is still pretty high on complexity budget for a body slot demonspawn facet (a mix of defense and offense, several different mutations). I would suggest something simpler, riffing on the seven league boots: Rampage but more range! Demon wings would be a 3 step mutation:

- Rampage
- Rampage+ (distance 2)
- Rampage++ (distance 3), blocks cloak

Undecided if it would give flight at level 1 or level 3, leaning towards 3. I know that @RojjaCebolla is really short on time these days, partly posting this to see what @semitonal might think about the idea before I dive into the implementation.

semitonal:
I feel like introducing multiple levels of rampaging (outside of just range 1 and the unrand's special unlimited range) would make it a bit confusing that normal rampaging items don't stack either with each other or with the mutation. Higher-range rampage also starts getting close to the issues with stabbing that the seven-league boots need a special case for (and I don't think enabling stacking would be good, for that same reason). I do think some kind of wing-flavoured cloak-blocking body slot would be nice and thematic to have, but I'm not sure we've managed to land on a suitable mechanic for it yet.

sharkjacobs:
- level 1: Rampage
- level 2: Lunge Rampage
- level 3: Flying

**Lunge Rampage:** a rampage that does a nominal amount of bonus damage like the Wu Jian Council Lunge.

This doesn't step on the toes of any existing Demonspawn mutations. It holds to the loose unspoken rule that body muts reduce defence rather than boosting it. The new ability has an existing design precedent. And I think it fits in under the complexity budget

Hellmonk:
Closing for now, as this pull request has become quite old. There are tentative plans for new body slot mutations in 0.34 which might use some of these concepts, either as part of a ds facet or part of a new species design.

PR #2438, opened 2022-02-22
fix: Restore launcher to hit info in monster desc

pdpol:
Before the recent launcher changes the targeter used to display the
chance to hit a monster but relied on a projectile. This attempts to
recreate that by leaning on the fake projectile logic used elsewhere for
launcher behavior in get_monster_desc.

I don't really know C++ so sorry if this is extremely gross lol. Happy to rework it in any way needed.

### DIFF
--- a/crawl-ref/source/throw.cc
+++ b/crawl-ref/source/throw.cc
@@ -133,8 +133,18 @@ void fire_target_behaviour::update_top_prompt(string* p_top_prompt)
 item_def* fire_target_behaviour::active_item()
 {
     const int slot = action.get_item();
-    if (slot == -1)
-        return nullptr;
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Thanks for finding the bug! This solution doesn't quite work, sorry. The problem is here:
```
            item_def fake_proj;
            item_def *thrown = &fake_proj;
            populate_fake_projectile(*launcher, *thrown);
            return thrown;
```

Here, you're allocating a new `item_def` on the stack within the current function, and then returning a pointer to that item. But when this function returns, its memory in the stack can be re-used by other functions. That pointer could end up pointing to anything!

I'll poke around with this and see if i can figure out what the right fix is.

PR #2442, opened 2022-02-24
Mummy reform: Decay ability & necro enhancers

PleasingFungus:
Mummies have long been notoriously awful, and recent versions have
done nothing to help them. That's fine to some extent - it's funny
to have a challenge species - but there's a limit, and it's with
planned changes to add many potions to different starts, mummies are
more left behind than ever.

So, to give em more distinctiveness and early game tactical options,
add a 'decay' ability that lets them plop miasma on their current
tile. This plays a bit like super conjure flame (a wall early, a
damage enhancer later), and is very funny in certain cases (vs
frenzied enemies, Robin, etc).

To limit the effect, Decay costs HP drain, currently about 75% of an
Apocalypse - it's not something you can use every fight. If it's still
too strong, we can raise the costs and/or adjust it to do stat drain
instead, which would line up with the effect of malmut on mummies.

Also move necro enhancers from XL 13 and 26 to 3 and 13, to make them
show up in a reasonably relevant timeframe.


ebering:
> That's fine to some extent - it's funny
to have a challenge species - but there's a limit, and it's with
planned changes to add many potions to different starts, mummies are
more left behind than ever.

Agree here, but I wonder if we can approach this without giving out Yet Another Species Active. (This is my aesthetic preference slipping in.) Lici, one of our top players, had suggested making Mu a bit of a foil to Dg, with the following two changes: give Mu innate Faith, and reduce the XP apt to -2. I had been considering them for 0.29 but have been slow to getting to the flow of 0.29 because the world.

What do you think about the faith switch (along with the change in schedule for necro boosts)? It's not an active so immediately tied to escape, instead it makes a resource available earlier. It would also help nudge Mu gameplay a bit away from Gozag (who cares not for the innate devotion that the mummification process brought about).

PleasingFungus:
> Agree here, but I wonder if we can approach this without giving out Yet Another Species Active. 

I see where you're coming from, but think about it this way. Since mummies can't use potions, they have considerably fewer tactical options cluttering up their brainspace (battlespace? Mind Palace?!). This frees up space for other actives.

Playing around with a faith gimmick for mummies seems sort of fine, but I don't think it has any significant effect until... maybe d:5 to d:9, depending on how the game goes? It doesn't seem like it fills the same niche that this active does (of giving players tactical options throughout the game).

ebering:
Some further feedback on this, playing Mu with `a-good-start` and aops: there was one place where I might have considered this ability over pillar-dancing that one hobgoblin to regen mp before trying to put it to sleep again on a MuEn, but I don't think I would have, even at 1/3rd chance, because the hobgoblin wasn't sufficiently threatening to be worth the drain cost.

I agree that mummies have some unused facets of their tactical hypercube but I don't think that is best filled with extra actives (see Demigods, Felids, Octopodes). Maybe there's a non-faith passive that expands tactical options throughout the game and fits with Mummy's theme (Harm?????? /j), and doesn't require a god to come online.

But I'm also not convinced the gap on D:1-3 is too big to prevent mummies from being fun, even with expanded kits. In addition to it being passive, I also like faith because it is dungeon-reactive and shifts the incentive created by -Potion away from Gozag and into the broader pantheon; putting these reasons out there as brainstorming seeds for the previous speculation.

PleasingFungus:
Fair enough. Will pull in necro enhancer changes separately and close this for now.

d-strickland:
I'm just a random player browsing PRs, but I really like this idea! What about changing it from an active to an on-hit effect. Could be %HP gated like the sanguine armor DS muatation, or just invoked when you get hit hard.

PleasingFungus:
It's definitely on my list of ideas to revisit :) Will see if it can land somewhere in some way at some point.

PR #2443, opened 2022-02-24
give monk a starting item that's usable in a fight

RojjaCebolla:


PleasingFungus:
What fight is a monk getting into that ?holy word will help with?!

PR #2445, opened 2022-02-24
At game start, downgrade unusable scrolls to identification scrolls

RojjaCebolla:
I think this only affects Warpers and Wanderers at present, but may become more broadly applicable if player start inventories are changed in the future.

### DIFF
--- a/crawl-ref/source/ng-setup.cc
+++ b/crawl-ref/source/ng-setup.cc
@@ -148,6 +148,15 @@ item_def* newgame_make_item(object_class_type base,
             item.sub_type = ARM_ROBE;
     }
 
+    // If the character can't use a given scroll, hand out a replacement instead.
+    if (item.base_type == OBJ_SCROLLS) {
+        if (item.sub_type == SCR_HOLY_WORD && you.undead_or_demonic())
+            item.sub_type = SCR_IDENTIFY;
+        else if (item.sub_type == SCR_BLINKING || item.sub_type == SCR_TELEPORTATION)
+            if (you.stasis())
+                item.sub_type = SCR_IDENTIFY;
+    }
+
     // Make sure we didn't get a stack of shields or such nonsense.
     ASSERT(item.quantity == 1 || is_stackable_item(item));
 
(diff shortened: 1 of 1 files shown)


ebering:
Why tho?

Not a fan of substitute items in backgrounds beyond size changes to meet species constraints.

RojjaCebolla:
If the new stuff in crawl:a-good-start comes down the pipe, the consumables that a background comes with become an even bigger part of that background. I figured species which can't use those consumables should get *something* rather than just silently be weaker. Is there a different replacement that'd sit better with you?

ebering:
You can see my commentary in `a-good-start` about how generous I think PF's suggested additions to the kits are.

Presently certain species miss out on big parts of existing backgrounds:
- Felids on a variety of things
- MuFi on the might potion
- FoWr on the blink scroll (and the Blink spell along with FoWz)
- etc.

To the extent that a species should be compensated for its limitations that is already baked into the species. In my testing `a-good-start` I found that baked-in comp sufficient. Are there specific combos you think are lacking?

RojjaCebolla:
this feels like maybe just an unresolvable difference of opinion - 

i believe that the mummy/formicid is intended *in general* to get by without certain consumables, but in *specific* backgrounds that hand out a strong tool to help survive an unlucky or rough patch in the early game, i feel an alternate item for the player who missed out is warranted. scroll of id can maybe help them quickly find something usable, but some other low-power scroll like fog would also make sense. 

and i'm gettin the feeling that your belief is instead that some jobs+species combos are fine to be missing out on something big. that's the spice of life or whatnot.

if that's the general mood of the devteam, i needn't argue further! :) i'm PRing these patches as quick "rough draft" ideas for revision/feedback. not super married to any of them at this time, i'll just finish the patch real quick with a case for potions, and if y'all choose to merge it or not that's groovy.

ebering:
> your belief is instead that some jobs+species combos are fine to be missing out on something big. that's the spice of life

Certainly my belief, with the addendum that if it's too spicy the seasoning to adjust is the species. 

At one point in the starting-consumable-kit there was some discussion of handing something (an id scroll or a cure pot) out across backgrounds, but per background flavored tweaks were much preferred. Your change would see almost every mummy start with an id scroll!

I'm still unconvinced that any fallback consumable is needed (MuFi and FoWr have been fine all this time); but if there are going to be fallbacks it would be much better for them to match the consumable they're replacing in some way.

ebering:
@semitonal had some similar remarks in IRC. We haven't done this historically and aren't looking to start.

PR #2449, opened 2022-02-26
Channeling: improve the worst cases, reduce the best cases

RojjaCebolla:
`"You aren't skilled enough to handle this power!"`
I dunno how direct one should be with hinty messages. I don't want someone to be misled into thinking that raising Intelligence or Spellcasting will help them wrangle an orb of energy, but it'd feel a bit lazy and unimmersive to outright say "You require more ~~pylons~~ Evocations skill" when the orb activates.

If a player equips the orb of energy with low evocations skill, they'll get constant backfires. Printing a message could help hint that it's not just bad luck, it's a bad idea.

When backfires do happen, due to low skill or low luck, refund the mana anyways - it's being gained at a rather steep and unsustainable price, after all.

Lastly, to reduce the chance of runaway situations with free mana, only refund slightly over half of the spell's mana cost.

### DIFF
--- a/crawl-ref/source/spl-cast.cc
+++ b/crawl-ref/source/spl-cast.cc
@@ -685,27 +685,39 @@ static void _handle_channeling(int cost)
     const int sources = 3 * player_equip_unrand(UNRAND_WUCAD_MU)
                         + 2 * you.wearing_ego(EQ_ALL_ARMOUR, SPARM_ENERGY);
 
+    // Channeling items test evo skill in order to activate.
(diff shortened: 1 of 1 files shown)


RojjaCebolla:
image
current failure rates for the orb of energy

RojjaCebolla:
if it's not desirable for energy to be available on common gloves, maybe it could be an unrand?

ebering:
I didn't take the gloves (for now I think we want orb egos to be orb exclusive), or the always refund on activation, but the message on guaranteed backfire is a good idea so I've added that. Thanks for the submission!

PR #2451, opened 2022-02-27
when felid has 2 lives, fix Experience Info to show actual died-now xp needed

elliottbernstein:
Currently, when a Felid has two lives, the related Experience Info message is off by one. E.g., at level 10 with 2 lives, it reads, "If you died right now, you'd get an extra life in 1 levels' worth of XP." If you died at that moment, however, you'd need to gain two levels to get another life.

This fixes the Experience Info to display the correct number of levels needed to gain another life following death when you currently have two lives.


ebering:
Thanks for the patch!

PR #2452, opened 2022-02-27
Hungry Hungry Ghouls

Monkooky:
This PR seeks to solve two issues: 
First, ghouls are extremely one dimensional, and their current playstyle is better implemented in trolls. 
Second, ghouls' current gimmick is not thematically coherent.

Ghouls are very one dimensional due to extremely low int, and aptitudes that push them strongly towards UC. 
While their current gimmick disables in-combat regen and gives them heal on kill instead, this serves a similar purpose to a troll's boosted regen, and in practice feels very similar.

Further, the current gimmick does not convey hunger or sating hunger effectively. NoRegen does not impart any urgency to combat, and the ability to 'eat' dying enemies on the far side of LOS breaks theme.

This PR attempts to fix these issues by granting Ghouls significantly boosted intelligence, but rapidly draining int while monsters are in line of sight. Enemies dying in melee range will restore drained int. 
The heal on kill has been switched to melee range and made twice as effective; this was done for convenience not by design and will change. 

The current implementation is very preliminary.


crawl-ref/source/timed-effects.cc
```diff
@@ -200,6 +200,18 @@ static void _abyss_speed(int /*time_delta*/)
         --you.abyss_speed;
 }
 
+//Drain int until at 1 base int; disregarding gear.
+//Rate of drain is proportional to potential remaining drain
+static void _ghoul_hunger(int /*time_delta*/)
```
ebering, line 205: This should not be implemented as a timed-effect, but in `player_reacts`, since it occurs only when monsters are in LOS and is a "player reaction" like regeneration etc. It will behave strangely if implemented as a timed effect.

Additionally it should scale the chance based on `time_taken` so that the chance doesn't vary with fast/slow actions.

ebering (changes requested):
> First, ghouls are extremely one dimensional, and their current playstyle is better implemented in trolls.
> Ghouls are very one dimensional due to extremely low int, and aptitudes that push them strongly towards UC.
> While their current gimmick disables in-combat regen and gives them heal on kill instead, this serves a similar purpose to a troll's boosted regen, and in practice feels very similar.

I disagree with this somewhat, I don't think they're so one-dimensional as you claim, and definitely aren't as close to trolls. To the extent that they're one-dimensionally pushed to melee, restricting their heal mechanic to melee range will increase this problem, not mitigate it.

> Further, the current gimmick does not convey hunger or sating hunger effectively. NoRegen does not impart any urgency to combat, and the ability to 'eat' dying enemies on the far side of LOS breaks theme.

Flavour is secondary to good mechanics in crawl dev land, so we're not so worried about that.

**However**, your PR has a *super cool* idea for a mechanic that fits the *rotting species hungering for the living*: stat rot when monsters are in LOS, restored by kills. 
I'd lean into this idea and take it all the way. Consider the following proposal for a new ghoul: ghouls rot any stat (the exact random model tbd) when (living only?) monsters are in LOS, and restore stat rot on kills (of living monsters). It might even be reasonable to do this and remove the NoRegen/HP on kill side of things entirely.

I've left some technical feedback on the details of implementing this, if you decide to go in this direction.

I think we're getting a bit close to 0.29 for a big species overhaul, but I do want to see more of this stat-rotting-ghoul idea.

Monkooky:
The current state of affairs is thus:

Stat rot occurs while monsters are in LOS. 
There is a grace period before stat rot begins.
On kill, this grace period is reset and some of the stat rot is healed.
When no monsters are in LOS, stats are rapidly restored. 


Some discussion of the whys:
The fundamental direction shift here is largely under ebering's advice, and doubles down on the incentive for quick kills and LOS control.

I experimented with two forms of stat healing- a multiplier to incoming XP for the purposes of healing drain, and flat stat restore on kill. Most of this experimenting was under a paradigm where ghouls only healed stats on kill. 

The former of these has serious comprehensibility issues, makes the precise mechanics of stat drain much more relevant than they should be (gaining skill_cost_level can be highly detrimental), and somewhat encourages stashing easy to kill high xp enemies e.g. ogres.
The latter _heavily_ encourages stashing rats. 
Making stat damage impermanent for ghouls seemed like the only way to avert these and other deep dwarf-esque issues.

The grace period exists for feels, and to mitigate how much popcorn impacts drain. With the hunger period, killing 4 enemies in 4 turns delays rot just as much as killing 1 enemy in 4 turns. 


In practice, the Hunger mechanic is most impactful when retreating long distance- without consumables, it is not at all a viable option unless you stop to kill occasionally. However, it also imposes a sense of urgency on fights- waiting for an enemy to come to you is significantly worse than going ham. I think it very successfully evokes the hungry ghoul feel.

I think, however, it might be extremely bad. Like, new lowest winrate bad. 


I'm going to try playing around with hellmonk's meme suggestion of int -> slay over time, see if that is more interesting- but I am rather happy with how the pr stands.

Monkooky:
Interest is low, and the systems here still require significant revision to be interesting and satisfying. 
As this has languished for an extended time, I will be closing it.

PR #2453, opened 2022-02-27
New unrand: Mindarmour of the Mutant Mage or..

RojjaCebolla:
...Mad Mauler or Magnificent Mentalist or whatever you wanna call it.

`+3 ac, +3 str, wild magic, inacc`
`+3 more ac and +Will after being targeted by a hostile hex`

~~In short: the Helmet of Magneto! Fly about, resist Professor X's mental manipulation, you know the drill. Slight change from magneto's helm: it covers your entire face, thus the Inacc property. You're seeing your surroundings with telepathy or other magnificent mental powers, presumably. Gives a bit extra AC, naturally. Might be flavorful to add "spells are harder to cast, but more powerful" to this.~~


ebering:
[Check out the unrand review design page](https://github.com/crawl/crawl/wiki/Unrand-Review)

We've already got a lot of unrand headgear, both of the "unique mechanic" kind and the "flavor" kind. 

This helm seems to be straight flavor!

(Also, when making a PR it's helpful to put the specifics of what's changing in the PR text. At least the item slug `+3 magnetron {Will+ SeeInv Inacc}`)

ebering:
The new worked version does do more than flavour, but like your first draft of demonic wings it goes way over on complexity budget. It's trying to do too much at once:

- be an wild magic unrand 
- react to will checks unrand
- also inacc for some reason

Wild Magic is a really strong effect, stacking with spell enhancers above the enhancer cap. A good comparison for this effect is archmagi, which is restricted to robes typically for a reason. Putting Wild Magic on a +3 helm is gilding a lily, before we even get to the other effects. There's probably space for a Wild Magic unrand, but that would want to be the design focus.

Without Wild Magic, the mindarmour effect is something that's "nice" and "niche" but doesn't seem like much of an interesting trade-off: I get a +3 helm that situationally is +6 Will+, which lands in the unambiguously good but not that interesting category of unrand. 

We do currently lack an unrand helm, so please open a new PR if you've got some ideas! A Wild Magic focused design, with some serious tradeoffs (-1? *Corr? Contam? Will-? something more intense with contam?) could be a great item. (but don't try to force the magneto flavour too much...)

PR #2456, opened 2022-03-01
add new unrand: Shortbow that fires miniature orbs of destruction

uzkbwza:
With the ranged weapon rework, I think it's a good time to start exploring new ideas for ranged unrands, as there are currently very few. I propose this unrand shortbow, which fires a moderately strong projectile that behaves similarly to Orb of Destruction. The damage of this is static (doesn't scale with Dex) and you cannot use it with Portal Projectile. Training bows will affect its delay the same as it would any other shortbow. 

Obviously the damage value can be tweaked, as well as other properties. I don't really know what a good damage level would be, I just tested a few different values in game before it started to feel not OP (nor useless) but I feel like it could be a little more lethal without being unbalanced. You could also adjust the "spell power" to affect the orb's accuracy of trajectory. Maybe it shouldn't make as much noise, either. I added SInv because I think it fits with the weapon and makes up slightly for some of its drawbacks, but that isn't essential to the concept, and could be whatever.

The reason I chose a bow instead of a hand xbow or sling is to prevent the "cheesy" strategy of using a shield + amulet of reflection, which would nullify one of the main drawbacks to OOD (that you can be damaged by your own orbs). There also aren't any artifact shortbows. Generally, two-handed weapons present the tradeoff of less SH for more damage, and I think that works well here. Again, this doesn't really affect the main concept of this artefact, and it could just as easily be a sling or hand xbow. Maybe a two-handed sling?

Point is, with some tweaks (and probably bug fixes), I think this can be a good find for the mid-game but by no means a no-brainer.

Currently it is using placeholder tiles, because I don't have the artistic ability to make one myself. I also haven't written a unique description for the orblet 'monster', but I'm unsure if leaving it identical to the OOD description is a good idea. My apologies if there's anything else I forgot to include.

Any suggestions for a more interesting name are welcome.

https://user-images.githubusercontent.com/43023911/156085306-c07b4491-00f8-48c9-bbdb-7f83e3194c0b.mp4

.


uzkbwza:
> I (speaking not as a dev, but as a player contributor) don't feel it needs Sinv, because OODs are already a great way to deal with invisible enemies
that's a great point, lol. honestly, it probably doesn't need any other enchantments, anyway, especially if the damage is particularly high.
One other problem is that, when your skill is high enough to attack at under 1.0AUT, the orbs tend to collide into each other and fizzle out. I intentionally prevented them from loudly exploding when they collided with each other, but I'm unsure how to prevent them from colliding at all (at least, when they're going in the same direction). Maybe increasing their move speed? I'll experiment and see what works.

Monkooky:
I might instead fiddle with the attack delay so that mindelay is >= 1. (It's probably best to not be machine-gunning oods, even if it would be funny.)

uzkbwza:
> I might instead fiddle with the attack delay so that mindelay is >= 1. (It's probably best to not be machine-gunning oods, even if it would be funny.)

I think this could work, but then it runs the risk of just becoming a worse version of OOD. Consider the autumn katana - another unrand that uses an existing spell. The difference between using it and just casting Manifold Assault with a big sword is two things: 
1. There is no MP requirement or usage.
2. You don't control when it activates - it's purely based on chance. 
In the same way, I would like to differentiate this artefact with OOD by making it behave somewhat differently, and not be just a "free but worse" version of the same spell. Though you could argue that's more or less what the autumn katana does.

Hellmonk:
I think the core conceit here (ranged weapon that fires a slow projectile) is actually quite good as an unrand concept. Unfortunately, we've allowed this pr to rot for 3 years, so it'd probably be easier to reimplement from scratch than update this.

PR #2458, opened 2022-03-02
arrival vaults

RojjaCebolla:
Teleporters generally only show up in dangerous scenarios; this puts
a pair of teleporters into an arrival vault for people to play with.
These can also be used to juke gnolls and panlords, in times of need.

(Inspired by the "throwing tutorial" key holder vault in Brogue, which
shows up early and often in dungeon generation, helpiing the player
learn that items can be thrown to trigger traps from a distance.)

### DIFF
--- a/crawl-ref/source/dat/des/arrival/simple.des
+++ b/crawl-ref/source/dat/des/arrival/simple.des
@@ -2508,3 +2508,38 @@ tt.........tt
  tt.......tt
   ttt@@@ttt
 ENDMAP
+
+##############################################################
+NAME:    spicycebolla_xxxplode
+TAGS:    arrival transparent
+ORIENT:  float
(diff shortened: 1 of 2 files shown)


RojjaCebolla:
a few permutations
image
image
image

ebering:
Since arrivals are decorative, unless there's a comment saying "wish we could do X" and X was subsequently added (like clear doors), we prefer to leave existing ones as is. The other two look neat, I'll merge them as is.

RojjaCebolla:
Fair! I thought it might restore some of the former glory of what was once a pretty chaotic vault: 7482a9b90ce0 But, fair that there's no need to mess with a fine vault. We can just make new ones! I wanted to see what y'all think about "no stress transporters" before cooking up a few more.

PR #2462, opened 2022-03-03
Shift gyre and gimbal to be early-game

RojjaCebolla:
The protection brand is particularly relevant when enemies may only be
swinging for 9 to 20 damage. In the same way, damage-dealing can be
less of a struggle for short blades when AC is lower.

In the late-game, most short blades that maintain relevance will be sporting different brands (or are rapiers). If one finds this weapon in Vaults, they're probably already wielding something else and unlikely to switch to Gyre.

Having this weapon crop up earlier will give players a chance to use it
during a time where it's not outclassed by as many other weapons.
Making the base item a short sword may help nudge people to look for
replacements. Also, there aren't any artifact short swords currently.

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -1400,9 +1400,9 @@ DBRAND:  It occasionally heals its wielder for a portion of the damage dealt
 DESCRIP: Its wielder's spells cost health as well as magical power.
 
 ENUM:    GYRE
-NAME:    pair of quick blades "Gyre" and "Gimble"
-OBJ:     OBJ_WEAPONS/WPN_QUICK_BLADE
-TYPE:    pair of quick blades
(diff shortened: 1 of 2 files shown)


acrobat3:
> If one finds this weapon in Vaults, they're probably already wielding something else and unlikely to switch to Gyre.

Disagree.
I think that Gyre&Gimble is generally better than the +9 rapier(quick blade) with damaging brand, except when you have orb of guile or invest a lot in shield skills.
It's pretty good to have two stabbing chance per 3 aut.
And also Gyre&Gimble does really high damage when you have enough slaying.

ebering:
Definitely don't want to make them short blades: double quick blades are Cool™ (also you need the balance for them to be swordchucks, which they are canonically).

I also think you're underestimating G&G as a weapons. They may not be "always switch character build" good, but they do manage to tempt some builds! They're in a good place.

PR #2465, opened 2022-03-04
tweak cast message for zenata's blade dance

RojjaCebolla:


ebering:
A variation included in e5966c3f1

PR #2467, opened 2022-03-05
silent ossuary map

hellmonk3:
All but the outer edge of the map is silenced. Typical ossuary monster set.

### DIFF
--- a/crawl-ref/source/dat/des/portals/ossuary.des
+++ b/crawl-ref/source/dat/des/portals/ossuary.des
@@ -1244,9 +1244,9 @@ WEIGHT: 10
 ORIENT: encompass
 TAGS:   no_item_gen no_monster_gen
 MONS:   silent spectre
-MONS:   wight / wraith w:3
+MONS:   wight / wraith w:2 / mummy w:3
 MONS:   centaur zombie w:20 / water moccasin zombie / scorpion zombie /\
-        bullfrog zombie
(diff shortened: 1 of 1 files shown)


crawl-ref/source/dat/des/portals/ossuary.des
```diff
@@ -1238,3 +1238,53 @@ ccccccccccccc+++ccccccccccccc
...
              ccc
 ENDMAP
+
+NAME:   ossuary_hellmonk_silent_tomb
+WEIGHT: 10
+ORIENT: encompass
+TAGS:   no_item_gen no_monster_gen
+MONS:   silent spectre
+MONS:   wight / wraith w:3
+MONS:   centaur zombie w:20 / water moccasin zombie / scorpion zombie /\
```
ebering, line 1248: Maybe include an actual scorpion chance in the "tougher monster" list?

ebering (commented):
Plays fine, very flavorful and spooky. Tested with HuFi, DrCj, OpTm, and DgSu.

In playing through with each of these I mostly didn't interact with the silence: the doors provided good choke points outside of the silence for my spellmans. Stepping into the silence to fight with Wereblood without getting swarmed was neat.

I've made some suggestions for adjusting the terrain to make the silence a bit more a part of the map. I played a version with gates between the chambers as a KoFE and the silence was much more felt.

PR #2470, opened 2022-03-06
Convert Ghoul devour-on-kill to a mutation

alexjurkiewicz:
This behaviour was previously hard-coded.

I considered an alternate implementation of creating US_HUNGRY_UNDEAD.
This ended up creating a lot more code and making undead state checks
more complex, so I discarded it.


wheals:
It would be great if this would include an update to mutations.txt with a description of the ghoul heal-on-kill mechanic (this is the biggest win of converting it from a fakemut to a real mut, IMO). Otherwise looks good.

alexjurkiewicz:
Done. I just noticed Ghouls can restore hit points from non-adjacent kills. Is this intentional? It doesn't really work with the flavouring, but seems nice mechanically to open up more play-styles.

PR #2477, opened 2022-03-07
Simplify job starting stats

RojjaCebolla:
this PR sorta in the realm of justforfun / thoughtexperiment, but hey, maybe it'd make evaluating roles easier for new folks?

> Frequently, experienced players will remind newer players that jobs /
> backgrounds are meant to be a starting kit, not a lifelong class choice
> commitment. One proviso to that is the starting stats. Most roles'
> str/int/dex numbers are in the 2-6 range, but there are a few big outliers.
> 
> A TrBe begins with 3 intelligence, putting them easily within range for
> death by brainlessness. A new player who chooses this combo may not
> realize that they'd counterintuitively be wise to increase their int.
> And even when the intensely skewed stats of Be or Cj don't put one in
> danger of stat-zero, they still result in the background having an
> impact that will last well beyond a "starting kit".
> 
> This commit greatly reduces the stat-based differences between roles,
> which will make it easier for players to "pivot" away from the starting
> kit in a way that I believe is more in tune with dcss's style. The risk
> of stat-zero will also be greatly reduced. If some roles lose too much
> early-game power from this shift, 1-2 points of extra stats could help.

_does kinda waggle its eyebrows in the direction of just giving everyone 4/4/4 or 5/5/5 though..._

### DIFF
--- a/crawl-ref/source/job-data.h
+++ b/crawl-ref/source/job-data.h
@@ -6,6 +6,15 @@
 
 using std::vector;
 
+///< convenience macros
+#define BASHER  6, 3, 3 ///< str focused
+#define CASTER  3, 6, 3 ///< int focused
+#define DODGER  3, 3, 6 ///< dex focused
+#define RANGER  2, 5, 5 ///< str averse
+#define DANCER  5, 2, 5 ///< int averse
+#define WARMAGE 5, 5, 2 ///< dex averse
+#define JACKER  4, 4, 4 ///< jack of all trades, get it!
+
 enum weapon_choice
 {
     WCHOICE_NONE,   ///< No weapon choice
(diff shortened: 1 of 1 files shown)


gammafunk:
I don't think starting stats are actually a proviso to experienced players' advice to new players: it's frequently their advice that players feel free to tailor their stat choices based on what their character is actually doing. The starting stats are chosen mostly to make the various roles balanced at the beginning of the game so they can be effective at what they naturally do. They don't demand that e.g. Cj never use a weapon, and we wouldn't want Cj to have much focus on weapon usage early on. If such a character does find a great melee weapon and other relevant equipment early, the stat level ups are part of the way they perform this pivot. Generally speaking, the stats they level into will be sufficient to make whatever transition they need, and their are other tools such as god choice, wizardry, slaying, etc. that help characters to their new role. Pre-investing points in less relevant stats just because e.g. a Cj *might* switch to become more melee heavy will generally lead to early game balance issues.

You are also greatly overstating the effects of 0-stat. To take your extreme case of TrBe, exactly 1 death with int <= 0 has occurred for current (0.28+) versions. While it's fairly inconsequential whether you eventually put a stat-up into int or not, equipment-based mitigation is the usual way to deal with low int. Often this happens naturally by wearing some randart that happens to also grant int. There are extended game branches where stat-0 becomes more relevant to something like a TrBe, but this is relevant to experienced players and is a great case where such a character might start to customize int. This is particularly because of additional aspects like a potential switch from Trog, thus enabling use of spells, and the fact that such extended characters like to start using some spells in general.

PR #2479, opened 2022-03-08
Add new monster speech, banishment, and death for Lodul, Grum, Erica, Nikola, Maggie, & Margery.

Rosstin:


crawl-ref/source/dat/database/monspeak.txt
```diff
@@ -5042,6 +5083,28 @@ _Maggie_rare_
 @The_monster@ shouts @at_foe@, "Leave me alone!"
 
 @The_monster@ mutters to herself, "Master Lance won't take me back. This is the way forward."
+%%%%
+Maggie killed
+
+w:10
```
RojjaCebolla, line 5089: don't need to write w:10, that's already the default weight of a speech line

crawl-ref/source/dat/database/monspeak.txt
```diff
@@ -5042,6 +5083,28 @@ _Maggie_rare_
...
+%%%%
+Maggie banished
+
+w:10
+__NONE
+
+w:1
+@The_monster@ groans, "Eek! Not the Abyss! That place is such a drag!"
+
+@The_monster@ says wistfully, "How many dragons are there in the Abyss, I wonder?"
```
RojjaCebolla, line 5107: and because weight 10 is the default, this line is just as likely as __NONE. Was w:1 your intention for all the spoken lines?

crawl-ref/source/dat/database/monspeak.txt
Rosstin: Thanks, I wasn’t aware

crawl-ref/source/dat/database/monspeak.txt
Rosstin: My intention was to make the death/banishment lines a lower likelihood than "no message"

Rosstin: Yes, w:1 for all of them - somehow I thought that each line below w:1 would have that weight applied. I'll change it.

Rosstin:
Merged these contributions with #2481

PR #2481, opened 2022-03-09
Add speech, death, banishment quotes for many uniques.

Rosstin:


ebering:
Some of these (particularly the death and banishment lines) were over the top or not really in-line with things. Death messages should be pretty unique and special. I kept a few, and added all of the regular dialog. Thanks for the submission!

Rosstin:
Cool, thank you. Appreciate the info and feedback, stoked to see some stuff
made it in.

PR #2485, opened 2022-03-10
Add many new welcome messages to miscname.txt.

Rosstin:
Add many new welcome messages. The original messages still have a higher weight of appearing.

reasons

### DIFF
--- a/crawl-ref/source/dat/database/miscname.txt
+++ b/crawl-ref/source/dat/database/miscname.txt
@@ -66,7 +66,225 @@ Nightmarish fiends assault your waking thoughts!
 Seething terrors besiege your sanity!
 
 Lucidity slithers from your feeble grasp.
+
 %%%%
+# Random relationships
+_random_relationship_
+
+parent
+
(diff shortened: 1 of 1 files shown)


ebering:
Entry messages are the very first thing a player sees. We want them to evoke the theme without saying too much at all, and let the player's imagination fill things in (see my comment on your victory message PR for some what not to include guidelines). There are also some affirmative "to include" guidelines: the messages we have now have content in one of two overlapping categories:
 - indicating that the orb is in the dungeon
 - indicating that the player is seeking the orb

An additional constraint for these messages: they should be 80 characters or less (which means that all adjectives for `@_great_adj_@` must be at most as long as the tag), so that we don't force a `--more--` on game open. A few Halloween messages break these rules, but those are very special and get a pass for that reason.

Ultimately I'll merge `fantabulous` from the new adjectives and an edit of the vizier message. Feel free to send in a new PR with new messages. Sorry it took so long to get to these reviews.

PR #2487, opened 2022-03-11
Draft: Add win messages based on your character's species, skills, and religion.

Rosstin:


crawl-ref/source/dat/database/miscname.txt
```diff
@@ -314,4 +314,112 @@ You hear your own voice laughing maniacally!
 
 You hear your own voice screaming in agony!
 
+
```
Rosstin, line 317: Perhaps I will put these in their own file instead of filling up miscname.txt. "epilogues.txt" maybe.

crawl-ref/source/escape.cc
```diff
@@ -0,0 +1,208 @@
...
+#include "describe.h"
+#include "english.h"
+#include "enum.h"
+#include "invent.h"
+#include "item-status-flag-type.h"
+#include "libutil.h"
+#include "message.h"
+#include "stringutil.h"
+#include "tag-version.h"
+#include "unicode.h"
```
Rosstin, line 27: I need to go through the list of includes and figure out what I actually need to include here

ebering:
Just to record some conversation that happened in ##crawl-dev regarding this:

Entry and exit messages are aimed at being very vague, we want to leave the character's story up to the player's imagination. Which means: not telling them why they're seeking the orb, not telling them what happens once they've got it, and not saying what the orb can and cannot do.

The draft messages in this very much go beyond these guidelines. If you're still interested in exit messages, think much smaller, similar to the death messages given by gods. One 80 character line, with an oblique nod to the god's flavour and not much more.

Feel free to submit a new PR with much more toned down death messages, or come by #crawl-dev to spitball ideas.

PR #2489, opened 2022-03-14
Good lugonu

sdynet:
Bend space
The higher the power, the higher the probability of moving the surrounding objects together, and at the same time, the higher the damage. However, this effect also applies to neutral objects, so players can become hostile to them if they use this power to escape from the existence of abyss that emerged through 'corrupt'. The higher the power, the higher the probability of Blink, and the higher the likelihood of them antagonizing the player. This bizarre interaction should have been eliminated sooner. Now, even if a neutral monsters is damaged through bend space, it is no longer hostile to the player.

Enter the Abyss
I want to distinguish Enter the Abyss from Sanctuary more clearly. If 2 MP is too low, 3 or 4 would be fine.

I want to change Banish, but I'm holding it for a while. I'm thinking about whether it's better to change this decision to HD-based or to leave the decision as it is, but change it to a smite method.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -461,7 +461,7 @@ static vector<ability_def> &_get_ability_list()
             7, scaling_cost::fixed(5), 10, -1, {fail_basis::invo, 70, 4, 25},
             abflag::none },
         { ABIL_LUGONU_ABYSS_ENTER, "Enter the Abyss",
-            10, 0, 28, -1, {fail_basis::invo, 80, 4, 25}, abflag::pain },
+            2, 0, 28, -1, {fail_basis::invo, 80, 4, 25}, abflag::pain },
         { ABIL_LUGONU_BLESS_WEAPON, "Brand Weapon With Distortion",
             0, 0, 0, -1, {fail_basis::invo}, abflag::none },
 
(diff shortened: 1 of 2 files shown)


PleasingFungus:
Very sorry for the belated response here!

I've implemented the Bend Space change you suggested as 1ff7896281. I'm very confused by your change to Enter the Abyss - isn't it already extremely different from Sanctuary? Why would we need to differentiate further?

sdynet:
> Very sorry for the belated response here!
> 
> I've implemented the Bend Space change you suggested as [1ff7896](https://github.com/crawl/crawl/commit/1ff789628195c3bd8f74f5e19654594e4aa70355). I'm very confused by your change to Enter the Abyss - isn't it already extremely different from Sanctuary? Why would we need to differentiate further?

Your patch is cool.

Oops, the explanation was too short. I'm sorry. I felt that this power needed some improvement in the latest version. The overall risk of dungeons has increased, and Abyss does not give any experience. And as L's followers actively use Banish, the risk of this power increases. I wanted to compensate for this incomplete risk.

PleasingFungus:
I see, thank you!

ebering:
The no-xp change has been reverted and the Bend Space changes made, so I'm going to close this PR.

PR #2497, opened 2022-03-20
Korean translation(gods.txt)

sdynet:
Korean translation(gods.txt)


PleasingFungus:
Looks good, thanks! Merged as b34f9e133a. Don't forget to run util/checkwhite in future :)

PR #2498, opened 2022-03-21
Display Irradiate damage in monster descriptions

nlavsky:
There is an inconsistency with damage numbers. According to the monster utility, radroaches can do `3d25 / 3d26` damage with Irradiate. Monster descriptions show only one formula, though. Half of the time it will be `3d25`, and in other cases it will be `3d26`, even for the same radroach.

What would be the best way to address this?

For posterity:
> PleasingFungus: correct fix is to pass a random bool to the damage func and do false for the ingame monster damage desc


PleasingFungus:
Nice fix, thanks!

PR #2507, opened 2022-03-28
feat: guarantee that AK starting exit generates in LOS

catern:
It's basically a spoiler that an exit always generates near the AK
starting spawn.  If the player doesn't know that, they'll waste a fair
bit of piety using Depart the Abyss.  That could be fine, except that
now that means their starting escape mechanism, Bend Space, will be
offline for the first few levels of the Dungeon, defeating the purpose
of bumping AK piety to make it available.

Guaranteeing that the exit generates in LOS requires that the player's
LOS be known at the time we're generating the pseudo-abyss, so I had
to move the initial call to update_vision_range.  (Otherwise it
wouldn't work for Kobolds)

### DIFF
--- a/crawl-ref/source/abyss.cc
+++ b/crawl-ref/source/abyss.cc
@@ -85,7 +85,7 @@ static coord_def _place_feature_near(const coord_def &centre,
                                      int radius,
                                      dungeon_feature_type candidate,
                                      dungeon_feature_type replacement,
-                                     int tries, bool not_seen = false)
+                                     int tries)
 {
     coord_def cp = INVALID_COORD;
     for (int i = 0; i < tries; ++i)
@@ -96,9 +96,6 @@ static coord_def _place_feature_near(const coord_def &centre,
         cp = centre + offset;
 
         if (cp == centre || !in_bounds(cp))
-            continue;
-
-        if (not_seen && cell_see_cell_nocache(cp, centre))
             continue;
 
         if (env.grid(cp) == candidate)
@@ -1542,15 +1539,6 @@ static void _abyss_generate_new_area()
     place_transiting_monsters();
 }
 
-// Check if there is a path between the abyss centre and an exit location.
-static bool _abyss_has_path(const coord_def &to)
-{
-    ASSERT(env.grid(to) == DNGN_EXIT_ABYSS);
-
-    monster_pathfind pf;
-    return pf.init_pathfind(ABYSS_CENTRE, to);
-}
-
 // Generate the initial (proto) Abyss level. The proto Abyss is where
 // the player lands when they arrive in the Abyss from elsewhere.
 // _generate_area generates all other Abyss areas.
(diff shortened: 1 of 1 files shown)


semitonal:
Probably a simpler and better approach is just to always place it at the player's feet, instead of having it depend on your species (since with this patch, the Abyss could presumably still shift and prevent you from getting to the exit that it placed).

RojjaCebolla:
just bikeshedding here, you could place the exit within 4 tiles and then it's quite unlikely that any species for years to come will have too small of a vision radius to see it, but the player still has to take a few steps before leaving.

crawl-ref/source/abyss.cc
```diff
@@ -1580,19 +1577,12 @@ void generate_abyss()
...
     // altar near-by.
     if (player_in_starting_abyss())
     {
-        env.grid(ABYSS_CENTRE) = DNGN_ALTAR_LUGONU;
-        const coord_def eloc = _place_feature_near(ABYSS_CENTRE, LOS_RADIUS + 2,
-                                                   DNGN_FLOOR, DNGN_EXIT_ABYSS,
-                                                   50, true);
-        // Now make sure there is a path from the abyss centre to the exit.
-        // If for some reason an exit could not be placed, don't bother.
-        if (eloc == INVALID_COORD || !_abyss_has_path(eloc))
```
Implojin, line 1594: If you're taking this approach I think you should also delete the body of _abyss_has_path() above, which would now be unused?

crawl-ref/source/abyss.cc
```diff
@@ -1580,19 +1577,12 @@ void generate_abyss()
...
-            goto retry;
```
Implojin, line 1595: Same thing for the retry: goto label, which would now be unused

Implojin:
Both of my comments above are resolved as of 50f236bb82, but apparently github's UI won't let me mark them resolved without authorship of the PR or write access to the repo. :/

For what it's worth, I agree with the reasoning behind this PR. I was live spectating the game that prompted this, where the player expressed frustration at having no options after exiting the abyss. Despoilering the starting AK exit seems reasonable, I think.

https://crawl.kelbi.org/crawl/morgue/TheMeInTeam/morgue-TheMeInTeam-20220325-170049.txt
https://old.reddit.com/r/dcss/comments/tqh2ka/just_complaining_about_shafts_and_malevolent/i2h3fnv/

ebering:
Looks good, thanks for the patch! (And thanks to @Implojin for helping out with some more eyes)

ebering:
Oh, and welcome to contributing to crawl! How would you like to be listed in the CREDITS?

catern:
"Edgar A. Bering IV" ***@***.***> writes:
> Oh, and welcome to contributing to crawl! How would you like to be
> listed in the CREDITS?

"Spencer Baugh" as you've already done works fine, thanks!

PR #2508, opened 2022-03-29
Draft: Permanent Transmutations

OctopodesWeekly:
UPDATE: I most likely won't have time (or ability) to actually complete this PR! Bit off more than I could chew.
---
Credit to bauth for the concept and inital implementation. Changes have been made to adapt to mainline.

For a curiously long time, Transmutations have acted like Charms; a limited duration cast that had to be reapplied constantly if you wanted to sustain them, and not lose a turn of combat. While there has been opposition to making Charms permanent, I don't think this applies to Transmutations, which all have noticable downsides. I believe (and may be wrong) that the main problem with perma-tmuts are as follows:
- If we don't gate by [casting power], players can cast them at 99% failure.
- If we do gate by [casting power], then we introduce arbitrary spellpower/skill thresholds.

This PR aims to address that.

All body armour melding spells, along with Beastly App. and Lichform, now last permanently with enough spellpower.  This spellpower threshold was created manually: deriving the spellpower used to get a <20% failure rate in robe, then rounding that spellpower up to a clean number. Historically, advice was to start reliably using spells at 10% rate, so this threshold goes somewhat above that. 

These power thresholds all line up at 20% (most spells) or 25% power (level 7+).

In exchange, many of these spells recieve new power scaling. Details:
- Beastly appendage deforms player, giving +3 AC if players can deform. Power cap is nerfed more. Minor malus to casting. (Flavor issues?)
- Spider Form recieves a mini acrobat (power/7 EV when moving). Power Cap, Dex boost, and spellcasting are nerfed. Spider is one of two EV-based tmuts, and acro works well with the poison. Implementation is kinda wonky.
- Ice Form & Dragon Form recieve better AC scaling: slightly lower base AC, higher max power AC. Ice Form gets some spellpower UC. There could be more interesting changes, but Ice Form is probably fine as another 'blander' form, and Dragon Form has breath to scale.
- Statue Form & Storm Form recieve no changes. Statform imo scales comfortably using AC. Storm Form's entire point was to scale with spellpower, and any changes would be to general balance over anything.
- Lichform is always permanent, but dramatically drains the player based on *spell failure* rate. +Wiz increases the drain, while swapping body armour ends the form.
You can still swap +Int items, but its not like you couldn't already do this, and i thinks ok to keep a strategic-level spell considering the weirdness of Necromutaiton as a spell.
- Blade Hands is made much much shorter, like the Summon Hydra of Transmutations. Buffed to compensate.

Gameplay Notes:
- There still is a spellpower threshold, but its fine that transmuters should have a less complex skill game (rougher skilling, esp in the early game. inherient complexity in unarmed scaling). Plus, again, there *is* scaling.  Right now, going from 50 to 60 turns isn't much of a benefit, so there still exists an improvement.
- Expected behavior is that the game redraws durations every time you swap gear, meaning that removing a +Int item sustaining spellpower removes the permaform after a short & random duration.
- More/global penalties for permaforms?

Ideas:
- Blade Hands: Very short duration,  Gains Gyre & Gimble effect, -aux attack, overall damage buff (clashes with Storm Form?)

This is my first PR to crawl, and my first commit to anything that isn't very trivial, so expect weird things and mostly unusable code. Moreso want feedback on the general concept.


sdynet:
It's a tough but worthwhile challenge. Good luck!

ebering:
It was a good effort with the draft. I've been out of the loop for a bit so you might have come by the dev chat for assistance with your PR. If not, in the future do stop by!

> While there has been opposition to making Charms permanent, I don't think this applies to Transmutations,

For me, "permanent spells" are a non-starter in crawl. A spell is something that should expend MP for a tactical scale decision, lasting the space of a fight or two (at most). Some of the new necromancy bleeds into a slightly larger scope, but never beyond a floor.

We've got a gameplay mechanic for "permanently exchanging some tradeoffs for some power over many floors": it's items.

So really we'd want to see transmutation changes that cut in the opposite direction: more power scaling for forms, more motivation to switch between forms as a fight changes, and more forms to chose among. It's a delicate act but if you're interested in contributing to crawl and transmutations in particular, that's more the direction to be thinking.

PR #2509, opened 2022-03-29
add descriptors on stormform 'insubstantial'ness

ruxi:
+ add stormform 'insubstantial' descriptor to innate abilities - "immune to petrification and being set on fire". 
 + added sticky flame resistance

Rationale:
 - Stormform is insubstantial according to public-facing wiki. Some misconception between wiki data and game data on what the resistances / immunities are. This change reconciles that by explicitly stating what they are. 

Future directions:
 - Immunity to constriction and rotting are not part of this PR, but are 'wisp'-like immunities which may be considered in the future.


ruxi:
Refence to the wiki entry that is cause for confusion:

 - http://crawl.chaosforge.org/index.php?title=Storm_Form&oldid=59685

> [Insubstantial](http://crawl.chaosforge.org/Insubstantial) (grants immunity to [petrification](http://crawl.chaosforge.org/Petrification), [constriction](http://crawl.chaosforge.org/Constriction), and being [set on fire](http://crawl.chaosforge.org/Sticky_Flame))

semitonal:
Storm form does already get constriction immunity, so that could probably be documented in the description too, although in general, the insubstantial property isn't really properly implemented for players - for monsters it already does a lot of other things, such as granting immunity to barbs, liquefied ground, webs and nets, and preventing them from drinking potions. Probably just sticking to the more limited set of resists for players and documenting those is fine though, rather than making it a bunch more complicated for wisp and storm form.

It's worth noting that the unofficial Chaosforge wiki shouldn't really be used a source, as it's quite frequently out of date or incorrect (although there's definitely room to improve the form description in this case).

ruxi:
@semitonal: Saw your fix on constriction for immune forms (https://github.com/crawl/crawl/commit/d16b76e6b15c8753530d5545675ccee2a0688e68), the descriptor in this patch has been edited to reflect.

Appreciate the feedback, will avoid referencing the wiki moving forward - didn't know it was unofficial.

ebering:
Thanks for the patch! How would you like to appear in the `CREDITS.txt`?

ruxi:
> Thanks for the patch! How would you like to appear in the `CREDITS.txt`?

I would like to appear as Lewis R. Liu. Thanks!

PR #2510, opened 2022-03-29
FIX Remove an erronous "the" from a message.

Aliscans:
The player can see invisible monsters which are bribed, charmed or hexed. When the status is lost, the monster may suddenly vanish.

This changes the message this generates from "You can no longer detect the Sigmund." to "You can no longer detect Sigmund.". The "the" is retained for non-unique monsters.

monster::add_enchantment_effect() is unaffected by this. It inserts an adjective before the monster name, so the "the" is acceptable.


semitonal:
Thanks for the fix!

PR #2513, opened 2022-04-01
(Help me!)New spell: Phase Shift(L1 Tloc)

sdynet:
Goal: Create useful 1 level spell while maintaining the theme of the school.
Main design: player can swap positions with a random nearby foe. And them and the foe will suffer some damage.

- [ ] Not allowed to use magic when caught in a net(and Constr)
- [ ] Add self-harm damage. (4% MaxHP?)
- [ ] Add spell to books with appropriate themes.

I created this spell by referring to the scorch code. The compilation was completed without any errors. However, the exe file does not respond after execution and ends. I don't know what's wrong...


sharkjacobs:
1. Really recommend you pick a new name instead of reusing "Phase Shift"
2. L1 is probably low, this seems better than blink
3. Shouldn't work against immovable enemies like plants and statues

PleasingFungus:
trivia: this spell existed in linley crawl! 3.4 release notes mention:

```
+        - added Swap, which allows the player to swap places with an
+          adjacent monster.  Isn't really that abusable as it uses the
+          same swap code that's used for friendly monsters and that
+          doesn't currently have insta-kill potential... that can be
+          fixed later, but right now I'm calling it a "feature".  Should
+          it be fixed, this spell could be changes to give a higher resist
+          and in insta-kill situations the monster could be allowed to
+          "scatter" to any of the immediate surrounding squares as a
+          save against death (player could get this too).
```

(Sorry for the extremely, ludicrously long delay in responding to this PR. Someday...)

sdynet:
> 퀴즈: 이 주문은 linley crawl에 존재했습니다! 3.4 릴리스 노트 언급:
> 
> ```
> +        - added Swap, which allows the player to swap places with an
> +          adjacent monster.  Isn't really that abusable as it uses the
> +          same swap code that's used for friendly monsters and that
> +          doesn't currently have insta-kill potential... that can be
> +          fixed later, but right now I'm calling it a "feature".  Should
> +          it be fixed, this spell could be changes to give a higher resist
> +          and in insta-kill situations the monster could be allowed to
> +          "scatter" to any of the immediate surrounding squares as a
> +          save against death (player could get this too).
> ```
> 
> (이 PR에 대한 응답이 극도로, 우스꽝스럽게 오래 지연되어 죄송합니다. 언젠가...)

wow! fine ;)

Hellmonk:
I am working on an updated version of this at level 2 fire/tloc. I believe the crash in the original pr is because the spell was not added to any books, which causes it not to be marked as a player spell.

PR #2516, opened 2022-04-04
Add new vault layouts

ClanchGames:
I added some vault layouts in simple.des and float.des.
This is my first contribution and first pull request in Github.
So please give me advices about anything in contribution and creating pull request.

### DIFF
--- a/crawl-ref/source/dat/des/arrival/simple.des
+++ b/crawl-ref/source/dat/des/arrival/simple.des
@@ -2512,21 +2512,21 @@ ENDMAP
 NAME:   reono_arrival_forest_river
 TAGS:   arrival no_monster_gen no_trap_gen no_pool_fixup
 ORIENT: float
-SUBST:  w = w:5 W:1 t:2
+SUBST:  . = t:1 w:1
 MAP
-ttttttttt
-twwwtwwwt
-twww{wwwt
-twwwWwwwt
-twwwWwwwt
-twwwWwwwt
-twwwWwwwt
-twwwWwwwt
-twwwWwwwt
-twwwWwwwt
-twwwWwwwt
-twwwWwwwt
-twww@wwwt
+@WtttttW@
+WW.....WW
+t.WWWWW.t
+t.WWWWW.t
+t.WWWWW.t
+t.WWWWW.t
+t.WW{WW.t
+t.WWWWW.t
+t.WWWWW.t
+t.WWWWW.t
+t.WWWWW.t
+WW.....WW
+@WtttttW@
 ENDMAP
 
 NAME:   reono_arrival_mine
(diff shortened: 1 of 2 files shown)


crawl-ref/source/dat/des/arrival/simple.des
```diff
@@ -2508,3 +2508,44 @@ tt.........tt
  tt.......tt
   ttt@@@ttt
 ENDMAP
+
+NAME:   reono_arrival_forest_river
+TAGS:   arrival no_monster_gen no_trap_gen no_pool_fixup
+ORIENT: float
+SUBST:  w = w:5 W:1 t:2
```
hellmonk3, line 2515: This subst can create pockets of shallow water that aren't accessible without flight or, rarely, inaccessible disconnected areas behind trees. You could try substituting only the glyphs next to the outer walls to make the vault feel less boxy without generating disconnects.

crawl-ref/source/dat/des/arrival/simple.des
```diff
@@ -2508,3 +2508,44 @@ tt.........tt
...
+twwwWwwwt
+twwwWwwwt
+twwwWwwwt
+twwwWwwwt
+twwwWwwwt
+twwwWwwwt
+twwwWwwwt
+twwwWwwwt
+twwwWwwwt
+twww@wwwt
```
hellmonk3, line 2529: The forest river vault is almost entirely water and has a single exit, which makes it quite hard to fight in. Ideally, entry vaults should give the player some tactical room to maneuver and/or multiple ways out of the vault in case they run into trouble immediately. To that end I would suggest adding a little bit more ground at the back to fight on and either a hatch downward or an additional pathway leading out in a different direction.

crawl-ref/source/dat/des/arrival/simple.des
```diff
@@ -2508,3 +2508,44 @@ tt.........tt
...
+xxxx.xxxx
+xxx.xxxxx
+xxxx.xxxx
+xxxxx.xxx
+xxxx.xxxx
+xxx.xxxxx
+xxxx.xxxx
+xxxxx.xxx
+xxxx@xxxx
+ENDMAP
```
hellmonk3, line 2549: Same thing for the mine arrival vault; there's really nothing for the player to work with here if they run into a monster right away. It needs to at least have an additional pathway out or a hatch.

crawl-ref/source/dat/des/variable/float.des
```diff
@@ -1411,6 +1411,22 @@ MAP
...
 
+NAME:   crystal_passage
+DEPTH:  D, Depths
+TAGS:   transparent extra no_monster_gen
+KITEM:   d = any potion
+MAP
+xxxb@bxxx
+xxxb.bxxx
+xxxb.bxxx
+bbbT.Tbbb
```
hellmonk3, line 1422: You can see the rock walls in the corners behind the fountains on crystal_passage. I'm not sure if this is intended; it's a very easy fix if not. Don't forget to credit yourself in the vault name!

hellmonk3 (commented):
Finally found some time to review this. Sorry it took so long!

ClanchGames:
Thank you for your review. I fixed my layouts according to your advice. I add some exits. 
How about is this? I await your reply.

ebering:
Thanks for the vaults! The edited ones were good. The crystal crossroads worked better as a mini_feature without loot, but it was merged basically as-is.

Crawl work waxes and wanes depending on how much time the various devs have. If you've been watching the repo you'll see I've recently had quite a bit and have been trying to cut through the PR backlog. I hope this hasn't discouraged you from sending us more in the future!

ebering:
Oh, and how would you like to be listed in `CREDITS.txt`?

ClanchGames:
Thanks and I would like to be listed in CREDITS.txt as "Reono".

PR #2526, opened 2022-04-12
fix: restore scroll-swinging speed (Mantis 12701)

nlavsky:
After a8e9c987b8, scroll-swinging takes 3 auts instead of 15 auts.
Also, the speed of tossing items depends on the wielded weapon
instead of being 15 auts.

This commit restores the old speeds.


crawl-ref/source/player-act.cc
```diff
@@ -277,7 +277,8 @@ random_var player::attack_delay_with(const item_def *projectile, bool rescale,
                                   skill(SK_UNARMED_COMBAT, 10);
         attk_delay = random_var(10) - div_rand_round(random_var(sk), 27*2);
     }
-    else if (weap)
+    else if (weap && (projectile ? projectile->base_type == OBJ_MISSILES
+                                 : is_melee_weapon(*weap)))
```
rawlins, line 281: I've only glanced at this, but it looks to me like this condition won't be taken on a ranged (no-longer-projectile) weapon, which I think is wrong?

I'd sort of recommend setting up a named var like `const bool non_weapon = ...` before the conditional, and then using that in the check, to make it clearer what is going on here. Or commenting this a bit.

nlavsky: Thanks for the review. I added some named vars and comments as suggested, so the code should be less cryptic now.

For context:

`player-act.cc:player::attack_delay_with()` has an `if -> else if -> else if` block, which handles the speed of different attacks.
Firstly, it handles throwing projectiles (but not item-tossing), then it handles the speed of unarmed combat, and the third part is for melee and ranged weapons.
Prior to a8e9c987b8, neither block handled item-tossing and scroll-swinging. These actions used the starting value, `random_var attk_delay(15);`, which is set at the top of `attack_delay_with()`.

Previously the conditional for melee and ranged weapons looked like this:
```cpp
else if (weap &&
                 (projectile ? projectile->launched_by(*weap)
                 : is_melee_weapon(*weap)))
```

where `launched_by()` was

```cpp
bool item_def::launched_by(const item_def &launcher) const
{
    if (base_type != OBJ_MISSILES)
        return false;
    const missile_type mt = fires_ammo_type(launcher);
    return sub_type == mt || (mt == MI_STONE && sub_type == MI_SLING_BULLET);
}
```

After a8e9c987b8, it's just `else if (weap)`, which means "the player has something in the weapon slot". As the result, scroll-tossing inherited the attack speed of an equipped weapon and scroll-swinging took 3 auts.

This PR restores the old conditional.

Ranged weapons fire fake projectiles now (`throw.cc:throw_it()` -> `item-prop.cc:populate_fake_projectile()`). So we can check the type of the projectile via `is_launcher_ammo()` still.

PR #2527, opened 2022-04-14
guarded vaults for some early-unrands

dillytin:
For Morg, Devastator, Throatcut


hellmonk3 (commented):
Did a quick review, overall these seem cool and I like them. In addition to the couple specific things I mentioned, I  suggest randomizing number of monsters placed at all depths for throatcutter and morg. It should make them feel a little more varied. Devastator is fine since it's only placing a few monsters.

PR #2535, opened 2022-04-28
fix: further fix grammar for artefact bardings with Regen

nlavsky:
There are two different attunement messages for items with Regen. The
first one is displayed when a character puts on such item while at full
health. When an injured character puts on an item with Regen and then
reaches full health, the second message is printed.

The first attunement message was fixed in 68df51542, and this commit
fixes the second one.


ebering:
Thanks for the patch!

PR #2538, opened 2022-05-10
A couple of tweaks for the &s wiz-mode command

nlavsky:


ebering:
Thanks for the patch!

PR #2546, opened 2022-05-19
Add several small vaults containing transporters to mini_features.des

nicolae-carpathia:
Add several small vaults which use transporters in ways besides just
setting off an optional murder zone. Some, with "diode" in their names,
allow the player to only move in one direction down one or more hallways.
The others just make weird intersections.


ebering:
Cool vaults! I transported them to master

PR #2547, opened 2022-05-19
Add some nostalgic vaults taken/borrowed from Nethack

nicolae-carpathia:
Even though I don't play it anymore these days, if I hadn't played Nethack I would never have gotten into roguelikes in the first place, and I would probably have never made even a single vault for DCSS. As tribute to the games of my past, I converted a few of the levels and vaults (so to speak) from Nethack into DCSS.


crawl-ref/source/dat/des/variable/large_abstract.des
```diff
@@ -1750,3 +1750,183 @@ x0xx.xx0x.x.0x%x.xxxx0x....x...xx..xx..xx..xx.....x.x..x...x..xx....+...+.000%%x
 x..0.0..x.x...........xxx.0.0x.0..xx0...0.xx0.....x0..0x.0..0...0.0.x000x000x%%x
 xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 ENDMAP
+
+# The nh_big_room_* vaults are a NetHack reference/homage, so the real credit is
+# to whichever NetHack dev made the Big Room levels. I wanted these vaults to be
+# ORIENT: centre, but they're too wide for that, so the game will just refuse to
+# load the maps at startup. So we'll fake it by making them ORIENT: north and
+# then using spacers to shove them back towards the middle. The big_room vaults
```
ebering, line 1758: Dubious... How over the limit are these on width for `ORIENT: centre`? Is it really impossible to change shape? 

Alternatively, what does the level builder complain about them exactly, and can we work around that restriction from the C++ side... (willing to dive into this, but not at my crawlputer now).

crawl-ref/source/dat/des/builder/shops.des
```diff
@@ -1301,6 +1301,36 @@ xx'''NbWbb..xx
    xxxxxxxx
 ENDMAP
 
+# A reference to the special Oracle level from NetHack.
+# Like in NetHack, players can purchase readings from the Oracle. In a sense.
+NAME:   nicolae_nh_oracle_at_delphi
+TAGS:   transparent no_rotate
+DEPTH:  D:5-9
+WEIGHT: 1
```
ebering, line 1309: Why the Depth restriction? I know it's part of the reference but with this tight range and weight 1 this vault is going to be impossible to place for the level builder... IMO D, Depths, Elf is a good range for something like this.

crawl-ref/source/dat/des/variable/mini_monsters.des
```diff
@@ -2550,6 +2550,28 @@ MAP
  ...
 ENDMAP
 
+# A shopkeeper has learned why there are no NetHack-style shops in this dungeon.
+NAME:   nicolae_nh_ransacked_shop
+TAGS:   transparent patrolling no_item_gen
+DEPTH:  D:4-10
+KITEM:  1d = any armour no_uniq w:90 / any weapon no_uniq / nothing w:25
```
ebering, line 2557: 20 items on average (even if mostly junk) seems a bit generous,  I'd switch the weight proportion to be 1/5 items and 4/5 floor. The band will have good items too (maybe they stole em??? could give the band gold to enhance the story).

crawl-ref/source/dat/des/variable/mini_monsters.des
```diff
@@ -3060,6 +3082,33 @@ xGD..D..1..E..EGx
 xxxxxxxxxxxxxxxxx
 ENDMAP
 
+# Another NetHack-inspired vault, based on the top floor of Vlad's Tower.
+NAME:  nicolae_nh_vlads_tower
+TAGS:  transparent patrolling
+DEPTH: D:10-, Crypt
```
ebering, line 3088: Unless you want to add overflow tagging and move this to altars (which IMO it shouldn't be), start the depth D:11-, altar placement in D:3-10 should only happen by overflow temples.

crawl-ref/source/dat/des/variable/mini_monsters.des
```diff
@@ -3060,6 +3082,33 @@ xGD..D..1..E..EGx
...
+KITEM: | = |
+KITEM: _ = *
+KITEM: 1 = %
+KMONS: _ = vampire mage / vampire knight
+KMONS: 1 = vampire / jiangshi
```
ebering, line 3093: this needs some depth conditioning for the proposed spawn range. Jiangshi and Vampire Knights are too much for D:11 en masse.

crawl-ref/source/dat/des/variable/large_abstract.des
nicolae-carpathia: > Dubious... How over the limit are these on width for ORIENT: centre? Is it really impossible to change shape?

It's not impossible, but just porting over the Nethack vault square-by-square makes them this wide. I suppoooose I could make some editorial changes to make them smaller but keeping the same basic shape...

> Alternatively, what does the level builder complain about them exactly, and can we work around that restriction from the C++ side... (willing to dive into this, but not at my crawlputer now).

It's [this bit in mapdef.cc](https://github.com/crawl/crawl/blob/5b5c5c9dfb8e383e54acd0718f9f1718c98ca31b/crawl-ref/source/mapdef.cc#L2934); for vaults oriented in the four cardinal directions, size-checking is only checked along one dimension. I'm generally reluctant to make too many changes to the code base just to support some goofy vault idea, just out of a sense of not wanting to overstep my bounds (so to speak), but if you think it's worthwhile I guess we can do that.

For what it's worth, I didn't notice any major issues during testing with the hack-around used here. (Although since I wasn't in debug mode, I guess if a bunch of levels got vetoed before the final one appeared, I wouldn't have noticed.)

crawl-ref/source/dat/des/builder/shops.des
nicolae-carpathia: > Why the Depth restriction?

Solely for the reference, but you're right. Even the players who get the reference aren't going to know this vault's depth range. I'll fix it up.

crawl-ref/source/dat/des/variable/mini_monsters.des
nicolae-carpathia: Noted! Giving the band members gold might be finicky, since if you specify gold explicitly you also have to make sure to explicitly specify their usual gear or else they just get the gold.

crawl-ref/source/dat/des/variable/mini_monsters.des
nicolae-carpathia: > this needs some depth conditioning for the proposed spawn range. Jiangshi and Vampire Knights are too much for D:11 en masse.

pfffft, shouldna been standin there

(I'll make the changes.)

crawl-ref/source/dat/des/variable/large_abstract.des
nicolae-carpathia: I updated based on the other comments, but I'm still not sure what to do with this one.

ebering: Hmmmm. I'll ping @rawlins about this, since he's done a lot of work on reducing vetoes. Reading the builder log this seems to work, I'm out of time now but I'll try to git blame the maximum dimensions and see if I can find some reasons for them to be set what they are.

PR #2558, opened 2022-05-30
Remove MF_SPECIAL_MINUS and minus_is_pageup().

Aliscans:
The remaining code which this code affected was removed in 95bfa34988.

I saw this flag, wondered if it was something I needed to know about, and found out that setting it doesn't do anything. If no-one has any plans involving it, I think it can just go.

### DIFF
--- a/crawl-ref/source/invent.cc
+++ b/crawl-ref/source/invent.cc
@@ -1798,8 +1798,6 @@ int prompt_invent_item(const char *prompt,
             const auto last_keyin = keyin;
             current_type_expected = keyin == '*' ? OSEL_ANY : type_expect;
             int mflags = MF_SINGLESELECT | MF_ANYPRINTABLE | MF_NO_SELECT_QTY;
-            if (other_valid_char == '-')
-                mflags |= MF_SPECIAL_MINUS;
 
             while (true)
             {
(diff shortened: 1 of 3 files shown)


rawlins:
Well, I do actually have a recent commit on a branch that uses this flag: https://github.com/crawl/crawl/commit/feb87438ddbfd68d3d591a19032c59dbf8774dc2

It's not absolutely necessary for what that commit does, but it was convenient to reuse in that context. I didn't check carefully but when making the change in branch I was wondering if the commit you referenced probably broke some instances of `-` as pageup?

(TBH I would actually prefer to remove +/- as paging controls for most menus if I could convince myself they aren't widely used. My sense is that people who went regular keyboard paging controls tend to use `<`/`>` and space.)

Aliscans:
If that flag seems like a good way to do something useful, then of course you should use it. I put this in because it wasn't doing anything, and I couldn't see what it would be useful for (if minus is special, where does that leave space?).

I don't know if that change broke anything or not. The minus_is_pageup() change was in Menu::set_more(), and I don't know if anything displays the default more string.

I'm not going to close this now because your changes are in a branch, and they may change before they're merged.

rawlins:
This is used again in a bunch of menus in current trunk, so I'm going to close this, sorry.

PR #2568, opened 2022-06-11
Linearize piety pips

hellmonk3:
Based on a discord conversation. The piety scale is really weird for presumably legacy reasons and could probably just be linear 25 piety per pip. Since this makes the first pip come earlier, I nerfed starting piety slightly to compensate. There are a few knock on effects worth considering; this effectively nerfs Ru by capping piety lower (reducing max ability power) and Uskayaw by lowering the piety floor (making it take longer to hit 3-4* for the juicy abilities). The first of these should be fine, but Uskayaw could maybe use a small adjustment. Ignis and Yred might also want some changes; I didn't mess with them at all yet. 
The last piety star encompassing 150-200 piety is still a bit weird, but adding an additional piety level would require a lot more adjusting and I'm not convinced it's worth it. Tbh the game clarity benefits of linearizing this scale aren't that big, but I thought I'd give it a try.

### DIFF
--- a/crawl-ref/source/god-passive.cc
+++ b/crawl-ref/source/god-passive.cc
@@ -478,11 +478,9 @@ int chei_stat_boost(int piety)
 {
     if (!have_passive(passive_t::stat_boost))
         return 0;
-    if (piety < piety_breakpoint(0))  // Since you've already begun to slow down.
-        return 1;
     if (piety >= piety_breakpoint(5))
         return 15;
-    return (piety - 10) / 10;
+    return max(piety / 10, 1);
 }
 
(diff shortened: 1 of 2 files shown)


regret-index:
Piety gain itself is non-linear, so I'm not really convinced that the display being linear yet the actual gain taking longer between each pip is much of a valuable smoothening out versus the various acknowledged knock-on effects this has on balance. I will defer to those with more knowledge on piety gain, though.

Hellmonk:
closing this, probably not necessary. Piety gain and pip scale is hella weird though, should probably be somewhat different.

PR #2571, opened 2022-06-12
Fix teleportitis-related messaging and status display

nlavsky:
This PR adds some indication that teleportitis is suppressed in the Abyss to the mutations and character overview screens.
Also, it marks teleportitis as inactive on the ``A`` and ``%`` screens in more cases.


ebering:
Thanks for the patches!

PR #2573, opened 2022-06-16
Add 3 tiers to the MUT_READ_SAFETY mutation

Rytisgit:
Add 3 tiers to the MUT_READ_SAFETY mutation to make it more in line with the other bad mutations. Ru's sacrifice is now the mutation at tier 3, which is the current severity.


Rytisgit:
As far as I tested there should be no incompatibility with saves. Characters who previously had the mutation or had sacrificed it through Ru would have the first tier now instead, so a slight buff for those who would carry over their saves.

crawl-ref/source/mutation-data.h
```diff
@@ -1848,12 +1848,18 @@ static const mutation_def mut_data[] =
...
-{ MUT_READ_SAFETY, 10, 1, mutflag::bad | mutflag::xom, false,
+{ MUT_READ_SAFETY, 7, 3, mutflag::bad | mutflag::xom, false,
   "inability to read while threatened",
 
-  {"You sometimes lose the ability to read scrolls when taking damage.", "", ""},
-  {"You lose the ability to read scrolls when threatened.", "", ""},
-  {"You can once more read scrolls while threatened.", "", ""},
+  {"You rarely lose the ability to read scrolls when taking damage.",
+   "You ocassionally lose the ability to read scrolls when taking damage.",
+   "You sometimes lose the ability to read scrolls when taking damage."},
```
hellmonk3, line 1856: Gammafunk and I suggest occasionally / sometimes / often as the adverbs here to match teleportitis adverbs.

crawl-ref/source/ouch.cc
```diff
@@ -807,16 +807,20 @@ int do_shave_damage(int dam)
...
     const int hp_fraction = you.hp * 100 / you.hp_max;
-    return damage_fraction_of_hp > 5
-            && hp_fraction <= 85
-            && (damage_fraction_of_hp + random2(20) >= 20
+    const int safe_damage_fraction = 5 * safety_level;
+    const int scary_damage_fraction = 20 * safety_level;
+    return damage_fraction_of_hp > safe_damage_fraction
+            && hp_fraction <= 100 - scary_damage_fraction + safe_damage_fraction
+            && (damage_fraction_of_hp + random2(20) >= scary_damage_fraction
                 || random2(100) > hp_fraction);
```
hellmonk3, line 824: The random2 part of this formula doesn't match the behavior I would expect. damage_fraction_of_hp + random2(20) >20 can pass for any damage fraction, but this can't occur for scary_damage_fraction of 40 or 60; you effectively have a higher safety threshold on that part of the argument bc you need a damage fraction of 21 or 41 to have any chance to pass it (can still pass the other random2 on hp_fraction of course). The random2 should probably be random2(20* mut_level) instead.

crawl-ref/source/ouch.cc
```diff
@@ -947,7 +951,8 @@ void ouch(int dam, kill_method_type death_type, mid_t source, const char *aux,
         // don't always trigger in unison when you have both.
         if (you.get_mutation_level(MUT_READ_SAFETY))
         {
-            if (_is_damage_threatening(damage_fraction_of_hp))
+            if (_is_damage_threatening(damage_fraction_of_hp,
+                                        you.get_mutation_level(MUT_READ_SAFETY)))
```
hellmonk3, line 955: If you feel like it, making MUT_DRINK_SAFETY explicitly pass 3 to the function rather than using a default value is slightly nicer. If you feel very ambitious and don't mind if other players get mad at you, certain developers have suggested giving MUT_DRINK_SAFETY the same 3 level mutation treatment and adding it to the general bad mutation pool.

hellmonk3 (commented):
I cannot believe you would neuter such a beautiful and well loved mutation smh

crawl-ref/source/ouch.cc
hellmonk3: to be clear, the way you handle MUT_READ_SAFETY is fine, this is just about where to put the MUT_DRINK_SAFETY parameter since it's not the same.

crawl-ref/source/mutation-data.h
Rytisgit: Good suggestion, my main thought was that increasing the frequency for the third tier might confuse players that the mutation effect is now more common. But since you suggest it, it works for me.

crawl-ref/source/ouch.cc
Rytisgit: I also was thinking that with these changes the no drink mut could be introduced, since it was also suggested here, I added it as a new regular bad mut.

semitonal:
Copying over some comments from discussion in Discord/IRC earlier: giving MUT_READ_SAFETY multiple levels for when it shows up randomly sounds reasonable, but in general 3 levels is probably too many for most random bad mutations, since the final level will almost never show up. 2 levels is a good balance, since it lets there still be a distinction between occasional/frequent triggers for this kind of chance-based mutation, and also works well with temporary mutations from wretched stars since they give out 2 levels at a time.

Teleportitis and berserkitis are similar chance-based bad mutations that would be more suited to being 2 levels (as deterioration and shoutitis already are).

Rytisgit:
2 tiers for mutations works well, it's true that the 3rd tier is very rare. Regarding the conflict mutations, I see the point about the conflict mutations being completely incompatible so I've removed it. However, the conflicts do serve a pretty nice gameplay purpose of allowing the current mutation potion to better manage those mutations which have conflicts and strong will was pretty nice thematically to counter being threatened.

ebering:
Thanks for the patch! I've combined these into a single commit and merged to master.

PR #2576, opened 2022-06-19
Add a new unrand: the lochaber axe

nicolae-carpathia:
Long long ago, bardiches used to be known as lochaber axes, until they
were renamed in commit 8b14a9, on December 13, 2007, presumably so
that players wouldn't think they were some kind of axe.

The lochaber axe was gone... but not forgotten...

It's back, reimplemented as an unrand, and this time, there's no problem
if players think it's some kind of axe. Because it is. The lochaber
axe has both reach and cleaving, and combines them both into a sort of
"super-cleave", aka "reave", which attacks everybody in a radius of 2
with each swing.

Additionally, due to its nature as a combination of polearm and axe,
either skill can be used to wield it -- the higher of your Polearms or
Axes skill will be used to determine its damage and delay.

Currently, it's statted out as a +9 bardiche with vorpal brand, -1 to
base accuracy, and +1 to base damage and base delay. These stats will
probably be changed at some point. Currently uses the old tile for
lochaber axes, although it will probably need something more menacing
before it's ready for prime-time.

Much thanks to PleasingFungus for coding assistance, and gammafunk for
discussions, the lochaber axe tile, and not immediately reverting this.


ebering:
This was merged by @PleasingFungus

PR #2579, opened 2022-06-20
fix: apply item_slot when seeing monsters use unid'd scrolls or potions

nlavsky:
When a monster uses a scroll or potion, it identifies the consumable and applies autoinscriptions. But unlike other methods of item identification, it didn't change inventory letters according to the ``item_slot`` option.


ebering:
Thanks for the patch!

PR #2581, opened 2022-06-25
Stat mutation description adjustment

Goratrix:
Add spaces between the stat names and the numbers (to be consistent with
all the other similar mutation descriptions) and change Int/Str to Str/Int
for consistency with the order of the stats in the ui and elsewhere.


PleasingFungus:
Looks good, thanks! :)

PR #2583, opened 2022-06-25
Allow adjusting of explore_greedy non autopickup item visit method

Rytisgit:
explore_greedy_visit is added as an rc file option to adjust explore_greedy behavior. By default greedy explore travels to piles of items on each square in addition to items that are eligible for autopickup. This setting allows to also/only go to enchanted items instead. By choosing "enchanted" greedy exploration will ignore piles of mundane items and only visit the squares which contain
enchanted items. Choosing "both" will visit piles and singular enchanted items.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -829,14 +829,20 @@ explore_greedy = true
         otherwise identical to regular explore. Explore greed is
         disabled if autopickup is off (Ctrl-A).
 
-explore_greedy_visit = (pile | enchanted | both)
-        By default greedy explore travels to piles of items on each
-        square in addition to items that are eligible for autopickup
-        this setting allows to also/only go to enchanted items instead.
-        By choosing "enchanted" greedy exploration will ignore piles
-        of mundane items and only visit the squares which contain
-        enchanted items. Choosing "both" will visit piles if they
-        contain enchanted items or not and singular enchanted items.
+explore_greedy_visit = artefacts,glowing_item
+explore_greedy_visit += stacks
+        (List option)
+        Greedy exploration will visit squares with items based on
+        these conditions.
+
+        stacks: makes explore_greedy visit stacks even if they don't
+        necessarily have target auto pickup items
+
+        glowing_items: makes explore_greedy visit glowing_items,
+        even if not in a stack
+
+        artefacts: makes explore_greedy visit artefacts, even
+        if not in a stack
 
 explore_stop  = items,stairs,shops,altars,portals,branches,runed_doors
 explore_stop += greedy_pickup_smart,greedy_visited_item_stack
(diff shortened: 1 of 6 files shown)


Rytisgit:
I tried to get it to only walk towards unidentified enchanted items by also comparing  si->flags & ISFLAG_IDENT_MASK and si->flags & (ISFLAG_IDENT_MASK & ~ISFLAG_KNOW_PROPERTIES) when iterating all items in pile, but it didn't work right and still went towards identified enchanted items, so it's not included in this pr and just walks to wards all enchanted items the first time.

crawl-ref/docs/options_guide.txt
```diff
@@ -829,6 +829,15 @@ explore_greedy = true
         otherwise identical to regular explore. Explore greed is
         disabled if autopickup is off (Ctrl-A).
 
+explore_greedy_visit = (pile | enchanted | both)
```
ebering, line 832: Use stacks instead of piles to match the other `explore_stop options`

ebering (changes requested):
Stack visiting is an accessibility/comfort feature more than anything (now that pile contents are visible from a distance), particularly for VI players it is nice to automatically filter through stacks of items this way.

I suggest that instead of  `enchanted` you offer `glowing` (matching the `glowing_items` `explore_stop` condition) and `artefact`, and make the option a list option with the following three choices with the described semantics:
- `stacks`: makes explore_greedy visit stacks even if they don't necessarily have target auto pickup items
- `glowing_items`: makes explore_greedy visit glowing_items, even if not in a stack
- `artefacts`: makes explore_greedy visit artefacts, even if not in a stack

With floor-id I think these are more consistent with offering a smooth experience for players with accessibility needs, while also looting the dungeon effectively for all players. I would set the default to `glowing_items, artefacts` and add `stacks` to the recommended accessibility options in the last section of the options guide.

crawl-ref/docs/options_guide.txt
Rytisgit: updated to use stacks

Rytisgit: In case you are going to be testing this, adding 
`dprf("Options.explore_greedy_visit %i stack_greed: %i, glowing_greed:%i, artefact_greed:%i", Options.explore_greedy_visit, stack_greed, glowing_greed, artefact_greed);`
to line 302 in stash.cc is pretty good for seeing that it works.

crawl-ref/docs/options_guide.txt
```diff
@@ -829,6 +829,21 @@ explore_greedy = true
         otherwise identical to regular explore. Explore greed is
         disabled if autopickup is off (Ctrl-A).
 
+explore_greedy_visit = artefacts,glowing_item
```
ebering, line 832:
```suggestion
explore_greedy_visit = artefacts,glowing_items
```

ebering (commented):
Nice, looks good!

I'm going to commit the one consistency fix from the web UI, give this a quick test, then squash and merge assume it all works right.

ebering (approved):

ebering:
Thanks for the patch and the quick turnaround on the changes!

PR #2590, opened 2022-06-28
fix: reorder skills on the `m` screen

nlavsky:
The skill order from 586177b653 does make sense:
* Throwing no longer crosstrains with Slings (now Ranged Weapons), so
  they don't have to be in the same group;
* both Throwing and Evocations rely on consumables.

On the other hand, Throwing and Invocations have nothing in common and should not be in the same group. It's better to have Throwing with all the other Dex-based skills, right below the Str-based skills. Also, this allows to have all the physical/non-magic skills on the left, and the magic/divine skills on the right. Evocations also belongs there, because any sufficiently advanced technology is indistinguishable from magic :).

Current Skills menu:
old

Suggested Skills menu:
new

^ Skills take the same number of lines, 17, as before the merging of ranged skills.


PleasingFungus:
Thanks for the PR!

> On the other hand, Throwing and Invocations have nothing in common and should not be in the same group...

Throwing, Invo and Evo are all quite similar. They're all auxiliary, limited-use ways to deal with unusually dangerous situations, as opposed to weapon skills, which are a default killdudes tool for 'normal' fights.

> It's better to have Throwing with all the other Dex-based skills, right below the Str-based skills.

Throwing is Str-based.

PleasingFungus:
Sorry - looking back, I think my closing message might have come off as dismissive! Not my intent, just terse while holding baby.

Lot of dev discussion about this - you might be interested in yday's crawl-dev logs.

nlavsky:
That's fine.

I've seen the discussion on #crawl-dev, and it looks like you and other devs support the idea that the "Str vs. Dex" distinction is worth clarifying. The skills menu seems like the right place for doing this.

There have be other ways to indicate "Str vs. Dex" without reordering the skills:
* Use different colours for Str-based and Dex-based skills. The problem is that the menu already uses colours to indicate reduced, enhanced, and fully trained skills.
* Add a new column. There's _probably_ not enough space for `Str`, `Dex`, and `Int` (and the "Attr" header), but `S`, `D`, and `I` might fit.

PR #2591, opened 2022-06-29
Only list gods you can ever worship in the dungeon overview & dump.

Aliscans:
Different gods forbid worship from different species (some despise undead, some abhor djinn, Beogh can only stand orcs and everyone hates demigods).

This change gives a way for the player to check which gods are available, and hides irrelevant information about whether an unusable altar has been found.

This list only changes during a game when a character sacrifices love to Ru. This seems fine to me.


ebering:
Thanks for the patch!

PR #2598, opened 2022-07-02
Correctly produce strings of the form "one hundred and one"

damerell:
In Australian English there is an "and" after "hundred" in numbers
of this kind. (Source: I asked all the Aussies I know.)


ebering:
Thanks for the very important fix!

PR #2603, opened 2022-07-05
Proposal/POC a unique salamander who casts ignite poison

pdpol:
I wanted to see if a unique designed around Ignite Poison would be a fun thing, and as far as I know doesn't already exist. @PleasingFungus remarked that it sounded like a salamander, so I went in that direction to try out the concept. At a broad level, I think one of the fun things about uniques is seeing popular playstyles reflected, and ignite poison struck me as a particularly popular one. As I've started to get into this I have doubts about how fun this actually is, or if there are thematic problems, but I figured I'd put it out here and hear some opinions about what might make this idea work.

Here's a brief description of the state of things and some of my thoughts and concerns:

* Currently, I've made Calida a unique who appears exclusively in snake pit with a band and have tried to base her stats largely off of Vashnia and Aizul.
* In addition to Ignite Poison she casts Mephitic Cloud and Poison Arrow.
  * One of my concerns is that I'm conflating salamanders and nagas by giving her poison spells. I wanted to make sure that there was potential for a player to both stand in poison clouds that get ignited, as well as become poisoned and ignited themselves.
  * At first I thought Mephitic Cloud was a natural fit since it's such a classic spell to use with this playstyle but with the band this just kind of works out to salamanders getting confused and killing each other. I thought some type of hex-like effect like the confusion would be important to make this encounter spicier but rP trivializes that for the player anyway. I've also tried Poisonous Clouds out to keep the clouds aspect in, not sure which I prefer and also shares my same general concern that poison spells might not be thematic for a salamander
  * I initially tested with Venom Bolt instead of Poison Arrow but that felt trivial for a well-prepared player too, so upgraded.
  * Spell order and frequency - the intent is that if ignite poison is a good spell to cast, Calida should have a pretty decent chance of doing so and try it before anything else. If not, there's still a pretty good chance we'll get clouds or a direct poison attack.
* Band has a salamander mystic, naga ritualist, and salamanders
  * Once again just looking at precedent it seems like I'm doing something out of place by putting a naga ritualist here. I thought it would help balance the encounter against prepared players by having virulence and OTR in the mix.
* Balance - I have no idea what I'm doing 😄 . In testing it seems like Calida is definitely manageable by realistically well-prepared characters, and if anything is slightly easier to kill by an o-tab spammer than Aizul and Vashnia if they have the right stuff. Other solid things like properly-leveled BVC + iron shot also make short work of her.

Other thoughts I've had since I started down this path
* A salamander unique could potentially have a similar effect without necessarily needing to cast poison or ignite it. Maybe a melee-oriented unique that has a good weapon and stats and is surrounded by a shroud of volcanic smoke clouds that it has a natural ability to ignite into flame clouds?
* Maybe something less threatening that ends up not being a unique but just a challenging monster, something that thematically can do poison and fire?
* Am I just trying to reinvent a fire crab?


pdpol:
Gotcha, thanks for the context @semitonal, good to know this path has been tried before. Do you think there would be interest in a different approach to a salamander unique? Not that I have anything more concrete than the half-baked idea about a melee-oriented one above.

semitonal:
I could certainly see a salamander unique being cool in theory but yeah, it'd need a bit more of a concept working up I think. Being surrounded by flame clouds is sort of similar to Bai Suzhen's thing, and might also be a little mean to melee-focused characters (Bai Suzhen can get away with it by being a fairly late unique where there's a lot of other tools players should have to help).

PleasingFungus:
To say nothing of Asmodeus!

pdpol:
Ahhh yeah, all good points. Well, I don't really have anything I'm attached to that's related to this anymore, fun experiment to try out and all. I think I'll close this and if inspiration strikes at some point come back with something else. Thanks for taking a look!

PleasingFungus:
Thanks for submitting & for being a good sport about the whole thing!

PR #2604, opened 2022-07-05
Give Makh 2 more greater servants

riverfiend:
The comment suggests this should go up to 6, but oldcode didn't permit this. Fixed this unjust deprival of Makhleb's wrath to puny mortal players!


riverfiend:
Actually, in about a week I'll have a whole bunch of changes like these for god-wrath.cc -- I'll push them all at once when I've got them all ready.

PR #2605, opened 2022-07-06
Jelly simulacrum tiles

4Hooves2Appendages:
These are a bunch of suggestions for jelly-type simulacrum tiles.

If any of them are useable, please let me know and I will add them to the game. Otherwise, feedback please!


regret-index:
Most of these tiles are just the regular slime tiles recoloured blue (or a white mustache effect being placed on azure jellies), which is fairly worse for accessibility and visibility's sake than the current one added in bdae368 having an immediately recognizably different silhouette from normal slimes:

simulacrum_slime

While having two tier or three "tiers" for slime simulacra would be a good idea due to the existence of `ice_cave_hellmonk_azure_passage`, none of these would quite make the standards for doing such, I am sorry to say.

PR #2607, opened 2022-07-10
tiles: New amulet of acrobat icon

CipHuK:
The "Acrobat" and "Magic regeneration" amulets have icons that are hard
to distinguish

I hope that two gymnastic rings are better able to show the
understanding of an acrobat


PleasingFungus:
Thanks for submitting this! Very sorry for the long delay in review. I've recoloured the old icon for now, but I've added your icon to our cache so that other devs can use it later.

PR #2608, opened 2022-07-12
fix: tweak shield offsets for several elves and spriggans

nlavsky:
This commit fixes shield offsets for the following monster tiles:

* deep elf pyromancer, zephyrmancer, and demonologist;
* spriggan berserker and air mage.

Before and after:
before_and_after


ebering:
Thanks for the tweaks!

PR #2614, opened 2022-07-15
fix: remove -scroll and -potion from Xom muts

Implojin:
Since 0d03ebdc04, characters taking Xom are signing up for a semi-persistent
dual -potion -scroll challenge conduct, on top of Xom's regular actions.

That's pretty extreme! Let's dial it back a bit.

This commit removes -scroll and -potion from Xom's mutation pool, aligning
these muts with Xom's behavior of not explicitly granting teleportitis or
devolution.

These muts can still be granted randomly by Xom with one_chance_in(1000),
but that's quite a bit less punishing than mutflag::xom's one_chance_in(5).

### DIFF
--- a/crawl-ref/source/mutation-data.h
+++ b/crawl-ref/source/mutation-data.h
@@ -1840,7 +1840,7 @@ static const mutation_def mut_data[] =
 },
 #endif
 
-{ MUT_DRINK_SAFETY, 7, 2, mutflag::bad | mutflag::xom, false,
+{ MUT_DRINK_SAFETY, 7, 2, mutflag::bad, false,
   "inability to drink while threatened",
 
   {"You occasionally lose the ability to drink potions when taking damage.",
@@ -1854,7 +1854,7 @@ static const mutation_def mut_data[] =
    ""},
 },
 
-{ MUT_READ_SAFETY, 7, 2, mutflag::bad | mutflag::xom, false,
+{ MUT_READ_SAFETY, 7, 2, mutflag::bad, false,
   "inability to read while threatened",
 
   {"You occasionally lose the ability to read scrolls when taking damage.",


Implojin:
I've force pushed my branch here to reword the commit message slightly, if anyone reviewing this wonders about the force push.

Implojin:
So after doing some git archaeology, I see that 9cdcd136b3 carried over the previous mutflag::xom behavior of blurry vision. 

Blurry vision's behavior changed over the years from a chance of read failure, to an added delay when reading, to -vision, but a digression into an originalist discussion about now-removed behavior isn't really germane to whether Xom should be handing out a dual noscroll nopotion conduct in 2022 Crawl.

The present behavior is really brutal to play through in-game, and I don't think that it should exist as it currently does.

I'm rewriting this commit message to remove the references to past behavior.

semitonal:
I don't think the commit message reflects what's actually happening here - the one_chance_in(1000) is to just bail out from the loop and not mutate at all, I believe? The approximate behaviour is that you have a 1/5 chance of a mutflag::xom mutation (with mutation weight factored in to which one is picked), and a 4/5 chance of a fully random mutation (ignoring mutation weight), as far as I can tell?

I do agree that some kind of change here could be reasonable. I'm not sure that the small pool of mutations that Xom is particularly likely to apply is necessarily interesting behaviour in the first place, and ignoring weight and rolling random2(NUM_MUTATIONS) also isn't a good way of giving out mutations. It might be reasonable just to have Xom's random mutation action use the standard RANDOM_MUTATION handling so that mutation weight is factored in, although possibly that pushes the "bad" version of Xom's mutation action too far towards applying a bunch of good mutations.

semitonal:
Adjusted this to instead remove Xom's special mutation pool, and just select mutations normally (but with a higher ratio of bad mutations). Thanks for the PR!

PR #2616, opened 2022-07-15
Xom-themed vaults for Crypt and Abyss

Skrybe:
Crypt end and Abyss rune vaults, both with Xom flavor.  Both have themes of chaos effects, mutations, demons, and dancing weapons.


regret-index:
The Xom Abyss rune vault is pretty cool, building off of evilmike_abyss_rune_tso_outpost as well as that rare Xom angel thing. I'd probably adjust the monster distribution and count a little bit- opening a door to dancing weapons and daevas, only for there to be even more groups a bit further in, is pretty tough on the threat count. A few like macabre_mess, seething_liches, and horror_hoard are somewhat comparable, but that's only a few of the 18 abyss rune vaults. Still, it's got a hook and a good shape.

The Xom Crypt end I've got a few concerns over. It's extremely monster dense and uses thin walls with open corridors, so it's going to outpour several dozen monsters from the very start. We do have one Crypt end that does something like that in lightli_unholy_cathedral, but that vault is pretty high up on my branch ends nerfing list. Also, well. Crawl inherited a kinda strange split of branches over time, where some are pretty loose agglomerations of multiple themes and some are pretty restrained by top-down flavour requirements (looking at you, Snake). Crypt falls into the latter category and thus has less variety than many branches (plus already containing necromantic demons), but I'm still not much of a fan of putting non-summoned neqoxecs or cacodemons in the pretty cohesively undead branch. Xom flavour only goes so far, and getting mutated by a neqoxec in a branch one wouldn't expect them in amongst a dozen derived undead isn't much fun.

If you're okay with it, I can tweak these vaults a fair bit (probably adjusting the latter into another large vault for Depths, which has both undead and demons plenty) once you give me permission so they may be finally merged in.

Skrybe:
Xom Abyss rune vault: As far as adjusting the monster count, the first thing I'd do is drop one of the apocalypse crabs, and push the remaining one closer to the rune.  Also, my placement of runelights was sort of arbitrary, that might need to be adjusted too.

Xom Crypt end: Looking at this now, the large swarm of derived undead on either side of the entrance should probably be removed.  The wave of zombies and etc is funny, but I'm sure there are more interesting ways to use the space.

Yes, feel free to adjust these as necessary.  I hope you'll be conservative with changes to the Abyss rune vault, though, since I like the way that one turned out.

PR #2620, opened 2022-07-18
fix: allow Poisonous Vapours to work under Still Winds

nlavsky:
Allow Poisonous Vapours to work under `-Clouds`, since as of 3f0e5c0cc there are no clouds involved.


PleasingFungus:
Good catch, thanks! :)

PR #2626, opened 2022-07-23
Compact throwable

sdynet:
Dispersal is a special ego that doesn't hurt, and I think it fits the theme of darts better.
Remove boomerang ego:dispersal
Add dart ego:dispersal

Reduce inventory pressure and leave Silver as a place for javelin(and Lajatang of Order). For the same reason, leave only one cool-est of the two types of javelin. How much should the drop table be adjusted?

As a passing story, I heard from someone that I wanted to get rid of the poison dart, leaving only darts that do not damage the enemy in the dart category and only give them bad status. The idea led me here. I personally wanted to change the poison dart to a poison boomerang, but I decided to put this change on hold after hearing that this would make enemies of the early games with the poison boomerang too powerful. Does anyone have a great idea for this?


kd7uiy:
I like this idea, the idea of a boomerang that doesn't come back always bothered me. It makes much more sense as a dart.

PleasingFungus:
I've merged the first two parts of this change, just a bit late for the 1-year anniversary. Still thinking about pt 3. (Very, very, very sorry for the slow response here.)

PR #2632, opened 2022-07-28
Svaults sfor sbranches

nicolae-carpathia:
A selection of vaults for the various S-branches, separated into commits by branch and then by entry/decor/monstriferous.


crawl-ref/source/dat/des/branches/shoals.des
```diff
@@ -221,6 +221,63 @@ MAP
...
+...xx`xx...
+ .........
+ENDMAP
+
+NAME:  nicolae_shoals_entry_how_ready_are_you
+TAGS:  shoals_entry transparent no_pool_fixup
+KMONS: K = kraken
+KFEAT: K = w
+KFEAT: O = enter_shoals
+SUBST: M = wwx, _ = x_
```
hellmonk3, line 248: I think technically there's a very small chance that the whole box is enclosed by rock? Not sure if this matters. Players can usually boringly kite the kraken with a ranged attack from out of its range but that's a problem with almost every kraken vault so whatever.

crawl-ref/source/dat/des/branches/shoals.des
```diff
@@ -2217,6 +2335,166 @@ swwwwwwwwwwwwwwww
 swwwwwwwwwwwwww
 ENDMAP
 
+NAME:  nicolae_shoals_bring_sunscreen
+TAGS:  transparent water_ok no_tide
+DEPTH: Shoals:2-
+KMONS: S = sun demon / sun moth
```
hellmonk3, line 2341: imo this is too many sun moths; they're a top threat in a different sbranch and this vault placed 4-5 of them when I was testing it. Maybe nsubst in like one of them and do sun demons plus branch native monsters? Cool theme though.

crawl-ref/source/dat/des/branches/shoals.des
```diff
@@ -2217,6 +2335,166 @@ swwwwwwwwwwwwwwww
...
+KITEM:  12 = nothing w:30 / any book / any scroll / any potion
+# Deep elves have to bring the high-strength sunscreen. At least one potion
+# of resistance is guaranteed in the changing room.
+KITEM:  s = potion of resistance ident:all
+# 1/3 chance of rudely walking in on a deep elf changing.
+KMONS:  p = deep elf knight ; nothing / nothing w:20
+KMONS:  1 = deep elf pyromancer w:20 / deep elf zephyrmancer w:20 / \
+            deep elf archer / deep elf knight / \
+            deep elf archer band w:5 / deep elf knight band w:5
+KMONS:  2 = deep elf death mage / deep elf sorcerer / \
```
hellmonk3, line 2425: Vaguely wondering if this vault could have different themed sets for other branch vacationers. The elf set seems tough but manageable.

crawl-ref/source/dat/des/branches/spider.des
```diff
@@ -259,6 +259,33 @@ xxx....x.x.x.x....xxx
         xxxxx
 ENDMAP
 
+NAME:    nicolae_spider_entry_lesser_pyramids
+TAGS:    spider_entry transparent patrolling
+KMONS:   Os = pharaoh ant
+KMONS:   S = wolf spider / redback / tarantella / jumping spider
+KFEAT:   O = enter_spider_nest
```
hellmonk3, line 266: This is probably a bit much for spider entrance; 1 pharaoh ant plus some of the non wolf spider guys is plenty.

crawl-ref/source/dat/des/branches/spider.des
```diff
@@ -2091,6 +2118,157 @@ xx...x...xx
...
+ MMWWWWMMSSxxSSMMWWWWMM
+ xxMWWWWWWMxxMWWWWWWMxx
+ xxSSMWWWWWMMWWWWMMSSxx
+ xx''xxMWWWWWWWWMxx''xx
+  xxxxxSSMMWWMMSSxxxxx
+   xxxx''xxMMxx''xxxx
+    xxxxxxxxxxxxxxxx
+ENDMAP
+
+NAME:   nicolae_spider_solar_radiation
```
hellmonk3, line 2210: All of the spider monster vaults seemed pretty nasty but ultimately fair. I think this one is probably closest to being "too much" just because it's small; probably ok but you could drop the monster density a bit imo.

crawl-ref/source/dat/des/branches/swamp.des
```diff
@@ -1260,6 +1285,175 @@ tt... PppPqqq    QqqQQqQqqwwwwww.t
            PPPP    PPPP       ttt
 ENDMAP
 
+# The various "reclaimed" decor vaults just place a pattern of rock walls over
+# the terrain, as if the swamps and forest had formed around the ruins.
+# Sometimes they end up barely visible, enclosed entirely in mangroves except
+# for a corner or something. This is acceptable.
+NAME:  nicolae_swamp_reclaimed_barracks
```
hellmonk3, line 1292: I think I mentioned this in discord but it would be cool to get a pattern like this working in orc with crystal walls.

crawl-ref/source/dat/des/branches/swamp.des
```diff
@@ -2856,6 +3050,63 @@ Ztt..Kttttt..ttZ
...
+ RR--`````--OO
+  RRQ-----PPP
+   RQQQQPPPP
+     QQQPP
+ENDMAP
+
+NAME:    nicolae_swamp_vermiculated
+TAGS:    transparent patrolling
+DEPTH:   Swamp
+KMONS:   ~ = steelbarb worm / swamp worm / tyrant leech
```
hellmonk3, line 3082: Maybe have steelbarb at half weight so it stays more in branch theme? The center of the vault seems quite dense but it's ok since it also provides a ton of corners for the player to abuse.

crawl-ref/source/dat/des/branches/swamp.des
```diff
@@ -2856,6 +3050,63 @@ Ztt..Kttttt..ttZ
 ZtrrrrtZZtqqqqtZ
 ENDMAP
 
+NAME:    nicolae_swamp_thats_no_hydra
+TAGS:    transparent no_pool_fixup
+DEPTH:   Swamp
+KMONS:   1 = kraken
```
hellmonk3, line 3056: I think this one needs something a bit "extra" for the theme to be worthwhile. Not sure what, my first thought is hydra skeletons.

crawl-ref/source/dat/des/branches/shoals.des
nicolae-carpathia: > I think technically there's a very small chance that the whole box is enclosed by rock?
Technically, although I guess in that case you could just call it a freebie. I might drop this one, though. I'm not super excited by it.

crawl-ref/source/dat/des/branches/shoals.des
nicolae-carpathia: > Vaguely wondering if this vault could have different themed sets for other branch vacationers.

It originally had nagas, although it was pointed out that that would somewhat dilute the distinctiveness of the s-branches, so I changed it to elves. Orcs would also work, but many players will have finished Orc before coming to the s-branches, so I'm not sure if that would be a sufficient challenge.

crawl-ref/source/dat/des/branches/shoals.des
```diff
@@ -221,6 +221,63 @@ MAP
 ....xxxxxxx...@
 ENDMAP
 
+NAME:  nicolae_shoals_entry_flooding
+TAGS:  shoals_entry patrolling transparent
+KMONS: ON = water nymph
+KFEAT: O = enter_shoals
+SUBST: x = x`, ` = W., N = N .:80
```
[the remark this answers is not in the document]
nicolae-carpathia: > I kind of expected them to become water.

That's what the nymphs are for! 

I've changed it to an NSUBST to place between 1 and 3.

crawl-ref/source/dat/des/branches/shoals.des
nicolae-carpathia: NSUBST now places 1-3 sun moths, with 1 guaranteed and 1/3 chance of each of the other two.

crawl-ref/source/dat/des/branches/shoals.des
```diff
@@ -221,11 +221,12 @@ MAP
...
-NAME:  nicolae_shoals_entry_flooding
-TAGS:  shoals_entry patrolling transparent
-KMONS: ON = water nymph
-KFEAT: O = enter_shoals
-SUBST: x = x`, ` = W., N = N .:80
+NAME:   nicolae_shoals_entry_flooding
+TAGS:   shoals_entry patrolling transparent
+KMONS:  ON = water nymph
+KFEAT:  O = enter_shoals
+NSUBST: N = 2=N` / `
```
gammafunk, line 228: I think I would go further and make this just 1-2 nymphs, and probably at least half weight. It's nasty to fight even one water nymph in Lair, which is fine for a vault, but shoals entries place in 50% of games, so this vault will place a lot. One nymph for 75% of the time and two for 25% with default weight would probably be fine as well. But if you like placing up to three, strongly suggest lowering weight, and it's pretty reasonable to give it lower weight regardless of the nymph count.

crawl-ref/source/dat/des/branches/swamp.des
```diff
@@ -1260,6 +1285,175 @@ tt... PppPqqq    QqqQQqQqqwwwwww.t
...
+      xx  xx  xx  xx
+xx                      xx
+xx                      xx
+_
+_
+xx  xx  xx  xx  xx  xx  xx
+xx  xx  xx  xx  xx  xx  xx
+ENDMAP
+
+NAME:  nicolae_swamp_reclaimed_house
```
gammafunk, line 1336: These vaults are cool and I like the concept, but I feel you need to scale down their size and test their placements to see that they're actually placing in a meaningful way. Tests with this vault, which isn't the largest one, show that it rarely places such that even 25% of the vault is visible, and even then it's pretty difficult to see the overall structure, which is a shame. It might be the scaling down the vaults to 66-50% of their preset size will be enough, but you might want to test placements of them individually to know which need this and to what degree.

crawl-ref/source/dat/des/branches/swamp.des
```diff
@@ -2856,6 +3050,65 @@ Ztt..Kttttt..ttZ
...
+RR-`DDDDDDD`-OO
+ RR-`DDDDD`-OO
+ RR--`````--OO
+  RRQ-----PPP
+   RQQQQPPPP
+     QQQPP
+ENDMAP
+
+NAME:    nicolae_swamp_vermiculated
+TAGS:    transparent patrolling
```
gammafunk, line 3082: Feels like this vault probably doesn't need patrolling. For one it has a lot of walls that will block noise and prevent worms from wandering out so freely. Secondly it's all pretty standard Swamp fodder that's not going to be confusing to see outside of the vault. And finally it's probably good to break up the swamp worms so they're not so penned in anyhow, since they need LOF to do their thing. Generally think it's better to avoid use of patrolling unless there's a compelling use case like a theme that requires monsters to be in place for it to register, especially in more open vaults.

gammafunk (approved):
Looks good overall, nothing huge in terms of vaults not working. I didn't look at much at the couple of more problematic vaults Hellmonk talked about, since I think you're addressing those. My suggestions are all minor tweaks or syntax suggestions.

crawl-ref/source/dat/des/branches/shoals.des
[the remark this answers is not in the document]
nicolae-carpathia: Changed to have 1 nymph and a 25% chance of a second, with default weight.

crawl-ref/source/dat/des/branches/shoals.des
```diff
@@ -2217,6 +2313,168 @@ swwwwwwwwwwwwwwww
...
+KITEM:  P = leather armour / ring mail / scale mail / chain mail, \
+             long sword / short sword / rapier / dagger, \
+             buckler / nothing w:20
+# Some people like having a little something to read or drink on the beach.
+KITEM:  12 = nothing w:30 / any book / any scroll / any potion
+# Deep elves have to bring the high-strength sunscreen. At least one potion
+# of resistance is guaranteed in the changing room.
+KITEM:  s = potion of resistance ident:all
+# 1/3 chance of rudely walking in on a deep elf changing.
+KMONS:  p = deep elf knight ; nothing / nothing w:20
```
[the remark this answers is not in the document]
nicolae-carpathia: The philosophy of "shouldna been standin there" applies to everyone, not just the player. I've changed the layout so that the changing elf is standing on the same tile as the gear, which hopefully will inspire them to put it back on if they hear something going down. Or not.

crawl-ref/source/dat/des/branches/slime.des
```diff
@@ -653,6 +668,116 @@ MAP
...
+  ''xxxxxxxxxxx'''xxxxxxxxxxx''
+    '''''''''''   '''''''''''
+ENDMAP
+
+NAME:   nicolae_slime_rock_hall_of_fame
+TAGS:   transparent
+DEPTH:  Slime:2-, !Slime:$
+KMONS:  R = rockslime
+SUBST:  { = {[(, - = x_
+NSUBST: D = 1=x / 2=x. / 1=R.. / ., E = 1=x / 2=x. / 1=R.. / ., \
```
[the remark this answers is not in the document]
nicolae-carpathia: I'm not sure I'm going to bother making the changes in this existing file, since slime.des (and spider.des, re: a later comment) don't have any other suggestions for changing them, but I'll keep it in mind for the future.

PR #2635, opened 2022-07-29
Add reduce_bream_redraw option to draw beams fast

Rytisgit:
Add reduce_bream_redraw option to draw beams for animation without many redraws

The reduce_beam_redraw option now decides whether to draw whole beam at once
instead of animating every tile seperately with screen redraws in between. To achieve
this the bolt::draw function is modified to return after setting the tile to the bolt but
before trying to redraw. The redraw is then done only once outside the loop in which
bolt::draw function is called to draw the whole beam at once.

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -798,6 +798,11 @@ void bolt::draw(const coord_def& p, bool force_refresh)
                                              : element_colour(colour);
     view_add_glyph_overlay(p, {glyph, c});
 #endif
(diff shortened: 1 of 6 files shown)


rawlins:
Thanks for the very helpful PR! I've merged this with some tweaks and a renaming of the option in the linked commit above. (I've also made this, at least temporarily, default to true for webtiles in [f98c3da03f33](https://github.com/crawl/crawl/commit/f98c3da03f33); we'll see how that goes...)

PR #2636, opened 2022-07-29
Add chessboard temple for 17 gods

Rytisgit:
Gods are randomised on the chessboard
image


ebering:
Merged with a slight tweak, thanks for the submission!

PR #2646, opened 2022-08-04
Allow harm to stack to 3 levels

Rytisgit:
Harm stacking has been sorely missing from crawl.

### DIFF
--- a/crawl-ref/source/actor.cc
+++ b/crawl-ref/source/actor.cc
@@ -280,10 +280,14 @@ bool actor::reflection(bool items) const
             || wearing_ego(EQ_ALL_ARMOUR, SPARM_REFLECTION));
 }
 
-bool actor::extra_harm(bool items) const
+int actor::extra_harm(bool items) const
 {
-    return items &&
-           (wearing_ego(EQ_CLOAK, SPARM_HARM) || scan_artefacts(ARTP_HARM));
+    if (!items)
+        return 0;
+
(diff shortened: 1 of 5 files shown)


semitonal:
If you're submitting a PR for a gameplay/design change, I recommend including a descriptive commit message and some explanation for the change - see the section on pull requests in docs/develop/contribution-process.md.

Funny though the idea of stacking harm is, it doesn't seem clear to me that it would actually be a good addition, considering how much of a risk it already is at a single level.

ebering:
I had some conversations with Senti about this in the main RL discord; as have gammafunk and PleasingFungus. I think gammafunk even has some design sketches of a Dith overhaul involving harm stacking (has mentioned them but I don't know the details).

@Rytisgit since I'm short on time and this has some controversy, could you elaborate on the gameplay goals of harm stacking (other than big number funni). I'm out of time at the moment but would like to take a look at this in 0.30

Rytisgit:
Sure, the arguments I have are:
* Allows for a strategic option which is high risk high reward
 * Increases artifact effects by making harm more similar to other bad effects on items by stacking, instead of being a dead stat after 1 is used.
 * Harm is interesting as it affects various character styles differently, with melee/summoner being more affected by the downside of taking damage and ranged/mage could be argued to be more affected by benefit of dealing more.

robbiemu:
for what its worth, a random former contributor and returning player here -- this is interesting 👍

PleasingFungus:
Merged with some small changes. Thanks for the PR! :)

PR #2650, opened 2022-08-12
Make greedy explore less greedy

nlavsky:
Autoexplore does a lot of unnecessary item piles revisits now. For example, if you kill a monster that has a dagger of freezing, a plain +0 hat, and an artefact, autoexplore could visit the resulting pile of items up to four times. These (re)visits happen after

1) you kill the monster,
2) you apport the dagger,
3) the monster's corpse decays,
4) a Jiyva jelly eats the hat.

This PR fixes unnecessary revisits 2), 3), and 4).

Also, it reduces the number of unnecessary autotravel interruptions by ignoring piles of useless items during greedy explore.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -835,7 +835,7 @@ explore_greedy_visit += stacks
         Greedy exploration will visit squares with items based on
         these conditions.
 
-        stacks: makes explore_greedy visit stacks even if they don't
+        stacks: makes explore_greedy visit all stacks even if they don't
         necessarily have target auto pickup items
 
         glowing_items: makes explore_greedy visit glowing_items,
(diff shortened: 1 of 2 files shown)


mgdelmonte:
As a Lua user with a very complex autopickup function, I'm delighted by seeing fewer visits to this function.  I actually implement an internal cache on it, because crawl calls it so unnecessarily frequently that it slows exploration to a crawl (pun intended).

However, I worry that the reductions already implemented, plus these, are too much.  Consider the situation where an item is not desirable on the first visit to autopickup (e.g., it's armour and you're not strong enough to wear it) but a few steps later it **is** desirable (e.g., you're now wearing a ring of strength).  The item pile hasn't changed.  **You** have.

My Lua internal cache solves this by creating a character signature (a hash) and re-checking the autopickup function on an item if the hash changes.  For better behaviour, crawl internally could (should) do the same.

PR #2655, opened 2022-08-15
Extract more user-customisation from webserver files

alexjurkiewicz:
1. Add support for loading config overrides from `config.yml`, so server operators don't have to modify `config.py`
2. Move all default game definitions to `games.d/base.yaml`, so server operators can disable all default games with one file rename

### DIFF
--- a/crawl-ref/source/webserver/webtiles/config.py
+++ b/crawl-ref/source/webserver/webtiles/config.py
@@ -85,7 +85,6 @@ defaults = {
     },
     'server_socket_path': None,
     'watch_socket_dirs': False,
-    'use_game_yaml': True,
     'milestone_file': [],
     'status_file_update_rate': 5,
     'lobby_update_rate': 2,
@@ -166,8 +165,7 @@ def load_game_data():
     # TODO: should the `load_games` module be refactored into config?
     global games
     games = get('games', collections.OrderedDict())
-    if get('use_game_yaml', False):
-        games = load_games.load_games(games)
+    games = load_games.load_games(games)
     # TODO: check_games here or in validate?
     if len(games) == 0:
         raise ValueError("No games defined!")
--- a/crawl-ref/source/webserver/webtiles/config.py
+++ b/crawl-ref/source/webserver/webtiles/config.py
@@ -183,7 +183,10 @@ def load_game_data():
     # TODO: should the `load_games` module be refactored into config?
     global games
     games = get('games', collections.OrderedDict())
-    games = load_games.load_games(games)
+    if get('use_game_yaml', False):
+        games = load_games.load_games(games)
+    else:
+        logging.warn("Config setting use_game_yaml is False/unset, won't load game definitions in games.d/")
     # TODO: check_games here or in validate?
     if len(games) == 0:
         raise ValueError("No games defined!")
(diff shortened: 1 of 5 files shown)


crawl-ref/source/webserver/webtiles/config.py
```diff
@@ -166,8 +165,7 @@ def load_game_data():
     # TODO: should the `load_games` module be refactored into config?
     global games
     games = get('games', collections.OrderedDict())
-    if get('use_game_yaml', False):
```
rawlins, line 169: This check or the equiv really needs to be there for dgamelaunch-config setups; they use a `cp -r` from the repo directory to install the webtiles server, and those files are untemplated, so the default behavior installs broken yml files that prevent webtiles startup on dgl-config if this conditional isn't there. Something could be changed in dgl-config of course but since those scripts are never updated automatically I'd rather keep this centralized. See bec1141c92d9.

alexjurkiewicz: Hm. So the default value for it (as defined further up in this file) is True. So the DGL servers have a config.py template that explicitly includes `use_game_yaml = False`? I'll keep the exact config name then for simplicity.

rawlins: hm, I originally defaulted it to `False` (with `True` in the repo config) but then c07941e8bc853 changed it to `True` (which would require an explicit setting); I'm not really sure if that was intentional on my part, but all webtiles servers have probably restarted since then, so I suppose admins must have fixed it. Probably `True` is the "right" default.

rawlins (approved):
This looks good, though we're in a feature freeze right now so I'm afraid I'd rather wait to merge this.

Btw, the original long-term intent was that yml config could be done by directly loading a dict-like that is used instead of a `ConfigModuleWrapper` object, from some function called instead of `init_config_from_module`. Potentially, with multiple config objects around. I think the approach in this PR is fine for now but something like that might be more robust if you keep working on this overall project.

alexjurkiewicz:
Thanks for the help. No worries ok waiting til after release, good luck
with that!

rawlins:
> hm, I originally defaulted it to False (with True in the repo config) but then https://github.com/crawl/crawl/commit/c07941e8bc8534165c9c29d7628e19cb84d1727a changed it to True (which would require an explicit setting); I'm not really sure if that was intentional on my part, but all webtiles servers have probably restarted since then, so I suppose admins must have fixed it. Probably True is the "right" default.

Unfortunately what seems to have happened is this: I originally implemented a way to disable games.d reading by setting this boolean, defaulted it to `False`, accidentally changed it to `True` in the commit above, forgot, and then reimplemented the feature in a second way by allowing setting the directory to `None` in dac63da4db1d. Which makes it hard to guarantee what any particular servers have going on. I'm now thinking that the following might be best for compatibility in the short term: keep the boolean with a complex default: False with `dgl_mode` enabled (which probably does include CPO, so you'd need to set it explicitly), True otherwise. (Of course, "short term" is optimistic.) I can adjust this when I merge.

alexjurkiewicz:
Sure. I'm all for explicit configuration. Implicit is what for us into this mess. Let me know if you need more work from me here

PR #2657, opened 2022-08-15
fix: don't offer Meteorae potion petitions with invisibility

nlavsky:
Since Meteorae can become invisible only via the shadow form, there is no point to offer them petitions with the invisibility potion.


PleasingFungus:
Looks good, thanks!

PR #2665, opened 2022-08-18
feat: new Okawaru overflow

hellmonk3:
Teaches the player what a real friend looks like.


gammafunk (commented):
Seems good. I would actually recommend putting the weapon before the altar, probably one tile to the south. It makes it easier to see what the weapon is straight away without having to `x` or `ctrl-x` and keeps the theme, but that's a minor issue and certainly up to you.

PR #2669, opened 2022-08-18
Add helpful data to the Lua `monster` object

mgdelmonte:
This PR adds some helpful data to the Lua `monster` object:

* `turns` -- the number of turns since you last saw this monster
* `items` -- a list of items the monster is carrying (the same list is described when you examine the monster)
* `range` -- the monster's maximum range with a weapon, spell, or wand; a helpful substitute for interrogating every attack method (weapon, quivering, wand, spells) to see if it can harm you from where it stands.


rawlins:
from the PR description this looks useful, but from the commit sequence (72 commits!) it looks like something might have gone wrong in generating the PR? Also, this PR changes submodules.

PR #2670, opened 2022-08-18
add `crawl.stack` function for Lua debugging

mgdelmonte:
This PR adds the Lua `crawl.stack` function, returning the current Lua stack.  This is sorely needed for debugging player scripts, because crawl exits Lua immediately on any exception and provides almost no information about the cause.


rawlins:
Similar note for this PR, I think something went wrong based on the commit sequence showing in github and the submodule changes.

This looks super useful for debugging, but my immediate reaction is to worry about whether the call stack should really be available on clua on normal builds (will it leak info, present security issues). Possibly this should be debug builds only or wizmode only.

PR #2675, opened 2022-08-19
Adds lua function `view.get_map`, which returns the entire known map.

mgdelmonte:
This is significantly more efficient than querying each cell from lua. It also provides more and better data, but never more than is visible to a normal player.

See function comments for a complete description.


mgdelmonte:
@PleasingFungus is there any interest in adding this?  I'd love to see it but it's been around for a couple of years with no action.

PleasingFungus:
Very sorry for the frankly ridiculous delay! This (and the other clua changes you wrote around the same time) look reasonable, I just haven't had time or energy to do a proper review of them. We're getting close to feature freeze, so I'm hoping to do a pass on PRs, but I really can't promise anything, sorry.

PleasingFungus:
Merged, thanks!

PR #2676, opened 2022-08-19
Add more info to the lua `monster` object

mgdelmonte:
Adds helpful information that's difficult or impossible to extract from the UI:

* `turns` -- the turn number when you last saw this monster (for visible monsters, this will be the current turn; for "remembered" monsters that are not currently visible, this is the turn number on which you saw the monster at that location)
* `range` -- the monster's maximum range with weapon, spell, or wand
* `items` -- the list of items the monster carries


crawl-ref/source/show.cc
```diff
@@ -432,6 +432,8 @@ static void _update_monster(monster* mons)
     {
         mons->ensure_has_client_id();
         monster_info mi(mons);
+        if ( you.can_see(*mons) )
+            mi.turns = you.num_turns;
```
PleasingFungus, line 436: I don't really understand how `turns` is supposed to work, sorry. The PR describes this as "the turn number when you first saw this monster", but monster info is regenerated all the time - it doesn't persist when monsters move, etc.

Is the intent that `turns` is the turn at which you *last* saw a monster *at a given location*? I'm moderately sure that's what is implemented here, though I haven't tested.

crawl-ref/source/mon-util.cc
```diff
@@ -2627,6 +2627,11 @@ mon_spellbook_type get_spellbook(const monster_info &mon)
 vector<mon_spell_slot> get_unique_spells(const monster_info &mi,
                                mon_spell_slot_flags flags)
 {
+    vector<mon_spell_slot> slots;
+    // the rest of this function crashes if mi has no spells
```
PleasingFungus, line 2631: Minor: we can cut the comment. It's fairly intuitive, and very easy to discover if someone tries to delete it :)

PleasingFungus (commented):
Thanks for working on this! I had a question about `turns` before I go further with this change.

crawl-ref/source/show.cc
mgdelmonte: Yes, sorry, the PR description was wrong -- you've got it right above, it's the last time you saw the monster at that location.  I updated the PR description to be more accurate.

PleasingFungus: Thanks! Will take another look at this by early 2025 at the latest.

PleasingFungus:
(Will aim to get this in for 0.32.)

crawl-ref/source/show.cc
PleasingFungus: OK, done right on time. This is frankly beyond any reasonable timeline and beyond any practical apology, so I can only thank you for your contributions and your patience.

PR #2677, opened 2022-08-19
Add `crawl.stack` function to return the current Lua stack

mgdelmonte:
This PR adds a `stack` function to the `crawl` clua object, which provides much better (and I would say necessary) debugging information than currently available.

@rawlins this and my other two PRs are do-overs.  My crawl fork was stale and the git history was a horrid mess.  These and future PRs should be clean.

Re: your earlier comment (lost when I deleted my fork) to the effect that `crawl.stack` is a nice idea but you worry about it leaking info or causing security issues:  If it's called in clua it can only expose the stack in that script, which I would think has no exposure risk.  But of course making it debug-only is one way to address that.

### DIFF
--- a/crawl-ref/source/l-crawl.cc
+++ b/crawl-ref/source/l-crawl.cc
@@ -141,6 +141,31 @@ LUAFN(crawl_dpr)
     return 0;
 }
 
+/*** Returns the current lua stack, for debugging.
+ * @treturn string channel name
+ * @function stack
+ */
+LUAFN(crawl_stack)
+{
+    string r;
+    struct lua_Debug dbg;
+    int i = 0;
+    while (lua_getstack(ls, i++, &dbg) == 1)
+    {
+        lua_getinfo(ls, "lnuS", &dbg);
+        char* file = strrchr(dbg.short_src, '/');
+        if (file == nullptr)
+            file = dbg.short_src;
+        else
+            file++;
+        char buf[1000];
+        sprintf(buf, "%s, function %s, line %d\n", file, dbg.name, dbg.currentline);
+        r += buf;
+    }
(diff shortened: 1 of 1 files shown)


mgdelmonte:
Any chance this can get merged?  Feels like an easy low-impact high-value modification.

gammafunk:
Definitely like the idea of this. To be honest, I've been very lazy about adding a better way to debug qw for a long time and have suffered at times from not having any good kind of stack trace. I've relied on other types of debug printing, which is helpful on its own, but doesn't zero in on the problematic line of lua nearly as quickly as a stack trace would. 

I had been thinking about using a lua debug library that would have to be loaded locally, and didn't even think about doing something on the C++ side. So I think this PR could be a great idea to merge, and I just want to take a look at how it works in practice and look over the implementation carefully. I'll get to it hopefully this week, thanks!

gammafunk:
Merged, thanks for the PR. I'm far from a lua security expert, but I think this implementation is fine from a safety standpoint, as I mentioned in your first commit. I extended `crawl.stack` to work with coroutines, so I can use this with qw, using an approach I found in the standard lua `debug` library.

Some notes for potential future improvements:

- I think the formatting of the stack trace could be improved to conditionally use the `namewhat` field of `lua_Debug` instead of just the `name` field, but I didn't want to mess with that for now. 
- We're unfortunately mangling the `short_src` string by searching for what comes after the last `/`. Ideally we'd not rely on `short_src` and would directly process `source` to remove pathing information, but this has to deal with cases where the source is lua execution of a string and our limited buffer size. 
- It might be good to simply skip activation records that have no line numbers, since those seem to happen mostly from functions making tail calls and I'm not sure they add anything useful to the stack trace.

PR #2681, opened 2022-08-22
Force more for more portal messages

DevJac:
Not all messages related to portals were forcing more. Force more for
the periodic "you hear the sound of" messages as well as the final
notification when the portal is closed.

A potential downside of this is that it will force many messages for each timed portal. I personally like the extra focus on the distance clues because they help me find the portal. We already force more multiple times for anyone stair dancing in and out of the level, or loading a save. This would force a few more messages so long as the timed portal is active.

I also force a final message once the timed portal is closed. I did a search through the code and believe it identified every "disappear" message and covered them all.

Thanks for your consideration. I know this PR was unbidden, so I welcome feedback, discussion, or just rejecting and closing if you want.

### DIFF
--- a/crawl-ref/source/dat/defaults/messages.txt
+++ b/crawl-ref/source/dat/defaults/messages.txt
@@ -122,6 +122,15 @@ force_more_message += sound of rushing water
 force_more_message += oppressive heat about you
 force_more_message += crackle of arcane power
 force_more_message += hear a distant wind
+force_more_message += The portal closes with a thud
+force_more_message += The archway melts and disappears
+force_more_message += The flow of magic halts
+force_more_message += The gateway collapses
+force_more_message += The gate to the bazaar disappears
+force_more_message += The hiss of flowing sand is gone
+force_more_message += The drain falls to bits
+force_more_message += a river of lava pours forth, sealing it permanently
+force_more_message += The walls and floor vibrate strangely for a moment
 
 # Religion
 force_more_message += press .* to convert to Beogh


Oneirical:
I am not a dev, but I want to say that I would not support this change. Portal vaults in the game are supposed to be very dangerous, high risk high reward areas with a few exceptions. On many characters, I will outright ignore these portals if I do not have the right resistances or do not feel confident in my strengths. The messages are outlined in yellow, an easy-to-see colour, and there are a _lot_ of them being printed before the gate finally disappears.

It's easy to add force-mores in the RC file on a player case-by-case basis, but I do not think it should be the default.

DevJac:
@Oneirical I see where your coming from. I'm open to discussion and if you're opinion is the consensus, then I'm happy with it.

We could also force_more for only the final closing message. That would be a middle ground.

Oneirical:
Yes, a force more for only the final message would work just fine without being annoying!

mgdelmonte:
@Oneirical FWIW the lua command `crawl.enable_more(false)` will prevent even forced more from happening.  Admittedly a bit heavy-handed, but it's there if you need it.

DevJac:
I sympathize with @Oneirical, forcing more for every single message is a lot. I often have to ignore 3 or 4 force_mores every time I press 5 until the portal expires. I changed the PR to only force_more when the portal disappears / closes.

I also closed this PR somehow, reopening.

mainiacjoe:
These rc file lines have done this for me for several versions now.  To be fair I usually do try for portals so I'm rarely in the position of wanting to ignore these messages.
```
force_more_message += portcullis
force_more_message += tolling
force_more_message += whistling
force_more_message += melting archway
force_more_message += avalanche of
force_more_message += rusting
force_more_message += magical portal
force_more_message += beating
```

mgdelmonte:
@DevJac examine lm_timed.lua and the `TimedMarker:timeout` code -- which shows a variety of ways in which the portal can disappear.  Easy to confim with testing that when the sewer disappears it says both 

```
TIMED_PORTAL You hear the drain falling apart. 
```
and 10 turns later

```
PLAIN The walls and floor vibrate strangely for a moment. 
```

You'll never see the "falling to bits" message unless you're looking at the drain.

DevJac:
I added a force_more for the catchall message @mgdelmonte mentioned. At this point the PR achieves its goal of giving a force_more whenever a timed portal closes. Whether or not that's something the devs want to do still needs to be decided.

PR #2682, opened 2022-08-22
Fix bugs with lua skill functions

mgdelmonte:
The lua skill functions (`you.skill`, `you.base_skill`, etc.) were all done variously, with some using `str_to_skill_safe` (which returns Fighting if the name can't be matched) and some using `str_to_skill`; and each was done longhand and with different styles of error handling.  This PR factors out the common code and uses `str_to_skill`, which is the only method that makes sense.  The `str_to_skill_safe` method gives totally false results if the wrong skill name is used.

And as a related matter, this PR fixes the changelog, which refers to a skill as Ranged Combat when it is actually called Ranged Weapons.


PleasingFungus:
Merged as 7b7c7d10e174da3468b9f925363d8f248ed660bb, thanks!

PR #2690, opened 2022-08-26
Update spells.txt

zhangjoke:


PleasingFungus:
Looks good. Very big update :) Thanks again!

PR #2691, opened 2022-08-26
Update branches.txt

zhangjoke:


PleasingFungus:
Looks good.

One thing I'm curious about - here and in the monster descriptions, Google Translate thinks that you've translated "divines", as in "the divines Sheza and Sargol", as "priests". When I wrote those descriptions, I used 'divines' to mean 'gods' - they were rivals to the good gods, and committed the forbidden act of having physical bodies on the earth. Is Google Translate getting confused?

Thanks!

PR #2694, opened 2022-08-27
Fix bug in lua `target_weapon` -- cannot select weapon

mgdelmonte:
Crawl currently lacks the ability to speculatively examine the to-hit value of a non-wielded weapon.  The underlying `melee_attack` and `ranged_attack` objects appear to allow weapon selection, but for a player, they always use the currently wielded weapon.  This commit removes the misleading `item` parameter and corrects the function description.

In future it would be nice to support this ability, for parity with spells, items, and thrown missiles -- all of which allow you to select your option and evaluate the success rate with the targetting system, without actually choosing or wielding anything.


PleasingFungus:
Merged, thanks!

PR #2702, opened 2022-09-02
Add `ac`, `ev`, and `sh` functions to the lua `you` object.

mgdelmonte:
The lua `you` object has almost everything else you can see in the HUD, but it's lacking `ac`, `ev`, and `sh`.  This PR adds those stats.


PleasingFungus:
Looks good, thanks! Apologies for the delay.

PR #2707, opened 2022-09-04
Add `view.timed_portal` function to Lua

mgdelmonte:
Following up on #2681, in which @DevJac notes the frustration of dealing with timed portal messages, a nice general solution is to add a `view.timed_portal` function to Lua, which tells you whether there's an active timed portal on the current level.  This is nice because now you don't have to watch the message window quite so closely.

(Ideally, visual cue on the HUD would also be nice, for players who don't use a lot of script support.)

### DIFF
--- a/crawl-ref/source/l-view.cc
+++ b/crawl-ref/source/l-view.cc
@@ -209,7 +209,8 @@ LUAFN(timed_portal)
     for (rectangle_iterator ri(BOUNDARY_BORDER - 1); ri; ++ri)
     {
         const coord_def p = *ri;
-        if (feat_is_portal_entrance(env.grid(p)))
+        dungeon_feature_type feat = env.grid(p);
+        if (feat_is_portal_entrance(feat) && feat != DNGN_ENTER_TROVE && feat != DNGN_ENTER_ZIGGURAT)
         {
             PLUARET(boolean, true);
             return 1;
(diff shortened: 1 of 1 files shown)


crawl-ref/source/l-view.cc
```diff
@@ -199,6 +200,27 @@ LUAFN(view_cell_see_cell)
...
+ * @treturn boolean
+ * @function timed_portal
+ */
+LUAFN(timed_portal)
+{
+    for (rectangle_iterator ri(BOUNDARY_BORDER - 1); ri; ++ri)
+    {
+        const coord_def p = *ri;
+        dungeon_feature_type feat = env.grid(p);
+        if (feat_is_portal_entrance(feat) && feat != DNGN_ENTER_TROVE && feat != DNGN_ENTER_ZIGGURAT)
```
PleasingFungus, line 213: Do we also need to check if there's an untimed bazaar? Would it be better to somehow look for the lua logic that runs these timers, rather than scanning for features?

mgdelmonte: Finding the list of timers would definitely be a better approach, but I looked through the dlua code that controls markers and timed messages, and I couldn't make any sense of it.  And there's nothing that indicates whether a branch or entrance is timed or not.

Unless there's someone who understands the timed message system well enough -- who could offer a suggestion as to how to figure out if there's a timer on the level -- maybe it's best to just close this PR.  Keeping the list of timed portals up-to-date by manually updating this function whenever portals change would be a pain to manage.

PleasingFungus: Closing seems fine for now. We can re-open if we find a better idea for how to manage this.

PR #2712, opened 2022-09-08
Fix infusion cost

pisastrish:
Currently, if you attack while wielding something with a check_unrand_effects effect, the damage bonus from infusion will be applied, but handle_phase_hit will exit before paying the infusion cost (this is especially exploitable with mad mage's maulers). This commit fixes that behavior.

This is a fairly conservative change, I think there is more weirdness with check_unrand_effects causing handle_phase_hit to exit early and return false rather than true, but I don't know what the consequences of changing that would be, and I think this commit fixes the only exploitable bug the current behavior causes.

### DIFF
--- a/crawl-ref/source/melee-attack.cc
+++ b/crawl-ref/source/melee-attack.cc
@@ -527,6 +527,16 @@ bool melee_attack::handle_phase_hit()
              defender_name(true).c_str(),
              attacker->is_player() ? "do" : "does");
     }
+    
+    if (attacker->is_player())
+    {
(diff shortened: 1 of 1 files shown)


kd7uiy:
Agree with this change, recommend merging.

PleasingFungus:
Thanks for catching this! The fix in this PR isn't quite right. It's still possible for infusion to incorrectly be free after this PR is applied, e.g. when beheading a hydra or when the defender is banished.

I've fixed this in 22508d0. Thanks again, and sorry for the delay in review! :)

PR #2718, opened 2022-09-09
adjust a lair entrance and add a new one

hellmonk3:
Place a fire crab instead of a fire dragon in minmay_lair_entry_lava, since that is a) native to the lair monster set and b) a little less difficult to deal with on D:8-11. Add a new icy themed entrance that places a rime drake, a few wolves / polar bears, and some ice beasts.


crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -295,7 +295,7 @@ NAME:   minmay_lair_entry_lava
 TAGS:   lair_entry no_monster_gen transparent
 ORIENT: float
 WEIGHT: 1
```
gammafunk, line 297: Could consider upping the weight to 2 or 4 now that it's a significantly easier fight. Either way, the end of an era...

crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -478,6 +478,39 @@ MAP
 .............
 ENDMAP
 
+NAME:   hellmonk_lair_entry_icy
+TAGS:   lair_entry transparent
+ORIENT: float
+MONS:   ice beast, wolf / polar bear w:3, rime drake
+NSUBST: 3 = 1:3 / * = 2.., . = 1:2 / 4 = 2... / *:.
```
gammafunk, line 485: Usual suggestion here of reduced NSUBST form
```
NSUBST: 3 = 3 / 2.., . = 2 / 4=2... / .
```
and the one two lines below can have the `*` term simplified as well.

gammafunk (approved):
Looks good.

PR #2721, opened 2022-09-11
New background: Reaver

PleasingFungus:
Another twist on the classic 'magic plus melee' formula, Reavers are warriors who have some conjurations to help em out in rough spots. These spells are limited enough in various ways to make them difficult or impractical to use as a primary offense, but they're powerful enough to make the difference in tough fights - if you skill for them!

Starting spells:
- Kiss of Death (L1 Conj/Necro): Does high draining damage to an adjacent foe, starting at 2d5 and rising to 2d13 at max (25) power. However, also inflicts heavy drain on the player, equivalent to a use of Ru's Apocalypse.
- Unstable Shard (L2 Conj/Tmut/Earth): A bit like a super stone arrow (higher damage, more accurate, longer range, bounces off walls for trick shots), with one big downside: launching it sprays magical contamination all over the caster, taking you to grey contam in one cast and yellow in two.
- Hailstorm (L3 Conj/Ice): This one isn't new. You know this one. It has a miss chance now, and it's one spell level lower than it was. Ice Elementalists lose it, gaining back Summon Ice Beast in exchange.

Reavers are intended to be a flexible background - shifting more toward melee,
casting, or neither (remaining a hybrid) as the game goes on.

Partially reverts https://github.com/crawl/crawl/commit/1cc687962fc2b680d43a89788ed88c7025514f91 (2011).


PleasingFungus:
Merged.

PR #2729, opened 2022-09-15
hermit pendant guarded unrand vault

hellmonk3:


crawl-ref/source/dat/des/variable/mini_monsters.des
```diff
@@ -8552,6 +8552,40 @@ a'a'd'a'a
   a'a'a
 ENDMAP
 
+NAME:   hellmonk_guarded_unrand_hermit
+TAGS:   no_monster_gen
+DEPTH:  D:3-8
+WEIGHT: 3
+KFEAT:  _ = altar_ecumenical
+KITEM:  _ = hermit's pendant no_pickup
+NSUBST: E = 1:. / *=xxx., H = 1:. / *=xxx., I = 1:. / *=xxx., J = 1:. / *=xxx.
```
gammafunk, line 8561: Recommend syntax like:
```
NSUBST: E = . / xxx., H = . / xxx., I = . / xxx., J = . / xxx.
```
It's better to use those implicit NSUBST assignments that effectively do `1=` for non-final entries and `*=` for the final entry.

gammafunk: Another minor point, `I` is actually taken by the orcish idol feature, so ideally you'd use a different glyph there. We should probably undefine that one as a default since it no longer has any special function.

gammafunk (approved):
Looks good, I didn't generate this in-game, which I should have, but it's simple and I assume you've done a couple times. If it loads fine for you, seems ready to merge.

PR #2735, opened 2022-09-18
Clean up spell power code

Monkooky:
calc_spell_power previously took 5 parameters; 4 of which were unneeded. 
apply_intel, fail_rate_check, and scale were used exclusively in raw_spell_fail to make calc_spell_power function as
_skill_power * scale / 100. 
This has been corrected by replacing the call with _skill_power * scale / 100.

cap_power was used in three calls:
spell failure, which no longer uses check_spell_power 
target_desc, which now applies the cap twice
and one of the two calls for spell_range passed uncapped spell power.

The only 'material' change is that the impact of schools on failure chance no longer depends on spell power cap. However, this does not have an impact unless you have 75 skill, and probably was not intended in the first place.

### DIFF
--- a/crawl-ref/source/spl-cast.cc
+++ b/crawl-ref/source/spl-cast.cc
@@ -471,7 +471,7 @@ int calc_spell_power(spell_type spell)
     if (you.divine_exegesis)
         power += you.skill(SK_INVOCATIONS, 300);
 
-   
+
     power = (power * you.intel()) / 10;
 
     // [dshaligram] Enhancers don't affect fail rates any more, only spell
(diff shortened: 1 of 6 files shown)


crawl-ref/source/spl-cast.cc
```diff
@@ -2689,7 +2671,9 @@ static int _spell_power(spell_type spell, bool evoked)
     if (cap == 0)
         return -1;
     const int pow = evoked ? wand_power()
-                           : calc_spell_power(spell, true, false, false);
+                           : calc_spell_power(spell);
+    //TODO: currently applies cap twice to nonwand spells.
+    //Double check that we can safely put cap in wand_power
```
PleasingFungus, line 2676: Should be fine, yeah. `wand_power()` is only used in one other place, `zap_wand`, where it's immediately passed to `your_spells`. All wand spells (listed in `spl-book.cc`) have non-zero power caps, and the only one with <200 power cap is throw flame, which is capped at 50. But so's the corresponding zap in zap-data.h, so I really doubt it matters.

PleasingFungus (commented):
Thanks for this! Nice cleanup.

PR #2736, opened 2022-09-18
Fix Scorch description to note that damage is reduced by armour.

mgdelmonte:
Scorch is misleading as it is a pure fire spell that is actually reduced by armour.  The wiki also describes it as doing "unavoidable fire damage" but in fact this is not the case.  Scorch damage is reduced by both armour and fire resistance.  Probably the wiki should be updated as well.


PleasingFungus:
Hi, thanks for reporting!

All damage is reduced by armour unless noted otherwise. Look at other spell descriptions to see examples.

Cheers,
PF

PR #2737, opened 2022-09-19
increase player regen and decrease max hp

hellmonk3:
Based on discord conversation with elliptic. Low base regen makes resting take forever, which feels pretty bad. To improve quality of life, increase the base regen rate substantially, and to prevent that from being a large player buff for long duration combat situations (and/or make it easier to spend a long time doing tedious things), reduce max hp and zot duration to compensate. Probably needs savecompat and should also do something about piety decay either as part of this branch or immediately afterward.

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -1164,10 +1164,7 @@ int player_regen()
     // Note: if some condition can set rr = 0, can't be rested off, and
     // would allow travel, please update is_sufficiently_rested.
 
-    int rr = you.hp_max / 3;
-
-    if (rr > 20)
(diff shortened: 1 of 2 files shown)


PleasingFungus:
Very unclear to me whether players hitting the zot clock during 'normal play' is caused by HP or MP regen. If the latter, this PR shouldn't touch zot times.

Did you have any thoughts about what we should do with AoOs? I sort of doubt they do anything now in this system - do you want to cut them, rework them to be more aggressive..?

kd7uiy:
My biggest concern with this is those species which do not regenerate during combat. This would effectively nerf those species, those species should come with a similar increase of health to counteract.

Issue #2503, opened 2022-03-24
Sp is allowed to cast swift if theyre in shallow water, does nothing


ardl:
**PLATFORM and OS or BROWSER:**
local 0.28.0

**DESCRIPTION:**
Swift normally doesnt effect spriggans and they arent allowed to cast it. if a sp stands in shallow water theyre allowed to cast swiftness, which then doesnt do anything.

semitonal:
Fixed in 6573418d013d1d29.

PR #2739, opened 2022-09-20
disregard water for determining if you are too fast for swiftness

Monkooky:
Fixes https://github.com/crawl/crawl/issues/2503

This is kinda hacky, but the only alternative I saw was passing a parameter to player_movement_speed so it would conditionally ignore water effects.


PleasingFungus:
> The only alternative I saw was passing a parameter to player_movement_speed so it would conditionally ignore water effects.

Agree! Just think that's a lot better than the hacky approach :)

Thanks for spotting & diagnosing this!

PR #2743, opened 2022-09-21
Documents insubstantial in the monster x-V view

Monkooky:
As per title.
Also documents Sticky Flame's inability to stick to insubstantial monsters.

This is a needed change because currently insubstantiality of monsters is highly spoilery, and I still need to check the wiki on occasion to see if a monster will be affected by e.g. roots.


PleasingFungus:
Looks good. I'll poke at improving targeters.

PR #2744, opened 2022-09-22
Update mutation.cc so ashenzari correctly blocks beaks

Monkooky:
image
I was robbed, and want no others to suffer as I have


Monkooky:
Right now blocks all beak levels!

PleasingFungus:
Nice catch, thanks! Sorry for the delay in review.

PR #2745, opened 2022-09-23
feat: Add new unrand gloves, the ghost crab claws

pdpol:
This is a proof of concept of an idea I pitched in the discord channel recently and got some feedback on. These gloves prevent a wearer from wielding a weapon or wearing a shield but in return, give boosts to int and MP, and have a 1/3 chance on successfully casting a spell to place a spectral mist centered on the player.

**Current implementation**
* I took a suggestion from @PleasingFungus to target early-game casters in terms of making this unrand appealing
* Chance to trigger, cloud radius, placement, and power are all flat and uninformed choices by me, really just to try to have something testable to get more feedback on if this approach (vs other ideas below) is a good path to keep going down, and ideas to make it feel more balanced/fun.

**Other ideas & open questions**
* The original idea for this unrand was that it would be oriented towards non-tmut UC users, but it quickly became apparent that the most likely candidates like trolls and some DS wouldn't even be able to wear the gloves.
* @hellmonk3 suggested making them the "+0 gloves of spectral mist" which would remove the weapon/shield restriction and probably just make these a simpler and very useful boon to any meleedude's kit, which I like, but I do still think the idea of wearing crab claws that make it too hard to hold things is funny.
* After working on this for a bit I thought maybe this could be fun as a lvl 4/5 tmut/necromancy spell that melds hands?

* Testing on a DrCj quickly made me realize this won't trigger on channeling. So for instance it might be annoying to an early Cj that channeling searing ray doesn't get you more cloud chances, but magic dart spam does. Couple questions on this point I have are:
  * I see potential for abuse with players spamming something like casting a form to get clouds going, although there are very practical drawbacks to doing this in combat, so maybe it self-corrects?
  * Should this trigger on channel?

Planning to do a little more testing soon and have some gifs or vids to share but messing around in early D has been fun so far. I think it needs a little more variance. Right now I was thinking about having the cloud placement vary, for instance I thought it would be cool if the cloud targeted the nearest monster instead of the player occasionally.
Screen Shot 2022-09-23 at 3 01 18 PM


sdynet:
That's interesting.

kd7uiy:
Interesting idea. It might work better if it also gives some kind of a UC/AC bonus, and maybe some natural shielding. But that starts to sounds like a weapon, which might work better overall for this.

Imagine a two-handed weapon that somehow used UC? Hmmm...

PR #2750, opened 2022-09-26
Fix wizlab wizards "come into view" by giving them M_UNIQUE

### LINKED ISSUE #2738: The headmaster in Yara's Duelist Academy has an indefinite article before the name when they show up
Local, Windows 10

0.29.1 stable

**DESCRIPTION:**
Loaded up a local game in wizmode, went into Yara's Duelist Academy, opened the door to the room with the headmaster, log says: "A naga mage and a Xathriw come into view."
Since the headmaster's name is a proper noun I figure this is not intentional.

**EXTRA FILES:**
Here's a screenshot.
crawl_2022-09-20_00-24-39

If there's something else that's necessary or something I can improve, just let me know.

SamanthaGobbo:
Wizlab wizards did not have the M_UNIQUE flag, which caused them to get displayed in "come into view" messsages as "a Name" instead of "Name". this only happened when they came into view along with other monsters.

This fixes the issue by giving them the M_UNIQUE flag.

Resolves #2738

image
(previous this would have said "A Xowim, a necromancer and an ogre mage come into view")

I tried to find out if there was a specific reason these 3 uniques did not already have this flag to begin with (which this commit might break), but could not find one other than that they did not have it either before 567207be75278f5cb21d86088f46c684a7274fbd, which is the commit which added them to mon-data.h


kd7uiy (approved):

rawlins:
Thanks for the PR! I'm not sure why this PR languished, but I've done a quick audit of where `mons_is_unique` is called and it looks like this shouldn't have unintended consequences. I believe those three weren't treated as uniques because they have randomized names and don't resemble normal uniques in a few other ways, but this commit hopefully won't interfere with that. (I do wonder if the check in `monster::is_named` might be wrong for these three types? It looks to me like they should have already been covered.)

Issue #2740, opened 2022-09-20
Dream sheep attack type


T-Minutiae:
Version: Windows local tiles 0.30-a0-134-gb40fab5c76

AT_HEADBUTT would make infinitely more sense than AT_BITE. It is technically possible to get bitten by a sheep but due to the fact that they have no teeth in the front part of the upper jaw and the angle of their heads it would be an extremely awkward means of attack. They'd use their incredibly hard heads to smash you instead.

And yes, I am fully aware that this is a fantasy video game. With murderous sheep that can use PSI. But we have to draw the line _somewhere_.

PleasingFungus:
Just to clarify, you're aware that AT_HEATBUTT does not exist?

PleasingFungus:
Ah, I see you tried to add it in #2761. Did you compile the game to test? Your change won't work as written - you also need to update `mon_attack_name`. (And, as kd7uiy noted, you should make that change in a separate PR.)

Sorry for the belated response on this. :)

Issue #2582, opened 2022-06-25
Shallow water has the tile of swamp wall


dillytin:
![image](https://user-images.githubusercontent.com/65288602/175771742-d1d0aad5-c6d4-4aaf-917a-ab51d85e52b6.png)

![image](https://user-images.githubusercontent.com/65288602/175771736-a3aee2b2-2684-4439-9b86-d140d384b876.png)

https://webzook.net/soup/morgue/trunk/dilly/morgue-dilly-20220625-074652.txt

Swamp1:cheibrodos_swamp_tower_of_moss

ebering:
This seems to be a bug with the interaction between the vault `TILE:` directive and the `ruin` tag, the correct fix is going to be a modification to tile handling in `_ruin_level` in dungeon.cc

regret-index:
Fixed in c73acac.

PR #2761, opened 2022-10-03
modify cheibrodos_swamp_tower_of_moss to replace wall with shallow_water

CorbinWeir:
Addresses issues #2740 and 2582

Fix tile texture
Change dream_sheep AT_BITE to AT_HEADBUTT


kd7uiy:
Typically you want to separate unrelated changes in to separate PRs. In this instance, the vault modification is a separate change from the sheep issue, you might want to divide those in to two separate PRs.

ebering:
Thank you for the submission. As mentioned above please split commits so that they're on a single topic. The AT_HEADBUTT (submitted in a separate PR) was not completely implemented.

I'm going to close this one because this is not the correct fix to issue #2582 I will add a comment on that issue as to its likely source

PR #2763, opened 2022-10-03
Stop momentum strike from using vhi's checks

Monkooky:
momentum strike case was missing a break;, break;ing it.


kd7uiy:
Reviewed change and looks like a minor, but needed, change. The change looks good to me.

PleasingFungus:
Nice catch, thanks!

PR #2770, opened 2022-10-05
Fixes info screen for draconians to show they can fly based on the type of draconian

### LINKED ISSUE #2752: Missing line about flight for black draconian knights
webtiles on chrome

0.30-a0-165-g1891117ce2

**DESCRIPTION:**
Black draconians and black draconian knights can fly, but only plain black draconians state this in their x v description.

**EXTRA FILES:**
knightflight

kd7uiy:
This small change will allow draconians with a base type that can fly to fly, and specifically to show it correctly in the information screen. Draconian's base type can't be determined purely from the mons.type, but instead should use draconian subspecies if the species is a draconian.

After toying with this a while, I have changed it in `mons_base_type` to return the draconian subspecies for draconians. 

This likely will cover many other similar bugs in the future, and in fact I can see several places in the code that an improved `mons_base_type` to cover draconians would result in less code. I did fix one of these, and will look for others when I get a chance. I would appreciate any assistance in identifying any negative consequences with those with more experience to the code.

Fixes #2752

### DIFF
--- a/crawl-ref/CREDITS.txt
+++ b/crawl-ref/CREDITS.txt
@@ -409,6 +409,7 @@
     Rob Young
     Zannick
     Zooko
+    Roadster Tracker
 
 The Dungeon Crawl Stone Soup development team is:
     advil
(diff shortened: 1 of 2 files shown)


rawlins (commented):
I'm afraid I'm going to fix this in a different way. I think this commit appears to work because the change to `monster_inherently_flies` has a bug, sorry (see the specific note there).

In general I would be wary of changing the semantics of a function like `mons_base_type`, because this is used in a *lot* of places without any assumption that the monster type will be changed by it unless the monster is undead. In a quick audit I spotted many such calls, which caused me to go with a much simpler fix (I'll link to this PR in the fix commit).

PR #2773, opened 2022-10-06
Adding two new rocket themed arrival vaults, zombie monster vaults

kd7uiy:
Two vaults of different sizes, roughly rocket shaped, complete with smoke and fire (lava) from the engine!


kd7uiy:
Looking at the crash report, it isn't with anything I touched, not even remotely. My best guess is that it changed the seed effectively in a way that the new random seed has an issue that wasn't there. The issue has something to do with demonspawn mutations, and crashes with mut_species.lua line 31.

regret-index:
The arrival vaults look solid. The trees feel a little silly (I guess the rockets are landing in from a forest outside?) and I might want to randomize that to sometimes place plants or rock instead, but they should be fine with maybe a little less smoke inside the rockets.

I'm not much of a fan of the Shoals boat vault. There can be any number of attempted theme justifications for why monsters might be placed anywhere, but if there's no real connection between a given branch and the vault's guests, then one is mostly just diluting both branch and monster identities to slowly put practically anything anywhere. I'd be much more supportive if this was reversed to put Shoals enemies (and maybe some other appropriate stuff, or with some random other collection of passengers?) into D and Depths rather than some humans going into Shoals- D and Depths already support a whole bunch of random themes in comparison.

The zombie capsules are a bit weird. They don't need "extra" tags with these many monsters and with any rewards, since minivaults can be plenty mini already. Crypt might already have a lot of harmless zombies around for some reason, but I'm not sure there's much interest in throwing around a lot of slow zombies that have reasonable odds to just be freed away from any parseable context by a mummy or wraith. Probably would be fine with a less muddled mix.

If you're fine with it, I can go through these and fix them up towards a merge-able state.

kd7uiy:
That is fine by me.

PR #2779, opened 2022-10-14
Lower the Dispersal spell's level

sdynet:
It's a good spell, but the level is rather high, so it's ambiguous to use in early-medium games. Therefore, I want to lower the level by one. However, considering that it has become easier to use, I will reduce the range by one.


regret-index:
From a design standpoint, Dispersal's level location is very awkward to shift. Moving it from 6 to 5 brings it much closer to the overlaps of Teleport Other at level 3 (willpower check to teleport) and Iskenderun's Mystic Blast at 4 (moving nearby monsters away). Making the spell weaker and more accessible thus only emphasizes the overlap between it and these earlier spells. It also hurts the school's level distribution even further, by widening the gap between it and Disjunction at level 8 if one's not picking up the summoning spells at levels 5 and 7.

Dispersal itself is admittedly at an awkward use point as currently stands, due to the high investment needed for a level 6 spell. There are currently rough dev plans to add more mid-level spells with less overlapping designs to Translocations and move Manifold Assault up higher, which will incidentally improve Dispersal's situation by providing more reasons and goals for getting higher translocations. That itself will probably do plenty for the spell.

PR #2783, opened 2022-10-16
fix: make Poisonous Vapours targeter ignore rPois enemies

nlavsky:
Otherwise the targeter always starts on the closest monster, even if the monster is resistant or immune to poison.


semitonal:
Merged, thanks for the fix!

PR #2784, opened 2022-10-17
fix: adjust monser placement in wad_wood_setup.

pdpol:
An absdepth conditional was placing a harder monster set at shallower depths. This flips the condition, and also makes the transition point L:3 instead of L:4, which I think seems in line with the recent reduction to 5 floors.


pdpol:
I have to imagine the build failure is transient:

```E: Failed to fetch http://apt.llvm.org/focal/pool/main/l/llvm-toolchain-15/lldb-15_15.0.3~++20221017083135+d35bc70e8251-1~exp1~20221017083227.78_amd64.deb  Cannot initiate the connection to apt.llvm.org:80 (2a04:4e42:77::561). - connect (101: Network is unreachable) Could not connect to apt.llvm.org:80 (146.75.30.49), connection timed out [IP: 146.75.30.49 80]```

PR #2794, opened 2022-10-28
4 lair minivaults

pdpol:
I'm hoping to add some vaults to some of the branches that don't have as many as D and Depths, and figured I'd start here. I'd like to do some more lair stuff but kind of felt like this would be a good stopping point to get some feedback - both because this is my first time trying to do anything with monster and item placement, and because I know that cabin in the woods and maybe dawg pound are a stretch in terms of theme/balance - but hopefully in the same way that many great vaults in the game are as well!

### DIFF
--- a/crawl-ref/source/dat/des/branches/lair.des
+++ b/crawl-ref/source/dat/des/branches/lair.des
@@ -1883,18 +1883,17 @@ ENDMAP
 
 # Someone hasn't taken care of their garden, but occasionally it takes care of itself.
 NAME:    pdpol_lair_neglected_garden
-TAGS:    extra decor transparent
+TAGS:    extra decor transparent no_trap_gen
 DEPTH:   Lair
-SHUFFLE: ac"`/pbbp
-SUBST:   ab = B, cp = P
+SHUFFLE: ac"` / pbbp
+SUBST:   ab = D, cp = P
 : if crawl.one_chance_in(5) then
-TAGS:    no_trap_gen
 MONS:    plant, bush
-SUBST:   V = T, B = 2, P = 1
+SUBST:   V = T, D = 2, P = 1
 FTILE:   ' = floor_dirt, 12`" = floor_grass
 :else
 MONS:    withered plant / plant w:1, fungus
-SUBST:   P = 1, B = 1:6 2
+SUBST:   P = 1, D = 1:6 2
 :end
 SUBST:   '`" = .
 MAP
@@ -1902,9 +1901,9 @@ MAP
 ..`P"'"P`..
 .`Pac'caP`.
 .``c'''c``.
-.`"''B''"`.
-@`"'BVB'"`@
-.`"''B''"`.
+.`"''D''"`.
+@`"'DVD'"`@
+.`"''D''"`.
 .``c'''c``.
 .`Pac'caP`.
 ..`P"'"P`..
@@ -2778,26 +2777,26 @@ MAP
   -------------------
 ENDMAP
 
-NAME: pdpol_lair_dawg_pound
-TAGS: no_monster_gen
+NAME:   pdpol_lair_dawg_pound
+TAGS:   no_monster_gen
 DEPTH:  Lair
 KMONS:  1 = wolf / warg w:5
 KMONS:  2 = hell hound / raiju
-KITEM: 2 = *
-SUBST: _ = x _:3, S = S:1 . %:2
+KITEM:  2 = *
+SUBST:  _ = x _:3, S = S:1 . %:2
 NSUBST: s = 1:S / *= . S:1, D = 2:2 / 4= 1 2:1 / *= . 1:2
 : dgn.delayed_decay(_G, 'S', "felid skeleton")
-CLEAR: _
+CLEAR:  _
 MAP
             ________
-          __xxxxxxx__
-          _xx.SS..x__
-      _____x..SS..x__
-    ___xxxxx....DDx__
- ____xxx..SS..DDxxx__
- _xxxx.....SxDxxxxx__
+          __xxxxxxx_
+          _xx.SS..x_
+      _____x..SS..x_
+    ___xxxxx....DDx_
+ ____xxx..SS..DDxxx_
+ _xxxx.....SxDxxxxx_
 _xxSS....x..xxxx____
-_xDSS..Dxxssxxxx__
+_xDSS..Dxxssxxxx_
 _xS..Dxxxssxx____
 _xxDxxxxssxxx_
 __xxxxx..xx_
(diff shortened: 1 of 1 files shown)


crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -2748,6 +2778,107 @@ MAP
   -------------------
 ENDMAP
 
+NAME:   pdpol_lair_dawg_pound
+TAGS:   no_monster_gen
+DEPTH:  Lair
+KMONS:  1 = wolf / warg w:5
+KMONS:  2 = hell hound / raiju
```
hellmonk3, line 2785: This vault steps on the dogs lair ending a bit imo. Maybe downweighting the raiju is enough but it feels like it should get something extra to make it feel more distinctive.

crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -2748,6 +2778,107 @@ MAP
...
+_xxSS....x..xxxx____
+_xDSS..Dxxssxxxx__
+_xS..Dxxxssxx____
+_xxDxxxxssxxx_
+__xxxxx..xx_
+    xx@@xx
+ENDMAP
+
+# Dead by dawn!
+NAME:    pdpol_lair_cabin_in_the_woods
```
hellmonk3, line 2808: Really like what this vault is going for. However, it's got a clear runed door that does not reveal the part of the vault that's actually dangerous - this is a pretty bad spoiler, especially given the monsters you're placing. You could fix this either by putting the runed door on the inside room that gets the scary monster (using some sufficiently tricky substs or shuffles) or by putting windows into every room on the outside (and adding enough space to walk all the way around it).

crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -2748,6 +2778,107 @@ MAP
...
+TAGS:    no_item_gen no_monster_gen
+DEPTH:   Lair
+# Trying to get some good horror move trope representation
+KMONS:   M = Rupert / reaper / vampire knight / phantasmal warrior / tentacled starspawn
```
hellmonk3, line 2812: This will place no monster if it rolls Rupert and he's already generated elsewhere. You can do M = Rupert, ghoul / reaper / etc to place a backup monster if he has already spawned (feel free to use a different monster of course). I'd also recommend using a vampire mage over a vampire knight, since that one stands out as more difficult than the other non-Rupert options imo.

crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -2748,6 +2778,107 @@ MAP
...
+   ________________________
+    ______________________
+ENDMAP
+
+NAME:   pdpol_lair_cavern_descent
+TAGS:   no_pool_fixup
+DEPTH:  Lair
+KMONS:  1 = hydra / boulder beetle / hornet / rime drake
+KMONS:  2 = cane toad / black mamba / komodo dragon / polar bear
+KMONS:  3 = water moccasin / scorpion / black bear / wolf / bullfrog
```
hellmonk3, line 2853: Scorpion and bullfrog seem kind of weak for this monster set.

crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -2748,6 +2778,107 @@ MAP
...
+KMONS:  4 = basilisk / torpor snail w:5 / manticore w:3
+KMONS:  5 = electric eel
+SUBST:  u = wW
+NSUBST: ' = 2:1 / 6= 1:2 2:3 3 / .
+NSUBST: W = 3:4
+NSUBST: w = 2:5
```
hellmonk3, line 2859: Recommend some amount of randomization on the number of monsters being placed here.

crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -2748,6 +2778,107 @@ MAP
...
+  xx'''x'''xx
+ xx''x'''x''xx
+ x''xxw'wxx''x
+xx'xxwwwwwx''x
+x''xwuuuuw'''x
+xx'xwwuWWwxx'x
+xx'''wuWWWx''x
+xx..wwuWWW%%xx
+x..xxuuW**Wxx
+xx..xxxw**xx
```
hellmonk3, line 2877: It's possible for a few of the 'u' glyphs to end up as disconnected shallow water.

hellmonk3 (commented):
Did a quick runthrough, here's some stuff I noticed. Guessing gammafunk or someone else will look more thoroughly later.

crawl-ref/source/dat/des/branches/lair.des
pdpol: Ahh yeah I thought that might be an issue... I'm definitely open to ideas for something extra, and honestly I'm not very attached to the dog theme either, I'm mostly trying to figure out fun ways to get some encounters with non-lair beast type monsters that aren't lair ends.

So a different kind of direction - do you think it would be too out there to have a hog themed minivault with a hell hog boss or two? That's the kind of thing I'm thinking, but I feel a bit out of my element when it comes to balance and reaching outside of the branch-appropriate monster set.

crawl-ref/source/dat/des/branches/lair.des
pdpol: Yeah totally fair, I thought it might be permissible because of a small handful of other spoilery ones (I think there's one where you can't quite see a boulder beetle in the back that I like) but it's probably pretty mean with these monsters and the size of this thing.

I'll try out both those suggestions, I wouldn't have thought to do the runed door on just the monster's room but that could be a fun way to create the horror sense I was going for by walking around a creepy cabin before you see the thing. But if I'm being real people are probably going to otab right through it anyway lol.

crawl-ref/source/dat/des/branches/lair.des
pdpol: Oh whoops that's good to know. So with that comma-separated syntax, will it always try to place Rupert first? Or would there still be a chance to place the other monsters even if Rupert wasn't placed elsewhere? Could I do something like `M = Rupert / ghoul / reaper / etc, ghoul / reaper / etc` so that there's an equal chance but then it handles the Rupert picked but unavailable case?

pdpol: Also will go ahead and change that vampire knight.

crawl-ref/source/dat/des/branches/lair.des
pdpol: Yeah, I'll just get rid of them and put wyvern in.

crawl-ref/source/dat/des/branches/lair.des
pdpol: Any thoughts on the loot distribution for this one (and the rest in general)? That's an area where I feel mostly like I don't know enough to get too creative so I kept it pretty generic.

crawl-ref/source/dat/des/branches/lair.des
pdpol: Oh whoops I'll try to clean that up.

crawl-ref/source/dat/des/branches/lair.des
hellmonk3: I think it actually should use a glyph to subst Rupert or his backup in with some chance and then do the rest of the kmons line normally. There's probably some one line syntax that works but I don't remember what it is, you'd have to test it. Definitely doesn't work the way I originally wrote it though, sorry about that.

crawl-ref/source/dat/des/branches/lair.des
hellmonk3: Single hell hog is maybe ok? They're speed 14 and have a 3d20 fireball, so it stretches the limit of what you can put in lair (compare fire dragon which has 3d24 breath and much better defenses, but is speed 10). This would let you add the fabled holy swine though...

crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -2748,6 +2778,107 @@ MAP
...
+tt.ttt.F"n`````b``gMMMMctt__
+ttT.tt...c`````b``gMMMMctt__
+ tTt.tt..ccccccccccccccctt__
+ @..ttttTttttttttttttttttt__
+   @ttttttttttttttttttttt__
+   ________________________
+    ______________________
+ENDMAP
+
+NAME:   pdpol_lair_cavern_descent
```
gammafunk, line 2848: Nice little monster vault that offers two paths to the same loot, with the "easier" path requiring flight or ambitiousness. Good type of vault to have in your first submission set, I approve!

gammafunk (approved):
I reviewed the two vaults not commented on already by hellmonk. Looks like a good first submission. I can follow up and merge in a couple of days if there are no outstanding issues and hellmonk doesn't get to this first. Might end up making a subsequent commit with some tweaks, but shouldn't be anything too huge. Thanks for the submission.

pdpol:
Thanks for the review @gammafunk! I'm hoping to finish up addressing the feedback on `pdpol_lair_cabin_in_the_woods` today, time pending, so hopefully I can get that in before one of you gets to merging.

As for replacing the dog vault with a hog one, I'll see what I can do but not sure if I'll get to it so soon. Is it helpful for you all if I keep changes for each vault in their own commits? Or is there anything I could remove (like the dog vault entirely) that would make squashing not a problem?

crawl-ref/source/dat/des/branches/lair.des
pdpol: I think I managed to find a way to do this just building `kmons` in a lua chunk with `you.uniques("Rupert")` to determine whether to throw him in or not, but let me know if that isn't the best way to do it.

crawl-ref/source/dat/des/branches/lair.des
pdpol: I think I got the runed doors just around the monster room working pretty nicely - it's kind of a mindbender (at least for me) but I think the result is good. One of the hard things about it was that originally I tried to do a line like `SHUFFLE: Maed/?ghf/Hgfh` to get all three rooms to swap in one line but I realized that can't work because we might end up with no `a` glyph and then an isolated room.

gammafunk:
@pdpol We will rebase and squash your commits so that it's likely one commit per vault. So feel free to make whatever type of follow-up commits are most convenient. Generally one follow up commit per vault would probably be best, or even one commit like "adjust lair vaults based on PR feedback" would be good. Many tiny commits all tweaking the same vault usually get squashed to avoid commit spam if they happen close together, especially when they have a logical grouping like a PR. 

We can merge the other vaults now and keep pr open for eventual hog vault, or ignore that vault and close pr if you prefer. Just let us know, thanks!

crawl-ref/source/dat/des/branches/lair.des
```diff
@@ -2748,6 +2778,107 @@ MAP
...
+DEPTH:  Lair
+KMONS:  1 = hydra / boulder beetle / hornet / rime drake
+KMONS:  2 = cane toad / black mamba / komodo dragon / polar bear
+KMONS:  3 = water moccasin / black bear / wolf / wyvern
+KMONS:  4 = basilisk / torpor snail w:5 / manticore w:3
+KMONS:  5 = electric eel
+SUBST:  u = wW
+NSUBST: ' = 2:1 / 6= 1:3 2:5 3 / .
+NSUBST: W = 2:4 / 4= 4:5 W
+NSUBST: w = 2:5 / *= 5:1 w:15
```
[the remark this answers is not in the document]
pdpol: I tried this out and the results surprised me:
Screen Shot 2022-12-11 at 10 35 36 PM

It seems like it might be trying to do the "place exactly 5 1 glyphs" thing?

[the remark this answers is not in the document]
gammafunk: @pdpol Yes, I seem to recall there being a bug with the implicit `*=` for the final term fails if there are spaces in the term. This is something that needs to be looked into more and hopefully fixed, but it seems you will need the explicit `* = ` portion like you originally had, for now.

gammafunk:
@pdpol Merged the three vaults we discussed with some changes in 42e7360. Most of the changes were simplifications to syntax or reglyphing or other cleanups, but the cavern descent vault did see a reduction in monster count. It looks ok to me, but if you feel that it's a bit too weak for what you were going for or would like a somewhat different distribution of monsters, feel free to make a follow-up commit of your own. In general it's good to not shove too many monsters into a small vault unless that's integral to the theme or level of challenge required.

The weight simplifications I made were just to use more common fraction values like 1/2, 1/3, 1/4, etc when those are sufficient, and those are also nice because they can be expressed with patterns like `DE`, `DEE`, and `DEEE` and mixed into patterns with larger number of glyphs for fractions  with a (reduced) denominator no higher than 8 or so. This helps keep statements simple to understand and reason about numerically.

Like I mentioned before, I'll keep this PR open in case you want to try rewriting the dog vault or submit some other adjustments.

pdpol:
@gammafunk Thanks again for the review and good to see those changes, I'll keep those in mind for the next one.

I think I'll actually close this PR out and start on a fresh branch for the hog idea that I was planning to replace the dog vault with and see if I can come up with a few other concepts.

PR #2795, opened 2022-10-28
Fix int params to crawl.sendkeys()

brianfaires:
Currently crawl.sendkeys(6) sends "6" instead of keycode 6.
Due to lua_isstring(int) returning true; ints treated as strings.
Moved number check else block ahead of lua_isstring() check.
Made number check explicit; lua_isnumber() can return True for strings.


rawlins:
Thanks! Sorry for the delay in merging this.

PR #2798, opened 2022-10-29
Allow explore_stop_pickup_ignore for all spellbooks

brianfaires:
Each explore_stop_pickup_ignore regex is checked against DESC_PLAIN.
Ignoring all spellbooks requires listing every book by name in RC file.
Added a check for values of "book"|"books"|"spellbook"|"spellbooks".
If found, spellbooks are ignored. Manuals are unaffected.

### DIFF
--- a/crawl-ref/source/items.cc
+++ b/crawl-ref/source/items.cc
@@ -3089,8 +3089,17 @@ static bool _interesting_explore_pickup(const item_def& item)
         const string name = item.name(DESC_PLAIN);
 
         for (const text_pattern &pat : ignores)
+        {
             if (pat.matches(name))
                 return false;
+
+            const string c = lowercase_string(pat.tostring());
+            if ((c == "book" || c == "books" || c == "spellbook" || c == "spellbooks")
+                && item_is_spellbook(item))
+            {
+                return false;
+            }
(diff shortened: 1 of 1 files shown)


gammafunk:
These special patterns shouldn't be necessary. Since the `DESC_PLAIN` format of spell books is "book of ..." even for randart books, a pattern for `explore_stop_pickup_ignore` that matches any spell book would simply be
```
explore_stop_pickup_ignore += ^book of
```
which matches any item whose name begins with "book of". This excludes manuals, which have a `DESC_PLAIN` starting with "manual of". I'd recommend trying this setting, and feel free to make an issue if something about this isn't working properly.

brianfaires:
@gammafunk - testing on 0.33, commit c1556f0eadec6d1d5c5b3a06b3b90344f75fd072 (May 20), I'm not seeing that behavior. `explore_stop_pickup_ignore` is only affecting books that start with "book of".  Using init.txt of:
```
explore_stop = greedy_pickup_smart
explore_stop_pickup_ignore += ^book of
```

also tried:
```
explore_stop = greedy_pickup_smart
explore_stop_pickup_ignore += book of
```

PR #2805, opened 2022-11-01
Add lua function item.inscribe(inscription="", append=true)

brianfaires:
Currently, no way to set inscriptions from lua aside from sendkeys().
That gets messy and spams the dialog.
Added function to append or replace an existing inscription.


brianfaires:
Whoops - force pushed, then force pushed back to original

mgdelmonte:
I'd love to see this merged.

PR #2808, opened 2022-11-02
Tweaking spider text labels to be more consistent

### LINKED ISSUE #2742: spider form attacking with "front legs"
Version: Windows local tiles 0.30-a0-134-gb40fab5c76

When in spider form and examining a monster with x-v, the game will claim that the monster has a X% to evade your front legs. This is consistent with other forms but it is rather akward in the case of spider form. It is also inconsistent with what the A, % and possibly other screens and messages tell the player. "Fangs" or something among those lines would make more sense.

spider form front legs

kd7uiy:
Fixed #2742, changing front legs to front pincers, which we could also call Pedipalps if we wanted to follow more closely with real anatomy. Venomous is the more commonly used term for things that can poison, and technically is the right term here, so let's be consistent.


PleasingFungus:
Looks good. Thanks! :)

PR #2811, opened 2022-11-03
Fix: Ref to scroll of ID may break in _identify()

brianfaires:
Break caused by auto_assign_item_slot() and swap_inv_slots(),
when the ID'd item swaps with the scroll of ID.
Parameter 'link' is already provided to accommodate this.
Checking validity of link in case scroll is on ground (link==-1).

Mainly, fixing this means scroll of ID will be assigned to its assigned item_slot if it was previously unknown.


PleasingFungus:
Thanks for the fix, sorry for the delay! Merged it in. :)

PR #2814, opened 2022-11-04
Resonance strike damage tweaks

nlavsky:
This PR caps the ally-dependent part of Resonance Strike damage and adds the damage numbers to the `xv` screen.

### DIFF
--- a/crawl-ref/source/mon-cast.cc
+++ b/crawl-ref/source/mon-cast.cc
@@ -5188,10 +5188,12 @@ static void _cast_resonance_strike(monster &caster, mon_spell_slot, bolt&)
     if (!target)
         return;
 
+    // base damage 3d(spell hd)
+    const int pow = caster.spell_hd(SPELL_RESONANCE_STRIKE);
+    dice_def dice = resonance_strike_base_damage(pow);
+    // + 1 die for every 2 adjacent constructs, up to a total of 7 dice when
+    // fully surrounded by 8 constructs
     const int constructs = _count_nearby_constructs(caster, target->pos());
-    // base damage 3d(spell hd) (probably 3d12)
-    // + 1 die for every 2 adjacent constructs (so at 4 constructs, 5dhd)
-    dice_def dice = resonance_strike_base_damage(caster);
     dice.num += div_rand_round(constructs, 2);
     const int dam = target->apply_ac(dice.roll());
     const string constructs_desc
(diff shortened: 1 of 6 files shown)


crawl-ref/source/mon-cast.cc
```diff
@@ -5189,10 +5189,12 @@ static void _cast_resonance_strike(monster &caster, mon_spell_slot, bolt&)
...
 
     const int constructs = _count_nearby_constructs(caster, target->pos());
-    // base damage 3d(spell hd) (probably 3d12)
+    // base damage 3d(spell hd)
     // + 1 die for every 2 adjacent constructs (so at 4 constructs, 5dhd)
-    dice_def dice = resonance_strike_base_damage(caster);
+    const int pow = caster.spell_hd(SPELL_RESONANCE_STRIKE);
+    dice_def dice = resonance_strike_base_damage(pow);
     dice.num += div_rand_round(constructs, 2);
+    dice.num = min(MAX_RESONANCE_STRIKE_DAMAGE_DICE, dice.num);
```
PleasingFungus, line 5197: I'm a bit confused by this cap. If the target is surrounded by a full 8 constructs, they'd take (3 + 8 / 2 = ) 7 dice of damage. But you set `MAX_RESONANCE_STRIKE_DAMAGE_DICE = 6`. What's the point of that? If you're going to cap it, wouldn't it make more sense to cap it earlier, at 5 dice? This cap seems like it mostly does nothing, which seems like a bad use of complexity.

crawl-ref/source/mon-cast.cc
```diff
@@ -5189,10 +5189,12 @@ static void _cast_resonance_strike(monster &caster, mon_spell_slot, bolt&)
         return;
 
     const int constructs = _count_nearby_constructs(caster, target->pos());
-    // base damage 3d(spell hd) (probably 3d12)
+    // base damage 3d(spell hd)
     // + 1 die for every 2 adjacent constructs (so at 4 constructs, 5dhd)
```
PleasingFungus, line 5193: Should update this comment, or add an ellipsis and add another comment above the `min()` call.

PleasingFungus (changes requested):
Thanks for writing this up! The damage display side of things looks good, but it seems like you should either remove the cap (for simplicity) or document it in the spell description in some way.

PR #2816, opened 2022-11-07
No shelter for projectiles

nlavsky:
Sheltering orbs of destruction from injuries does nothing, but ironbound preservers do it anyway.

Friendly guardian golems are so friendly, they protect even hostile OoDs. This happens because of how `mons_aligned()` treats all projectiles.


PleasingFungus:
Thanks! @semitonal will like this one.

PR #2819, opened 2022-11-10
Added worn armor info button to action panel

tnobley:
Added known spells info button to action panel
Added makefile to automatically compile WEBTILES and run webserver (tornado+python3)


rawlins:
This seems to be a duplicate of part of #2831, so I will close on that basis.

PR #2822, opened 2022-11-11
Task b and d

LeftwardOwl:


rawlins:
Hi! Please try to double check that you are clicking the correct branch info when submitting PRs. I'm sort of guessing (because this is the second one this week) that this may have been intended as a PR against https://github.com/NCPE227/DC-SS-0x0b.

PR #2823, opened 2022-11-11
clarify keystroke in tutorial

duonghoang2001:


rawlins:
Thanks for the PR! I'm not sure if this was really intended for this repository, as it seems that PRs involving [laurareese03/crawl](https://github.com/laurareese03/crawl) may relate to a class and we've gotten several incomplete PRs from this. Nonetheless I looked at this one.

* Re the tutorial clarification: the way the tutorial is implemented, what is shown is robust to the up/down keys being rebound to something else. So a change like this would also need to handle this. I'm also not sure that this clarification is really needed though?
* In the second commit, I have merged the css fixes, thanks for spotting these! (1e1ebe7c0743)
* However, I think the style change to the "play now" div is too intrusive (and this would go live on servers that format the lobby in quite different ways, so I'm not sure it would have the intended impact across them) -- so I didn't merge this. I'm not opposed to improving the formatting though.

PR #2827, opened 2022-11-12
Update style.css

cgelu:
Created dark and light themes


rawlins:
This looks like a PR that may have been intended for another repository as part of a larger project? (E.g. for a class.) Adding light mode vs dark mode themes for parts of webtiles would definitely be reasonable to consider, but what's in this PR looks fairly incomplete to me. Closing.

PR #2828, opened 2022-11-13
rework guardian golem

hellmonk3:
Guaranteed inner flame made the spell good, but took away from the intended theme. It became a spell about making explosions, not protecting summons. There are other perfectly good explosive spells - let guardian golem provide extra summon support with minor healing and heal other and cut the bomb entirely.


Oneirical:
Guardian Golem in its current state is not exactly guarding what it is supposed to protect - it's much the opposite - but it is one of the biggest hallmarks of positional magic in the game. It strongly encourages using the environment to block enemies from swerving around it, and stepping in such a way to bait as many foes as possible in the blast zone without being detonated yourself.

I understand the reasoning from a thematic standpoint, but it's sad to see a big risk-reward, strategical gameplay element being gone.

Ash-R:
Re:
https://github.com/crawl/crawl/commit/799ed1240d53ce75d0245ea8acd5b5beeaeea1a9
https://github.com/crawl/crawl/commit/6492fb5f8eb4a734494002d192a87f9fb3927446

In addition to hellmonk's original argument regarding intended theme (of the summoning in general, not just the name of the spell) I'd argue that in it's current state, Guardian Golem has most certainly qualifies for "If "ready" access to suicide bombers becomes an issue" because "players are not going around casting inner flame on allies in place of using fireball" in effect this is exactly what they are doing to the detriment every other spell in the entire Summoning school.

robbiemu:
I think the theme of this change makes sense. But then summon school will be woefully underpowered in the early game. It could fully heal small animals every turn and still be underpowered. Perhaps adding a companion change like increasing the number of canines you can summon would balance out the damage loss? Easier than a whole new spell.

PleasingFungus:
Closing in favor of 7aaa3406e030c0ce06dc74f308086128971e6b10. Thanks for submitting!

PR #2831, opened 2022-11-17
Rest pac

tnobley:


crawl-ref/Makefile
```diff
@@ -0,0 +1,3 @@
+make_webtiles:
```
rawlins, line 1: we won't merge this makefile, can you remove it from the PR? I understand it may be convenient, but it's really ad hoc, and sudo *very very much* should not be present in a Makefile

crawl-ref/source/tileweb.cc
```diff
@@ -603,7 +603,39 @@ wint_t TilesFramework::_handle_control_message(sockaddr_un addr, string data)
         // (possibly just as a string, like the lua API for this)
         process_command(CMD_GAME_MENU);
     }
-
+    else if (msgtype == "list_worn_armor")
```
rawlins, line 606: I don't think it's great to have a new set of message names mapping in an ad hoc way to commands under a different naming scheme. It would be much better to have a single message type with command names used directly as the message data, and then process this in the crawl binary in a less brute-force way. There's already code to map command names (e.g. strings like "CMD_LIST_ARMOUR") to enum values (see `name_to_command` in macro.cc)

crawl-ref/source/webserver/game_data/static/action_panel.js
```diff
@@ -157,26 +157,52 @@ function ($, comm, client, cr, enums, options, player, icons, gui, main,
         }
         $tooltip.css({top: y + 10 + "px",
                      left: x + 10 + "px"});
-        if (slot == -2)
-        {
-            $tooltip.html("<span>Left click: minimize</span><br />"
+
+        switch (slot + NUM_RESERVED_BUTTONS) {
```
rawlins, line 161: I would also prefer something a lot more general here, though the exact form I'm not as sure of. One consideration is that it may be good to allow players to hide these extra buttons.

In some cases I'm not entirely sure of the rationale for why these and not other possibilities, e.g. why show spell library rather than cast. Probably the ability menu is another candidate for this. (In the long run, probably what would be ideal is allowing arbitrary commands to appear in this menu.) Some motivation in commit messages would definitely help.

crawl-ref/source/webserver/game_data/static/action_panel.js
```diff
@@ -451,7 +507,24 @@ function ($, comm, client, cr, enums, options, player, icons, gui, main,
         // TODO: select tile via something like c++ `tileidx_command`
         draw_action(gui, gui.PROMPT_NO, null, 0, adjusted_scale, selected == 0);
         draw_action(gui, gui.CMD_GAME_MENU, null, inc, adjusted_scale,
-                    selected == 1);
+            selected == 1);
```
rawlins, line 510: many (or maybe all) of the relevant commands have icons already for local tiles buttons, it'd be better to use those. Some of these will be fairly confusing in this use, e.g. the airstrike icon doesn't cast airstrike, etc.

chucksellick:
Since there seems to be no movement since the original feedback from rawlins 3 years ago, it is maybe best to close this. I'll leave it for a while to see if the original contributor or anyone else has anything to update. Possibly several of the UI improvements here would be good if the feedback were taken into consideration, but I'm also not clear what the improvements here are exactly as there's no writeup.

ClippyClipster:
Adding the original author

++Samuel

Implojin:
Because this is a 3 year old stale PR with no response to advil's reviews and no commit messages, I'm closing this.

PR #2836, opened 2022-11-20
Update unrand.txt

zhangjoke:


PleasingFungus:
Nice, lots of new descriptions added in here! :)

PR #2837, opened 2022-11-20
Rework Guardian Golem into Unstable Golem

Oneirical:
This PR is a proposed alternative to [this Guardian Golem rework](https://github.com/crawl/crawl/pull/2828).

I suggest the opposite, removing the injury bond mechanic and keeping only the explosion, and also renaming Guardian Golem to Unstable Golem for flavor.

Spell power still increases the Golem's HP, which is relevant because it prevents it from getting oneshotted and not getting to use its blast.

Why?

* It is one of the biggest hallmarks of positional magic in the game. It strongly encourages using the environment to block enemies from swerving around it, and stepping in such a way to bait as many foes as possible in the blast zone without being detonated yourself.

* It makes the player think through each encounter differently - gameplay with a healing-centric Guardian Golem would encourage the player to place down every single summon in every battle, which is a "no-brainer". The explosive mechanic makes the Golem an anti-synergy with other summons, and makes the player ask themselves which strategy is preferable with questions such as "Is the space open enough to make using Unstable Golem safe?" Each early game battle can be handled with Mammal/Imp/Hound, or with the risk-reward high damage of a Golem.

* Explosions are fun and there can never be enough of them in this game.


PleasingFungus:
Closing in favor of 7aaa3406e030c0ce06dc74f308086128971e6b10. Thanks for submitting!

PR #2849, opened 2022-11-26
fix fsim with ranged weapons

RypoFalem:
Use fake projectiles rather than whatever you're quivering (eg using your crossbow as a missile for your crossbow). Bug exists in 0.29.1 too.


PleasingFungus:
Nice fix! Thanks for finding this and sorting it out. Code looks clean (except for the fake projectile hack itself, but that's my fault :P).

PR #2858, opened 2022-12-03
nerf randarts

hellmonk3:
Based on some discord feedback. Try to make high plus weapons and high stat properties slightly less common. No idea if this will be noticeable.


PleasingFungus:
Merged as 7904f2a0a232c5667fcc6d37865a4e3568357422.

PR #2861, opened 2022-12-05
fix c++ warning

Jbudone:
Fix compile warning from deprecated implicit declaration of operator overload

```
dgn-proclayouts.cc: In constructor ‘LevelLayout::LevelLayout(level_id, uint32_t, const ProceduralLayout&)’:
dgn-proclayouts.cc:237:33: warning: implicitly-declared ‘FixedArray<dungeon_feature_type, 80, 70>& FixedArray<dungeon_feature_type, 80, 70>::operator=(const FixedArray<dungeon_feature_type, 80, 70>&)’ is deprecated [-Wdeprecated-copy]
  237 |     grid = feature_grid(env.grid);
      |                                 ^
In file included from externs.h:29,
                 from AppHdr.h:316,
                 from dgn-proclayouts.cc:5:
fixedarray.h:43:5: note: because ‘FixedArray<dungeon_feature_type, 80, 70>’ has user-provided ‘FixedArray<TYPE, WIDTH, HEIGHT>::FixedArray(const FixedArray<TYPE, WIDTH, HEIGHT>&) [with TYPE = dungeon_feature_type; int WIDTH = 80; int HEIGHT = 70]’
   43 |     FixedArray(const FixedArray &other)
      |     ^~~~~~~~~~
```

To avoid duplicating code I introduced a copy method and call that from both copy and assignment operators

### DIFF
--- a/crawl-ref/source/defines.h
+++ b/crawl-ref/source/defines.h
@@ -194,12 +194,16 @@ const int AGILITY_BONUS = 5;
 
 #define MAX_MONSTER_HP 10000
 
+// Many, MANY places currently hard-code this to 8 bits, but we need to
+// expand it. Please use colour_t in new code.
+typedef uint8_t colour_t;
+
 // colours, such pretty colours ...
 // The order is important (IRGB bit patterns).
-enum COLOURS
+enum COLOURS : colour_t
 {
-    COLOUR_INHERIT = -1,
-    BLACK,
+    COLOUR_INHERIT = (colour_t)-1,
+    BLACK = 0,
     COLOUR_UNDEF = BLACK,
     BLUE,
     GREEN,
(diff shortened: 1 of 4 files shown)


Jbudone:
Also fixed warning from conditional type mismatch; changed colour type to the new colour_t

crawl-ref/source/defines.h
```diff
@@ -194,12 +194,16 @@ const int AGILITY_BONUS = 5;
...
+typedef uint8_t colour_t;
+
 // colours, such pretty colours ...
 // The order is important (IRGB bit patterns).
-enum COLOURS
+enum COLOURS : colour_t
 {
-    COLOUR_INHERIT = -1,
-    BLACK,
+    COLOUR_INHERIT = (colour_t)-1,
```
rawlins, line 205: I'm a bit worried about this because COLOUR_INHERIT will no longer test as `< 0`, one would need confidence that its sign is never used (i.e. color values are never tested with `>= 0` -- did you check this? (It's kind of hard to search for unfortunately)

PR #2865, opened 2022-12-07
Update monsters.txt

zhangjoke:
To make the words more uniform and the sentences more smooth, and adapt to line character limit.


PleasingFungus:
Thanks for all of these! I appreciate your hard work. :)

PR #2873, opened 2022-12-10
Allow GameOption options to be set in the game

Aliscans:
This implements an options menu for Crawl (accessed from the "main menu") which let the player set options based on the GameOption class. The list appears in the order in which they appear in options_guide.txt, arena.txt or fight_simulator.txt, with section headings to break up the list.

You can:
- View and change option settings.
- Bring up help text for each option.
- Search for options where the name matches a pattern.
- Save options to the .crawlrc file used for that game.

You can't yet:
- Interact with options which weren't already implemented as GameOptions.
- Tell from the UI whether changes made with it will affect the current game.
- Select where options should be saved.
- Access the options menu before starting the game.

In addition:
- Some of the option-setting functions are still a bit basic.
- There are many options which do not use `GameOption`, and so are unaffected.
- There isn't an icon for the options menu.
- There is no documentation for `tile_use_small_layout` in `options_guide.txt`.

I intend to work further on some of these issues, but I think this is useful as things stand.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -1774,6 +1774,7 @@ menu_colour ^= <match>:<colour>:<regex>, <colour>:<regex>, ...
            pickup        (specific to pickup menus)
            shop          (shop menus)
            notes         (the ?: screen)
+           option        (the options menu - match option names but not values)
            resists       (the % screen)
            spell         (the Z and I screens)
            stash         (the results from Ctrl-F)
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -18,7 +18,8 @@ The contents of this text are:
                 name, remember_name, weapon, species, background, combo,
                 restart_after_game, restart_after_save, newgame_after_quit,
                 name_bypasses_menu, default_manual_training,
-                autopickup_starting_ammo, game_seed, pregen_dungeon
+                autopickup_starting_ammo, game_seed, pregen_dungeon,
+                suppress_startup_errors
 2-  File System and Sound.
                 crawl_dir, morgue_dir, save_dir, macro_dir, sound, hold_sound,
                 sound_file_path, one_SDL_sound_channel
@@ -43,11 +44,11 @@ The contents of this text are:
                 scroll_margin, always_show_exclusions
 3-f     Travel and Exploration.
                 travel_delay, explore_delay, rest_delay, travel_avoid_terrain,
-                explore_greedy, explore_greedy_visit, explore_stop,
-                explore_stop_pickup_ignore, explore_wall_bias, travel_key_stop,
-                travel_one_unsafe_move, tc_reachable, tc_dangerous,
-                tc_disconnected, tc_excluded, tc_exclude_circle,
-                runrest_ignore_message, runrest_stop_message,
+                explore_greedy, explore_greedy_visit, explore_item_greed,
+                explore_stop, explore_stop_pickup_ignore, explore_wall_bias,
+                travel_key_stop, travel_one_unsafe_move, tc_reachable,
+                tc_dangerous, tc_disconnected, tc_excluded, tc_exclude_circle,
+                tc_forbidden, runrest_ignore_message, runrest_stop_message,
                 interrupt_<delay>, delay_safe_poison, runrest_ignore_monster,
                 rest_wait_both, rest_wait_ancestor, rest_wait_percent,
                 explore_auto_rest, auto_exclude, travel_open_doors
@@ -150,7 +151,7 @@ The contents of this text are:
                 mouse_input, wiz_mode, explore_mode, char_set, colour,
                 display_char, feature, mon_glyph, item_glyph,
                 use_fake_player_cursor, show_player_species,
-                use_modifier_prefix_keys, language, fake_lang,
+                use_modifier_prefix_keys, language, fake_lang, messaging
                 read_persist_options
 
 5-b     DOS and Windows.
@@ -459,6 +460,15 @@ pregen_dungeon = incremental
         level entry, as was the rule before 0.23. Dungeons will not be stable
         given a seed with this option.
 
+suppress_startup_errors = false
+        If this is false, and an error was detected as the game first started
+        (such as a mistake in a configuration file), bring up a screen before
+        loading to display these.
+
+        The main menu will contain a warning of errors regardless of this
+        setting, and it does not affect errors which are detected as a
+        character is loaded.
+
 2-  File System.
 ================
 
@@ -487,8 +497,11 @@ macro_dir = <path>
         It should end with the path delimiter. The default value for this
         is dependent on system and build type.
 
-sound ^= <regex>:<path to sound file>, <regex>:<path>,
+sounds_on = true
         (Requires "Sound support"; check your version info)
+        If true, plays sound effects in various situations.
+
+sound ^= <regex>:<path to sound file>, <regex>:<path>,
         (Ordered list option)
         Plays the sound file if a message contains regex. In case of
         multiple matching regexes, only the sound of the last regex
@@ -698,7 +711,7 @@ feature_item_highlight = reverse
         hidden by items. If you use this option, the item glyph(s) on the
         square are hidden by the feature symbol.
 
-trap_item_highlights = reverse
+trap_item_highlight = reverse
         Highlights traps that would otherwise be hidden by items. If you
         use this option, the item glyph(s) on the square are hidden by the trap
         symbol (^).
@@ -846,6 +859,11 @@ explore_greedy_visit += stacks
         artefacts: makes explore_greedy visit artefacts, even
         if not in a stack
 
+explore_item_greed = 10
+        Greedy exploration treats items as if they were this much closer (or
+        further away, if negative) when deciding whether to visit a square with
+        an item or to explore a new square.
+
 explore_stop  = items,stairs,shops,altars,portals,branches,runed_doors
 explore_stop += greedy_pickup_smart,greedy_visited_item_stack
         (List option)
@@ -927,7 +945,8 @@ tc_dangerous      = cyan
 tc_disconnected   = darkgrey
 tc_excluded       = lightmagenta
 tc_exclude_circle = red
-        The above five settle the colouring of the level map ('X').
+tc_forbidden      = lightcyan
+        The above six settle the colouring of the level map ('X').
         They are
           reachable: all squares safely reachable (without leaving the
                      level)
@@ -938,6 +957,8 @@ tc_exclude_circle = red
           excluded: the colour for the centre of travel exclusions ('e')
           excluded_circle: the colour for travel exclusions apart from
                            centre
+          forbidden: squares which cannot be reached without crossing excluded
+                     squares.
 
 runrest_ignore_message ^= <regex>, <regex>, ...
 runrest_stop_message ^= <regex>, <regex>, ...
@@ -2885,6 +2906,10 @@ fake_lang = <lang1>[,<lang2>[,...]]
 
         Options are: (dwarven|jagerkin|kraut|runes|wide|grunt|butt:<n>)
         Experiment to find out what they do!
+
+messaging = true
+        If set to true, this allows you to receive DGL messages sent
+        by other players. (This setting only applies on the online servers).
 
 read_persist_options = false
         When set to true, the game will read additional options from
--- a/crawl-ref/source/dat/defaults/menu_colours.txt
+++ b/crawl-ref/source/dat/defaults/menu_colours.txt
@@ -8,52 +8,52 @@ menu += darkgrey:(melded)
 menu += darkgrey:(inert)
 
 # Bad items
-menu += lightred:.*bad_item.*
+menu += lightred:.*\bbad_item\b.*
 
 # Useless items, comes here to override artefacts etc.
-menu += darkgrey:.*useless_item.*
+menu += darkgrey:.*\buseless_item\b.*
 
 # Items disliked by your god.
-menu += $forbidden:.*forbidden.*
+menu += $forbidden:.*\bforbidden\b.*
 
 # Handle cursed and equipped items early to override other colour settings.
-menu += lightred:.*equipped.* cursed
-menu += red: (a|the) cursed
-menu += inventory:lightgreen:.*equipped.*
+menu += lightred:.*\bequipped\b.* cursed\b
+menu += red: \b(a|the)\b \bcursed\b
+menu += inventory:lightgreen:.*\bequipped\b.*
 
 # Important game items
 #
-menu += lightmagenta:.*misc.*rune of Zot
-menu += lightmagenta:.*orb.*Zot
+menu += lightmagenta:.*\bmisc\b.*\brune of Zot\b
+menu += lightmagenta:.*\borb\b.*\bZot\b
 
 # Artefacts
 #
-menu += lightcyan:.*unrand.*
-menu += white:.*artefact.*
+menu += lightcyan:.*\bunrand\b.*
+menu += white:.*\bartefact\b.*
 
-#menu += white:.*identified.*artefact.*
-#menu += lightblue:.*unidentified.*artefact.*
+#menu += white:.*\bidentified\b.*\bartefact\b.*
+#menu += lightblue:.*\bunidentified\b.*\bartefact\b.*
 
 # Possible egos
-menu += lightblue:^unidentified .*weapon.*(runed|glowing)
-menu += lightblue:^unidentified .*armour.*(runed|glowing|embroidered|shiny|dyed)
+menu += lightblue:^unidentified .*\bweapon\b.*\b(runed|glowing)\b
+menu += lightblue:^unidentified .*\barmour\b.*\b(runed|glowing|embroidered|shiny|dyed)\b
 
 # Enchanté
-menu += green:enchanted
-menu += green:staff of
-menu += green:ego
+menu += green:\benchanted\b
+menu += green:\bstaff of\b
+menu += green:\bego\b
 
 # Emergency items
-menu += yellow:.*emergency_item.*
+menu += yellow:.*\bemergency_item\b.*
 
 # Good items
-menu += cyan:.*good_item.*
+menu += cyan:.*\bgood_item\b.*
 
 # Dangerous (but still useful) items
-menu += $dangerous:.*dangerous_item.*
+menu += $dangerous:.*\bdangerous_item\b.*
 
 # Defaults for normal items
-menu += lightblue:unidentified.*(potion|scroll|wand|jewellery|magical staff)
+menu += lightblue:\bunidentified\b.*(potion|scroll|wand|jewellery|magical staff)
 
 
 # Colouring of autoinscribed god gifts
@@ -66,4 +66,4 @@ menu_colour += inventory:white:\w \*\s
 
 
 # Not really menu.
-menu_colour += notes:white:Reached XP level
+menu_colour += notes:white:\bReached XP level\b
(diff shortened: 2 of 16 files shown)


rawlins:
FYI I am hopefully going to have time to look at this in detail next week.

crawl-ref/source/scroller.cc
```diff
@@ -176,11 +178,23 @@ bool formatted_scroller::process_key(int ch)
...
+        y -= UI_SCROLLER_SHADE_SIZE/2; // Skip the shaded border at the top.
+#elif defined(USE_TILE_WEB)
+    if (!use_shade && y)
+        y--; // 1 line should be enough.
     if (from_webtiles)
         m_scroller->set_scroll(y);
     else
+#else
+    (void)use_shade;
+#endif
```
PleasingFungus, line 197: Seems maybe marginally nicer to put this conditional earlier. Something like:

```
#ifdef USE_TILE_LOCAL
void formatted_scroller::scroll_to_line(int y, bool use_shade)
#else
void formatted_scroller::scroll_to_line(int y, bool /*use_shade*/)
#endif
```

But also I think we can probably just cut this `#else` - do we really have a linter that complains if we don't use a function parameter?

crawl-ref/docs/options_guide.txt
```diff
@@ -459,6 +460,14 @@ pregen_dungeon = incremental
         level entry, as was the rule before 0.23. Dungeons will not be stable
         given a seed with this option.
 
+suppress_startup_errors = false
```
PleasingFungus, line 463: Thanks for fixing up these various missing/out of date option descriptions! In future, it'd probably be better to put these types of fixes in a separate, smaller PR, to make review easier.

crawl-ref/source/dat/defaults/menu_colours.txt
```diff
@@ -8,52 +8,52 @@ menu += darkgrey:(melded)
 menu += darkgrey:(inert)
 
 # Bad items
-menu += lightred:.*bad_item.*
+menu += lightred:.*\bbad_item\b.*
```
PleasingFungus, line 11: I don't really understand the changes to this file, sorry. Can you explain?

crawl-ref/source/game-options.cc
```diff
@@ -7,9 +7,38 @@
...
+    static map<unsigned, string> list;
+    if (list.empty())
+    {
+        // This is the same list as in _curses_attribute.
+        list =
+        {
+            {CHATTR_NORMAL, "none"}, {CHATTR_STANDOUT, "standout"},
+            {CHATTR_BOLD, "bold"}, {CHATTR_BLINK, "blink"},
+            {CHATTR_UNDERLINE, "underline"}, {CHATTR_REVERSE, "reverse"},
+            {CHATTR_DIM, "dim"}
```
PleasingFungus, line 31: Probably more readable if there's one line per entry, rather than try to cram multiple `CHATTR_` entries onto one line.

crawl-ref/source/game-options.cc
```diff
@@ -7,9 +7,38 @@
...
+#include "prompt.h"
+#include "scroller.h"
 #include "tiles-build-specific.h"
 
+// Return a list of the valid field values for _curses_attribute(), excluding
+// hi: and hilite: values, which are all duplicates.
+// This is called after initialisation to ensure that colour_to_str() works.
+static map<unsigned, string>& _curses_attribute_map()
+{
+    static map<unsigned, string> list;
```
PleasingFungus, line 22: `list` is a pretty unfortunate name here - it's not a list, it's a map, and the name `list` gives the reader no information about what's in it. Perhaps call this `attrs_to_strings` or similar.

crawl-ref/source/game-options.cc
```diff
@@ -66,6 +95,108 @@ bool read_bool(const string &field, bool def_value)
...
+        : Menu(_flags), caller(_caller) { }
+    int pre_process(int key) override
+    {
+        if ('?' == key)
+        {
+            caller->show_help();
+            return static_cast<int>(CK_NO_KEY);
+        }
+        else
+            return key;
```
PleasingFungus, line 111: DCSS generally uses early returns. (There are exceptions, because this project is ridiculous, but that's our modern style.)

```suggestion
        return key;
```

crawl-ref/source/game-options.cc
```diff
@@ -66,6 +95,108 @@ bool read_bool(const string &field, bool def_value)
...
+/// @param[in,out] caller Option to edit. Reads the name and current value.
+///                       Writes the new value.
+/// @returns       True if something is chosen, false otherwise.
+bool load_string_from_UI(GameOption *caller)
+{
+    string prompt = string("Enter a value for ")+caller->name()+":";
+    /// XXX - shouldn't use a fixed length string here.
+    char select[1024] = "";
+    string old = caller->str();
+    while (1)
```
PleasingFungus, line 175:
```suggestion
    while (true)
```

crawl-ref/source/initfile.cc
```diff
@@ -2067,25 +2079,58 @@ newgame_def read_startup_prefs()
...
+        auto date = TIME_FN(&date0);
+        fprintf(f, "\n# Options saved for %s at %02d:%02d:%02d on %d %s %d.\n",
+                you.your_name.c_str(), date->tm_hour, date->tm_min,
+                date->tm_sec, date->tm_mday, mons[date->tm_mon],
+                1900+date->tm_year);
+        for (auto g : Options.get_sorted_options())
+        {
+            if (!g->name().empty()) // not GameOptionHeading
+                fprintf(f, "%s = %s\n", g->name().c_str(), g->str().c_str());
+        }
```
PleasingFungus, line 2125: Can delete these braces, right? (If you run `util/unbrace`, won't it do that automatically?)

crawl-ref/source/initfile.cc
```diff
@@ -5670,6 +5718,184 @@ bool parse_args(int argc, char **argv, bool rc_only)
...
+    }
+
+    void filter_items()
+    {
+        string prompt = "Search for what? (regex, leave blank to show all)";
+        char select[1024];
+        auto old_search_pat = search_pat;
+        int new_hovered = -1;
+        vector<MenuEntry*> chosen_items;
+        while (1)
```
PleasingFungus, line 5763:
```suggestion
        while (true)
```

crawl-ref/source/initfile.cc
```diff
@@ -5670,6 +5718,184 @@ bool parse_args(int argc, char **argv, bool rc_only)
...
+                if (!g)
+                    heading = entry;
+                else if (search_pat.empty() || search_pat.matches(g->name()))
+                {
+                    if (heading)
+                        chosen_items.emplace_back(heading);
+                    heading = nullptr;
+                    if (found_hover && new_hovered < 0)
+                        new_hovered = chosen_items.size();
+                    entry->hotkeys[0] = index_to_letter((i++)%52);
```
PleasingFungus, line 5790: Cheeky. Instead of being cute, let's explicitly increment `i` on a separate line.

crawl-ref/source/initfile.cc
```diff
@@ -5670,6 +5718,184 @@ bool parse_args(int argc, char **argv, bool rc_only)
...
+            bool found_hover = false;
+            int i = 0;
+            for (auto entry : all_items)
+            {
+                if (entry == items[last_hovered])
+                    found_hover = true;
+                auto g = static_cast<GameOption*>(entry->data);
+                if (!g)
+                    heading = entry;
+                else if (search_pat.empty() || search_pat.matches(g->name()))
```
PleasingFungus, line 5783: Let's remove a layer of indentation, to make this a bit more readable.

```suggestion
                if (!g)
                {
                    heading = entry;
                    continue;
                }
                if (!search_pat.empty() && !search_pat.matches(g->name()))
                    continue;
```

crawl-ref/source/initfile.cc
```diff
@@ -5670,6 +5718,184 @@ bool parse_args(int argc, char **argv, bool rc_only)
...
+    menu.set_tag("option");
+
+    int i = 0;
+    for (const auto option : list)
+    {
+        EGP_MenuEntry *entry;
+        if (option->name().empty())
+            entry = new EGP_MenuEntry(option->str(), MEL_SUBTITLE);
+        else
+        {
```
PleasingFungus, line 5876: Same comment here re dedenting.

```suggestion
        if (option->name().empty())
        {
            entry = new EGP_MenuEntry(option->str(), MEL_SUBTITLE);
            continue;
        }
```

But honestly, it'd be much nicer to turn this whole `for` loop into a helper function that returns a `EGP_MenuEntry *entry`. I haven't dug into the code to see how feasible that is.

PleasingFungus (commented):
Thanks for writing this! I'm extremely excited to see progress on in-game options - we've been talking about this for as long as I can remember.

Comments below are mostly stylistic. I haven't had time to dig into the guts of the changes, sorry - hopefully this is still helpful!

crawl-ref/source/scroller.cc
Aliscans: 1. It would show that it _is_ unused, 

On the other hand, it would need an extra proprocessor condition as it has two alternatives rather than three. That takes  the proportion of the code which comprises preprocessor stuff from 20% to 27%, which to me just makes it harder to read.

Another way to remove that line would be to have three versions of the whole scroll_to_line() function  They only share the last two code lines, so it may not need an extra function to handle those.

2. Yes, `-Wextra` brings in `-Wmissing-declarations` on G++.

I haven't changed anything for now.

crawl-ref/docs/options_guide.txt
Aliscans: Would it help to split it now? I only put them there because one of the other changes draws attention to the missing options.

crawl-ref/source/dat/defaults/menu_colours.txt
Aliscans: With PCRE, or with GNU's POSIX implementation, it means that they don't match options such as "bad_item_prompt". There's no equivalent to \b in the POSIX standard, so I've changed it to use space now.

It would be easier if these all used a tag, but this works for now.

crawl-ref/source/game-options.cc
Aliscans: I've renamed the variable (to "ret" - the function's just a constructor, so that seemed like the best name for the thing it returns), and rearranged the initialiser list as you said above.

crawl-ref/source/game-options.cc
Aliscans: I think I based that on EGP_Menu::pre_process(), which has more cases. I've changed it now.

crawl-ref/source/game-options.cc
Aliscans: I've changed this.

crawl-ref/source/initfile.cc
Aliscans: I don't think `unbrace` is quite that clever. I've removed them now.

crawl-ref/source/initfile.cc
Aliscans: I changed it to `while (chosen_items.empty())` instead.

crawl-ref/source/initfile.cc
Aliscans: I'm not sure what the first bit means, but I've changed it.

crawl-ref/source/initfile.cc
Aliscans: I've changed that.

crawl-ref/source/initfile.cc
Aliscans: That change wouldn't work as the loop doesn't end with the else block. It can go into a separate function, however, so I've done that.

crawl-ref/docs/options_guide.txt
PleasingFungus: I think it'd be nice - I suspect we'd be able to merge the options fixes much faster than the rest of this commit. Don't do it if it's a ton of work for you, though.

crawl-ref/source/initfile.cc
PleasingFungus: Sorry for confusion - I generally feel that using postincrement/postdecrement as part of a larger expression decreases readability.

crawl-ref/source/scroller.cc
PleasingFungus: Ah, I hadn't spotted the intervening `#elif defined(USE_TILE_WEB)` - that complicates this for sure. Yeah, three separate versions of the function could be reasonable. I'd try it out and see how it looks?

rawlins:
I've been playing around with this. It's great, congratulations on getting this all to work! I've put a bunch of testing notes below (that were without referencing the code). Some of this is probably stuff that can't be addressed in the scope of a single PR. My biggest concerns right now (that would hold me back from merging) are re how saving is handled.

**ui**

* text box should probably be bigger, for many options this doesn't matter, but for long ones it's somewhat painful to view/edit them. Example: `morgue_dir`. I also found this to be true in webtiles/tiles.
* Probably not tractable as part of this PR: it would be good to have console popups not take over the entire screen.
* I disprefer dark blue for headers because it's often very hard to read on console. Perhaps consider cyan or lightcyan.
* Superficial: would probably call these "options" rather than "preferences" for consistency with docs.
* The help/search in options_guide is great! I wonder if we could even parse options_guide so that whatever help there is is visible directly on the edit popup (not needed for this PR). Is there a way to get help for textbox options when you are in their popup? Also, positioning of the search results seemed to be completely not working on local tiles. I didn't look into this too carefully, but it seemed off by some positive amount that increased through the file. (Tested on hidpi, this sounds potentially like a dpi bug)
* Maybe: right click activates option help (similar to item menus)
* Options organization: the long list with headers is good for now, but I think it could be good to eventually have this with more of a submenu structure. The question is, what options do players *really* want to be able to edit? At a more superficial level, I think it might be better to alphabetize within sections.
* Could the search use the header search code rather than a popup? Maybe there's a reason it uses the popup though. Also might be good if esc with a search active exited the search, rather than the menu.
* In webtiles/local tiles, the validation error popups would only exit on esc/click (or ? and then esc), even though it says "any other key to continue."
* I didn't find it intuitive that changes aren't saved beyond the game unless ctrl-s is hit. Maybe if this doesn't autosave, it should indicate unsaved state somehow? (Though see below for notes on saving)
* Maybe want to cue that `,` will work to skip sections

**function**

* some explanation of errors would be helpful (e.g. `Bad pickup_menu_limit: "asdf"` could say what type is expected, I assume these are derived from the GameOption objects)
* Probably not all options should be shown, though I imagine this could take some iteration. Example: fsim options shouldn't appear in a game that can't access fsim. Arena options probably shouldn't appear at all. Webtiles-specific options probably shouldn't appear from a console client (and similarly for console-specific options), or if they appear, maybe relegated to a submenu.
* There are some options that I'm worried are unsafe to expose in this way because editing them should only be done by someone who knows exactly what they're doing, e.g. `morgue_dir`, sdl font locations (maybe any path). These might be ok if there's a "reset to default" thing though.
* for c++ options not currently implemented, adding GameOptions is probably what we do. But a certain number of options are handled in lua, not immediately sure what to do with them. These are ones that may need more ui, as well (e.g. `force_more_message`).
* Need a way to reset the saved options in the game (I originally wrote this before really looking into the details of saving, but it's probably still true). Also, and this is possibly harder, would be good to allow resetting specific options to default. I've had on my todo list better handling of option defaults for a while.
* I think probably the webtiles options should be pushed immediately after exiting the menu, and aren't right now. Example, when I change `tile_display_mode` in webtiles nothing happens, but when a spectator connects (something that happens to trigger an options sync) it gets fixed up.

**saving**

I have some reservations about the current approach to saving. We do not have any programmatic editing of the player's rc file and I'm not sure adding this is a good idea. I'm not even 100% sure that servers will have the permissions set up for this (only the webtiles server / dgl would need to write to that file before this, not the crawl binary). I can see that it may be useful to have this as a manual option for players who are never going to manually edit their rc, and seems better offline as well.
* Awkward interaction with manually edited parts of the file: options set this way override it completely, because they occur later in the file.
* Bad interaction with the autosave preferences file: anything in the rc file overrides this completely, so any option saved in both is going to be ignored from the pref file until the player rc is manually resaved (at which point they will be in sync). Example: `quiver_menu_focus`, action panel settings, `default_manual_training`. So using the save feature ever just seems to break the prefs file.
* Speaking of this: why not save in the pref file? This is what it's designed for. This might also need some further work on how that file is managed of course.
* When trying this out locally, saving just seems to concatenate a new set of options at the end of the existing rc file? Maybe I'm doing something wrong. But this probably isn't the behavior we want.

Further testing needed:
* There are definitely options that are designed to change value in `Options` in only a very controlled way or will need followup code. Example: right now changing `tile_display_mode` in local tiles changes what tiles are used correctly and immediately, but not the tile size (adjusting the ratio between square and glyph sized). (In this case I know that the necessary code is called when dragging the window between screens, but I would need to look into exactly what is missing.) I don't have a sense for how many of these there are.

crawl-ref/source/scroller.cc
Aliscans: I've changed this now.

crawl-ref/docs/options_guide.txt
Aliscans: I've done this now.

robertxgray:
This is a much needed feature, thank you.

> Save options to the .crawlrc file used for that game.

This file will be read-only in many cases. I can provide fixes for Linux and Android though.

> There is no documentation for tile_use_small_layout in options_guide.txt.

The Android branch introduced two new options: tile_use_small_layout and tile_window_limit_size. It seems you started your branch while the merge was on progress. They are documented in the main branch.

It may be a good idea to warn players that the game needs to be restarted after an options change.

Aliscans:
> This is a much needed feature, thank you.
> 
> > Save options to the .crawlrc file used for that game.
> 
> This file will be read-only in many cases. I can provide fixes for Linux and Android though.

Are you talking about what happens if Crawl is run without a user-provided preference file? I can see how `find_crawlrc()` would not be useful there, and I'll look into changing how it chooses a file. Or are you thinking of another problem?

> > There is no documentation for tile_use_small_layout in options_guide.txt.
> 
> The Android branch introduced two new options: tile_use_small_layout and tile_window_limit_size. It seems you started your branch while the merge was on progress. They are documented in the main branch.

That's good. It did stand out a bit when every other GameOption was in one of the doc files.

> It may be a good idea to warn players that the game needs to be restarted after an options change.

It varies from one option to another. I haven't tried to categorise them all, so at the moment I don't know if a restart will help.

I'm a bit wary of advising people to restart Crawl, anyway, as they could have Lua scripts which change the settings to something completely different to whatever they just set.

robertxgray:
> Are you talking about what happens if Crawl is run without a user-provided preference file? I can see how `find_crawlrc()` would not be useful there, and I'll look into changing how it chooses a file. Or are you thinking of another problem?

Yes, I mean the user-provided preference file. Users who don't know about that file are those that would benefit more from the options menu. I plan to auto-generate an empty init.txt for Android in the next update because it will feature a launcher for editing it, similar to the webtiles version.

crawl-ref/source/dat/defaults/menu_colours.txt
rawlins: I think the fix here should be to prevent `menu_colour` from applying to this menu -- despite the name it is really designed for items, not menus in general, and the default values reflect that. Though it looks like this was done intentionally in `EGP_MenuEntry::highlight_colour` so maybe there's a reason I'm not seeing?

rawlins: > The docs say menu_colour covers a lot more than just inventory menus, so I thought it was reasonable to apply it here. I notice that the code doesn't match this, but I don't know why this is.

The docs are wrong :-). Or rather, they are written in a confusing way, from a time where there essentially were no menus that didn't have something to do with items (so I believe this wording would be most likely referring to e.g. stash tracking). I just don't see a great reason to apply regex highlighting to options menus, just like it doesn't apply to the game menu, or the macro menu.

PR #2880, opened 2022-12-13
Use more Australian spelling

Aliscans:
Change how some words are spelt in places where the user can see them:
armor => armour
defense => defence
offense => offence
alternate => alternative

It does a bit more with _describe_action() because "armour" wouldn't fit as it was, and it made sense to me to unify the "make sure the action name is the right length" code as I was changing it anyway.


crawl-ref/docs/develop/arena.txt
```diff
@@ -149,6 +149,6 @@ example:
     crawl -arena "arena:smoke random v random"
 
 will cause the monsters to fight in an arena which is filled with smoke. The
-available alternate terrains are in source/dat/arena.des. If an arena you
+available alternative terrains are in source/dat/arena.des. If an arena you
```
rawlins, line 152: We don't enforce the aus English thing in developer documentation. Also, fwiw this (really any change that tries to avoid adjectival "alternate") reads really awkwardly in us English. I would reword this to avoid the word altogether, in this case probably simply dropping the word: "The available alternative terrains are in source/dat/arena.des."

crawl-ref/docs/develop/background_creation.txt
```diff
@@ -85,7 +85,7 @@ herd-watcher:
 
         // and just a couple skill points in stuff you might do around a field.
         { { SK_FIGHTING, 6 }, { SK_SUMMONING, 8 }, { SK_SPELLCASTING, 3 },
-          { SK_DODGING, 3 }, { SK_ARMOR, 4 } { SK_STEALTH, 1 }, },
+          { SK_DODGING, 3 }, { SK_ARMOUR, 4 } { SK_STEALTH, 1 }, },
```
rawlins, line 88: despite prev comment about developer docs, this one doesn't match the code so should be changed

crawl-ref/docs/macros_guide.txt
```diff
@@ -54,7 +54,7 @@ The macro.txt file
 
 Macros and keymaps are stored in a file called macro.txt in your
 settings/ directory. You can change where the file is read from and
-written to by specifying an alternate directory with the macro_dir
+written to by specifying an alternative directory with the macro_dir
```
rawlins, line 57: I would prefer to use "different" instead of "alternative"

crawl-ref/docs/options_guide.txt
```diff
@@ -69,7 +69,7 @@ The contents of this text are:
                 status_caption_colour, enemy_hp_colour, clear_messages,
                 show_more, small_more, show_newturn_mark, show_game_time,
                 equip_bar, animate_equip_bar, item_stack_summary_minimum,
-                mlist_min_height, mlist_allow_alternate_layout,
+                mlist_min_height, mlist_allow_alternative_layout,
```
rawlins, line 72: not entirely in favor of changing opts to satisfy the aus english thing, but you did at least provide backwards compatibility

crawl-ref/source/dat/descript/quotes.txt
```diff
@@ -1526,14 +1526,14 @@ fire dragon scales
...
- armor scarce enabled them to contend on equal terms with the uncouth but
+ defence, [the Germans] still found themselves at a disadvantage; for their
+ armour scarce enabled them to contend on equal terms with the uncouth but
  formidable weapons of their adversaries. The Bohemians were armed with long
  iron flails, which they swung with prodigious force. They seldom failed to
  hit, and when they did so, the flail crashed through brazen helmet, skull and
  all.”
     -James A. Wylie, _The History of Protestantism_, vol. I, book 3, ch. 15
-     “Jon Huss and the Hussite Wars”. 1878.
+     “John Huss and the Hussite Wars”. 1878.
```
rawlins, line 1536: good catch, the book's name in the reference appears to have been incorrect (seems independent of aus though)

rawlins (commented):
I won't merge this as I can't really assess the australianness of these changes, but here's some comments.

crawl-ref/docs/macros_guide.txt
Aliscans: I've changed that now.

crawl-ref/docs/develop/arena.txt
Aliscans: There were a few places where I felt the word didn't add anything to the sentence it was in, but I didn't want to make changes I couldn't objectively justify. I've dropped this now and corrected the file location on that line.

I do wonder whether this could be merged with `docs/arena.txt`, though.

crawl-ref/source/dat/descript/quotes.txt
Aliscans: The name seemed as unlikely to me as the American spellings, to be honest. I _was_ impressed to find a scan of the book online to check against.

crawl-ref/docs/options_guide.txt
Aliscans: Crawl expects people to write option names down, and silently ignores mistakes. It seems to me like the place where spelling matters the most.

Compatibility is easier now than it was when this option was last renamed, so I had no reason not to.

semitonal:
Merged, thanks!

PR #2887, opened 2022-12-21
fix: make autoexplore wait for movement restrictions to expire

nlavsky:
If you wear an amulet of magic regeneration, it's possible to cast Momentum Strike and then restore MP before -Move times out. Pressing `o` will just print "You cannot move." and you'll have to use `5` or `.` to rest -Move off.

Fix this by making autoexplore wait for -Move to expire before attempting to move. Tree form and Stuck status from Translocations miscasts also prevent autoexplore, so let's fix these cases too.


PleasingFungus:
Looks good. Nice fix! (Hopefully we don't add some form of non-motility that you *can't* rest off later.)

PR #2889, opened 2022-12-22
new simple depths entry

hellmonk3:


nicolae-carpathia:
depths... deep troll... it's like poetry, it rhymes

PR #2894, opened 2022-12-23
feat: Filter out useless altars from stash search

pdpol:
Currently altars are always included in stash search results as "useful" even if the player could never use them, i.e. by playing a Demigod.

This change simply adds a check to the dedup/useless filter to filter these out. Since the predicate got so big I also thought it merited refactoring out into its own helper function.

altar_filter


PleasingFungus:
Looks good, thanks! (Nice gif!) I'm going to adjust formatting a little in a follow-up commit.

PR #2898, opened 2022-12-25
Edit layout_misc_corridors in Lair

Skrybe:
Make layout_misc_corridors sometimes use boxy=false when in Lair.  This provides a more "natural" look than the square rooms it generates otherwise, and works well with Lair ruination.  As a bonus, it also creates more crevices for the player to hide in.


ebering:
Thanks for the patch!

PR #2899, opened 2022-12-26
fix: local tiles sidebar height

Antern:
Screenshot 2022-12-26 at 12 24 30
Screenshot 2022-12-26 at 12 23 52

Andoird branch merge broke some menu height calculations. Options.tile_sidebar_pixels is 32px by default, so wouldn't affect rendering in all cases but then tile_sidebar_pixels is actually changed.


PleasingFungus:
Looks good, thanks! :)

PR #2900, opened 2022-12-26
fix: Make a flavor gimmick wand no_pickup

pdpol:
Imagine my delight when I stumbled upon my very own vault in an online game, and then my abject horror when the wand, which really ties the whole room together, was missing!

I tested the game seed (4877243199622027174) locally and found that the wand had been placed correctly, so unless there are some possibilities I don't know about I think it's likely that a monster waltzed in and walked away with it.


PleasingFungus:
Fair enough! :)

PR #2904, opened 2022-12-27
Compiler warnings

Aliscans:
Remove two types of compiler warning (which have no connection to one another):

1. compile out view_add_glyph_overlay() when USE_TILE_LOCAL is set.

2. Suppress two specific libPNG warnings from tilegen.elf.

There may be other ways of achieving the second thing (I know one error message goes if you call `png_set_interlace_handling(png_ptr)` before `png_read_update_info()`), but user_warning_fn should have no impact on the image files the program produces.

### DIFF
--- a/crawl-ref/source/rltiles/tool/tile.cc
+++ b/crawl-ref/source/rltiles/tool/tile.cc
@@ -371,6 +371,18 @@ bool tile::texture(const tile &img)
     return true;
 }
 
+#ifdef USE_TILE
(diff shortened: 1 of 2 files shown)


PleasingFungus:
Thanks! Very happy to see an end to these libpng warnings - they've been confusing folks for many many years.

Aliscans:
Thanks. And sorry for missing <cstring>. My only reason for using strcmp there rather than std::string was that the parameter wasn't a string.

PR #2906, opened 2022-12-31
Refactor knockback

Monkooky:
Performing knockback of an actor is currently found in two locations: One in beam.cc for most spells, and one in evoke.cc for gust of wind specifically. The former function is predicated on properties that only make sense for beams (in particular origin.cc), and the latter on performing the full Gust of Wind effect. 

This change refactors knockback to be a standalone function called by both beam.cc and evoke.cc. This should improve maintainability for knockback effects, as well as make it much easier to apply knockbacks with non-spell effects.

This commit incidentally changes the behaviour of knockback in the following ways:
Knockbacks can no longer be 'aimed' (e.g. controlling the knockback direction of Phial of Floods by aiming past a target)
Subtle changes in Gust of Wind's knockback after a target would hit a wall
Gust of Wind behaves oddly with dispersal traps, in the same manner IMB already did.


crawl-ref/source/actor.cc
```diff
@@ -1032,6 +1033,107 @@ void actor::collide(coord_def newpos, const actor *agent, int pow)
...
+ *        using repeated one-tile knockbacks. Any LOS requirements
+ *        must be checked by the calling function.
+ * @param cause The actor responsible for the knockback.
+ * @param source The tile this actor is being knocked directly away from.
+ * @param distance The number of single tile knockbacks attempted.
+ * @param strength Each attempt has probability size/strength of failing.
+ *                 If strength is 0, never fail.
+ * @param pow Cause 2d(pow/10 + 1) damage on collision.
+ * @param message The message to display on successfully pushing this actor.
+ *                If message is null, display no message.
```
PleasingFungus, line 1047: Generally better to pass around `std::string` and check whether the string is empty, rather than using `const char*` as the parameter type. The latter can result in really subtle memory corruption bugs.

crawl-ref/source/evoke.cc
```diff
@@ -415,72 +415,10 @@ void wind_blast(actor* agent, int pow, coord_def target)
...
-        }
-
-        if (pushed)
+        if (agent->can_see(*act))
         {
-            if (act->is_monster())
-            {
-                act->as_monster()->speed_increment -= random2(6) + 4;
-                if (you.can_see(*act))
-                    affected_monsters.push_back(act->as_monster());
```
PleasingFungus, line 480: This got lost, meaning that fan of gales will never print anything about monsters being blown back. (`player_affected` is never set, either.)

crawl-ref/source/evoke.cc
```diff
@@ -415,72 +415,10 @@ void wind_blast(actor* agent, int pow, coord_def target)
...
-
-                    // If no luck, they slam into something.
-                    if (!success)
-                        collisions.insert(make_pair(act, newpos));
-                }
-            }
-        }
-
-        if (pushed)
+        if (agent->can_see(*act))
```
PleasingFungus, line 418: This is wrong. We should knock back actors regardless of whether we can see them. `can_see` should only affect whether they're added into `affected_monsters`.

crawl-ref/source/beam.cc
```diff
@@ -4623,74 +4624,17 @@ void bolt::knockback_actor(actor *act, int dam)
...
-            break;
-        }
+    const char* message = origin_spell == SPELL_CHILLING_BREATH
+                ? make_stringf("%s %s blown backwards by the freezing wind.",
+                    act->name(DESC_THE).c_str(),
+                    act->conj_verb("are").c_str()).c_str()
+                : make_stringf("%s %s knocked back by the %s.",
+                    act->name(DESC_THE).c_str(),
+                    act->conj_verb("are").c_str(),
+                    name.c_str()).c_str();
```
PleasingFungus, line 4635: Let's simplify this to `%s %s knocked back by the %s` and only pass the last element ("freezing wind" / beam.name) into `knockback()`.

crawl-ref/source/evoke.cc
```diff
@@ -574,36 +512,6 @@ void wind_blast(actor* agent, int pow, coord_def target)
         else
             mpr("The monsters around you are blown away!");
     }
```
PleasingFungus, line 514: The messaging is broken now. We get:
```
You slam into a wall!
You are blown backwards!
```

Which is, appropriately, backwards. :)

This is actually quite tricky to fix properly. We have a few options:
- Defer collisions and handle them separately. This is a bit scary once dispersal traps get involved, but it does mean that we can also remove the message-printing logic entirely.
- Rewrite `::collide()` to be optionally silent and handle printing later. This is a bit hairy as wel.
- Just let wind_blast print messages for each individually affected creature, removing its special printing logic and letting `::knockback()` handle everything. This is fine for monsters - they rarely blow back much more than the player. It would be a bit worse for the Storm card, since that's more likely to blow back many things (and hence be quite message-spammy), but that's going to be spammy anyway once collisions get involved. So let's just do that.

crawl-ref/source/actor.cc
```diff
@@ -1032,6 +1033,107 @@ void actor::collide(coord_def newpos, const actor *agent, int pow)
     }
 }
 
+/**
+ * @brief Attempts to knock this actor away from a specific tile,
+ *        using repeated one-tile knockbacks. Any LOS requirements
+ *        must be checked by the calling function.
+ * @param cause The actor responsible for the knockback.
+ * @param source The tile this actor is being knocked directly away from.
```
PleasingFungus, line 1041: This is the same as `cause->pos()` in all cases, so don't bother passing it separately.

crawl-ref/source/actor.cc
```diff
@@ -1032,6 +1033,107 @@ void actor::collide(coord_def newpos, const actor *agent, int pow)
     }
 }
 
+/**
+ * @brief Attempts to knock this actor away from a specific tile,
+ *        using repeated one-tile knockbacks. Any LOS requirements
+ *        must be checked by the calling function.
+ * @param cause The actor responsible for the knockback.
```
PleasingFungus, line 1040: Pass `cause` as a reference (&) not as a pointer (*). It's always non-null, so we don't need to invite bugs by allowing it to be nullable.

crawl-ref/source/actor.cc
```diff
@@ -1032,6 +1033,107 @@ void actor::collide(coord_def newpos, const actor *agent, int pow)
...
+ * @param source The tile this actor is being knocked directly away from.
+ * @param distance The number of single tile knockbacks attempted.
+ * @param strength Each attempt has probability size/strength of failing.
+ *                 If strength is 0, never fail.
+ * @param pow Cause 2d(pow/10 + 1) damage on collision.
```
PleasingFungus, line 1045: This comment is doomed to become obsolete. The caller does not need to know this math, and anyone who changes the behavior of collide() will never, ever remember to update this.

Instead, let's say something like `Determines damage done to us if we hit something.`

PleasingFungus (commented):
Thanks for writing this up! I left some comments for future reference, but made fixes myself while I was in the area.

PR #2911, opened 2023-01-01
Add indicator for rampage (Local tiles)

Monkooky:
Adds an icon indicating which directions will trigger a rampage for local tiles. This icon is on by default, and can be disabled with show_rampage_indicator = false in your settings.

This also pulls the 'can rampage' logic out of the 'do rampage' function.


Implojin:
As written, I think this indicator is going to be incorrect in the presence of beholders and fearmongers?

Also, the tracer_target assert in _rampage_forward should definitely be moved into get_rampage_target.

Monkooky:
Both good comments, both resolved! 
Thanks!

crawl-ref/source/movement.cc
```diff
@@ -576,52 +576,32 @@ bool prompt_dangerous_portal(dungeon_feature_type ftype)
...
- */
-static spret _rampage_forward(coord_def move)
-{
-    ASSERT(!crawl_state.game_is_arena());
-
-    // Assert if the requested move is not a move delta
-    // this would throw off our tracer_target.
-    ASSERT(abs(move.x) <= 1 && abs(move.y) <= 1);
-
+coord_def get_rampage_destination(coord_def move, monster* target){
```
PleasingFungus, line 579:
```suggestion
coord_def get_rampage_destination(coord_def move, monster* target)
{
```

crawl-ref/source/tileview.cc
```diff
@@ -1392,6 +1393,12 @@ void tile_apply_properties(const coord_def &gc, packed_cell &cell)
             cell.bg |= TILE_FLAG_TRAV_EXCL;
     }
 
+    if(you.rampaging() && adjacent(you.pos(), gc) && Options.show_rampage_indicator
```
PleasingFungus, line 1396: Trivia: this is wrong! `adjacent` returns true even if `you.pos() == gc` - we need `grid_distance(you.pos(), gc) == 1`.

crawl-ref/source/initfile.cc
```diff
@@ -250,6 +250,7 @@ const vector<GameOption*> game_options::build_options_list()
         new BoolGameOption(SIMPLE_NAME(show_travel_trail), USING_DGL),
         new BoolGameOption(SIMPLE_NAME(use_fake_cursor), USING_UNIX ),
         new BoolGameOption(SIMPLE_NAME(use_fake_player_cursor), true),
+        new BoolGameOption(SIMPLE_NAME(show_rampage_indicator), true),
```
PleasingFungus, line 253: I'm going to cut this until someone actually asks for it. Better fewer options than more :)

crawl-ref/source/tileview.cc
```diff
@@ -1392,6 +1393,12 @@ void tile_apply_properties(const coord_def &gc, packed_cell &cell)
...
+       && get_rampage_target(gc - you.pos()) != nullptr)
+    {
+        cell.bg |= TILE_FLAG_RAMPAGE;
+    }
+
     if (!map_bounds(gc))
```
PleasingFungus, line 1402: As written, we'll crash (I suspect!) if we move too close to the map edge with rampage on. Let's try to avoid that and move our new logic below this `map_bounds` check.

PleasingFungus (approved):
Nice refactor, nice feature! I'll merge this

PR #2920, opened 2023-01-08
New unrand: toga "Victory"

Implojin:
This is an early unrand with a variable bonus based on kills:

As the player kills threatening monsters, this robe gradually gains AC and Slay, capping out at +10 each.

Beware -- Quaffing potions or reading scrolls resets these bonuses to zero!

(This still needs art: It's just using the robe of clouds tile for now.)

-----

I'm not 100% this is the best idea for an unrand (it's an earlygame generic wyrmbane-as-armour that resets its bonus on consumable use), but talking with PF about unrands gave me the motivation to push this. Merge it if you like, reject it if not, maybe offer some feedback if you think it would work better with changes?

No worries if this PR doesn't land! I just thought, hey, maybe it would be fun?


Implojin:
Ah, I've just realized that I didn't write this to ramp-up from kills using ranged attacks or magic and the description doesn't mention this -- this is what I get for bringing back something I'd half forgotten about on a whim.

I'm going to mark this as draft for a bit while I take another look at this and get a tile together.

Implojin:
Ok, I've fixed up the on-kill bonus to properly apply bonuses from kills with ranged weapons and spells, and I've given this a tile that will hopefully be replaced by something better.

Any thoughts?

semitonal:
(I made some comments in #crawl-dev before noticing you'd left, so just copying those over to here!)

`<Discord|> <k​ate> it sort of has an unfortunate interaction with scrolls of enchant weapon/armour and acquirement etc, assuming the intention is that you're penalised for using scrolls/potions to help you in battle
<Discord|> <k​ate> having a bunch of specific potion/scroll exceptions would probably be too messy though
<Discord|> <k​ate> i'd also definitely cut the chance to upgrade it on killing an MTHRT_EASY monster`

I don't have any objections to the general design though, it could be interesting!

crawl-ref/source/dat/descript/unrand.txt
```diff
@@ -768,3 +768,10 @@ Force Lance
 A plain iron lance, imbued with supernatural gravitas via the artifice of the
 mage Gell. It strikes with a force many times stronger than that imparted by
 the wielder.
+%%%%
+toga "Victory"
+
+An ancient artefact of Okawaru, these woolen robes were lost when a former
+champion failed a duel. Its defense and damage sometimes increase when the
+wearer emerges victorious from threatening combat -- but beware! Its bonuses
+reset to zero if the wearer reads any scrolls, or quaffs any potions.
```
semitonal, line 777: A minor style note is that this could maybe be reworded slightly to tend more towards flavour than mechanics (since the mechanics are explicitly spelled out in the DBRAND) - maybe just something like "Its enchantments increase in power when the wearer emerges victorious from threatening combat, but resorting to the use of scrolls or potions for assistance causes them to fade away again."

Implojin: Ah, I hadn't seen this comment while writing the latest commit. No worries, that's easy to adjust.

semitonal:
I've merged this, thanks for the PR! Have also added some prompts on trying to read/quaff with enemies in LOS, since that seems like an easy mistake to make.

Implojin:
Thanks kate!

PR #2923, opened 2023-01-09
Fix estimation of the minimum usable window size

robertxgray:
We were using the m_fonts array to get the font sizes. This is wrong because caching leads to unpredictable array contents.

Now we are using proper char width and height. It also has the benefit of improved precision.

The bug was reported in reddit:
https://www.reddit.com/r/dcss/comments/107d1tj/what_has_changed_with_the_ui_trunk_compiled/


PleasingFungus:
Thanks!

PR #2926, opened 2023-01-10
Update description of Pain

Monkooky:
Pain still exists as a monster spell, and is inaccurately described as causing a minor backlash.

This fixes that.


PleasingFungus:
Good branch name :)

PR #2934, opened 2023-01-10
Rework Qazlal

sdynet:
Qaz is a god with an interesting theme. But because it's been a long time since it was built, the mechanism is very old. In particular, the storm noise given to believers from the early game was themeally cool, but it was viciously difficult to control the balance. So I'm going to update the mechanism and at the same time move back the section where the player gets the noise.
Key objectives:
①Reduce Qaz's noise in early games. However, in later games, similar noises are made.
②More consistent refinement of themes (destruction and protection)

**Overview of Rework Qaz**
0*: Immune Cloud
1*: Storm shield(Harm, RMsl, +3 AC)
2*: Upheaval
4*: Upgrade storm shield(additional AC bonus, elemental adaptation, cloud generation)
5*: Disaster area
* The dust cloud is replaced by a sand storm cloud. It causes physical damage.
* Cloud generation works in a similar way to wereblood.

**Why do you give Qaz a Harm?**
①To compensate for the weakness of Qaz's active power when the skill level is low.
②Harm is notorious. Because of that, there are quite a few people who are scared without using it. I want to give them a unique experience. (Most of the test participants still hated Harm, but a few changed their minds. Yay!)

**Upgrade storm shield**
_How it works:_
If you kill an enemy, you get a stacked buff like wereblood. This figure(N) increases up to 30. (Storm x N)
This number is initialized to zero when you move the floor, and it doesn't decrease naturally within the same floor.
Increases the trigger probability of the following effects in proportion to the number:
①The rate of cloud generation
②AC bonus (+1 per 5 stacks. That is, up to +6. You have +3 on stack 0 by default, so you get +9 ac on stack 30.)
③Increasing the probability of element adaptation trigger

_(If you feel this is good.) Why is it so powerful?:_
When the stack is 30, the overall performance of the storm shield is set to be 30-50% higher than the existing Qaz.
Dev teams tend to induce players to use other tactics rather than stair dance tactics. One attempt is the Yred Rework. Also I expect players to boldly fight their enemies hot and fierce instead of dancing on the stairs. So to encourage players to continue stacking without using stairs, the stack bonus of the storm shield was set high.

_Why did you use this method?:_
The uncontrollable storm noise was wonderful, but it gave a nightmare experience to players experiencing this in the early games and abyss. So I made the storm shield noise free when the figure is zero, like wereblood. I have also devised several ways to maintain the "unstoppable noise" that continues from the initial game, but I have not been able to find a suitable return for this huge risk.


acrobat3:
I played this for about 15 hours with various combo (total 7 games). And it felt well-organized and very interesting.

my favorite aspects of this rework

1. passive harm creates an interesting play in dangerous situations.
In my experience, the best play in dangerous situations in dcss is usually to try to escape. But passive harm discourages trying to escape (receive 20% more damage while run away) and encourages killing enemies before player die (deal 30% more damage). So this made me more aggressive in dangerous situations and it was a very interesting experience.

2. stacked buff(storm buff) discourage stairdancing.
Stairdancing is the optimal play in many situations and quite boring at the same time. Storm buff, which works somewhat similar to Yred's piety, greatly discouraged stairdancing by asking me "Would you take a bunch of ac and cloud or would you give up all and do stairdance?". This made me take more risks and have more fun.

PR #2935, opened 2023-01-10
Update divine vigour to apply multiplicatively

### LINKED ISSUE #2910: Ely Divine Vigour ignores the increase and decrease of hp and mp by items and mutations
PLATFORM and OS or BROWSER:
local tiles

DCSS VERSION AND GAME INFORMATION:
0.30-a0-368-g86fb4627a9

DESCRIPTION:
Ely divine vigour ignores the increase and decrease of hp and mp by items(ring of magical power, ring of octopus king, faerie dragon scale...) and mutations(robust, fragile, high mp, low mp...)

example 1
You have 50 mp without items and 27 invo skills, then you get 25mp from divine vigour.
You have 68 mp with two ring of magical power and 27 invo skills, then you should get 34mp but only get 25mp from diven vigour.

example 2
You have 100 hp without mutations and 27 invo skills, then you get 50hp from divine vigour.
You have 110 hp with robust 1 and 27 invo skills, then you should get 55hp but only get 50hp from divine vigour.

This bug has a particularly big impact when you are wearing mp+ items.

Monkooky:
Update divine vigour to apply multiplicatively with all other effects.

Fixes #2910

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -3892,7 +3892,6 @@ int get_real_hp(bool trans, bool drained)
 
     // Mutations that increase HP by a percentage
     hitp *= 100 + (you.get_mutation_level(MUT_ROBUST) * 10)
-                + (you.attribute[ATTR_DIVINE_VIGOUR] * 5)
                 + (you.get_mutation_level(MUT_RUGGED_BROWN_SCALES) ?
                    you.get_mutation_level(MUT_RUGGED_BROWN_SCALES) * 2 + 1 : 0)
                 - (you.get_mutation_level(MUT_FRAIL) * 10)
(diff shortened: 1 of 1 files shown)


Monkooky:
Fixes #2910

crawl-ref/source/player.cc
```diff
@@ -3916,6 +3915,10 @@ int get_real_hp(bool trans, bool drained)
             hitp = hitp * 3 / 2;
     }
 
+    //Divine vigour should apply multiplicatively with all other effects
+    hitp *= 100 + (you.attribute[ATTR_DIVINE_VIGOUR] * 5);
```
PleasingFungus, line 3919: You don't need these parens. PEMDAS, etc.

crawl-ref/source/player.cc
```diff
@@ -3916,6 +3915,10 @@ int get_real_hp(bool trans, bool drained)
             hitp = hitp * 3 / 2;
     }
 
+    //Divine vigour should apply multiplicatively with all other effects
```
PleasingFungus, line 3918: Don't think these comments are needed. If we wanted something like this, we'd note `// The below effects multiply each other, rather than applying additively` - but I think it's pretty clear from context. We definitely don't need a comment just for divine vigour.

Also, put a space after the `//`.

PleasingFungus (commented):
Thanks! Merged with some small tweaks.

PR #2944, opened 2023-01-17
Record Wu Jian's Lunge and Whirlwind as CACT_ABIL, not CACT_INVOKE.

Aliscans:
Previously, Lunge, Whirlwind, Serpents Lash and Heavenly Storm were recorded with CACT_INVOKE, and Wall Jump with CACT_ABIL.

This change means that Wu Jian's powers are now split into martial attacks (abilities which do not consume piety or fail if you are silenced) and other powers (invocations which do both).


semitonal:
Thanks for the patch!

PR #2948, opened 2023-01-18
Acrospider

OctopodesWeekly:
Spider Form has historically been seen as a weak start, and AOO have done it no favors. Let's buff it.

Spider form now gives a (pow/10) Acrobat boost. Thematic with spider,, and synergizes well with poison in general. Because of current AOO, there is still an active choice in whenever you want to wait/move or attack.

It has also been reduced to Level 2, and pow cap is reduced to 50.

Taken from my previous transmutation PR (#2508 ), but just spider form.


OctopodesWeekly:
I didn't mean to push this into the actual crawl repo (yet), just my own fork. Whoops!

PR #2959, opened 2023-01-25
Refactor inaccuracy

Monkooky:
adjust inaccuracy to
1: return actual inaccuracy modifier rather than an indicator of 'degree'
2: use superclass code in player::inaccuracy() rather than duplicating it
3: not use heinous implicit typing


PleasingFungus (commented):
Pushed a slightly different version of this.

PR #2960, opened 2023-01-26
docs: update jiyva powers description

Implojin:
Jiyva's powers description on ^ was mildly incomplete when referring to jellies consuming items: Most eyes also consume items under Jiyva, but as far as I could tell this wasn't documented anywhere in-game. Eyes and jellies use different console glyphs and internal genus, so this seems worth disambiguating; this commit clarifies the description slightly.


semitonal:
Thanks! I adjusted to instead refer to them as "fellow slimes".

PR #2961, opened 2023-01-26
Connect item sets: {vane and {Tremorstones

sdynet:
Compared to other categories, Misc has many items with similar roles. Therefore, I think it is a suitable target to apply item sets. However, since this is taking one tactical means away from the player, I'm going to increase the number of charge by one(vane) and two(tremor).


yrdzrfxndfvh:
i think there are some instances you replaced that would not make sense thematically to change (first i saw was nzn_swamp_blast_mining)

semitonal:
Thanks for the PR! I've merged this with some fixes to vaults, but have kept the charges at the same level as current - they're extremely strong items and it's not a case where it'd be necessary to compensate for one type no longer being available.

PR #2962, opened 2023-01-27
Check if the player wants to use a "martial attack" with an odd weapon.

Aliscans:
Previously, the game asked for confirmation before attacking with (e.g.) a weapon inscribed with !a. It also did this before rampaging into a monster, but not for Wu Jian's "martial attacks".

It now carries out the same checks before Wu Jian's lunge and whirlwind attacks, and does it before a wall jump if this attacks a monster.

### DIFF
--- a/crawl-ref/source/god-abil.cc
+++ b/crawl-ref/source/god-abil.cc
@@ -5673,7 +5673,9 @@ bool wu_jian_do_wall_jump(coord_def targ)
     auto wall_jump_direction = (you.pos() - targ).sgn();
     auto wall_jump_landing_spot = (you.pos() + wall_jump_direction
                                    + wall_jump_direction);
-    if (!check_moveto(wall_jump_landing_spot, "wall jump"))
+    if ((wu_jian_wall_jump_triggers_attacks(wall_jump_landing_spot)
+         && !wielded_weapon_check(you.weapon())
+        || !check_moveto(wall_jump_landing_spot, "wall jump")))
     {
         you.turn_is_over = false;
         return false;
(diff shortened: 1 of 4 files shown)


PleasingFungus:
I don't understand this PR, sorry. Why would we want to prompt the player before they move around near a monster? That seems like it'd be very commonly a good thing to do even if you're carrying a 'weird weapon'.

Aliscans:
There's another way of looking at it. 

You press a key which would ordinarily make your character attack a monster, but your current weapon has a `!a` inscription. Why would you want the game to make your character attack without further confirmation?

Lunge and Whirlwind do nothing if you're wielding (say) a sling, so the main effect of this change is on things the player has explicitly asked the game to prompt before attacking with.

PleasingFungus:
Ah, gotcha. That makes more sense. Will this prompt if you attempt to move while unarmed & without UC skill?

rawlins:
I think it's anything that would trigger `wielded_weapon_check`: https://github.com/crawl/crawl/blob/master/crawl-ref/source/fight.cc#L706

(From the branch name I'm guessing what @Aliscans had in mind with "weird weapons" was really nonweapons / breadswinging, which may also trigger a warning here I think?)

Aliscans:
I was originally thinking of non-weapons or missile weapons, but then I realised that lunge and whirlwind don't work with those. Wall Jump does, so it's not a total misnomer.

I actually came across this when trying to play a hybrid, and being annoyed by the number of times I unintentionally bashed things with a magic staff I had inscribed with `!a`.

I've just moved the extra code down the list a bit because I'd realised that it could trigger in some situations where movement was impossible. I still don't have much feel for what should be checked first, though, so it could still be in the wrong place.

PleasingFungus:
There's some tricky parts of this (checking invis, checking confuse), but it looks like you've gotten everything right. Fingers crossed! :)

PR #2963, opened 2023-01-27
Sling of Outrageous Fortune

Monkooky:
Adds a new unrand 'sling' with very poor accuracy that is not affected by slay/enchantment bonuses, but whose damage scales 3x with slay/enchantment bonuses. 

The purpose of this sling is, admittedly, the joke. 

The design is meant to scratch the same itch as the Dark Maul through a completely separate vector- silly high damage at the cost of usability, but achieved with poor accuracy rather than high delay. 
The sling's effect encourages seeking slay, while also utilizing non-slay based means of increasing accuracy. Ideally it should hit a balance point where the sling is not worth using for a character with a decently enchanted ranged weapon but no slay, means of constriction, or source of halo, and is competitive with a especially strong randart hand crossbow for a character with all of the above.

The description could likely use rewriting, and the effect is probably in need of tweaks.

### DIFF
--- a/crawl-ref/source/actor.cc
+++ b/crawl-ref/source/actor.cc
@@ -207,8 +207,11 @@ void actor::shield_block_succeeded()
 
 int actor::inaccuracy() const
 {
+    int degree = 0;
     const item_def *amu = slot_item(EQ_AMULET);
-    return amu && is_unrandom_artefact(*amu, UNRAND_AIR);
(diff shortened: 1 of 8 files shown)


Monkooky:
On returning to this idea I don't find it as interesting mechanically as I would like, and I have not seen a path to improving it.

PleasingFungus:
;_;

PR #2975, opened 2023-01-30
added leading space to text message on uselessness

spiros:
Add space to fix description text.

image


PleasingFungus:
Nice catch, thanks! :)

PR #2981, opened 2023-02-02
Lower longbow base delay to match previously lowered mindelay

Rytisgit:
In e572f607256a the mindelay of longbows was lowered to 0.6, but base delay remained unchanged. This makes the longbow just as weak as it was before until the character manages to get past 20 skill. This buff brings longbows more in line with other ranged weapons.


acrobat3:
>  the mindelay of longbows was lowered to 0.6, but base delay remained unchanged.

base delay was also lowered  from 1.8 to 1.7 unless I read something wrong

Rytisgit:
longbow in trunk:
image
longbow in .28: 
image

if it ever was 1.8 delay it was between commits in trunk.

PleasingFungus:
Longbows have 12 base damage and base delay 1.7. Glaives have 15 base damage and base delay 1.7. A longbow has fullscreen range; a glaive has range 2. (And, of course, a longbow has a lower mindelay!)

I'm happy to discuss nerfing other ranged weapons if that seems needed.

PR #2987, opened 2023-02-05
bugfix: Restore "original" (unsorted) as the default ordering for shops, to unbreak Lua

mgdelmonte:
Sorting the shopping menu by type, name, or price is neat, but it breaks the Lua integration because the sorted menu scrambles item letters, so that `items.shopping_list` has no relationship to any menu you can currently see in the game.  That is, `items.letter_to_index` fails to work properly.

This PR restores the original "stock" order as the default, which means that `items.letter_to_index` will work again.

In practice, the stock order is very often exactly the same as the type order (the former default) -- and it's always the same for shops with only one type of inventory.  So the PR unbreaks Lua and should have an almost unnoticeable effect on regular play.

### DIFF
--- a/crawl-ref/source/shopping.cc
+++ b/crawl-ref/source/shopping.cc
@@ -809,16 +809,17 @@ static bool _purchase(shop_struct& shop, const level_pos& pos, int index)
 
 enum shopping_order
 {
-    ORDER_TYPE,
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Seems reasonable. Merged, thanks!

PR #2998, opened 2023-02-16
[accessibility] Fix draconian confusion for red-green colour-blind people

powerbf:
Like about 8% of males, I'm red-green colour-blind. Currently, it's very hard to distinguish the yellow draconian from the green draconian. I brightened to skin of the yellow draconian to make it more obviously yellow.


PleasingFungus:
Looks good. Thanks!

PR #2999, opened 2023-02-18
Remove the "You had/have # spell level(s) left." line from Dj morgues

antalsz:
Since Djinn don't memorize spells or keep track of spell levels, I found this line confusing in my morgues.  (This PR strips out the "You couldn't/cannot memorise any spells." line for the same reason, since it's the same line.)

If this breaks some sort of parsing or is otherwise not desired, I could also tweak it to always say "You couldn't/cannot memorise any spells." instead.


PleasingFungus:
Nice fix, thanks!

PR #3001, opened 2023-02-19
Clear the training target before the message, rather than afterwards.

Aliscans:
This means that a lua hook which catches the message can change the skill target for the relevant skill (imagine a script which sets an Axes target at 3, and sets it to 4 when it reaches 3). Previously, the game set the target to 0 after the message returned, whether or not the skill was still at its target.


PleasingFungus:
Sure, makes sense. Apologies for the delay in responding to this. :)

PR #3002, opened 2023-02-20
Add ghostly pirate ship vault.

beargit-crawl:
Add a pirate ship vault containing various spooky inhabitants. Including the undead captain with his cutlass. Also, the feared Kraken in the lore is present. Play testing this vault seems to have been fine but it may require more tuning.

### DIFF
--- a/crawl-ref/source/dat/des/variable/mini_monsters.des
+++ b/crawl-ref/source/dat/des/variable/mini_monsters.des
@@ -8735,6 +8735,75 @@ xxNNNxx
 xxx@xxx
 ENDMAP
 
+# A monster will have this equipped rather than it being on the ground
+# Yo ho ho and a bottle o' rum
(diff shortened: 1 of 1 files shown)


regret-index:
The vault concept is pretty solid, and I like the efforts to scale it for such a broad number of branches. There's a lot of different monsters that are pretty harmless for such a deep-placing vault, though (wights, phantoms, electric eels, Depths flying skulls); while some minor degree of flavour spawns are fine for a vault, particularly higher numbers of them can both feel tedious and make one miss any actual threats in the vault. The header's a bit disorganized, too (like a pile of KITEM definitions that could be just one ITEM line), and some trial placements in Depths clearly generates some teleport islands of shallow water amongst deep water.

If you're okay with it, I'm willing to go through and fix this up myself, whenever you confirm such, to make this vaults merge-able.

PR #3004, opened 2023-02-20
Mail order catalogue and Alliterative Armaments

Monkooky:
Adds two new shop vaults:
the Mail Order catalogue stocks mail and order. There's not much else to it.

Alliterative Armaments stocks melee weapons with alliterative brands; e.g. flaming falchion or short sword of speed.
This includes the otherwise unobtainable spectral/speed sacred scourge and eudemon blade of electrocution, which I suspect devs will remove.
Weapons are weighted so that equally rare base types are equally rare in-shop, and rarer base types are rarer in-shop.


bobross419:
heavy is in trunk.  Heavy handaxe is pretty decent

Picklenugget1:
Spectral spear?

PR #3008, opened 2023-02-22
Have Wall Jump's targeter only highlight monsters it could affect.

Aliscans:
As things stood, the targeter for Wall Jump included every monster next to the landing location, revealing invisible monsters and making it harder to tell what the ability did. It now only highlights monsters the action would try to attack.


semitonal:
Thanks for the patch!

PR #3010, opened 2023-02-22
Fix fear.

### LINKED ISSUE #2421: Fear does not breeak for half health monsters
Webtiles 

.28

**DESCRIPTION:**
Read a scroll of fear, hitting a monkey and an ogre. Then, quaffID'd attraction. Hit the monkey. Got informed it snapped out of fear.
However, monkey was still frightened.
It may be relevant that the monkey was poisoned?

**EXTRA FILES:**
image

Monkooky:
Monsters would previously snap out of fear even when hit, so long as they were below half health and were not cornered. This fixes that.

Fixes #2421

### DIFF
--- a/crawl-ref/source/mon-behv.cc
+++ b/crawl-ref/source/mon-behv.cc
@@ -743,7 +743,7 @@ void handle_behaviour(monster* mon)
 
         case BEH_FLEE:
             // Check for healed.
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Haha, excellent find :)

PR #3017, opened 2023-02-25
Monster invis """rework"""

hellmonk3:
This branch reveals invisible monster positions always, allowing autofight to function, spell targeters to not lie to the player, and autopickup to not turn on and off constantly. Information about the monster (type, remaining hp, status conditions) is not revealed. In some sense this is a design flaw b/c it can still encourage out of game tracking of monster positions, hex success chance, etc, but I think it's more palatable as a compromise than cutting monster invis altogether.

This does cost some interesting behavior with invis monsters (since ofc players don't "really" always know their positions) but I think the interface improvement is worth it.


mgdelmonte:
👍 I think this is a great idea.  FWIW I usually pick up a ring of see invisible mainly to avoid the ugly UI/UX complications, and not to improve in-game survivability.

SentientSupper:
I want this because it greatly simplifies implementing #3705.

Hellmonk:
closing this pending other devteam invis plans in 0.34 (real)

PR #3019, opened 2023-02-27
Update ability.txt

zhangjoke:
Update this file and make some of the sentences smoother.


PleasingFungus:
Thanks again for these! :)

PR #3026, opened 2023-02-27
feat: New Unique Regal and New Monster Spell Charming Aura

bobross419:
Regal is intended as a late game monster that provides a different ally counter than mass abjuration.  The front rank of allies have a chance to get charmed which turns them against the back ranks.  Summoners can then decide to recast hoping for it to be closer (the charmed summons still exists as a hostile).  I'm hoping the dynamic makes the fight more tactical for the summoner.  Without the entourage he ends up being rather bland and only an interesting fight for dedicated summoners, but with the Klown and cob it should even push the blaster caster away from the otherwise no brainer servitor into a more dynamic fight.  The inclusion of an item from the Octopus King set is to flavor the octopod unique and this late in the game should not be overpowering.
 
New Unique: Regal

A unique Octopode prince carrying one unique Octopus King item and able to turn most foes whose gaze falls directly on him to his side.  He spawns with a small entourage of a Killer Klown and a Death Cob to increase the threat level.

New Monster Spell: Charming Aura
Attempts to charm everything in direct line of fire that isn't blocked by other monsters.  This is not a full screen effect like Olgreb's Toxic Radiance; it only charms the front rank of monsters by firing tracers against all monsters in LOS_NO_TRANS and discarding failures.

The below is not included in commit message (but I can upon request).

I tried to make a LOS_NO_ACTOR los_type but my c chops aren't that good.   If I'm understanding opacity, I think it is probably a better way to approach it since it would make the monster iterator logic a lot cleaner.  As it stands, I make a beam for each , test fire the tracer, then if the tracer hits I fire the beam, then I have to duplicate a bunch of the charm logic.  I was expecting setting the type BEAM_CHARM to handle the monster attitudes and enchantments for me because testing with a wand of charming on a bad guy worked no problem.  If someone is able to walk me through making BEAM_CHARM handle the attitude and status logic or implementing LOS_NO_ACTOR I'd be grateful for the assistance.

I would also appreciate input on the placement of both the unique and the spell in the various lists.  I tried to go off what the dev docs said for monster and spell creation, but they didn't directly address uniques and the lists I added to don't seem to be specifically ordered in any way.  

I've included a test_sprint.des map to test the logic.  I've shipped it for testing convenience, but I intend to remove it before acceptance and merge.

Thanks for the opportunity to contribute, and I look forward to working with you all on this.


bobross419:
I've been poking at some of the open issues that involve targeting and I realize why LOS_NO_ACTOR wouldn't work here.

semitonal:
Hi, thanks for the pull request! The idea of an Octopode unique with the artifacts is very flavourful, but I think it would need a different approach in terms of the mechanics.

It's certainly fine for some monsters to be more difficult or less difficult for certain character builds, but I don't think that so narrowly targeting summoning-focused players is a good approach to make for an interesting challenge generally. Additionally, we try not to place Zot-only monsters outside of Zot, so death cobs and killer klowns wouldn't be an appropriate band to use.

bobross419:
@semitonal thanks for taking a look!  I hadn't even considered that I never see the Zot monsters outside zot.   I was hoping the high level characters would offset the summoner-centric nature of the unique.  

Would you recommend replacing the band with more appropriate monsters?  Or maybe replacing the band altogether and give the unique a more robust spell set?

PR #3028, opened 2023-02-28
new vault in crypt and depths

yrdzrfxndfvh:
summary of the crypt one i wrote in its commit thingy (im not very good at github): a big circular vault with quite a few monsters and the leviathan (kraken) at the center + a good chance of getting a nice quickblade

the depths one: just a small little aviary with a shrike room and a tengu room

note about the second one: i saw there was already a vault named aviary so if that name is unavailable you can change it to birdhouse


regret-index:
Well, this submission got me to unearth Devil Daggers from my backlog so I could understand it some more, so that's already quite distinct already.

- Since in that game one is bound to an arena, I think the vault invoking such should be entirely enclosed with either stairs or a teleporter to enter it. This would also prevent such specific zombies very rapidly wandering out of the vault and also prevent easily sniping the non-regenerating kraken zombie.
- The provided quick blades are cute (and probably safe rewards by Crypt depth), but to incentivize bothering with the vault for those not using short blades, it could also do with some scrolls of enchant weapon, like the weapon-upgrading gems from the game.
- Octopode zombies are verboten, alas- octopode monsters were cut a very long time ago, and only humans are allowed as non-decorative non-natively-spawning player species monsters. What might work instead are spectral tentacled monstrosities, which might need some renumbering (honestly, kraken zombies in general aren't very scary by Crypt depth so they could swap) but would also make the vault distinct and are entirely valid things to throw around.

On the Depths aviary birdhouse- plain tengu were also culled heavily from vaults. The vault should be fine just placing more tengu warriors and conjurers, anyway.

If you're fine with it, I'm willing to do these changes (and clean up the headers a little) myself, whenever you confirm such, to make these vaults merge-able.

yrdzrfxndfvh:
lol the devil daggers vault is terrible but yah you can fix them up if you want👍

regret-index:
Handled in 469dc525a82f345.

PR #3029, opened 2023-02-28
feat: Add menu_item_pad option

Antern:
Allows to disable padding in UI Menu for compact view.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -81,7 +81,7 @@ The contents of this text are:
                 action_panel, action_panel_filter,
                 action_panel_show_unidentified, action_panel_scale,
                 action_panel_orientation, action_panel_font_family,
-                action_panel_font_size
+                action_panel_font_size, menu_item_pad
 3-i     Colours (messages and menus).
                 menu_colour, message_colour
 3-j     Quivers, firing, and ammo.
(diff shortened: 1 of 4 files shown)


Antern:
Here is menu_items_pad = false
Screenshot 2023-02-28 at 03 35 07

rawlins:
Good idea! Was this accidentally closed? Some suggestions based on reading the code:

* By convention, this option should have a name like `tiles_menu_item_pad` and be organized in the `3-n     Tiles Options.` section of the options guide (since it'll never make sense to have this option on console builds).
* As a boolean, I would rather give it a more intuitive name like `tiles_compact_menus` (with corresponding inverted semantics) -- the current name makes me think you could set the exact pad value. (Which would be a bit fine-grained to expose to the user.)
* If you haven't yet, it's worth testing this on various non-inventory menus and with other settings, especially `tile_menu_icons = false`
* Implementation: I think it might be simpler just to conditionally set `item_pad` to 0 instead of adding conditionals everywhere that `item_pad` is used, though maybe I'm missing some details that this would mess up.

Antern:
@rawlins I've closed it, because it doesn't make much difference. 4 pixels per row in pretty specific conditions.
Thank you for the comments, I consider the tiles_compact_menus option with some addition tweaks, and will come back later with another PR if them will work.

rawlins:
ah, got it. Yeah, on just testing removing the pad I do agree it's subtly noticeable but minor. I wonder if what you'd want to do for this idea is scale down the icons? I think people may already use `tile_menu_icons = false` as a defacto compact mode, so if you could get that spacing just by rendering the icons at half size, it might do what you're aiming for here.

PR #3030, opened 2023-03-01
Updated contrast of floor tiles

belock-c:
Hello all,
New to the community here, though I've been playing dcss for around a decade now.

It came to my attention through webtiles that the contrast of all floor tiles are off by a value of 1 (too high).

You can plainly see from this comparison image that floor tiles should have lower contrast.

contrast-change

For your consideration.
Thanks.


PleasingFungus:
These screenshots are identical. Did you accidentally upload the same one twice?

rawlins:
We are a bit confused by this -- with some effort, we convinced ourselves that the uploaded tiles are minutely different, but it does not seem to be in a perceptible way (about 1-2 intensity, tending towards very dark colors that are not very distinguishable). Could you say more about why you think these tiles were "wrong" in the first place, and why/under what circumstances a change so small would make a difference?

belock-c:
Other than it affects core game graphics over hundreds of tiles? I'm not sure. I think asking how small a difference it makes is short-sighted. If it's wrong, fix it.

PleasingFungus:
I think it's already right. Thanks for the submission! :)

rawlins:
> If it's wrong, fix it.

Just one more followup, since it seems like there may still be some miscommunication here: you haven't said what you think makes these tiles wrong -- relative to what?

From my perspective, a tile is "right" if it's usable: is easily visible under a wide range of display circumstances, is easy for players to identify as what it is suppsoed to convey, looks good, for something like floor tiles in particular it isn't too dominant (doesn't make wall/monster tiles hard to spot), and other criteria along these lines. That's not to say that all the floor tiles are "right" by these criteria*, which may be quite subjective, but a non-perceptible difference in intensity doesn't affect any of these, and I don't see a reason why the current tiles could be argued to be wrong in terms of the change in this commit -- that is, I'm not sure of any other standard for rightness/wrongness that could be applied here.

* We definitely welcome commits like ba51bde10e9e30f49a99645756f77d466c46deb2 that do attempt to improve tiles along these dimensions!

belock-c:
> relative to what?

The other tiles.

belock-c:
For instance, look at the left edge of the mace. In the top image, it's the same 'depth' as the floor tile just to the left of it.
In the lower image, it sits on top.

rawlins:
Thanks for the response. Not exactly sure what you mean by relative to the other tiles, but I'm afraid I simply can't see the difference you're describing, the maces (probably you mean flails?) in that screenshot appear identical on any displays I've tried, and an image diff tool isn't showing anything either. Sorry..

PR #3032, opened 2023-03-01
Give monsters made of water insubstantial

Picklenugget1:
Gives water elementals and wellsprings insubstantial. It's quite difficult to grapple water, is it not?

My first pull request, I hope I got this right.


chucksellick:
Insubstantial monsters are generally ghosts, things with no substance at all: also I feel like the effects of being insubstantial are at odds with the effect of being SIZE_LARGE; and some of the more thematic properties that insubstantial gives you are already covered by being will invuln or nonliving.

Given the age of this PR and that monster flags have now moved to .yaml anyway I suggest we close this unless anyone disagrees.

regret-index:
The solution to this is to convert the current `mons_genus(foo) = MONS_JELLY` checks for barbs, nets, and constriction into a generalized `M_AMORPHOUS` flag that water elementals and elemental wellsprings could also use. Water elementals should not be particularly immune to petrification clouds or mud, out of the main differences between such a status and the insubstantial flag. Since this is not what this PR does, and since monster data indeed is now handled completely differently, I'm closing this PR (and leaving myself a note to look into doing such in the near future.)

PR #3044, opened 2023-03-09
fix: a few victory bugs

Implojin:
Some assorted Victory bugfixes that I ran into when looking into PF's meld bug; please see the commits for details.

fix: don't reset victory stats on meld
fix: reset victory stats when quaffing lig
fix: don't reset victory for cancelled scrolls

### DIFF
--- a/crawl-ref/source/item-use.cc
+++ b/crawl-ref/source/item-use.cc
@@ -2978,9 +2978,10 @@ bool drink(item_def* potion)
         return false;
     }
 
-    if (player_equip_unrand(UNRAND_VICTORY))
+    if (player_equip_unrand(UNRAND_VICTORY, true)
+        && !you.props.exists(VICTORY_CONDUCT_KEY))
     {
-        item_def *item = you.slot_item(EQ_BODY_ARMOUR);
+        item_def *item = you.slot_item(EQ_BODY_ARMOUR, true);
         string unrand_prompt = make_stringf("Really quaff with monsters nearby "
                                             "while wearing %s?",
                                             item->name(DESC_THE, false, true,
@@ -3037,7 +3038,7 @@ bool drink(item_def* potion)
     }
 
     // Drinking with hostile visible mons nearby resets unrand "Victory" stats.
-    if (player_equip_unrand(UNRAND_VICTORY)
+    if (player_equip_unrand(UNRAND_VICTORY, true)
         && there_are_monsters_nearby(true, true, false))
     {
         you.props[VICTORY_CONDUCT_KEY] = true;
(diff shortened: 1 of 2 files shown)


semitonal:
I wonder if the stats should reset on intentional meld, but not on unintentional meld? It feels like swapping into spider form to use consumables and then cancelling the form might be a weird tech - it does mean quaffing/reading that way would take extra turns, but that would definitely let you use blinking/teleportation pre-emptively without losing the bonuses for example.

crawl-ref/source/item-use.cc
```diff
@@ -3012,6 +3012,14 @@ bool drink(item_def* potion)
...
     bool heal_on_id = player_equip_unrand(UNRAND_DELATRAS_GLOVES);
 
+    // Drinking with hostile visible mons nearby resets unrand "Victory" stats.
+    // We need to check for this before potentially melding it.
+    if (player_equip_unrand(UNRAND_VICTORY)
+        && there_are_monsters_nearby(true, true, false))
+    {
+        you.props[VICTORY_CONDUCT_KEY] = true;
+    }
+
```
semitonal, line 3022: Hmm, looking into this a bit further it also looks like this introduces some new inconsistencies - if you abort quaffing a potion for some reason (e.g. after being prompted to go berserk while wielding a non-weapon, or to go invisible while backlit), it'll now reset the unrand stats due to this being moved to before `quaff_potion()`.

Implojin: Thanks for the feedback kate, I'll take another look at this.

Implojin:
> I wonder if the stats should reset on intentional meld, but not on unintentional meld? It feels like swapping into spider form to use consumables and then cancelling the form might be a weird tech - it does mean quaffing/reading that way would take extra turns, but that would definitely let you use blinking/teleportation pre-emptively without losing the bonuses for example.

So, I'd originally hoped to allow drinking/reading while melded -- it seemed like there was potential for fun discoverable techs there, and it would have softened the conduct slightly. The issues above convinced me that's not worth it, though.

semitonal:
I think there's another new weird interaction now unfortunately - quaffing lignification with monsters visible does prompt you and set `VICTORY_CONDUCT_KEY` correctly, but since the robe gets melded, it stops running its `world_reacts` func and doesn't reset the plus until you take another action once it unmelds (and this also means that you'll keep getting prompted to read/quaff things while lignified, even though it's due to lose the pluses). Feasibly the `world_reacts` funcs could also have a meld param and just return early for other unrands? Feels a bit messy but it does seem possible that other unrands might also still want to do something while melded (like applying god conducts if necessary?)

PleasingFungus:
I'd rather not add `meld` params for all unrands (for `world_reacts` or for `unequip`) until we need them for at least three unrands. If this is the only unrand that needs to have logic for acting while melded, it's much lower impact to special-case it.

semitonal:
The meld param for `unequip` feels like it'd already be useful for various unrands to reduce message spam (Condemnation, Olgreb, Firestarter - anything where the messaging is purely flavour rather than describing you gaining/losing an ability like Dragonskin, Kryia's, Night).

Special-casing lignification in PotionLignify::quaff() seems like it'd be the other way to go to make sure it resets the pluses at the right time.

Implojin:
Thanks again for the reviews folks, I really appreciate it.

Gonna go ahead and merge this, if any other issues crop up, lmk.

PleasingFungus:
Sounds good. I haven't re-reviewed but I have faith. 🙏

PleasingFungus:
Minor: please try to use rebases for this sort of simple PR in future. It keeps the commit history cleaner and easier to understand.

Implojin:
Yup, sorry about that!

PR #3046, opened 2023-03-10
Fix and improve CI builds

robertxgray:
### GENERAL

- Fix Github Actions deprecation warnings. 
  * The current ccache action cannot be updated, it has been replaced with the most starred one. This obsoletes the cache action and the  manual ccache stats management.
  * Update the checkout action to v3.
  * Update the setup-python action to v4.
  * Update the codecov action to v3.
  * Replace the obsolete set-outputs commands with the new GITHUB_OUTPUT environment file.

### ANDROID

- Update the Android Gradle plugin and use a compatible Gradle version.  This improves the build times and fixes the SDK XML related warnings.
- Ccache doesn't get hits on Android because the compiler's mtime  is different in every build. Change the compiler check strategy to  content.
- Proper fix for 093205a. I'm very sorry for that.
- Use the Maven Central repository instead of the deprecated JCenter.

### APPIMAGE
- Fix a typo in the cache config for the AppImage builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     name: Code Linting
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Checkwhite
         run: util/checkwhite -n
         working-directory: crawl-ref/source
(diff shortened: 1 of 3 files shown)


robertxgray:
> Just to check, what do you mean by this?

I mean a newer version of the action, compatible with Node.js 16, doesn't exist. I didn't know updating it ourselves could have been an option.

rawlins:
Got it, thanks. It's probably ok to move to something else rather than doing it ourselves, and I doubt @alexjurkiewicz will mind (in fact I believe I'd promised to move it a while ago and never acted, so this would resolve that issue...)

alexjurkiewicz:
Definitely no hard feelings on my part!

PR #3047, opened 2023-03-11
Convert draconian Tail Whip to a mutation

Monkooky:
Changes Draconian's tail whip attack to be realized as a mutation, allowing it's details to be viewed in the mutations menu.


regret-index:
Fixed in its own fashion in ade8af2 amongst the rest of the major form + mutation refactoring.

PR #3051, opened 2023-03-17
Android launcher and improved support for old devices

robertxgray:
**Android launcher**

A launcher for the Android port supporting new features.

- Simple editor for the init file.
- Morgue browser. View and download your morgue files.
- Keyboard selector.
  - Classic keyboard, same as before.
  - Transparent keyboard, classic with no background drawn on top of the game.
  - Android keyboard, whatever you have configured in the device by default.
  - No keyboard, handy option when a hardware keyboard is available.
- Optional extra mini keyboard with directional keys that can be placed on a corner.
- Configure the keyboard size, well, height only because we're using all the width available already.

**Improved support for old devices**

In theory, DCSS was playable on any device supported by SDL2, that's every version from Android 4.1. However, I tested it using virtual devices and it crashes. This release fixes those crashes and uses the AppCompat library and themes for proper rendering on any supported version.

There are other minor changes, please see the commit descriptions for details.

https://user-images.githubusercontent.com/9676118/225873293-e85e9dd1-c632-42a7-82aa-926ee2a88ee0.mp4


crawl-ref/source/package.cc
```diff
@@ -676,7 +676,9 @@ plen_t package::get_chunk_compressed_length(const string &name)
 chunk_writer::chunk_writer(package *parent, const string &_name)
     : first_block(0), cur_block(0), block_len(0)
 {
+#ifndef __ANDROID__
```
rawlins, line 679: conditionally defining this assert seems iffy, since if it fails subsequent code will probably segfault...

If it's really ok for this to be called with `nullptr` parent, probably it should have an early return or following code (e.g. `pkg` on line 690) needs something else

crawl-ref/source/tags.h
```diff
@@ -65,7 +65,9 @@ class writer
         : _filename(), _file(0), _chunk(0), _ignore_errors(false),
           failed(false)
     {
+#ifndef __ANDROID__
```
rawlins, line 68: same worry here, the next line will segfault if this assert fails

rawlins (commented):
just a couple of questions about impacts on package.cc

crawl-ref/source/package.cc
robertxgray: You are right, sorry. I shouldn't have accepted the first solution I found for the crash logs. After further testing I found the root cause of the problem. New solution in 6d99694.

robertxgray:
The launcher has been available on Google Play for more than 15 days and the feedback is very positive. I'll merge it soon if nobody objects.

rawlins:
Sorry, I didn't get back to this before but merging it looks good to me!

PR #3053, opened 2023-03-22
Fix and refactor thunderclouds

### LINKED ISSUE #3050: Thunder Cloud damage scales inversely with action speed
After observing very anemic thunder cloud damage while using Zephyr, I dove the code to try and argue for vane changes, and instead found what I suspect is a bug rendering thunder clouds bad. 

Thunderclouds are meant to trigger roughly once every 30 auts. The mechanism for this is checking 
```x_chance_in_y(you.time_taken, aut_per_lightning)```
(cloud.cc 1284)
when clouds would deal damage, and only doing damage if this passes.

Monsters suffer cloud damage when they act. (Mon-act.cc 2361, 1705, 3635)

Thus, if a monster is acting once every 10 auts, and the player is acting once every 3, then every 10 auts the monster will have a 3/30 chance of taking thunder damage, averaging out to one strike every 100 auts. 

Testing this with a juggernaut vs a quickblade had two lightning strikes occur in 270 auts

### PR DESCRIPTION
**Pull request by Monkooky**
Fixes #3050

Refactors thunderclouds to minimize their special casing. Adjusts thundercloud damage to on average equal normal cloud damage.

### DIFF
--- a/crawl-ref/source/cloud.cc
+++ b/crawl-ref/source/cloud.cc
@@ -1253,7 +1253,7 @@ static int _actor_cloud_damage(const actor *act,
     return timescale_damage(act, final_damage);
 }
 
-void _actor_apply_cloud(actor *act, cloud_struct cloud)
+void _actor_apply_cloud(actor *act, cloud_struct &cloud)
 {
     const bool player = act->is_player();
     monster *mons = act->as_monster();
--- a/crawl-ref/source/cloud.cc
+++ b/crawl-ref/source/cloud.cc
@@ -254,7 +254,7 @@ static const cloud_data clouds[] = {
       ETC_DARK,                                 // colour
       { TILE_CLOUD_STORM, CTVARY_RANDOM },      // tile
       BEAM_ELECTRICITY,                         // beam_effect
-      {12, 12},
+      {19, 18},
     },
     // CLOUD_NEGATIVE_ENERGY,
     { "negative energy", nullptr,               // terse, verbose name
@@ -1053,20 +1053,6 @@ static bool _actor_apply_cloud_side_effects(actor *act,
     case CLOUD_STEAM:
         if (player)
             maybe_melt_player_enchantments(BEAM_FIRE, final_damage);
-    case CLOUD_RAIN:
-    case CLOUD_STORM:
-        if (act->is_fiery() && final_damage > 0)
-        {
-            if (you.can_see(*act))
-            {
-                mprf("%s %s in the rain.",
-                     act->name(DESC_THE).c_str(),
-                     act->conj_verb(silenced(act->pos())?
-                                    "steam" : "sizzle").c_str());
-            }
-        }
-        break;
-
     case CLOUD_MEPHITIC:
     {
         if (player)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: crawl-ref/source/cloud.cc [hunk 1]
```diff
@@ -1254,91 +1240,27 @@ static int _actor_cloud_damage(const actor *act,
...
         break;
     }
 
     return timescale_damage(act, final_damage);
 }
 
-// Applies damage and side effects for an actor in a cloud and returns
-// the damage dealt.
-int actor_apply_cloud(actor *act)
+void _actor_apply_cloud(actor *act, cloud_struct cloud)
```
PleasingFungus, line 1256: Will do a full review of this later but a quick note: you're passing this struct by value, meaning you copy the entire thing whenever you call this function. You want to pass it by reference instead. (`cloud_struct &cloud`).

Monkooky: Corrected, thanks!

crawl-ref/source/cloud.cc
```diff
@@ -1254,91 +1240,27 @@ static int _actor_cloud_damage(const actor *act,
...
         break;
     }
 
     return timescale_damage(act, final_damage);
 }
 
-// Applies damage and side effects for an actor in a cloud and returns
-// the damage dealt.
-int actor_apply_cloud(actor *act)
+void _actor_apply_cloud(actor *act, cloud_struct &cloud)
```
PleasingFungus, line 1256: Functions that are only used in one file, like this one, should be declared `static`. Otherwise, they'll be global and usable in all files, which can result in really baffling errors.

crawl-ref/source/cloud.cc
```diff
@@ -1383,10 +1303,27 @@ int actor_apply_cloud(actor *act)
...
+// the damage dealt.
+void actor_apply_cloud(actor *act)
+{
+    cloud_struct* cl = cloud_at(act->pos());
+    if (!cl)
+        return;
+
+    cloud_struct cloud = *cl;
+
+    if (cloud.type == CLOUD_STORM){
```
PleasingFungus, line 1318: Someone forgot to run `util/unbrace` 😛. (Or `util/checkwhite`, for that matter...)

PleasingFungus, line 1318: Minor: we use [Allman](https://en.wikipedia.org/wiki/Indentation_style#Allman_style) as our house style. Braces on the next line. (Not my preference, but we gotta have a standard; otherwise it's chaos!)

crawl-ref/source/cloud.cc
```diff
@@ -1383,10 +1303,27 @@ int actor_apply_cloud(actor *act)
...
-    return final_damage;
+// Applies damage and side effects for an actor in a cloud and returns
+// the damage dealt.
+void actor_apply_cloud(actor *act)
+{
+    cloud_struct* cl = cloud_at(act->pos());
+    if (!cl)
+        return;
+
+    cloud_struct cloud = *cl;
```
PleasingFungus, line 1316: This does a copy of the entire cloud struct every time we make this call. Seems unnecessary. See 298cdec9ad for how I avoided this.

PleasingFungus (approved):
Thanks for cleaning this up! Apologies for the delay in the full review - I've had zero dev time all week.

PR #3061, opened 2023-03-27
Revert Nonsensical Gender Removals

Darkpixel797:
Starting small with my first commit, just some flavortext and tagging changes. When browsing the Crawl Wiki, I noticed that a few monsters had their genders removed in 0.24 for seemingly no reason (except maybe Dissolution as he became a slime). This reverses the gender removals (male to neutral and female to neutral) in commit ec079b9.


rawlins:
I have no idea why you think we'd accept this

Darkpixel797:
I thought it was worth a shot given the guy who originally made the changes seemed to drop off the face of the earth.

Darkpixel797:
That's bad advice. Its nearly impossible to change people's minds, this goes tenfold for people you don't know irl or people in positions of power. Asking anyone about it was never gonna do anything except make arguments that go nowhere. I was just checking with this PR to see that after the original guy who made the changes is gone if anyone still stood by said changes, and apparently they do.

PleasingFungus:
If you re-read my post, at no point did I suggest that you "change people's minds". Nor did I suggest you start arguments. My suggestion was purely that you try to understand the change which you described as "nonsensical" before you tried to revert it. Perhaps that is too much to ask. :)

Darkpixel797:
I'm pretty sure I do understand why said change was made initially, it was either to test out the newly revamped genderless tag at the time (in which case some of the monsters he applied it to didn't make sense as candidates for the tag and those are the ones I reverted, Nessos literally references that fact that he likes mares in another dialogue line for example). The other possibility is that it is some sort of political statement, and I disagree with political statements in games that don't enhance the experience. So either way I don't agree with the reasons, aka that they don't make sense, hence the word "nonsensical". If you don't agree with me that's fine, you're the one with push access after all, I hold no power here, but don't pretend that if I'd approached things differently it could have gone through. I was just checking if there was agreement or not, and there is not.

PleasingFungus:
Neither of your hypotheses are correct, unfortunately.

> So either way I don't agree with the reasons, aka that they don't make sense, hence the word "nonsensical".

This PR description claims that the changes were made for "seemingly no reason", not that you disagreed with the reasons you hypothesized. :)

> ...don't pretend that if I'd approached things differently it could have gone through...

I have never claimed anything of the sort.

Darkpixel797:
Enlighten me to the true reasons then.

Fine, I should have put "no good reason" instead.

Then what was your first comment about, were you telling me to spend more time talking with people about it to get rejected anyway? I'd rather just get the answer immediately, and ideally professionally instead of sarcastically but oh well.

PR #3067, opened 2023-04-01
Three fixes for release CI workflows

### LINKED ISSUE #1344: local tiles: Don't Hardcode Font
local tiles, OS: Linux

**DCSS VERSION:**
0.24.1

**DESCRIPTION:**
Is there a reason why DCSS tiles is hard-coded to dejavu mono? Can it use fontconfig instead like most other applications do these days? There are fonts that are metric-compatible with dejavu, like bitstream-vera, and DCSS hard-refuses to launch tiles without dejavu.

robertxgray:
1. I made a fix for an old issue: #1344.  The implementation is very similar to the one used in the Fedora packages. While testing, I found two errors in release workflows.

2. AppImage builds can't run linuxdeploy because FUSE 2 is not available by default. The GitHub runners must have changed, they're probably using FUSE 3 now. The solution is easy because v2 is still in the repos and can be installed alongside v3.

3. The migration from the obsolete set-output commands in 94e0f26 was incorrect. My fault, sorry :disappointed:

Closes #1344

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -364,8 +364,8 @@ jobs:
             source='crawl-ref/source/stone_soup-latest-win32-installer.exe'
             dest='dcss-$tag-win32-installer.exe'
           fi
-          echo "name=source::$source" >> $GITHUB_OUTPUT
-          echo "name=dest::$dest" >> $GITHUB_OUTPUT
+          echo "source=${source}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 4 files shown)


robertxgray:
I forgot, this is the latest test:
https://github.com/robertxgray/crawl/actions/runs/4579905032

And this is the output:
https://github.com/robertxgray/crawl/releases

robertxgray:
I'm going to go ahead and merge this because the fixes are necessary and they have been tested. We can blame the newbie dev if something breaks.

PR #3068, opened 2023-04-01
Check if the cursor is enabled before saving it's position

### LINKED ISSUE #1497: ASSERT(valid_cursor_pos(pos.x, pos.y, region)) in 'libconsole.h' at line 58
Linux 3.16.0 console

**DCSS VERSION:**
0.25.1 (using the .deb from http://crawl.develz.org)

**DESCRIPTION:**
I start the game, and load the character up. I select "Maximize vertical". This makes the window about 80x90, but the game crashes with the message:

ASSERT(valid_cursor_pos(pos.x, pos.y, region)) in 'libconsole.h' at line 58 failed. (invalid cursor position 40,-56 in region 1)

If I start the game with an 80x90 window, it opens correctly, but making the window smaller, and then larger again, also makes it crash. If I make the window narrower or wider, even if I make it higher at the same time, it doesn't crash.

**EXTRA FILES:**
[crash.zip](https://github.com/crawl/crawl/files/4976624/crash.zip)
This contains the save file and the crash report.

robertxgray:
This avoids crashes on old terminal emulators like xterm or urxvt.

Closes #1497

### DIFF
--- a/crawl-ref/source/libutil.cc
+++ b/crawl-ref/source/libutil.cc
@@ -447,8 +447,11 @@ void set_cursor_region(GotoRegion region)
 
 void clrscr()
 {
-    save_cursor_pos save; // is this generally correct? alternative, set region
-                          // to GOTO_CRT.
+    if (is_cursor_enabled())
+    {
+        save_cursor_pos save; // is this generally correct? alternative, set region
+                              // to GOTO_CRT.
+    }
     clrscr_sys();
 
 #ifdef USE_TILE_WEB


crawl-ref/source/libutil.cc
```diff
@@ -447,8 +447,11 @@ void set_cursor_region(GotoRegion region)
 
 void clrscr()
 {
-    save_cursor_pos save; // is this generally correct? alternative, set region
-                          // to GOTO_CRT.
+    if (is_cursor_enabled())
+    {
+        save_cursor_pos save; // is this generally correct? alternative, set region
```
rawlins, line 452: This won't do what you want; the `save` object is created and goes out of scope within this conditional (so position is never restored after `clrscr_sys`). In general, for conditional RAII you would need to duplicate the calls that the RAII object would go out of scope after, or modify the RAII object to handle conditions itself. Probably the former is easier here.

(I should say I'm not entirely sure if this is the right solution: usually crashes in this object are a symptom rather than a cause. But I haven't done any debugging of this particular case.)

robertxgray: Oh my, It worked like a charm in my tests. Let me check it again.

robertxgray:
Thank you very much for the review. We have another conversation in the issue so I'm closing this one.

PR #3071, opened 2023-04-01
make barachi cold blooded

yrdzrfxndfvh:
it just makes sense thats all i gotta say


dillytin:
should make barachim's base movement speed faster too!

semitonal:
Closing - realism isn't an aim here, and cold-bloodedness isn't a particularly good mechanic to add to more player species.

PR #3080, opened 2023-04-08
Add some new shops to shops.des

nicolae-carpathia:
nicolae, making shop vaults? How surprising!

Most of these are simple generic shop vaults, intended to help increase the amount of simple, non-gimmicky shops to increase variety a little. Also increase the weight on the very basic "shop" vault, so that simple one-tile shops get a little more representation. There are also a couple of multi-shop vaults.

I told myself I was just going to stick to simple, non-themed vaults, with no specially-defined inventories, but during a discussion where memes of antiquity came up, I was inspired. (And then after I broke the "only simple vaults" rule, I was inspired again, this time by one of my other shops.)

### DIFF
--- a/crawl-ref/source/dat/des/builder/shops.des
+++ b/crawl-ref/source/dat/des/builder/shops.des
@@ -815,7 +815,7 @@ ENDMAP
 
 NAME:   nicolae_shop_wares_on_display
 TAGS:   transparent shop extra allow_dup
-DEPTHS: !D:1-3
+DEPTH: !D:1-3
 WEIGHT: 2
 KFEAT:  S = general shop
 MAP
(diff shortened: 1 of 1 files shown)


crawl-ref/source/dat/des/builder/shops.des
```diff
@@ -500,6 +499,333 @@ x-----x
...
+NSUBST: w = @ / w
+MAP
+www
+wSw
+www
+ENDMAP
+
+NAME:   nicolae_shop_wares_on_display
+TAGS:   transparent shop extra allow_dup
+DEPTH: !D:1-3
```
geekosaur, line 818: Does this need to be something like `D, !D:1--3`?

PR #3088, opened 2023-04-13
Add some new Slime Pit endings

nicolae-carpathia:
slime_pit: The main change was splitting out the various functions for making the clear walls disappear when TRJ dies. I also removed some downstairs glyphs, which were in there for some reason. I removed the "no_dump" tag, since now that there are variants, players might want to see which one they got. I also trimmed some excess walls off the edges to get the width less than 70 tiles, allowing the vault to be rotated, and accordingly removed the "no_rotate" tag.

slime_pit_nicolae_amoeboid: A variation on the basic slime_pit theme -- stairs near the outer edge of slime walls, with stone, loot, and TRJ in the center. The slime walls and loot boxes are of course shaped differently, and the corridors in between them here aren't as narrow as in slime_pit original recipe.

slime_pit_nicolae_biohazard: A concept of 3-towards-1, in which there are three distinct paths to the center, and going from one upstair to another requires going through the TRJ zone. This specific layout is inspired by the negative space in the biohazard symbol; hence the name.

slime_pit_nicolae_inversion: Switching it around a little: stairs near an outer edge of stone walls, with loot and other decorative bits around the outer edge. The center holds slime walls and TRJ. As the comment notes, there is usually an acidless path to the center, but it's not guaranteed. But standing near an acid wall doesn't have long-term effects anymore, so it'll be fine. Probably. The distance between the loot vaults will no doubt make it harder for Jiyvites to get the good stuff before it gets munched. I do not consider this a problem.

slime_pit_nicolae_jellycorners: A smaller version of slime_pit_nicolae_rhizomes. Both layouts are built around the same idea: instead of starting near the edge and working one's way inwards, you start at the center and work your way outwards. Here, TRJ is lurking in one of the square stone ruins in the corners. The decaying stone walls might end up providing a little niche to hide in, but also maybe not.

slime_pit_nicolae_rhizomes: Made before slime_pit_nicolae_jellycorners, on the same basic concept. At first I thought it might be too large to work, since the player has to hunt all around the edge for TRJ, but reconsidered and thought it might be fine. We'll see.

slime_pit_nicolae_royal_road_to_ruin: For a bit, I considered a layout that was mostly just an A-to-B path, rather than mostly open areas the player can wander through; this concept was rejected and the map I'd made was turned into nicolae_slime_spelunking. The loot/TRJ section from that was kept and reused here. Here, the three stairs start in a mostly-open area, and then converge on the center together, turning into a short A-to-B path. Like slime_pit_nicolae_jellycorners, players will walk through decaying ruins of the pre-Slime civilization, which may or may not provide fightin' niches.

slime_pit_nicolae_trefoil: Another variation of the 3-towards-1 theme from slime_pit_nicolae_biohazard. Here, there's quite a bit more loot, so there are more monsters to compensate, and not many good places to hole up.

For all the new vaults, I mostly kept the same amount and ratios of monsters (types, numbers) and loot (* relative to |) as in slime_pit. That stuff is a pain to balance, and I figured it'd be easiest to go with what I know. I also didn't bother weighting any of the vaults; that can come later with feedback.

### DIFF
--- a/crawl-ref/source/dat/des/branches/slime.des
+++ b/crawl-ref/source/dat/des/branches/slime.des
@@ -38,6 +38,50 @@ function fixup_slime_vaults(data, triggerable, triggerer, marker, ev)
 
   triggerable:remove(marker)
 end
+
+function setup_slime_pit_ending(e)
+  e.place("Slime:$")
+  e.orient("encompass")
+  e.mons("the Royal Jelly")
+  e.mons("acid blob")
+  e.mons("great orb of eyes / nothing")
+  e.kfeat("Z = altar_jiyva")
+  e.kitem("O = slimy rune of Zot")
+
+-- Attach function to the Royal Jelly itself. This will follow it wherever
+-- it goes, even through polymorphs, and set dgn.persist.fix_slime_vaults
+-- when it dies through means other than being banished.
+  local function trj_die(monster, killer_type, killer_index, silent, wizard)
+      if killer_type == "banished" then
+        crawl.mpr("You feel a great sense of loss, and a brush of the Abyss.")
+      else
+        dgn.persist.fix_slime_vaults = true
+      end
+  end
+
+  local wall_fixup_marker = TriggerableFunction:new( {
+      func="fixup_slime_vaults",
+      repeated=true
+  })
+
+  wall_fixup_marker:add_triggerer(DgnTriggerer:new {
+    type="monster_dies",
+    target="any"
+  })
+
+  wall_fixup_marker:add_triggerer(DgnTriggerer:new {
+    type="entered_level"
+  })
+
+  e.lua_marker("1", MonPropsMarker:new {monster_dies_lua_key = trj_die})
+  e.lua_marker("Z", wall_fixup_marker)
+  e.kmask("|O* = no_monster_gen")
+  e.kprop("|O* = no_jiyva")
+  e.kprop("|O* = no_tele_into")
+  e.shuffle("{[(")
+  e.set_feature_name("stone_wall", "rune-carved stone wall")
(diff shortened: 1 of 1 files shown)


PleasingFungus:
image

image

image

image

image

LiciTheCrawler:
If we say 1 is too easy, 5 is the classic slime end and 10 is very hard.

slime_pit_nicolae_amoeboid: 7.5 - TRJ comes out easily and quickly, there are no chokes to fight in and TRJ can exit LoS quickly becasues of many corners

slime_pit_nicolae_biohazard:  TRJ came out of the middle by the time I cleared out the monsters outside of the vault, the layout makes it really costly to use other stairs (need teleport, digging or good lucky with other slime level layouts). used torment strats for the kill. Maybe a 9/10

slime_pit_nicolae_inversion: mostly the same as the old one except for the lack of an easy choke point for fighting. 6.5/10

slime_pit_nicolae_jellycorners: player can make nice killholes with only a few corr stacks in the middle, TRJ also doesn't get dragged to the player quickly so the player can make the floor safer before the fight. 5.5-6/10

slime_pit_nicolae_rhizomes: same as the one above!

slime_pit_nicolae_royal_road_to_ruin:  easy to make a safe space outside of the vault, the fight in the vault can be messy, but there are some places where the player get's only a level or two of corrosion from the walls, but getting to them fast enough might be hard.

slime_pit_nicolae_trefoil: the triangles make for a bit safer place to fight, but the open layout can put the player into some bad spots and making the level safe before the big fight might be hard. 8/10

Overall I feel like strategies like lig + torment + immolation with stair camping might become more desirable with the new layouts over fighting TRJ head on. The testing was done on a MiFi^Hep.

PR #3089, opened 2023-04-14
FIX: Put both DescMenu classes in unnamed namespaces.

### LINKED ISSUE #2824: Lookup descriptions submenus crash on exit
Linux 3.16.0 console using XTerm(358)

0.30-a0-359-gd0147acb1a

**DESCRIPTION:**
If I start crawl, and type `? / g Escape` (so go into one of the help menus and back out again) the game crashes. It does the same thing if I have a character loaded, or if I look at one of the god descriptions before leaving the menu.

Replacing `DescMenu` with `InvMenu` in `_full_describe_menu()` stops it from happening. I've attached a patch to show what I mean, but I'm sure there's a better way to correct this problem.

On an unrelated note, I notice that the crash message states that a save file called `.cs` has been created, but none actually is.

**EXTRA FILES:**
[crash log](https://github.com/crawl/crawl/files/9993651/crash.zip)
[patch to replace DescMenu](https://github.com/crawl/crawl/files/9993410/dif.txt)

Aliscans:
This resolves any confusion between them, and fixes #2824.

This is apparently what you do if you only want a class to be used in its own .cc file.

I also deleted some code which a comment said was to do with DescMenu being an InvMenu. It shouldn't be needed as that DescMenu isn't an InvMenu.

### DIFF
--- a/crawl-ref/source/directn.cc
+++ b/crawl-ref/source/directn.cc
@@ -542,23 +542,26 @@ public:
     }
 };
 
-// XX this probably shouldn't use InvMenu, why does it?
-class DescMenu : public InvMenu
+namespace
 {
-public:
-    DescMenu()
-        : InvMenu(MF_SINGLESELECT | MF_ANYPRINTABLE
-                        | MF_ALLOW_FORMATTING | MF_SELECT_BY_PAGE
-                        | MF_INIT_HOVER)
-    { }
-
-    // TODO: move more stuff into this class
-    bool skip_process_command(int) override
+    // XX this probably shouldn't use InvMenu, why does it?
+    class DescMenu : public InvMenu
     {
-        // override InvMenu behavior
-        return false;
-    }
-};
+    public:
+        DescMenu()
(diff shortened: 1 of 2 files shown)


robertxgray:
Have you considered renaming one of the classes instead? I think it's a simpler solution.

Aliscans:
> Have you considered renaming one of the classes instead? I think it's a simpler solution.

Either would work. I chose this option because it seemed like a better technical solution. With this approach, you can easily add more types of `DescMenu` in the future without reintroducing the problem. Changing the name just adds another name you need to avoid for any new types.

Would you prefer it if those namespace sections weren't indented? The convention for indenting namespaces isn't written down, so I could easily say that a namespace clause with one thing in it doesn't need it.

Neither approach solves the underlying problem, i.e. that there's no warning if you reuse a type name in the global namespace. ctags does most of the work to identify reused type names, so I wonder if there's a lint tool you could add to the github workflow which would look for this sort of thing.

robertxgray:
Thank you very much for the contribution.

PR #3107, opened 2023-04-26
Fix Lua bug in `view.cell_see_cell`

mgdelmonte:
The lua function `view.cell_see_cell` uses what appears to be an old and also currently invalid test for visibility, using `exists_ray`.  This test fails for situations such as when there are two plants obscuring a cell (the "half-opaque" rule).

For example, in the image below, the function currently returns "true" when asked whether any cell at (0,0), (0,-1), (0,-2), or (0,-3) can see the cell at (0,6), which is behind two plants.

image

This PR fixes the bug by calling the proper `cell_see_cell` function from `losglobal.h`.

It also removes the unnecessary bounds checks, as these are already done by `cell_see_cell`.


PleasingFungus:
Looks good, thanks! Merged. Vaguely worried about potential info leaks here, but those exist in the old code as well.

Apologies for the delay :)

PR #3120, opened 2023-05-05
Fix some cases of autopickup not being toggled on and off correctly

WizardIke:


crawl-ref/source/traps.cc
```diff
@@ -740,6 +741,17 @@ void trap_def::trigger(actor& triggerer)
...
             know_trap_destroyed = true;
             trap_destroyed = true;
+
+            // If we shaft an invisible monster reactivate autopickup.
+            // We need to check for actual invisibility rather than
+            // whether we can see the monster. There are several edge
+            // cases where a monster is visible to the player but we
+            // still need to turn autopickup back on, such as
+            // TSO's halo or sticky flame.
+            if (triggerer_was_invisible_monster)
```
wheals, line 751: Is there a reason this doesn't check `!you.can_see_invisible` like the other cases?

WizardIke: Auto toggling autopickup back on shouldn't do anything if you can see invisible as autopickup will either be on or manually turned off. I check `!you.can_see_invisible` when removing an enchantment so the message that the monster appears from thin air is correct. I could remove the check when polymorphing a monster as it is unnecessary or add it hear if it makes the code clearer.

wheals: I think removing the check when polymorphing would be best.

WizardIke: ok, removed the unnecessary check

wheals:
Thank you, merged in https://github.com/crawl/crawl/compare/aad2c9efd125...926718a23ee4 ! How would you like to be credited in CREDITS.txt?

WizardIke:
Credit me as Wizard Ike

wheals:
Done!

PR #3123, opened 2023-05-07
Changes in release CI workflows

robertxgray:
- Debian packages will be included in the standard release as a single .tar.xz file.
- Allow any version format in Debian packages so the workflow can be executed in beta releases.
- New workflow added for source packages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,29 @@ jobs:
         run: util/unbrace -n
         working-directory: crawl-ref/source
 
+  build_source:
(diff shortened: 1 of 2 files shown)


robertxgray:
Latest test:
https://github.com/robertxgray/crawl/releases/tag/0.0-b1
https://github.com/robertxgray/crawl/actions/runs/4908356116

robertxgray:
The CI checks now work here, but the new source build broke with the rebase. I'm going to close this so I can continue testing without generating noise.

PR #3130, opened 2023-05-12
Fix: Loss of precision when calculating max zoom

robertxgray:
While playing 0.30 with Barachi on my phone, I noticed the max zoom has been reduced from previous versions. It seems the refactor to FixedpGameOption introduced a loss of precision in the division.


robertxgray:
My mistake, I was cleaning old branches.

PR #3134, opened 2023-05-15
make tyrant leech cold blooded

yrdzrfxndfvh:
justification: they should be slowed by cold because they are watery squishy worms


PleasingFungus:
...but worms and swamp worms shouldn't be..? Very confused.

PR #3142, opened 2023-05-21
Fix the description of walls with unseen changes (take two)

robertxgray:
Temporary terrain changes produced by the summon forest spell can
affect dungeon features outside of the player's LOS. Those features
lose the flavour when described or if the dungeon view is reloaded.
The description is also wrong for the summoned trees after the spell
finished.
    
The first version of the patch had to be reverted because it crashed
on some vaults with loot on top of custom floor.


robertxgray:
This bug is still present on trunk, I've solved the conflicts.

Steps to reproduce:

1. Go to a wizlab with fancy walls.
2. Cast the summon forest spell.
3. Examine walls outside LOS. The tile is wrong if they're now trees.

Captura de pantalla -2024-03-29 10-57-34

PR #3143, opened 2023-05-21
Fix zoom out

robertxgray:
Prevents the current scale to be much higher than the maximum zoom. Otherwise zoom out stops working until a lower value is reached again.


rawlins:
Thanks, this looks right to me! I'm going to go ahead and merge this..

Issue #3139, opened 2023-05-20
strange machines do not get auto excluded


yrdzrfxndfvh:
**PLATFORM and OS or BROWSER:**
webtiles (likely not relevant)

**DCSS VERSION AND GAME INFORMATION:**
0.30

**DESCRIPTION:**
strange machines (the things in tukimas studio) do not get marked as unsafe like oklob plants and other statues

PleasingFungus:
Fixed in https://github.com/crawl/crawl/commit/2e3c275290d99f6ced86436532b9e5b7ed0a80f4, thanks!

PR #3148, opened 2023-05-25
Auto-exclude Strange Machine by default

sandertyu:
Resolves issue #3139. This makes the default auto-exclude behavior for strange machines consistent with other dangerous stationary monsters like oklob plants, statues, etc.


PleasingFungus:
Thanks!

PR #3149, opened 2023-05-25
Fix "I'll leave it outside" shop message.

crystalsolenoid:
When purchasing items from a shop when you don't have sufficient inventory space, a message is supposed to appear that notifies you that the shop keeper will put the items on the ground for you.

This message was not appearing because it was being immediately reset with line 1328 (new line numbering) before ever getting drawn.

I've written a fix that changes the behavior of update_help depending on if the last command caused items to be left outside. I'm not so sure that that's the cleanest way to fix this, but does work. If there's a better way to do this, I'm happy to work on the changes with guidance.

### DIFF
--- a/crawl-ref/source/shopping.cc
+++ b/crawl-ref/source/shopping.cc
@@ -816,6 +816,9 @@ class ShopMenu : public InvMenu
     level_pos pos;
     bool can_purchase;
 
+    int outside_items;
+    vector<int> bought_indices;
+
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Very sorry for the belated review! I agree that I'm not sure this is the cleanest fix, but I don't trivially see a better approach. Merged as https://github.com/crawl/crawl/commit/e0ecf0ced4cd3b2e50f3ff8d1ad904b41a912d72 - thanks!

Issue #3133, opened 2023-05-15
tile_shield_offsets not working


senlim:
**PLATFORM and OS or BROWSER:**
CPO webtiles on Google Chrome.

**DCSS VERSION AND GAME INFORMATION:**
Version 0.30

**DESCRIPTION:**
The tile_shield_offsets rc file option doesn't appear to be working. Shields appear in the default position no matter what values are entered. I asked in the Roguelikes Discord, and gammafunk couldn't seem to get it to work for him either.

sandertyu:
I've bisected the history and found that the problematic code is somewhere in between commit [14641e6](https://github.com/crawl/crawl/commit/55b71c9664861fa08b86f4190ed464a9fba3c635) and commit [55b71c9](https://github.com/crawl/crawl/commit/14641e6071dc92e3300ad0fe1399028fbe3a9ab1). The `tile_shields_offsets` value functions normally in commit 14641e6, and then does not function in 55b71c9. 

The problem is that none of the commits in between these ones seem to compile. I'll continue to try some things, but I think someone more experienced should probably take a look.

PR #3150, opened 2023-05-26
Fix tile_shield_offsets option

sandertyu:
Resolves issue #3133 . I think there was simply a typo/oversight to put `tile_weapon_offsets_option` instead of `tile_shield_offsets_option`. Either way, the `tile_shield_offsets` option now works as far as I can tell.

Here is an example with `tile_weapon_offsets = 20,10` and `tile_shield_offsets = -20,-10`

player tile with weapon in the bottom right and shield in the top left


PleasingFungus:
Nice fix! Apologies for the delay in merging.

PR #3152, opened 2023-05-28
Bugfix: Roots require solid ground

bobross419:
Wand of roots and the grasping roots spell are allowed to constrict on non-solid ground.  The wand's description explicitly states the effect requires solid ground.  I tested locally with the wand, but no matter how long I stood there Jorgrun would not cast it at me standing over deep water.  Jorgrun was observed rooting me over deep water in 0.31-a0-108-gdf935468f2 which prompted this PR.

Screen Shot 2023-05-27 at 8 24 39 AM

### DIFF
--- a/crawl-ref/source/actor.cc
+++ b/crawl-ref/source/actor.cc
@@ -708,7 +708,7 @@ bool actor::can_constrict(const actor &defender, constrict_type typ) const
     if (!see_cell_no_trans(defender.pos()))
         return false;
 
-    return typ != CONSTRICT_BVC
+    return (typ != CONSTRICT_BVC && typ != CONSTRICT_ROOTS)
            || feat_has_solid_floor(env.grid(defender.pos()));
 }
 


crawl-ref/source/actor.cc
```diff
@@ -708,7 +708,7 @@ bool actor::can_constrict(const actor &defender, constrict_type typ) const
     if (!see_cell_no_trans(defender.pos()))
         return false;
 
-    return typ != CONSTRICT_BVC
+    return (typ != CONSTRICT_BVC && typ != CONSTRICT_ROOTS)
```
PleasingFungus, line 711: There are three constrict types: MELEE, BVC, and ROOTS. We know this isn't MELEE because of lines 702-706, so `typ != CONSTRICT_BVC && typ != CONSTRICT_ROOTS` will never return true. This is equivalent to `return false`.

The correct fix is to delete the `typ` check. This returns the relevant logic to what it was before 57a6a0d306a.

PleasingFungus (commented):
Hi, thanks for reporting!

PleasingFungus:
Apologies for the delay in review. I've merged a fix.

PR #3155, opened 2023-06-01
build: workaround for an actions/checkout bug

rawlins:
a9a6e9e7af1 and ea43df2a32 removed a step that was probably intended to work around https://github.com/actions/checkout/issues/290 and https://github.com/actions/checkout/issues/882, but broke the CI build if anyone used a lightweight tag (because these fetch lines actually shallow the tag depth to 1). This commit attempts to thread the needle by making this workaround conditional only to releases, where (if the person doing the release has followed the steps correctly), there is guaranteed to be an annotated tag at depth 1 in the tag history. (If I'm parsing the checkout bugs correctly, they only trigger on a tag action anyways.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,8 @@ jobs:
         with:
           fetch-depth: 0 # all history
           submodules: true
+      - name: Fix tags for release # work around https://github.com/actions/checkout/issues/882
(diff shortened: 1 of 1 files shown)


rawlins:
Fake test release: https://github.com/rawlins/crawl/releases/tag/0.99.2, everything seems to work as intended

(@robertxgray not sure if you wanted to look this over)

robertxgray:
It looks good to me, thanks! :thumbsup:

PR #3176, opened 2023-07-01
Tweak forms

PleasingFungus:
This change is intended to allow more opportunities for players to shift into or out of a 'transmuter' playstyle, to improve the UI of forms, and to improve some miscellaneous issues, e.g. Lichform being useless in 3-rune games. For more context, see https://github.com/crawl/crawl/wiki/Transmutations-Reform.

Throughout, balance is a very rough sketch. I expect many things will need to be buffed, others will need to be nerfed, and some will need to be replaced entirely. This is a grand experiment, not a final state.

Talismans
---------

The largest change is that forms are no longer entered via spells. Instead, special items called 'talismans' must be found and evoked. Once entered, these 'talisman forms' last indefinitely.

Further notes on talismans:
- Talismans scale only on Transmutations skill. They do not care about Int, Spellcasting, other spell schools, wizardry, or encumbrance. (That is, they aren't spells.)
- Talisman forms have a 'minimum skill'. Below that skill, entering the talisman form will reduce the user's maximum HP (while in the form). This is intended to roughly mimic the inability to effectively cast spells at low skills/high fail% - it provides a space in which an 'early' form can be better than a 'later' one, even if you've found both.
- Talisman forms have a 'maximum skill'. Above that skill, no further scaling applies. This is intended to roughly mimic max spellpower - it makes it more obvious that later-game forms will end up outscaling earlier ones.
- It takes 5 turns to enter or leave a talisman form, exactly as with armour or amulets. Use of a talisman form is intended to be a strategic decision, again like wearing armour, rather than something swapped in each fight.
- Talisman forms are not removed by cancellation or dispelling. Given that they now take many turns to re-enter, it'd be far too brutal otherwise.
- If the player enters a temporary form (Dithmenos's shadow form, Kirke's pigform, a /poly form...) while in a talisman form, when that temporary form ends, the player will re-enter their old talisman form.
- Talismans don't need to be held after they're used. You can evoke them from the floor and leave them there. This avoids inventory pressure.
- Talismans can be used with Sacrifice Artifice, since they don't use Evo.
- Zin instantly excommunicates users of a talisman. Take that, nerds!

Art for talismans is pending.

Forms
-----

The following forms exist:

*Beast*: This is the starting form for the Transmuter background. It melds all aux armours in exchange for a Slaying bonus (ala Wereblood) - +2 at 0 skill, +8 at 13 skill (max).

This is intended to provide a bonus which is compelling early game (when no or few aux armours have been found) but more tenuous later, especially for non-transmuters. It's also intended to provide a bridge between Tmut and weapon use, since a transmuter who finds a great weapon can switch from UC to that weapon without giving up their form and Tmut training.

Beast form allows use of body armour so that it can present a reasonable slay-for-AC tradeoff without becoming overly strong for 'dex-based' characters, who wouldn't mind losing body armour nearly as much.

*Anaconda*: This is a tier 2 form. Anaconda form turns you into a giant anaconda. All your items meld, you can constrict, you get some AC and an HP bonus...

This is intended to replace Ice Form, a form to help transmuters transition into the mid-game. The rF- of Ice Form is less appropriate for early-game characters who can no longer switch between forms, and Ice Form is not evocative - no one gets Ice Beasts. On the other hand, turning into a snake... everyone gets that. That's the dream. Limbs are for dorks. Ssssss

*Maw*: This is a tier 2 form. Maw form melds the body slot, transforming it into a giant mouth, ala the Brazilian Mapinguari. The maw provides an aux attack with damage that scales on Transmutations skill. It also has the old Hydra form devour-on-kill-for-hp gimmick, since everyone loved that.

This is intended to be a way that Transmuters can transition into the mid-game, especially transmuters who use weapons. It's probably a bit too strong for quick blade users at present - perhaps I'll give it +str -dex, or something. (It may also just be too strong in general - numbers are WIP!)

*Blade*: This is a tier 2 form. It's blade hands. UC damage now scales slightly with Tmut skill. Not much to say otherwise.

This is intended to be another way that Transmuters can transition into the mid-game. It probably needs some kind of extra nerf to compensate for not caring about body armour penalties anymore - TBD.

*Statue*: This is a tier 3 form. It's statue form. Intended to be a way for transmuters to head into late-game while still being able to use weapons, if desired. Might need to be a bit stronger for weapon users.

*Dragon*: This is a tier 3 form. It's dragon form. AC and UC damage now scale slightly with Tmut skill. Intended to be a way for transmuters to head into late-game. Possible this should be tier 4 and Storm should be tier 3 - dragons are cool! Dragons should be the best!

*Storm*: This is a tier 4 form. It's storm form. Intended for players who want to dump ludicrous amount of skill XP into tmut.

*Death*: This is a tier 4 form. Replacing Necromutation/Lich Form, Death Form makes you dead (no drinking potions, holy wrath/dispel undead vuln, rC, rTorm, rPois, etc) and also gives you an assortment of spicy powers. On hit (with melee/UC), victims get slowed, weakened, and heavily drained. There's also a new active, Siphon Essence, which costs 20 (!) MP, halves the HP of all enemies in radius 2, and heals you based on damage dealt and Tmut skill. (That works on all non-MH_NONLIVING enemies, as do the debuffs.)

It no longer provides innate AC or Will, nor does it give a necro enhancer. Its UC damage is now significantly higher, comparable to blade hands, though still much lower than Statue, Dragon or Storm.

This is intended to be a way for players who want to spend huge skill XP on tmut to do so, including those who use tmuts + weapons. It's intended to feel a bit different from other forms while still being competitive in melee. Other forms have huge base damage - Death Form has lower damage but very strong debuffs. Other forms have AC (Statue), HP (Dragon) or EV (Storm) - Death Form gives Siphon Essence as a very powerful survival tool.

Other Notes
-----------

Transmuter no longer starts with any spells, and their stats have been adjusted accordingly.

Various books have been merged and consolidated to make up for the removal of eight spells. It *might* make sense to drop the book generation rate slightly, but I haven't done this yet.

TODOs
-----

Before this is merged, I'll need to do the following:
- Get autotraining working with talismans.
- Update sprints.
- Add save compat for old transmuters, esp those with Beastly Appendage on.
- Fix up transmuting into/out of anaconda form while constricting (if that's possible? Maybe with Xom involved?)
- Make Xom poly you again.
- Add a targeter for Siphon Essence.

Later changes
-------------

Transmutations skill is currently in a transitional state. At some point (probably after this is merged), I'd like to split apart the functions of 'powering tmut spells' (Irradiate, Yara's, etc) and 'using talismans'. This should reduce the extent to which Tmut has an 'identity crisis' - the form playstyle isn't really a casting playstyle (it wasn't before, either!), but it still has an odd and awkward overlap with various useful spells.

To do this, tmut spells would either be dispersed into different schools (Poison, etc), or talisman use could become a new, separate 'Shapeshifting' skill. (I suppose the Tmuter background would also need a rename?) This is a large enough change that I'm not rushing into it - talismans are massive enough on their own.

Also...

In the future, artefact talismans (i.e. randarts) could be interesting - to provide more excitement for rare finds. The randarts would have the usual panoply of properties (rF+, Dex-2, etc), which would apply while the player was in the relevant form.

Possibly Wanderers should get a chance to start with beast talisman?


alexjurkiewicz:
Rather than talisman items you have to return to, should picking up a talisman grant you the (a)bility to enter and exit the form? Perhaps the talisman item should be another form of dungeon feature, like altars/shops.

PleasingFungus:
Good questions!

> Rather than talisman items you have to return to, should picking up a talisman grant you the (a)bility to enter and exit the form? 

Common suggestion. I'm hoping to add randart talismans, which an (a)-menu UI wouldn't work with. If that falls through, we can reconsider it.

> Perhaps the talisman item could be another form of dungeon feature, like altars/shops.

Also suggested before. Same answer re: randarts.

anderspapitto:
> Common suggestion. I'm hoping to add randart talismans, which an (a)-menu UI wouldn't work with. If that falls through, we can reconsider it.

> In the future, artefact talismans (i.e. randarts) could be interesting - to provide more excitement for rare finds. The randarts would have the usual panoply of properties (rF+, Dex-2, etc), which would apply while the player was in the relevant form.

each randart talisman could simply have its own entry in the a-menu (with an interface to hide/forget them to combat clutter, like the hide menu for spells). The transmutation interface could be almost identical to the spell interface, extended to treat randart talismans like "randart spells"

PleasingFungus:
This is a terrible misuse of the term 'simply' :P

PleasingFungus:
Merged as 18800231877e12caceb48c2f929f842d55aac934.

PR #3178, opened 2023-07-02
One arrival vault

mainiacjoe:
An oval room with oval walls in the middle, with decorations around the entrance inside. 

This vault was the sample project in MainiacJoe's "The Beginner's Guide to Contributing Vaults to Dungeon Crawl: Stone Soup"


nicolae-carpathia:
I like the vault, but it's a little too complicated for simple.des; it's got quite a few header lines. It would fit better in small.des.

PR #3187, opened 2023-07-09
New init file option: lua_max_memory

robertxgray:
The max memory in MB allowed for user Lua scripts is only configurable via command line options. Setting this value in the init file can be useful in platforms where the command line is not available, e.g. Android.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -162,6 +162,7 @@ The contents of this text are:
                 background_colour, foreground_colour, use_fake_cursor
 
 6-  Lua.
+                lua_max_memory
 6-a     Including lua files.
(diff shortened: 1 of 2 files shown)


gammafunk:
One issue with this is that it would need to be disabled for online server builds. We can't let an online user override the lua memory limit with their rc file. 

I think the approach used for the `-throttle`/`-no-throttle` options might work. Those options exist depending on whether `DGAMELAUNCH` is defined or not. Similarly we might disallow the option from being read/used based on that define.

robertxgray:
Thanks, updated.

robertxgray:
Updated again to display the same errors as other numeric options.

PleasingFungus:
Looks reasonable to me, thanks!

PR #3192, opened 2023-07-11
Better numeric keypad support in Linux

### LINKED ISSUE #3174: SHIFT+NUMPAD3 and SHIFT+NUMPAD9 do not work in console version (Linux)
Console, Linux

Any version (old bug)

**DESCRIPTION:**
`SHIFT+NUMPAD3` and `SHIFT+NUMPAD9` are supposed to make the character run in the respective direction, but they instead log `_Unknown command.` in the message log. The weird thing is that other `SHIFT+NUMPADX` keys do work, it is just these two (`SHIFT+NUMPAD3` and `SHIFT+NUMPAD9`) that don't.

The bug is reproduced in the tutorial and in the normal game. I'm using the default DCSS config (nothing overridden) and turning `NUMLOCK` on or off has no effect.

**EXTRA FILES:**
[Also reported here on Mantis](https://crawl.develz.org/mantis/view.php?id=2486)

Screenshot from the tutorial when I try to run with `SHIFT+NUMPAD3` (notice `_Unknown command.` in the message log):
Screenshot from 2023-06-27 09-59-20

robertxgray:
* Fixes #3174
* Fixes the insert, delete and clear keys
* Fixes the control variations
* Fixes some /+ and *+ combinations
* Reduces duplicate code, unixcurses_get_vi_keys is not more

### DIFF
--- a/crawl-ref/source/cio.cc
+++ b/crawl-ref/source/cio.cc
@@ -28,15 +28,16 @@
 // they work?
 static keycode_type _numpad2vi(keycode_type key)
 {
-#if defined(UNIX) && !defined(USE_TILE_LOCAL)
-    key = unixcurses_get_vi_key(key);
-#endif
     switch (key)
     {
-    case CK_UP:    key = 'k'; break;
-    case CK_DOWN:  key = 'j'; break;
-    case CK_LEFT:  key = 'h'; break;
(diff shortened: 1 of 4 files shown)


pooka109:
It's in now. Thank you.

PR #3193, opened 2023-07-11
Make artefact orbs use slightly orbier names

Monkooky:
The case for 'orb' was missing from _get_artefact_type, meaning it's name would not reference it's orbiness, defaulting to it's artefactitude instead.

This corrects that, even though it is perhaps the world's most trivial omission.


PleasingFungus:
Thanks!

PR #3206, opened 2023-07-19
Add mace and sword Bailey types 

mainiacjoe:
**NB: This code doesn't crash Crawl, but I don't know how to create sample Baileys in wizmode so I haven't tested how it looks in-game.**

This pull attempt to accomplish three things:

1. Enable Baileys to be made with mace or sword themes also instead of being restricted to only axes or polearms.
2. Edit each entry vault and Bailey vault in the vault file so every layout is compatible with any of the four weapon types, instead of a given layout being associated with a single weapon type.  The exception is flooded_kennysheep which clearly leverages polearm reach in the layout.
3. To that end rewrite the monster generation functions to make the function names be agnostic of the Bailey type. These edit also formalize five equipment tiers that make the types of foes made for a given weapon type to be roughly similar with one another.


regret-index:
This needs another checkwhite. The vampire bailey also had the weapons ripped out of zombie hands, since zombies technically can but currently aren't animated as keeping their equipment.

Personally... I'm not much of a fan of broadening these from the most obviously mechanically distinct duo (polearms matter all the time, axes notably threatens early summoners and necromancers) to a still arbitrary selection (not that staves would work for a bailey and short blades would be far too rewarding early for the characters who care about them), but I can understand the desire to make more characters care about broader weapon drops. Making every bailey try to support every weapon is awkward outside of that, though- I'd argue bailey_nasty_gnolls's chokepoints want for reaching as much as bailey_polearm_flooded_kennysheep, and at least one other dev has argued most of the polearm bailies are nerfed while the axe bailies are buffed without much examining over the vault contents. I'd personally loathe the asymmetry of, say, 50% chance for polearm bailies and a 16% chance each for axes / long blades / maces, so I'd rather instead designate two bailies each per type (the orc warlord duo for axes to correspond to hill orc apts, then as mechanically distinct to compensate for plainer weapons in monster hands the vampire and elf ones for long blades, and the spriggan / ogre plus the vaults guards ones for maces), and also adjust the counts to place a little less or more versus weapon type. This would still add plenty to weapon versus map fitting variety (7 agnostic, 8 split evenly to be exclusive to a weapon type), even if we could do with another map or three afterwards.

There's also some concerns about bailey rewards in general being so uneven, with the warlords bailies having their acquirement and experience while random floor items sit in others. This isn't addressed much by this pull request but is somewhat pressing if any further work was done on bailies. (Also, I'd really rather have different enemy category names than the ones provided (maybe just a simple easy -> medium -> hard would be more flexible for the future), and the itemspec lists could be a lot shorter without losing much readability by putting two entries per line and not writing out "item" and "weight" every time.)

I'm not particularly enthusiastic about this sprawling project enough to work on it for 0.31-a, but it shouldn't be too bad if another vault dev picks it up with said concerns addressed. At some point for 0.32-a I've got an interest in making hardmode post-Lair bailies (bastions?) and some new maps for such, though, which means I might eventually come around and execute on all of these listed matters eventually myself.

mainiacjoe:
I will abandon this project. If anyone wants to finish it they may.

Hellmonk:
Believe that regret-index has some plans for baileys in future (0.34). Closing for now.

PR #3209, opened 2023-07-20
query item for property

Monkooky:
Adds a function which checks the value of the property associated with an artp provided by an item.

This is distinct from artefact_property() in that it is unified across all item types; for example Fire Dragon Scales of Fire Resistance would not be a valid input to artefact_property(), where this method would return the number of pips of rF provided.

The primary use case I have in mind for this is for querying STR or DEX provided by an item, however it could also simplify monster resistances and likely uselessness checks.

This method is less efficient than a purpose built query for e.g. fire resistance, due to the lookup tables for finding the equivalent SPARM and armor flag.

### DIFF
--- a/crawl-ref/source/artefact.cc
+++ b/crawl-ref/source/artefact.cc
@@ -466,6 +466,113 @@ static void _populate_item_intrinsic_artps(const item_def &item,
     }
 }
 
+static int _get_jewelry_property(const item_def &item, artefact_prop_type prop)
+{
+    ASSERT(item.base_type == OBJ_JEWELLERY);
+    ASSERT(!is_artefact(item));
+
+    const jewellery_type jewel = (jewellery_type)item.sub_type;
+    vector<jewellery_fake_artp> *props = map_find(jewellery_artps, jewel);
+    for (const auto &fake_artp : *props)
+    {
+        if(prop == fake_artp.artp)
+            return fake_artp.plus;
+    }
+    return 0;
+}
+
+
+/// The artefact properties corresponding to a given armor ego
+/// Does not register intrinsic properties e.g. dragon scales
+struct armour_fake_artp
+{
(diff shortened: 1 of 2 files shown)


Monkooky:
The goal starting this pull request was to make a unified function to query 'what does item provide me', since such a thing would be frequently convenient. However, this is a little bit stymied by the four different ways an item can provide the same property. 
You could, for example, obtain rF from an item if:
it has rF+ as an artefact property (ARTP_FIRE enum)
it has rF as an intrinsic property (SPARM_FIRE_RESISTANCE enum)
it has rF in it's flags (ARMF_RES_FIRE enum)
it is an rF ring (RING_PROTECTION_FROM_FIRE enum)

I am of the opinion that the special_armour_type and jewellery_type should be refactored to use the artefact_property enum- having one enumerator for 'property being provided' would remove any need for a method that takes, for example, a jewellery subtype enum and looks up the artefact property which is directly equivalent to that jewellery subtype. Such a refactoring would make get_item_property much cleaner, and likely as or more efficient than the current solution for e.g. player_res_fire(). Further, the functions get_armour_res_XX could be removed in favor of calling get_item_property.

In addition to being a great many changes (SPARM is checked roughly 500 times throughout the code), there are serious design level issues that I at present don't have a solution for.
- There is no clear equivalent to SPARM_NORMAL (-1 may function?)
- There is no clear elegant solution to mimicing SPARM_RESISTANCE; it is possible that bitwise magic could be done here? 
- The enums for NUM_SPECIAL_ARMOURS et al would need to be maintained or an alternative found
- str_to_ego would need to be rewritten wholesale, but should be anyways.
- The other problems that I haven't forseen would also exist.

Monkooky:
I will not realistically have the time to work on this, I suspect it's too stale to be worth the merge conflicts, and it does not seem like anyone else has time to take up the banner.

PR #3219, opened 2023-07-29
fix plasma beam damage

hellmonk3:
Pull out the lightning component to a new beam so it's not dependent on lightning bolt.


PleasingFungus:
I vaguely wonder about going back to your original approach of just having one zappy() and then tweaking appropriate variables manually for the second shot. This seems fine for now, though.

PR #3223, opened 2023-08-01
Add some self-published spell books

mainiacjoe:
- "Amateur Aromatherapy" has stinky clouds.
- "Elemental Confusion" is about how Tornado and Absolute Zero swapped spell schools.
- "Closer Than Before" has spells that used to be ranged but are now range 1.
- "Servitor Owner's Manual" has the highest spells that Spellforged Servitor can cast.
- "Guide to Peace and Solitude" has spells that get rid of annoyances.

### DIFF
--- a/crawl-ref/source/dat/des/builder/custombooks.des
+++ b/crawl-ref/source/dat/des/builder/custombooks.des
@@ -32,6 +32,7 @@ local books = {
 "Basic_Undead_Handling numspells:2 spells:animate_dead&dispel_undead",
 "Beguiling owner:Ashenzari numspells:5 spells:momentum_strike&dazzling_flash&"
   .. "portal_projectile&fulminant_prism&iskenderun's_mystic_blast",
+"Closer_Than_Before numspells:3 spells:agony&confusing_touch&dispel_undead",
 "Crowd_Control numspells:3 spells:gell's_gravitas&arcjolt&ignition",
 "Curse_Words numspells:5 spells:slow&anguish&silence&enfeeble&discord",
 "Earth_from_Core_to_Peak numspells:8 spells:sandblast&passwall&stone_arrow&"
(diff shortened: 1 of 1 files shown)


nicolae-carpathia:
I like all the themes here, although there are a few issues:
Amateur Aromatherapy: 
1) The spell is called Poisonous Vapours now. 
2) Is the intended theme "things which mechanically create clouds" (which would exclude Poisonous Vapours and Eringya's Toxic Bog), "things which create clouds or vapours flavor-wise" (which I think still excludes Bog), or just "things which probably smell real bad"? I'm assuming the latter, based on the name.

Elemental Confusion: I like it! Powerful spells and a deep-cut meta theme.

Closer than Before: 
1) Customized randbooks are named along the lines of [<optional Author Name/provided Owner Name>'s] <Noun for Book> of <Provided Title> -- currently, there's no syntax for just specifying the entire name of the book at once. So this book would show up with a name like "Folio of Closer than Before" or "Sif Muna's Compendium of Closer than Before", etc, which is awkward phrasing. It should be a noun or gerund of some kind. (This threw me off at first too.)
2) If the theme is "spells that used to be longer-range", does Confusing Touch really fit? I get that it's sort of a range-1 version of the removed Confuse, but it was still a different spell even then.

Servitor Owner's Manual: I like it, but it has the same naming issue I mentioned for Closer than Before; as-is, it'd end up being called, e.g., "Precepts of Servitor Owner's Manual". Something like "Servitor_Ownership" or "Spellforging_Servitors" would work better.

Guide to Peace and Solitude: Nice theme! Likewise, though, the name will already include a noun for a book, so it would be better specified as Peace_and_Solitude to avoid getting "Plog's Octavo of Guide to Peace and Solitude".

mainiacjoe:
I've fixed things.  Thanks for the prompt feedback. Also, I know it wouldn't fly so I didn't include it, but I want you to know that I initially added to "Peace and Solitude" "owner:Grumpy_Cat".

PR #3224, opened 2023-08-02
Fix Anaconda status description, remove unused variable

seandewar:
Mousing-over Anaconda status would show "No description found" on WebTiles as it had the wrong name in the db (could still be found via ?/t as it would match the contents).

Also remove an unused variable in `get_unusual_items()` after commit 128b3467.


PleasingFungus:
Thanks!

PR #3229, opened 2023-08-04
Two Depths entries

mainiacjoe:
"Scary Arena" has very good loot and very, very scary monsters, both stashed behind iron grates.  There are also bloodthirsty spectators hoping for foolishly greedy adventurers.
"Labyrinth Nostalgia" is exactly that.  A fixed maze, a minotaur with a window to warn of it, and one piece of very good loot.

### DIFF
--- a/crawl-ref/source/dat/des/branches/depths.des
+++ b/crawl-ref/source/dat/des/branches/depths.des
@@ -227,3 +227,34 @@ cA--Ac   cA--Ac
 cA--Ac   cA--Ac
 cB@@Bc   cB@@Bc
 ENDMAP
+
+NAME:    mainiacjoe_depths_entry_scary_arena
+TAGS:    transparent patrolling
+FTILE:   ` = floor_sand
+MONS:    human
+MONS:    caustic shrike, pearl dragon, juggernaut, shard shrike
+SHUFFLE: 2345
+# spending a dig charge is a greater hurdle to impulsiveness than a runed door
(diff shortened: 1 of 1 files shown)


regret-index:
...Both of these entries are pretty awkward.

The arena vault should use glass over grates because a handful of spells otherwise make the rewards completely free (e.g. airstrike, freezing cloud). It also sure is placing monsters that only appear extremely rarely in a slim number of Depths vaults (pearl dragons, shard shrikes) instead of much more Depths-connected hyper threats (maybe they should instead be quicksilver dragons, liches, or walking tomes?). Even after such points are addressed, I'm not sure how good it is that a category normally containing D-capstone vaults should instead be decorative until much much later in the game, and we already have ghost vaults for backtracking bait.

The labyrinth vault is deeply confusing in general. We haven't had minotaurs be normal Depths enemies since its very start nearly a decade ago, so there's no obvious reason why this one's a Depths guard at all except to randomly shift the focus point of D:15. (There was that while when labyrinths had random earth enemies scattered within, but I'm not sure we need more DTEE and stone giant entries.) There could be some conversation about branch entries on ORIENT: float by default reducing per-game variety, but I don't think slipping a vault into the wrong category is the right way about it.

regret-index:
A whiles after a talk with the submitter on Discord, I revamped the vaults for use as regular late D vaults instead of entries, which addresses most of the problems I had with them, and thus merged them in in 9dce7d3444c4a7889e.

PR #3238, opened 2023-08-05
Two Ossuary entries

mainiacjoe:
These were placed near similar entry vaults instead of both at the end.
- "Dusty Tomb" is empty with webs and sarcophaguses (sarcophagi?)
- "Find the Door" has eight doors to check to find the portal.  A perimeter hallway ensures access to all doors once the vault is found.


regret-index:
Moved the two submissions to the bottom of the file. Decorative versus threatening entry sorting is separate work to do later. Merged in 10ed6823588d9.

PR #3239, opened 2023-08-06
Add tile for player under Glaive of Prune curse

jjjingleheimer:
It would be helpful for some visual to see when the prune curse kicks in, & adds fun flavor.

Includes a few different tile versions I couldn't choose between :P

Tile colors were taken directly from the original Glaive of Prune tile. Also gave some versions an icy "curse" aura, to match Glaive of Prune freezing brand.

This is my 1st commit ever, so I hope this is in the right place, & apologize in advance for any formatting errors!

It's also my 1st time spriting, so I am very happy for critique, & will edit accordingly. (Even if that critique is, "yeah, no" :P)


regret-index:
Thank you for your submission! We're quite sorry it took so long to comment on this. It's actually quite a nice set of prunes, though I have a concern or two.

While I understand the idea of keeping the palette of the weapon matching the transformation tile, and while Crawl employs cartoon aesthetics plenty, the deployment of such on a large wrinkled lump mostly makes it come off like one has turned into a purple version of a glowing orange brain rather than a prune. Actual prunes themselves are fairly more desaturated, and I think having this be marginally darker but with glossy highlighting (with the current closest to this being the fourth row, 1d through 4d) would be more parseable as a prune.

My own preference also would be to use something in the style of the fourth column (thus, 4d), as the second and third columns look far more like some sort of ghostly fire rather than ice, though the cut-off of blue might look strange in-game.

PR #3242, opened 2023-08-08
Stop monsters applying Flux enchant to each other (fix issue 3241)

NephilaWeaver:
It should now only apply if the player is the attacker. Even though it was really funny.

### DIFF
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectModuleManager">
+    <modules>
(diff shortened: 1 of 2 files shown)


NephilaWeaver:
*bounces excitedly*

PleasingFungus:
BTW, are you in the credits already? And if not, how should we credit you?

NephilaWeaver:
I'm not in the credits, no!  
I didn't know something like this would be enough for it, but it'd be an honor <3

PleasingFungus:
We try to credit anyone who contributes any code. (But sometimes we forget.) Thanks again!

NephilaWeaver:
Thanks a lot. This makes me happy. I'll try to contribute when I can!

PR #3246, opened 2023-08-09
Leafy Arrival Vault

orjb1:
So the plan was for it to look kinda like;

image

Then i realised I was a bit limited to match that colour palette.  Trees and bushes kinda work, but I'm not sure I've figured out the right way to make the floor grassy - it did not work in wizmode but possible the floor was already there:

image

also it's chuffing huge.  Probably too big?

Anyway, feedback would be good if you have thoughts, and I can re-assess

### DIFF
--- a/crawl-ref/source/dat/des/arrival/small.des
+++ b/crawl-ref/source/dat/des/arrival/small.des
@@ -3857,3 +3857,37 @@ MAP
 .......l.l.......
 .................
 ENDMAP
+
+# Trying to make a pretty leaf of a starting vault, but may be too big and not sure the floor_grass thing works?
+NAME: orjb1_clovarrival
+TAGS: arrival transparent no_monster_gen no_trap_gen
+ORIENT: float
+FTILE: - = floor_grass
+MONS: bush
+MAP
+.......tttt...tttt
+......t____t.t____t
+.....t______t______t
+.....t_____________t
+.....t_____________t
+..tttWt____11_____tWttt
+.t___tWt__1111___tWt___t
+t_____tWt1111111tWt_____t
+t______tWt11111tWt______t
+t______1tWt111tWt1______t
+t______11tWtttWt111_____t
+.t____1111tWWWt11111___t
+..t__11111tW<Wt11111__t
+.t___11111tWWWt1111____t
+t_____111tWtttWt11______t
+t______1tWt111tWt1______t
+t______tWt11111tWt______t
+t_____tWt1111111tWt_____t
+.t___tWt__1111___tWt___t
+..tttWt____11_____tWttt
+.....t_____________t
+.....t_____________t
+.....t______t______t
+......t____t.t____t
+.......tttt...tttt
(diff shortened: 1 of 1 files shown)


PleasingFungus:
You've set `FTILE: - = floor_grass`, but you've used `_` for the floor.

orjb1:
> You've set `FTILE: - = floor_grass`, but you've used `_` for the floor.

picard_face_palm.png

orjb1:
image

That worked, thank you, but is it over large?

PleasingFungus:
Yeah, it's a bit big. I sort of feel like this violates the spirit of our [guidelines](https://github.com/crawl/crawl/blob/master/crawl-ref/docs/develop/levels/guidelines.md#some-guidelines-for-arrival-vault-creators) in terms of being basically ASCII art, but maybe other devs (@nicolae-carpathia ? @gammafunk ?) disagree.

FWIW, you probably want to set FTILE on the trees and bushes too - otherwise it looks quite awkward, as you can see in your latest screenshot.

orjb1:
image

I don't think it's ever going to not look incongruous with the surrounding dungeon :-/

Thank you for looking though!

PleasingFungus:
Sure thing! I think it's OK to be a little incongruous, we have plenty of vaults that look out of place, esp entry vaults. Hope you submit more vaults soon! :)

nicolae-carpathia:
> Yeah, it's a bit big. I sort of feel like this violates the spirit of our [guidelines](https://github.com/crawl/crawl/blob/master/crawl-ref/docs/develop/levels/guidelines.md#some-guidelines-for-arrival-vault-creators) in terms of being basically ASCII art, but maybe other devs (@nicolae-carpathia ? @gammafunk ?) disagree.
> 
> FWIW, you probably want to set FTILE on the trees and bushes too - otherwise it looks quite awkward, as you can see in your latest screenshot.

IMO the design is simple enough that it could just be considered a geometric shape rather than full-on ASCII art. I figure the ban on ASCII art is more for complicated things, like drawing a detailed face or something. (I should point out that I've got at least two vaults that could be considered ASCII art and they haven't gotten culled yet. And then of course there's the skull and bones Crypt ending.) 

It is big, but a look through arrival/large.des shows it wouldn't be the biggest, and bigger ones have been around for a long time.

I think the basic idea is perfectly fine as an arrival vault. It does need a few fixes before it's ready for prime time, though:
1. The upstairs in arrival vaults should be a `{` glyph, not a `<`, which places an escape hatch.
2. You don't need the `.` glyphs on the left hand side and in the top/bottom center niches; you can just use empty spaces for spacing out the glyphs.
3. The fixes for the grassy floor tiles don't appear to have actually been committed.
4. If you're going to use FTILE to recolor stuff, it can be nice to use an analogous COLOUR statement so the console players can get some of the effect. (Of course, there are also tons of vaults that don't bother doing anything for console players, so I suppose it's up to you.)

PR #3249, opened 2023-08-11
feat: option to make autoexplore wait while listed durations are active

### LINKED ISSUE #3210: Feature request: option to make autoexplore wait until expiration of negative effects
CUE webtiles

0.31-a0-343-g456c268693 

**DESCRIPTION:**
Setting explore_auto_rest to true makes autoexplore wait until the character finishes regenerating HP/MP before starting actual exploration. This is usually convenient, since autoexploring with low health could lead to potentially deadly situations. 

Similarly, autoexploring with `Slow` status effect can also be very dangerous, but currently there's no option to ensure that negative status effects are rested off before autoexplore begins. 

Ideally, there would be an option that lets you select if you want autoexplore to only wait until HP/MP is regenerated or to also wait for expiration of negative status effects.

Cureja:
Resolves #3210 from (laserbat)

Added two new rc options

- explore_auto_rest_status (LIST)
  - Checks list and converts to known durations based on short_text and name_text.
- explore_auto_rest_status_set (STRING) 
  - Sets explore_auto_rest_status based on all statuses of selected type. Current options are none, all_neg_status, all_cooldowns, and all.

Added two new duration_flags that signify negative effects and cooldowns that affect the explore_auto_rest_status_set option

Options are default off right now. 

Further improvements on functionality could be:
- [ ] adding glow-level
- [ ] optimizing travel check to be based on callbacks
- [ ] more robust checking of waits that are too long (durations that are set manually to be on infinitely like transformations will not be caught typically, I think?)

Naming for the both option title and settings could be improved.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -51,7 +51,9 @@ The contents of this text are:
                 tc_forbidden, runrest_ignore_message, runrest_stop_message,
                 interrupt_<delay>, delay_safe_poison, runrest_ignore_monster,
                 rest_wait_both, rest_wait_ancestor, rest_wait_percent,
-                explore_auto_rest, auto_exclude, travel_open_doors, fear_zot
+                explore_auto_rest, 
+                explore_auto_rest_status, explore_auto_rest_status_set, 
+                auto_exclude, travel_open_doors, fear_zot
 3-g     Command Enhancements.
                 auto_switch, easy_unequip, equip_unequip, jewellery_prompt,
                 easy_confirm, simple_targeting, force_spell_targeter,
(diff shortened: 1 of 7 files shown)


jejorda2:
I put  in my options file on CBR2, running 0.31-a0-667-g39726597b9:
explore_auto_rest_status_set = "all_neg_status"

I started a berzerker, berzerked, and now "Slow -Berserk" is in my status bar.
I press o to auto-explore, expecting that this option will delay the travel until after the statuses expire.
Instead, I move into view of an enemy with both negative statuses still in effect.

Am I using this wrong?

Cureja:
**I don't believe this change is live in any server currently as it is still waiting for review to be merged.** 

I double checked and ran it locally. With those settings, when the PR is merged, you will only move when slow disappears, which is as intended! 

If you wanted to wait for the "-berserk" cooldown, you would need to use `

> `explore_auto_rest_status_set = "all"`

or manually add it using 

> `explore_auto_rest_status += "-berserk"`

DracoOmega:
I know this is an awfully long time later, but I've just folded this into trunk (with a number of edits/changes) as of 118629b. Thanks a bunch!

PR #3250, opened 2023-08-13
feat: Auto-travel to traversable squares next to non-traversable ones

Cureja:
Added UI improvement that makes it so if the player selects a non-traversable target with a traversable cell adjacent, prompt before failing and allow the user to go to the closest adjacent cell instead. 

This feature doesn't affect controls that wouldn't have invoked the "Sorry, I don't know how to traverse that place." msg. This is intentional. While using touch controls or within the current view, it should be following the current idea to not spam the user with prompts and messages on easy misclicks.

This feature could also be expanded to have an options for always saying yes to the prompt, increased radius for the check (and corresponding option for it).

This feature request originated from the wiki on [UI Improvements](https://github.com/crawl/crawl/wiki/User-Interface-Improvements#auto-travel-to-traversable-squares-adjacent-to-non-traversable-ones-accessibility-improvement)

### DIFF
--- a/crawl-ref/source/travel.cc
+++ b/crawl-ref/source/travel.cc
@@ -3282,7 +3282,38 @@ static bool _find_transtravel_square(const level_pos &target, bool verbose)
             || target.pos.x != -1 && target.pos != you.pos())
         {
             if (!maybe_traversable)
+            {
+                // check if there is a spot adjacent to the place that is reachable
+                coord_def closest_alt = you.pos();
+                int closest_alt_dist = INT_MAX;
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Thanks! This feels really nice - merged as ec6b7dd1af40546f33bf6023e5f7a4f093c9dd75. (Apologies for the delay in review.)

How would you like to be credited?

Cureja:
Awesome! It was a joy to see and use in Trunk. 

Can I be credited as Cureja please~

PleasingFungus:
Done!

PR #3255, opened 2023-08-14
Add names to player octopode tiles

cityiguana:
Add names to player octopode tiles so that users can specify their preferred tile in their rc file


cityiguana:
Take a look at my morgue history to see how desperate I am to be able to select the blue octopode tile!

image

PleasingFungus:
Unfortunately, this breaks randomization of player tiles.

Try using the `tile_player_tile` rcfile option instead, though sadly that doesn't play nicely with forms.

PR #3273, opened 2023-08-26
Remove Storm Form breakpoints

Monkooky:
Previously storm form would breakpoint EV bonus at integer skill. This fixes that.


PleasingFungus:
Thanks!

PR #3276, opened 2023-08-28
Spellforged Talisman

Monkooky:
Adds a new talisman, the spellforging talisman, which turns your skin to spellforged steel.

The goal of this PR is to provide a talisman which is appealing to caster backgrounds who want to pivot to shapeshifting,  shapeshifters who intend to go hybrid, and shapeshifters who for one reason or another have strong capabilities for casting.

The form provides bonus AC equal to 2/3 of Int, and superpowered infuse-on-hit, but causes Int to not reduce spell fail, provides an EV penalty, and melds body armour and gloves.
The Int-based AC should be the primary motivator for this form- a DeCj pivoting in would likely immediately get ~18 AC. However, they would also likely be unable to cast spells of as high level (though the spellpower would be unchanged). The numbers and scaling here are very likely to need adjustment; I did not have good intuition what acceptable tradeoffs would be.

The infuse-on-hit spends 2 mana for up to 16 bonus damage at minimum skill, going to up to 40 bonus damage at 19 skill.
This is about 2.66 times as strong as Maw for a character with no strength, but significantly less sustainable. (This still may want toning down- infusion punch as a primary strategy should not be a no-brain decision for pivoting casters)

The EV Penalty exists to make this form less valuable defensively at minimum skill, and particularly so that casters using this form for defense have a reason to train Sh past 10.

The form melds body armour and gloves- melding body armour is not strictly necessary for the form's design, but it's justifiable for the theme and further biases the form towards casters. Melding gloves is strictly necessary unless you're willing to accept Mad Mage's Maulers being extremely funny (or special cased to be nonfunctional).

I still need to make the description for this talisman more informative


PleasingFungus:
Relatively quick thoughts:

> The goal of this PR is to provide a talisman which is appealing to caster backgrounds who want to pivot to shapeshifting, shapeshifters who intend to go hybrid, and shapeshifters who for one reason or another have strong capabilities for casting.

As discussed, the fundamental goal here seems good.

> The form provides bonus AC equal to 2/3 of Int, and superpowered infuse-on-hit, but causes Int to not reduce spell fail, provides an EV penalty, and melds body armour and gloves.

I appreciate that this is fairly straightforward to describe.

It's not immediately obvious to me how large an impact on spell success the removal of int would have. Do you have any example numbers, or should I putter around and experiment?

It's interesting to compare this to wearing, say, medium armour. You get a presumably comparable EV penalty, gain a probably comparable amount of AC, get a perhaps comparable casting penalty... upside is you get infuse, downside is you need to train shapeshift to get into this. That seems quite possible to balance, but I wonder a bit if it will feel different from just playing a medium-armoured caster. What do you think?

> The infuse-on-hit spends 2 mana for up to 16 bonus damage at minimum skill, going to up to 40 bonus damage at 19 skill.

One of the concerns I have about infuse in general is that it can cause a lot of resting after even fairly trivial fights, which puts pressure on the zot clock. Do you think that could be a problem here?

Monkooky:
> > The form provides bonus AC equal to 2/3 of Int, and superpowered infuse-on-hit, but causes Int to not reduce spell fail, provides an EV penalty, and melds body armour and gloves.
> 
> I appreciate that this is fairly straightforward to describe.
> 
> It's not immediately obvious to me how large an impact on spell success the removal of int would have. Do you have any example numbers, or should I putter around and experiment?

Each Int is 2% raw spell fail; when spell fail is in the roughly 30-95% range that's pretty close to 2% actual spell fail
 
> It's interesting to compare this to wearing, say, medium armour. You get a presumably comparable EV penalty, gain a probably comparable amount of AC, get a perhaps comparable casting penalty... upside is you get infuse, downside is you need to train shapeshift to get into this. That seems quite possible to balance, but I wonder a bit if it will feel different from just playing a medium-armoured caster. What do you think?

I had to ponder this for a fair bit, and for characters who stay focused on casting, it will likely be different from medium armour only by degree and availability. Notably, for a 4 Str DeCj, swapping to spellforged talisman was nearly exactly equivalent to swapping to ring mail, save for providing an actually acceptable amount of AC in return for the ~40% extra spell fail. 
For characters with less lopsided strength and with functional body armour slots, it's likely only appealing if they have abnormally high intelligence and intend to transition to melee, or intend to transition to statue form.
I'm not opposed to trying to change this to make it more divergent, but I don't have solid ideas at the moment.

> 
> > The infuse-on-hit spends 2 mana for up to 16 bonus damage at minimum skill, going to up to 40 bonus damage at 19 skill.
> 
> One of the concerns I have about infuse in general is that it can cause a lot of resting after even fairly trivial fights, which puts pressure on the zot clock. Do you think that could be a problem here?

Fortunately, I don't think this is a serious concern- the infuse is boosted enough that it's not too different from casting a level 2 spell for each trivial fight.

Hellmonk:
Closing pending new forms coming later in 0.33.

PR #3286, opened 2023-09-07
Rollpage heal

Monkooky:
Grants armataurs 5% mhp healing for each tile they roll while rampaging.

Minimal effort has been done in descriptions, and absolutely no effort was given towards balance.

### DIFF
--- a/crawl-ref/source/dat/descript/passives.txt
+++ b/crawl-ref/source/dat/descript/passives.txt
@@ -58,7 +58,8 @@ You take an additional free step when moving towards enemies.
 %%%%
 roll passive
 
-<rampage passive>
+You take an additional free step when moving towards enemies, and heal while
+doing so.
 %%%%
 faith passive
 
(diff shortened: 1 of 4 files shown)


semitonal:
Why?

Will leave open for comments for now, but otherwise will close this later - no explanation and a self-professed no attempt at balance doesn't make for a good PR.

Implojin:
There was some discussion of this in the #dcss channel of the roguelikes discord preceding this PR:

I had suggested that heal on rampage move could work with the talisman idea Monkooky was working on (since part of that talisman idea already involved rampage), but I think Monkooky decided to go in a different direction there?

PF then suggested heal on rampage move could work elsewhere, possibly on Armataur at ~XL7, or somewhere else (possibly being added to the demonic wings idea in #2437).

I like the idea of healing on rampage moves conceptually, because it would add a dynamic incentive to repositioning, that would change from turn-to-turn, depending on how hurt the character is and where monsters are positioned.

Totally agree that the commit message should include some of the reasoning behind this.

Implojin:
Just a heads up since you've marked this draft, I'm currently working on cleaning this up and merging it with some other Armataur changes we've discussed -- you don't need to do anything more with this at the moment!

Monkooky:
Good to know!

And apologies, I should have had this marked as a draft in the first place- it was never intended to be merged as-is.

PR #3295, opened 2023-09-10
Reallow gozag title for excessive spenders

Monkooky:
While it's not technically penance, Descent's debt mechanic is close enough to justify giving out Gozag's penance title.


PleasingFungus:
Lol.

PR #3296, opened 2023-09-10
Publish a cookbook

Monkooky:
Adds a cooking themed spellbook to Self-Published Books.


PleasingFungus:
Pushed with a small tweak, thanks!

PR #3299, opened 2023-09-12
feat: Note when Maurice steals items/gold

RypoFalem:
Maurice's sleight of hand may go unnoticed by a player; leaving a player who escaped confused by their mysteriously light pockets. Note the appropriation in the dump/morgue to help players find their missing loot.

Having things stolen is funny so recording this also makes for more interesting morgues.


RypoFalem:
It might also be worth adding a force_more_message `.+steals.+!` but I dunno if maybe that's too annoying for long time players who know the drill.

PleasingFungus:
Nice, thanks!

PR #3302, opened 2023-09-13
Charlatan orb

Monkooky:
Adds a new unrand which boosts evocations, but delays teleports and reduces stealth in the same way as the Orb of Zot.

This unrand serves two purposes besides thematic appeal; firstly it provides a preview of the less obvious effects of the Orb status, and secondly encourages a uniquely evocation focused playstyle; a niche not supported by any existing unrand.


pooka109:
Quick note: source/dat/descript/status.txt is going to need an "Orb?" entry. Probably something like:

```
%%%%
Orb? status

You are holding the Charlatan's Orb, whose energies disrupt translocation
magic. Teleportation will take longer.
```

(Based on current trunk's Orb status, which no longer has the obsolete Passage of Golubria reference in it.)

pooka109:
In trunk, along with the "Orb?" status. Thank you.

PR #3303, opened 2023-09-14
Make Ozo's Fridge ignore firewood for huddling

Monkooky:
Ozocubu's Refridgeration deals reduced damage to actors that are adjacent to another actor. This previously included firewood.

If fridge is your primary damage source (by a sufficiently large margin), you are encouraged to shoot firewood with a sling then wait for it to die when it comes into LOS, so that monsters could not huddle against it. This is an extremely unfun play pattern.

While this issue occurs to some degree with plant blocked missiles, it is not as severe in those cases, nor as easily fixable. In the case of missiles, retreating until an enemy enters line of fire is sufficient. In the case of Ozocubu's, you must retreat until the enemy is a full tile beyond the plant, which may be several turns where the enemy has LOF but can still huddle.

### DIFF
--- a/crawl-ref/source/spl-damage.cc
+++ b/crawl-ref/source/spl-damage.cc
@@ -668,11 +668,13 @@ static int _count_adj_actors(coord_def pos)
     for (adjacent_iterator ai(pos); ai; ++ai)
     {
         const actor* act = actor_at(*ai);
-        if(!act)
+        if (!act)
             continue;
+
         const monster* mon = act->as_monster();
-        if(mon && mons_is_firewood(*mon))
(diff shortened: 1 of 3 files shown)


PleasingFungus:
Seems like an OK change. I'm not convinced by the optimality argument - as you said, the same could apply to ranged characters - but seems reasonable not to count firewood anyway.

Please run checkwhite and update target.cc:targeter_refrig::is_affected, not necessarily in that order.

PleasingFungus:
Merged, thanks! :)

Very minor note: for future, please don't use quotes (single or double) in your branch names. It makes them a little fiddly to work with on the command line.

PR #3306, opened 2023-09-16
Add Alchemical

Monkooky:
Adds a new species, Alchemicals.

These sapient golems are powered by magic and are skilled in artifice and alchemy.

Alchemicals have 80% HP by default and an extremely poor fighting apt; however they are empowered by potions, gaining a decaying +50% Max HP. (like vigour, this increases current HP proportionally) This grants them considerable frailty by default, but the ability to become highly defensible if they correctly identify danger before dying to it.

Their aptitudes are somewhat poor for most reliable offensive options, however their aptitudes for poison, transmutations, throwing, invocations and evocations are notably high- ideally encouraging a toolkit of many specialized or limited options. Some changes- possibly significant ones- are likely warranted here.

To further cement the above goal, and also for additional flavor, Alchemicals have MP-powered wands as a species mutation, making their wand based evocations extremely powerful.


PleasingFungus:
Wondering if we should close this post-Oni. WDYT?

Monkooky:
As you say, with the addition of Oni, Alchemical's gimmick is a lot less unique. I agree with closing.

PR #3313, opened 2023-09-21
Fix torpor snail slow

Monkooky:
Torpor snails currently allows you to clear any other slow, by walking into the snail's LOS then leaving it.

This fixes that.


Implojin:
Looks good to me, thanks for the fix!

PR #3321, opened 2023-09-27
Hopefully deduplicate portals

Rytisgit:
grab some code from TimedMarker:timeout to set timeout to be faster, so portals
 can't be re-entered.

(Descent Mode Bug)


Hellmonk:
I think I fixed this in the last descent update. Please resubmit if it's still an issue.

PR #3325, opened 2023-10-08
Add an assorcment of vaults for the Orcish Mines.

nicolae-carpathia:


crawl-ref/source/dat/des/branches/orc.des
```diff
@@ -1399,6 +1399,433 @@ x1...1...1x
...
+xx-''-x-xx
+ xxxxxxxx
+ENDMAP
+
+NAME:  nicolae_orc_mage_tower
+TAGS:  transparent
+DEPTH: Orc
+KITEM: d = any book
+KMONS: S = orc sorcerer
+KMONS: M = orc wizard
```
regret-index, line 1589: Three free books is overly generous for one sorcerer and some harmless chaff in Orc, considering how many sorcerers Orc:2 might contain. Even with additional sorcerers there either should be a lower book count, a lower randbook numspell count, or the books shouldn't overlap much (sorcerers do summon, conjure, and hex all at once, after all).

crawl-ref/source/dat/des/branches/orc.des
```diff
@@ -1399,6 +1399,433 @@ x1...1...1x
...
+@.l!1}1!l.@
+..lIc1cIl..
+..llI!Ill..
+...lllll...
+ .........
+  ...@...
+ENDMAP
+
+NAME:    nicolae_orc_beogh_chapel
+TAGS:    transparent
```
regret-index, line 1429: See https://github.com/crawl/crawl/pull/3326 for the refresher write-up on extra and decor.

This, nicolae_orc_crystal_seam, nicolae_orc_found_a_spring, nicolae_orc_hostile_takeover, nicolae_orc_quarried_blocks, and nicolae_orc_stalacmites should all be tagged extra. With the other current vaults that makes for 16, a decently close if not sufficient enough point to bother with decor semi-guarantees.

crawl-ref/source/dat/des/branches/orc.des
```diff
@@ -1399,6 +1399,433 @@ x1...1...1x
...
+: if you.depth() == 1 then
+SUBST:   { = }, [ = ], ( = ), < = >
+: end
+MAP
+ccccc
+c...c       ...
+cR{.cccccccc.R..
+c...+.R.R.Rc.....
+cD(.+R.R.R.+..R..
+c...+.R.R.Rc.....
```
regret-index, line 1799: Adding slightly more width to the stairs section so the deep troll isn't directly beside the stairs when one goes down possibly the only non-teleport / dig staircase to Orc:2 is probably for the better. Even the most dangerous of the various ambush vaults don't place the threats directly adjacent to the stairs.

regret-index (changes requested):
Finally got around to the other large nicolae- vault batch for 0.31 to review. More tagwork and house style stuff, with a few vaults singled out for threat or reward or jank. Neither of these PRs would take too much effort to actually wrap up after addressing their reviews, Do have to wonder if they'll make it in before 0.31's release if it's just left on the tracker here, of course...

regret-index:
As a3383ba mentions, nicolae- has been preoccupied during 0.31's development and assented in the main developer channel that I could add in vaults that'd need extremely minimal editing before feature freeze. The following vaults didn't make it in: nicolae_orc_big_mining_drill, nicolae_orc_combat_training, nicolae_orc_ghostly_orcs, nicolae_orc_mage_tower, nicolae_orc_ogretown, nicolae_orc_sixteen_tons, nicolae_orc_troll_toll, nicolae_orc_vermin_nest. All but the last of these don't need much work beyond what I've already mentioned in the above feedback, but nicolae- previously stated he'd prefer to do more vault clean-up himself for pull requests like these, and also this branch got a monster and will get more since the time when the PR was made. This will stay open until then, unless he wants to start those vaults from scratch.

regret-index:
Nicolae has stated within a quick personal message he will start a new pull request if and when he returns to these vaults, so this state of this pull request is closed.

PR #3327, opened 2023-10-11
prevent a Grand Finale crash

geekosaur:
If there are lost souls in LOS when Grand Finale is used, one may create a new monster where the exploded one had been, causing a crash. Catch this and take the fallback path.


geekosaur:
As I think more about this, I'm not sure lost souls should be able to block god abilities like this. The alternative is probably to make lost soul revivification a fineff, but that's beyond me at present.

geekosaur:
Or maybe to flag a monster as the target for Grand Finale so it's not considered for revivification? (cf. mon-abil.cc:_lost_soul_affectable)

PleasingFungus:
Looks good for now, thanks! Agree long term we might want to do something else.

PR #3331, opened 2023-10-19
fix: implement a correction for imp-liments

RypoFalem:
lich -> death form. Imps can compliment the player again.

I needed to properly submit this pr since the commit title was very important


semitonal:
Thanks for the fix!

PR #3332, opened 2023-10-24
Remove Upheaval friendly fire

Monkooky:
Makes Upheaval (and consequently Disaster area) do no damage to Elemental Forces. Other allies are still affected, for thematic and consistency reasons.

This change has two benefits; firstly, it removes the incentive to remain at 16.66 invocations (the level at which Upheaval AoE is consistently 1, allowing reliable upheavals against adjacent enemies), and secondly it removes the awkwardness of using elemental force and disaster area in conjunction.


pooka109:
The updated version is in now. Thank you.

PR #3334, opened 2023-10-25
Make upheaval targetter more accurate

Monkooky:
Make upheaval targetter more accurately display it's area of effect


PleasingFungus:
I'm merging this in a minute (with some fixes), but I don't understand what the intended change was. Can you explain what was wrong with the old logic?

PleasingFungus:
Ah, I see, it marks the inner ring as guaranteed once you get an outer ring. Nice.

PR #3336, opened 2023-10-27
teach the player to [Q]uiver in tutorial2

RypoFalem:
The tutorial asked the player to [f]ire or auto-[p]hire boomerangs. Neither of these commands work unless the player already has an item quivered.

Teach the player about [Q] first and also mention [F]ire-no-quiver.

Wording is admittedly a bit awkward.

It looks like this in console
image

like this in local tiles (the awkward new line after "If quivered, you" is only present in the ctrl+p message log. Other tutorial messages have this issue too)
image

### DIFF
--- a/crawl-ref/source/dat/descript/tutorial.txt
+++ b/crawl-ref/source/dat/descript/tutorial.txt
@@ -215,15 +215,18 @@ tutorial2 boomerangs
 
 Now, for ranged combat! Pick up these boomerangs with
 <input>$cmd[CMD_PICKUP]</input> or <input>g</input><localtiles>, or by
-<input>mouseclick</input>,</localtiles> and continue.
+<input>mouseclick</input></localtiles>.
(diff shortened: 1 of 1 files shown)


RypoFalem:
I forgot to mention tutorial 2 does mention [Q]uiver already but only at the end of the level, after it is needed.

PleasingFungus:
I've fixed autoquivering per Advil's suggestion above in 9787180647, but I agree it seems nice to mention (Q)uivering somewhere. Maybe a bit later?

Might make sense to rework this PR to add a tutorial message somewhere else (give some nets/javelins and tell them to quiver those..?), or could close this PR and open a new one when we're ready. I'm open to w/e.

RypoFalem:
The boomerangs are auto-quivered now so you no longer need to teach how to quiver items at this stage. The end of tutorial 2 displays the quiver key and tutorial 4 has you quivering different spells. I don't think this is needed anymore!

PR #3338, opened 2023-10-29
[DRAFT] Lua 5.4 support

Zaba:
I've taken a stab at this out of idle curiosity, it turned out to be fairly straightforward once I figured out the substitute for `setfenv`, so I figured I'd put this here in case anybody else has an interest in this.

I haven't tested this extensibly at all, so it's probably buggy as sin. But it builds, it starts, it's able to load all the vaults, so at least DLua is presumably working more or less as intended (or at least, none of the problems cause a crash). And most importantly, none of the changes look particularly scary. I'm sure a better replacement for `luaL_openlib` than what I'd quickly thrown together can be figured out if needed, too. I have not tested CLua at all, beyond whatever the game does in a default configuration.

The only build configuration I tried is `make LUA_PACKAGE=lua-5.4 NO_PKGCONFIG=` on macOS with Lua 5.4 from Homebrew and other dependencies from `contrib`, so the build system changes are most definitely not complete either.

### DIFF
--- a/crawl-ref/source/dat/clua/delays.lua
+++ b/crawl-ref/source/dat/clua/delays.lua
@@ -121,7 +121,7 @@ function rr_add_message(s, v, mode)
         end
     else
         local entry = { filter = filter, value = v }
-        local position = table.getn(g_rr_ignored) + 1
+        local position = #g_rr_ignored + 1
         if mode > 1 then
             position = 1
         end
@@ -182,7 +182,7 @@ function rr_add_monster(mons_table, s, mode)
             end
         end
     else
-        local position = table.getn(mons_table[1]) + 1
+        local position = #(mons_table[1]) + 1
         if mode > 1 then
             position = 1
         end
--- a/crawl-ref/source/dat/clua/kills.lua
+++ b/crawl-ref/source/dat/clua/kills.lua
@@ -47,7 +47,7 @@ function group_kills(a, namemap, keys, selector)
                 return selector(key, k)
             end
         )
-        if table.getn(selected) > 0 then
+        if #selected > 0 then
             if count > 0 then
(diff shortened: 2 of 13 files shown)


geekosaur:
The big problem with this isn't `setfenv`; it's that later versions of Lua don't have a good way for a host to limit memory allocations, which is necessary on the public servers.

rawlins:
As far as I'm concerned the first major blocker for this most definitely has been the lack of setfenv in 5.2+! So it's great if you were able to get this to work. I haven't looked into the memory limit issues at all though, it wasn't even worth doing given the setfenv issues.

crawl-ref/source/l-crawl.cc
```diff
@@ -1781,6 +1781,12 @@ LUAWRAP(crawl_clear_message_store, clear_message_store())
  */
 LUARET1(crawl_seen_hups, number, crawl_state.seen_hups)
 
+LUAFN(crawl_setfenv)
+{
+    lua_setupvalue(ls, 1, 1);
```
rawlins, line 1786: Having taken a quick glance (and also in the past spent a bunch of time googling the issue), it's not entirely obvious to me how/if this is a full equivalent to setfenv; could you document this at all? (Was setfenv really doing nothing more than setting the upvalues? Most of the code I saw for trying to emulate setfenv was more involved, though a lot of what I found was also on the lua side. This is a fairly crucial issue as clua sandboxing relies on setfenv...)

gammafunk:
Thanks for the work on this. I've used it as a basis for my draft PR #4895, which contains further changes necessary for the build on all platforms, among other changes. I will discuss the questions of your `crawl.setfenv()` implementation and the custom allocator etc.

PR #3339, opened 2023-10-30
Make Lightning Rods use half AC

Monkooky:
Makes lightning rod check against half AC rather than proportional AC. This change follows in a push to make lightning damage always ignore half AC, and also removes an inappropriate usage of proportional AC.

This is a buff, but it's unclear to me to what degree, as lightning rod damage code is fairly opaque.


PleasingFungus:
Makes sense, thanks!

PR #3340, opened 2023-10-31
Enable a confirmation prompt for "bump" attacks with ranged weapons.

### LINKED ISSUE #3052: Move-attacking with a ranged weapon doesn't honour inscriptions
Web console, underhound.eu

Trunk as of ~13th March

**DESCRIPTION:**
Wield a shortbow. Inscribe it to prevent attacking with it (using !a or even !*). Try and move into an adjacent creature - you fire a shot at it without a confirmation prompt.

Aliscans:
If the user attacks by moving towards a monster holding a ranged weapon, this change causes the player to be asked to confirm the attack in some situations.

A "!a" inscription might prompt a "Really attack ...?" message, which is temporarily suppressed after a "yes" response. A "!f" inscription might prompt a "Really fire ...?" message which appears for each attack. This mirrors how the "attack with wielded item" command works.

This should fix #3052.

### DIFF
--- a/crawl-ref/source/fight.cc
+++ b/crawl-ref/source/fight.cc
@@ -252,16 +252,13 @@ static bool _can_shoot_with(const item_def *weapon)
         && !you.berserk();
 }
 
-static bool _autofire_at(actor *defender)
+static void _autofire_at(actor *defender)
 {
-    if (!_can_shoot_with(you.weapon()) || you.duration[DUR_CONFUSING_TOUCH])
-        return false;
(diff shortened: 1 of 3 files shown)


WizardIke:
I would love to merge this PR, but it has been left sitting for 3 years and a lot of the code it changes is different now

CrawlOdds:
Fixed in 4eb44f5 (I'm glad to say it got much easier in the intervening years!)

PR #3352, opened 2023-11-18
Rework lightning rod damage

Monkooky:
Lightning rod damage was previously divided by the arclength+2 at a particular distance; e.g. if your cone hit two squares at range two, the damage for enemies at range 2 was divided by 4.

This had several issues: it was extremely unintuitive, it was not communicated sufficiently by the description, and it was very difficult to reason about in terms of effectiveness.

This PR changes lightning rod damage to be divided by arclength + 2 where arclength is computed at range 5, rather than computing a distinct arclength for each range from 1 to 5.

### DIFF
--- a/crawl-ref/source/spl-damage.cc
+++ b/crawl-ref/source/spl-damage.cc
@@ -3025,7 +3025,7 @@ spret cast_thunderbolt(actor *caster, int pow, coord_def aim, bool fail)
 
     int arc = hitfunc.arclen;
     
-    //ASSERT(arc > 0);
+    ASSERT(arc > 0);
     dprf("arc length is %d", arc);
 
     for (const auto &entry : hitfunc.zapped)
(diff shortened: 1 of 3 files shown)


Monkooky:
This PR introduces some new unusual behaviours in exchange for the old unusual behaviours. 
Firstly, you can shoot through walls you can see around:
image

Under old rules you would often be able to achieve this on the second blast of the rod, but not the first. It is unclear if this behaviour is an improvement.


Secondly, there are cone problems with the targetting; the following two zaps have the same arclength; the second is strictly superior under the new system. I do not believe these problems are new, they are merely comprehensible now.
image
image


Finally, it is worth noting that this is a considerable nerf versus nearby enemies; taking a half-screen sweep from 1/7x damage against adjacent foes to 1/23x damage. This should be easy to fix with numbers changes.

Monkooky:
An explanation for the geometry will go here when I wake up

CrawlOdds:
We're planning to go a slightly different way with this in #5305, so closing this one.

PR #3353, opened 2023-11-19
Update ZH Translation

zhangjoke:
Update Chinese translation. By the way, make some sentences more coherent.


PleasingFungus:
Thanks as always!

PR #3356, opened 2023-11-21
new slime ending

yrdzrfxndfvh:
add a new slime ending in the shape of a (very rough) mandelbrot set
FEATURES:
- corrosive slime pillars in the main area
- azure jelly friends waiting for you in a pool
- potentially trapped entrance to trj area (can dig around it if its too scary)
- loot spindle

### DIFF
--- a/crawl-ref/source/dat/des/branches/slime.des
+++ b/crawl-ref/source/dat/des/branches/slime.des
@@ -1414,8 +1414,9 @@ ENDMAP
 NAME:   slime_pit_tekkud_mandelbrot
 KMONS:  Y = azure jelly
 KFEAT:  Y = shallow_water
-KFEAT:  ^ = zot trap w:5 / dispersal trap w:15 / alarm trap / net trap
-SUBST:  ' = x., D = ^..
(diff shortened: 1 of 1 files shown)


PleasingFungus:
This is a very cool idea :)

dillytin:
Are jellies able to activate net trap? If not, almost no point of putting a net trap at D

regret-index:
The reason this vault doesn't fully build is that it needs to use "floor" and not "nothing".

I'd need to do pretty drastic redesign surgery to make this vault be good enough for 0.31's extremely impending feature freeze, and I'd discourage trying to push for that under the current circumstances. It's a pretty cute notion, but the layout is rather small compared to the current slime end set and has a very non-randomized component to TRJ's chamber being very full and also very small with only two real "sections" to a rather narrowed level. I'd rather reverse the positions of the stairs and TRJ (as fighting TRJ in the lower chamber would be far more variable between games), and would probably consider using two mandelbrots side-by-side pointed in opposite directions rather than the literalist approach of just placing a mandelbrot for the whole floor.

yrdzrfxndfvh:
someone already made a mandelzot

PR #3357, opened 2023-11-22
62 Inventory Slots

Rytisgit:
The partial dropping mechanic is mostly unused and @ inscription functionality is a small sacrifice for a whole 10 slots in the inventory. 

Partly tested and coded so far, seems to work with new characters and crashes with older saves due to assert about unchanging inventory size, which is being changed.

The assert is the main thing I am seeking advice with this draft PR.


Rytisgit:
weird failure to build by the pipelines, works locally

PleasingFungus:
Thanks for working on this! I'm hoping that @rawlins or @semitonal will have more time to chat about their concerns; I'm going to hold off on review for at least a little while to give them time.

PleasingFungus:
One belated note about this - this probably causes problems for folks who macro eg `1` to `za`, right? Not saying that's a blocker necessarily, just wanted to call it out explicitly.

Rytisgit:
Hi, there is no problems with macroing the numbers due to this change. `1` to `za` still works just as it did.

PleasingFungus:
Ah, I'd misunderstood. I'd thought that if you pressed `q1` to drink the potion in item slot 1, you'd end up triggering the macro - that is, you'd end up drinking `z` and then entering the `a`bilities menu. But it looks like macros don't trigger when you're in a prompt.

Thanks!

Rytisgit:
I cleaned up my commits, rebased my changes on master to support the latest changes, I tested with coglin, and also added new fixes for jewellery_prompt = true with octopode since it required special handling, so this should be completely working at this point. For me this worked to build, I assume it is failing due to the minor tag upgrade.

I also hosted this on my server to test out in case anyone wants to see it working - http://xom.montres.org.uk:8080/ Select TestCrawl to test out this branch with 62 inventory slots.

Hellmonk:
closing as superseded by 928708b. sorry for the delay

PR #3358, opened 2023-11-24
removes ability to start repeat sequence with armour equip commands

Rytisgit:
The repeat command doesn't allow other multi-turn commands such as wield weapon or wear jewelry, so this makes the behaviour unified.


wheals:
Good catch, thank you! Merged in abc238812c

PR #3361, opened 2023-11-29
Print a special message when an item reaches its enchantment limit.

Aliscans:
Change what happens when weaponry or armour in enchanted to its maximum.

1. Print "while" rather than "moment" in the message, to give the player a hint that something unusual has happened.

2. If caused by a scroll, cancel repeated actions. The next scroll can't enchant this item again, but the same keypresses could enchant something else.

Also, print the message at the end of enchant_armour(). This matches enchant_weapon(), and means that any lua function which is triggered by the message sees the item in its final state.

### DIFF
--- a/crawl-ref/source/item-use.cc
+++ b/crawl-ref/source/item-use.cc
@@ -3281,7 +3281,7 @@ bool enchant_weapon(item_def &wpn, bool quiet)
     bool success = false;
 
     // Get item name now before changing enchantment.
-    string iname = _item_name(wpn);
+    string iname = quiet ? "" : _item_name(wpn);
 
     if (is_weapon(wpn)
         && !is_artefact(wpn)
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Thanks! Merged with some minor refactoring.

crawl-ref/source/item-use.cc
```diff
@@ -3369,29 +3376,30 @@ bool enchant_armour(int &ac_change, bool quiet, item_def &arm)
...
             canned_msg(MSG_NOTHING_HAPPENS);
         return false;
     }
 
+    string name = quiet ? "" : _item_name(arm);
+
+    arm.plus++;
+    ac_change = 1;
+
     // Output message before changing enchantment and curse status.
```
PleasingFungus, line 3393: RIP curse status :)

PR #3362, opened 2023-11-30
Note when felids lose or gain lives

RypoFalem:
This should make it much easier to keep track of how many lives you had at what point in the game. (Currently there is nothing for gained lives when you lose lives, it's hard to search for because the noted death message uses different verbs)

Here's an example: 
```
     6 | D:1      | Reached XP level 4. HP: 21/21 MP: 5/5
     6 | D:1      | Gained a life (2 lives remaining)
     8 | D:1      | Mangled by an emperor scorpion
     8 | D:1      | Lost a life (1 life remaining)
```


semitonal:
Merged, thanks for the PR!

PR #3363, opened 2023-12-02
Add more monster speech lines for Donald and Crazy Yiuf.

pooka109:
Here are a few more speech lines I thought of for Donald (complaints regarding gods, especially gods he currently has few complaints about) and Crazy Yiuf (one more line of Xom-related philosophy).


PleasingFungus:
Nice, thanks!

PR #3364, opened 2023-12-02
fix order of monster sets in ziggurat.lua

yrdzrfxndfvh:
gehenna/cocytus and negative energy are out of order


regret-index:
Dealt with in a different way in 5654b9b.

PR #3368, opened 2023-12-04
Tell the player about MF_SECONDARY_SCROLL in menus.

Aliscans:
If `MF_SECONDARY_SCROLL` is set, state in any menu "subtitles" that object class keys (such as ? or /) move to an item in its section. This was not previously communicated.

Add `MF_SECONDARY_SCROLL` to equip/unequip and identify menus.

The previous situation was:
The wield command showed weapons and staffs, but there was no one-key way to go to the staves.
The menu displayed if you read a scroll of identify showed scrolls and potions. If you typed `!` in that menu it would identify the first unidentified potion.

With this change:
Pressing `|` in the wield menu moves the cursor to the first staff.
Pressing `!` in the identify menu moves the cursor to the first potion.
Both keys are displayed within the menu.

### DIFF
--- a/crawl-ref/source/invent.cc
+++ b/crawl-ref/source/invent.cc
@@ -891,10 +891,14 @@ menu_letter InvMenu::load_items(const vector<const item_def*> &mitems,
         if (!inv_class[i])
             continue;
 
-        string subtitle = item_class_name(i);
+        string subtitle = item_class_name(i), select_all;
 
         // Mention the class selection shortcuts.
-        if (is_set(MF_MULTISELECT))
+        if (is_set(MF_SECONDARY_SCROLL))
+            select_all = "go to first";
+        else if (is_set(MF_MULTISELECT))
+            select_all = "select all";
+        if (!select_all.empty())
         {
             vector<char> glyphs;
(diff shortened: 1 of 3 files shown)


wheals:
The ID behavior is intentional (by popular demand), as mentioned in 190321e, so if we want it to stay that menu shouldn't have MF_SECONDARY_SCROLL set. The rest of the commit looks good.

Aliscans:
> The ID behavior is intentional (by popular demand), as mentioned in [190321e](https://github.com/crawl/crawl/commit/190321e9dd265d5756422dd396f933f44fe0b71a), so if we want it to stay that menu shouldn't have MF_SECONDARY_SCROLL set. The rest of the commit looks good.

That's fine; my main concern was that the keys did something (and a different something depending on which prompt it was, at that), but the game gave no hint as to what pressing `?` would do.

It does make things more complicated, as the game can't tell (at the point at which these headers are added) if the menu will still have category hotkeys by the time it's displayed, and the `MF_` flags don't cover it. I've made "use secondary hotkeys" a parameter to InvMenu::load_items() to address this.

wheals:
Sorry it took me a while to come back to this, looks good.

PR #3375, opened 2023-12-05
Add Dithmenos to the list of evil gods that Zin's Recite talks about.

pooka109:


PleasingFungus:
Nice catch. I wonder if we should move this to C.

PR #3377, opened 2023-12-09
Update ZH translation

zhangjoke:
Update Chinese translation. By the way, Modify some expressions.


PleasingFungus:
Thanks, as always! :)

PR #3378, opened 2023-12-11
Felid: update manual about Felids

CipHuK:
Felids aren't fast any more

See 406407be8511ac9846d21bd4a51990ab4ce1dbb7


PleasingFungus:
Good catch, thanks! I did look for this...

PR #3380, opened 2023-12-12
Remove spurious sticky flame reference in _ugly_thing_resists().

pooka109:
Since red very ugly things don't have sticky flame anymore, it shouldn't be referenced when giving them their resists, either.


PleasingFungus:
Looks good, thanks!

PR #3381, opened 2023-12-12
Fix a few obsolete references to MR (magic resistance).

pooka109:


PleasingFungus:
Seems reasonable, but do you mind fixing your commit messages to fit our [conventions](https://github.com/crawl/crawl/blob/master/crawl-ref/docs/develop/coding_conventions.md#1---commit-conventions)? I'll be happy to merge after that.

pooka109:
Thank you. Is this better?

PleasingFungus:
Yes, these are fine. Thank you! :)

(For future reference, I might have titled them something like "Fix name of ENCH_LOWERED_WL", "Fix name of _foe_mr_lower_goodness", and "Fix hex_success_chance prototype".)

PleasingFungus:
Actually, looking at them again, these commit titles are 65 characters long, and we usually want those to be 50 characters or less. Otherwise, various bits of UI break. Not the end of the world, though. :)

PR #3383, opened 2023-12-13
Fix warning message if Beogh couldn't give a follower an orcbow.

pooka109:


pooka109:
I tried changing the commit message to 50 characters and rebasing, and I seem to have screwed up the latter.

PR #3384, opened 2023-12-13
Rename "mr" (magic resistance) to "wl" (willpower) in more places.

pooka109:
The species YAML files and associated documentation on species creation now reference "wl/willpower" instead of the obsolete "mr/magic resistance". I went with that because WL is also how aptitudes.txt abbreviates Willpower in its footer and header.

Now, the only references to "mr/magic resistance" left seem to be in tags.cc (which encompasses the magic resistance era, so that makes sense) and source/test/stress/qw.rc (which is apparently maintained elsewhere).


pooka109:
Updated the commit message to fit the length limit.

I think I updated it properly, but I'm not certain.

pooka109:
After looking into it more, and learning how updates are supposed to be done, apparently I screwed this up too.

PR #3387, opened 2023-12-14
Explicitly disable polymorphing into ugly things.

pooka109:
Monsters currently can't be polymorphed into ghost demon-type monsters anyway, but given how ugly things can never be polymorphed into other monsters, it makes sense to disable polymorphing other monsters into them as well.


PleasingFungus:
It seems fine to allow monsters to be trapped in a hellish, inescapable existence as an ugly thing. Maybe not morally fine, but when has that ever stopped players?

pooka109:
Okay. I suppose it does add to the existential horror of ugly things.

PR #3392, opened 2023-12-15
Use a PromptMenu for various selections in wizard mode.

Aliscans:
Various wizmode functions functions printed a menu in the message window, but didn't check if it would fit, and thereby hid some of the options.

Make some use PromptMenu instead. This displays the menu in a separate popup if it's larger than the message window, so doesn't share this issue.

There's nothing wizard-specific about the class I added; I only included the #ifdef because I haven't seen a similar issue anywhere else.


wheals:
Merged in https://github.com/crawl/crawl/compare/70f7560b1c4f...aad2c9efd125

PR #3394, opened 2023-12-16
Increase the inventory size to 104 stacks.

Aliscans:
The player accesses the first 52 items in the same way as now. The last 52 (all denoted by a "letter" of `&`) can be accessed by scrolling through. Pressing `&` goes to the next one on the list.

This is intended mainly to give an idea of what effect expanding the inventory would have on gameplay. The UI may not be the best possible one (it would involve fewer keypresses if you could type `&a` or `&Z`, for instance), but it should show how it plays.

There are a couple of things I can think of which I haven't explained within the game.

- What the `&` key does in an inventory list is not explained.
- The game reassigns `&` items by default if you drop them and pick them up again. This is intentional, but doing that with the entire inventory could lead to other items also being reassigned.

### DIFF
--- a/crawl-ref/docs/crawl_manual.rst
+++ b/crawl-ref/docs/crawl_manual.rst
@@ -472,7 +472,7 @@ auto-explore will usually save real time.
 Stashes and Searching
 ========================================
 
-Since you can only carry 52 items, you will occasionally want to stash things
+Since you can only carry 104 items, you will occasionally want to stash things
 away (by dropping them with the 'd' command). When you want to search for
 something in your stashes, you can do this with the Find command 'Ctrl-F'. The
 parser even accepts regular expressions, although you will mostly just need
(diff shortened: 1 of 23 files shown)


PleasingFungus:
Thanks for taking a shot at this! However, it seems like here the cure is worse than the disease. Adding an extra keypress to interact with half of your inventory is exactly the sort of thing we've kept the inventory limit at 52 to avoid, and making that keypress the very awkward `&` seems worse.

Happy holidays!

Aliscans:
No problem. Do you have any comment on what I wrote about it? Thanks.

PleasingFungus:
Hm, not sure exactly what you want me to comment on, sorry!

> This is intended mainly to give an idea of what effect expanding the inventory would have on gameplay. 

This is a fine and good idea - testing 104 item gameplay by whatever means. However, it doesn't seem like something we'd ever *merge* in that state, so it doesn't need to be an open PR. Does that make sense?

PR #3398, opened 2023-12-18
Fix inconsistencies in vitrify messages.

pooka109:
When the player is vitrified more by BEAM_VITRIFY_GAZE, there's a message for that. However, there isn't one for when the player is vitrified more by BEAM_VITRIFY. This fixes that.

Also, the monster messages for becoming as fragile as glass end in "!" in one place (Killer Klowns' clear moon pies), and "." in two others (the vitrify beam and the vitrifying gaze beam). This makes them all use "!", especially since vitrification does more damage now.


PleasingFungus:
Looks good, thanks!

PR #3403, opened 2023-12-19
dreamshard necklace quote

yrdzrfxndfvh:
i think i put it in the right place👍


pooka109:
Thank you. It's in.

(It needed to be in the "items" section of quotes.txt, so I moved it there, put fancy Unicode quotes around it, and added an alias for the necklace's new name once it's saved your life.)

PR #3407, opened 2023-12-20
Zin's Imprison improve

sdynet:
Recite relies heavily on skill level and has no effect on beasts. In the early game, the player's skill level is very low, and half of the monsters that appear are beasts. Therefore, I think Recite is a poor choice for a first power. So I want to bump Recite to 3*, which doesn't help me right away, and I'm eyeing Imprison as an alternative (1* power). This is because while there are a lot of different powers in DCSS - damage, healing, buffs, debuffs, allies, etc. - the concept of "temporarily removing an enemy from the battlefield" is very rare. I think Zin's focus on Imprison rather than Recite could open up a new playstyle, so I'm going to experiment with reducing the cost of Imprison and increasing its duration to see what it would look like to actively use this power.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -391,13 +391,13 @@ static vector<ability_def> &_get_ability_list()
 
         // INVOCATIONS:
         // Zin
-        { ABIL_ZIN_RECITE, "Recite",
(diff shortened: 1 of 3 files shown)


Hellmonk:
I would like some Zin changes as well but I think imprison would be too strong at 1*. I will see what I can do in the next version.

Issue #3408, opened 2023-12-21
Player can be "confused by a divine experience" from Xom wrath even when Xom does nothing


pooka109:
PLATFORM and OS or BROWSER:
local tiles, Ubuntu 22.04.3 LTS

DCSS VERSION AND GAME INFORMATION:
0.31-a0-1327-g8c40ea976f

DESCRIPTION:
It's possible for Xom's wrath to do nothing, and yet for the player to be "confused by the divine experience" anyway, unlike all the other gods' wrath. I was able to reproduce this twice with a certain seed, by dumping Xom for Gozag via wizard mode; I've attached a morgue of it as proof.

EXTRA FILES:
[morgue-Vorloth-20231220-202308.txt](https://github.com/crawl/crawl/files/13734553/morgue-Vorloth-20231220-202308.txt)

PR #3409, opened 2023-12-21
Disable divine experiences for Xom non-wrath (#3408).

pooka109:


PleasingFungus:
Thanks! Wrote this a little differently.

PR #3414, opened 2023-12-22
Missing status descriptions and Bullseye, PProj fixups

seandewar:
- Remove the old "PProj" status description, and replace it with an updated one
  for "Bullseye" (I don't know Chinese, so I just removed that one).

- As Portal Projectile is now a monster-only spell, update its description to
  attempt to more accurately reflect its behaviour when casted by monsters.

- Remove `action::affected_by_pproj()`, which is unused.

- Remove `action::is_bullseye_active()` and its uses.
  This function incorrectly treats each shot as costing 1 MP, and implies that
  bullseye is unusable while confused; both cases are no longer true.
  Remove uses from `action::uses_mp()` overrides and from `fill_status_info()`.

- Also update some out-of-date and missing status descriptions that I noticed.

I wouldn't really consider myself much of a wordsmith, so I hope my descriptions
aren't too terrible. :laughing:


PleasingFungus:
Thanks! Nice fixes. :)

PleasingFungus:
Just realized you aren't in the CREDITS (unless it's under another name) - how would you like us to credit you?

PleasingFungus:
Done! :)

PR #3416, opened 2023-12-23
Make burial acolytes actually worship Kiku.

pooka109:
As in, set their monster god properly.


PleasingFungus:
Thanks! I implemented this slightly differently.

PR #3418, opened 2023-12-25
Allow cloning of uglies and dancing weapons.

pooka109:
Ghost demon monsters can now be cloned. As for the others, named monsters still aren't clonable, which rules out Pandemonium lords, player ghosts, and player illusions; conjured monsters still aren't clonable, which rules out spectral weapons; and inugami are explicitly not clonable because you don't get the usual penalty if the inugami clone dies.

(I couldn't find a reason why ghost demons were barred from being cloned; git blame for that chunk of code goes back to 2010 when Mara's creating player illusions was added. Allowing clones of dancing weapons might be controversial, since a player could just cast Tukima's Dance and then use a phantom mirror on the resulting weapon, but I don't know if that's game-breaking.)

### DIFF
--- a/crawl-ref/source/mon-clone.cc
+++ b/crawl-ref/source/mon-clone.cc
@@ -220,11 +220,12 @@ void mons_summon_illusion_from(monster* mons, actor *foe,
 
 bool mons_clonable(const monster* mon, bool needs_adjacent)
 {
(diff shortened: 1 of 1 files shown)


pooka109:
Then again, you can cast any summoning spell (other than Call Canine Familiar. anyway) and clone the resulting summons, so maybe it isn't that game-breaking.

PR #3420, opened 2023-12-25
Allow polymorphing monsters into uglies.

pooka109:
Polymorphing Pan lords into other monsters is still disallowed.

### DIFF
--- a/crawl-ref/source/mon-info.cc
+++ b/crawl-ref/source/mon-info.cc
@@ -1210,7 +1210,12 @@ string monster_info::common_name(description_level_type desc) const
 
 bool monster_info::has_proper_name() const
 {
-    return !mname.empty() && !mons_is_ghost_demon(type)
+    // Some ghost demon monsters (Pan lords, player ghosts, and player
+    // illusions) have name overrides.
+    return !mname.empty()
+            && type != MONS_PANDEMONIUM_LORD
+            && type != MONS_PLAYER_GHOST
+            && type != MONS_PLAYER_ILLUSION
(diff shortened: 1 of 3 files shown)


wheals:
It seems a little odd to me to allow polymorphing into pan lords. While they're not technically uniques (well, except for the ones that are), they feel special in the same way that player ghosts do.

pooka109:
I just figured out that polymorphing uglies won't really work, as there's no trivial way to preserve the names of uniques polymorphed into uglies. And I tried updating the branch, but I think I screwed it up again. So polymorphing into Pan lords is all that's left in the patch.

As for it's being odd, I did waffle on the concept, and it might be too evil.

pooka109:
And I finally updated it properly.

pooka109:
Updated again. I finally got polymorphing into uglies working properly, and keeping uniques' names.

The old attempt missed monster_info::has_proper_name(), which was skipping over all ghost demons instead of just the ones with name overrides (Pan lords. player ghosts, and player illusions). The others all seem to have new names work without the skipping: the uglies keep new names if they were polymorphed, dancing and spectral weapons can't be polymorphed into in order to get new names anyway, and inugami keep new names if they were polymorphed (which I verified by disabling their M_NO_POLY flag temporarily).

So the patch allows polymorphing into both again. If Pan lord polymorphs are that controversial, I can easily revise it to just allow polymorphing into uglies.

pooka109:
One last update. I put back one chunk of code that was supposed to block polymorphing monsters from Pan lords instead of to them. (It seemed redundant, what with two other checks blocking it, but allowing polymorphs to new things is enough of a change here.)

pooka109:
After some thought, I've updated it to only allow polymorphing into uglies. (The code to set the name of a polymorphed-into Pan lord was something of a hack, anyway.)

PR #3423, opened 2023-12-26
Improve how search handles artefacts.

### LINKED ISSUE #3350: Stash search doesn't search artefacts correctly.
Linux 6.1.0 console using XTerm(358)

0.31-a0-1009-g454b1f1e0a

**DESCRIPTION:**
If I am carrying a suit of ring mail, and search (Ctrl-f command) for "ring && !!mail", it doesn't find it.
If I have an artefact suit of ring mail, it does.
The "stash search" help text says it should not find ring mail. so I would have expected it to find neither.

This happens because the game searches for non-artefacts using the name together with any stash search prefixes.
It does this for artefacts, but also searches `chardump_desc(item)`. This contains a longer description of the item's properties, but  no name. It contains "ring" (as part of the word "wearing"), but not "mail".

If, instead, it carried out one search which included all of the relevant text this wouldn't happen.

**EXTRA FILES:**
[x.zip](https://github.com/crawl/crawl/files/13362475/x.zip)
Save file with two ring mails, one an artefact.

Aliscans:
Instead of returning artefacts where the name & prefixes match, or where the artefact description matches, return artefacts (from inventory, floor or shops) which match with all three things combined.

Should fix #3350.

### DIFF
--- a/crawl-ref/source/stash.cc
+++ b/crawl-ref/source/stash.cc
@@ -399,8 +399,10 @@ vector<stash_search_result> Stash::matches_search(
     {
         const string s   = stash_item_name(item);
         const string ann = stash_annotate_item(STASH_LUA_SEARCH_ANNOTATE, &item);
-        if (search.matches(prefix + " " + ann + " " + s)
-            || is_dumpable_artefact(item) && search.matches(chardump_desc(item)))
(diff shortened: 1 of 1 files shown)


wheals:
Merged in 7fc57aa741

PR #3433, opened 2023-12-28
Update Armataur manual description.

pooka109:


Implojin:
Ah dang, I apologize, I didn't see this PR when I pushed e2d6d42b48.

Thanks anyway!

pooka109:
No problem.

However, the Armataur description still doesn't mention that rolling regenerates MP as well as HP. I can still open another one for that :)

Implojin:
> At higher levels they also regenerate health when rolling

Hm, maybe it would be more clear the way you've written it?

I could merge that if you'd like, or you could push it yourself? Either way sounds fine.

Issue #3430, opened 2023-12-28
Error "Unknown skill `Summoning`" when entering a trove


WizardIke:
**PLATFORM and OS or BROWSER:**

webtiles

**DCSS VERSION AND GAME INFORMATION:**

Version: 0.31-a0-1444-g3ad46bb
Game Seed: 11376649488515153250

**DESCRIPTION:**

I entered a trove and the error shown in the screenshot appeared in the message log. No other effects were noticeable. There is also a Summonings skill instead of Summoning in the skill menu but all the spells use the Summoning skill (this doesn't cause any problems when casting though).

**EXTRA FILES:**

![Screenshot (980) 2](https://github.com/crawl/crawl/assets/26264063/4ecac631-4b04-46ae-acc0-f762971627dd)
Morgue: https://crawl.project357.org/morgue/Wizard1ke/morgue-Wizard1ke-20231227-033103.txt

PR #3434, opened 2023-12-28
Fix trove.spell_skills logic for trove.des

CipHuK:
fix https://github.com/crawl/crawl/issues/3430
additionally fix a couple occasions of using Summoning in docs


gammafunk:
Thanks for the fix! I discovered your PR after a different user reported the bug in Discord and had made my own fix, which made a few more changes to the relevant functions.

PR #3435, opened 2023-12-28
Mention Armataur rolling MP-regen in the manual.

pooka109:


Implojin:
Seems fine to me, although I wonder if it might be better phrased as "they regenerate both health and magic when rolling"?

pooka109:
Thank you. And, yes, that phrasing is a bit better. I'll add it.

PR #3446, opened 2023-12-30
Trove: stop asking for dragon armour toll from draconian

CipHuK:
As draconians do not use dragon armour, it is not a big deal for them to pay such small price to enter trove


PleasingFungus:
Looks good, thanks!

PR #3452, opened 2024-01-05
Four new god-related death messages

Oneirical:
I always loved the death messages linked to some gods like "Your body crumbles into a pile of gold." for Gozag, so here are four new ones.


pooka109:
Implemented, albeit in a different way, in pull request #3462. Thanks.

PR #3462, opened 2024-01-06
Move god death messages to the god speech database (displaying specific ones after the speech), and add some new ones.

pooka109:


pooka109:
Updated. All the gods can use the god speech database for death messages now.

The Beogh heretic message (extended for apostates who left Beogh), the Nemelex message (now done only if you left a corpse), the Gozag gold message, and the Okawaru duel message are now all displayed after the god-specific database death messages.

pooka109:
Addendum: Xom's death messages are the same as they were, but Xom's a weird case by definition.

pooka109:
And Xom's message is now displayed after the database death messages, too.

PR #3470, opened 2024-01-12
Very basic guarded unrand vault

orjb1:
I had the idea of quicksilver dragons guarding arga (since I feel they're thematically linked).

image

Obviously hoping for feedback, my current thoughts:

- Could slim down to 3 wide, but I like arga in the middle of three tiles at the back
- possible want no_item_gen
- 4 dragons reasonable/too hard?  Is it worth scaling with 2/4 depending on depth.

### DIFF
--- a/crawl-ref/source/webserver/webtiles/server.py
+++ b/crawl-ref/source/webserver/webtiles/server.py
@@ -195,7 +195,7 @@ def stop_everything(shutdown_event):
     def do_stop(tries=0):
         if len(ws_handler.sockets) == 0:
             shutdown_event.set()
-        elif tries > 10: # 60: # 30s
+        elif tries > 60: # 30s
             # try something a bit stronger
             logging.error("Stop failed after 30s, cancelling remaining tasks.")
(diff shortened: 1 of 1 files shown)


regret-index:
I appreciate the straightforward nature of the vault matching the nature of the artefact, and the thematic match-up of speed, fancy mineral, and +willpower seems reasonably appropriate. There's a couple of concerns I have, though, aside from minor cleanliness concerns like using KMONS over MONS. The vault is also too long to let people without spoilers know the vault actually contains Arga- you could shorten it, or you could use glass in the back to directly allow autoexplore to eventually find out what's actually in the vault. I'd also prefer to lower the quicksilver count to 3, myself, and I don't think living dragons really fit Crypt enough to justify it placing there compared to Vaults and Depths already having quicksilvers alongside other dragons.

With your approval, I could clean this up and tweak it myself before merging it in.

orjb1:
Assuming you don't mind, I can try and do the tidy up myself this week and push?

I'm keen to get to the point where my PRs do not need work from devs bar a quick merge?  (I have _more_ ideas...)

PR #3471, opened 2024-01-14
Fix missing descriptions for Okawaru's "Receive {Weapon,Armour}" ability

seandewar:
Fixes "No description found" when examining these with "a?" or "?/a".


wheals:
Thanks for the catch!

PR #3484, opened 2024-01-22
Update ZH translation

zhangjoke:
This version has a lot of updated content, so it took more time to translate the content than before.


PleasingFungus:
Thanks for this! Looks like it took a lot of work. :)

We're mostly in code freeze right now during the tournament, but I'll merge this afterward.

PR #3497, opened 2024-01-29
fix: Update game.html tags

alexjurkiewicz:
There were a number of `<div>` and `<span>` tags that were self-closed (eg `<div/>`). Trailing slashes in opening tags are ignored by HTML parsers and by convention only used for "void" tags (those that don't support closing tags).

This led to HTML parsers in some cases misparsing the document, where eg `<div/><div/>` would be interpreted as `<div><div></div></div>` instead of the intended `<div></div><div></div>`.

Read more:
https://developer.mozilla.org/en-US/docs/Glossary/Void_element#self-closing_tags

Convert all `<div>` and `<span>` to standards compliance.

Also, make the following (mostly cosmetic) changes with [Prettier](https://prettier.io/):

* Consistent 2-space indentation
* Add a trailing slash to all `<link/>` tags
* Don't use HTML comments to keep sequential `<span>`s on the same line
* Clean up inline style strings
* Consistent JS quoting

### DIFF
--- a/crawl-ref/source/webserver/game_data/templates/game.html
+++ b/crawl-ref/source/webserver/game_data/templates/game.html
@@ -1,107 +1,160 @@
 <script>
   require.config({
-    paths: { "game-{{ version }}": "/gamedata/{{ version }}" }
+    paths: { "game-{{ version }}": "/gamedata/{{ version }}" },
   });
-  require(['jquery', 'client'], function ($, client) {
+  require(["jquery", "client"], function ($, client) {
     client.inhibit_messages();
     window.game_loading = true;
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
I've been playing with the client renderer. I'm not sure how things actually work now 😳 In my local testing, there are lots of rendering errors, like the minimap being park of the RHS stats panel. Regardless, I've tested these changes and they don't appear to break anything, and they should prevent weird issues in future.

PleasingFungus:
Looks good! Hopefully this doesn't break anything, yep. :)

Apologies for the long delay in response.

PR #3504, opened 2024-02-05
remove tentacle submerge, and the rest of submerge

Monkooky:
Submerge as a mechanic introduces very unique special cases that must be guarded against throughout the code. The untargettability it grants requires a special check in several targetters, and the ability to temporarily coexist with monsters makes removing the enchantment very complicated.

Right now the only monsters that can submerge are tentacle segments, for whom the ability is fairly uninteresting. Preserving submerge for tentacle segments is decidedly not worth the technical debt it incurs.

This does not touch the tile drawing code; it may still be possible to see 'submerged' tentacles if a tentacle somehow gets disconnected. There may be vestigial code for drawing other submerged entities.


crawl-ref/source/tilebuf.cc
```diff
@@ -242,6 +242,7 @@ void SubmergedTileBuffer::add(tileidx_t idx, int x, int y, int z, bool submerged
     int alpha_top = ghost ? 100 : 255;
     int alpha_bottom = ghost ? 0 : 40;
 
+    //TODO: this seems to be unreachable; see if it can be removed
```
PleasingFungus, line 245: This is needed for any case in which a player or monster is standing in water.

PleasingFungus (approved):
Nice PR! Very happy to see the end of this code. :)

PleasingFungus:
See 465344b4efd168a361204d1d8331f8f906e98e94 for some additional small cleanup. Overall, quite tidy and thorough, as far as I could tell. :)

PR #3515, opened 2024-02-05
Make rCorr halve corrosion stacks on the player instead of a coinflip chance to prevent

SentientSupper:
Protecting yourself against corrosion with rCorr has feelsbad moments where you get corroded for the full amount. It's not noticeable when it prevents corrosion because some sources of corrosion aren't guaranteed to begin with.

Each stack of corrosion on the players reduces AC and slay by 1. Quadrupled most sources of corrosion to compensate. Having corrosion resistance will halve the number of stacks applied to the player instead of a coinflip chance to prevent the status entirely. This should hopefully make it feel more consistent. 

Corrosion from slime walls and Dis are not affected by this change, though this change makes it easier to halve the corrosion from slime walls if need be. Corrosion against monsters is not touched as well.

### DIFF
--- a/crawl-ref/source/attack.cc
+++ b/crawl-ref/source/attack.cc
@@ -1096,7 +1096,7 @@ int attack::player_apply_slaying_bonuses(int damage, bool aux)
                         || (weapon && is_range_weapon(*weapon)
                                    && using_weapon());
     damage_plus += slaying_bonus(throwing);
-    damage_plus -= 4 * you.corrosion_amount();
+    damage_plus -= you.corrosion_amount();
 
     // XXX: should this also trigger on auxes?
     if (!aux && !ranged)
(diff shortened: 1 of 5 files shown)


PleasingFungus:
Thanks for the nice PR! Looks quite clean.

We talked about this in IRC last week. There were some concerns that this would lead to players resting off corrosion more often, since you're more likely to get corroded even when you have rCorr. This is especially annoying with oklobs.

My feeling is that these are real issues, but smaller than the better 'playfeel' of having rCorr consistently halve corrosion. I'm willing to give it a try. :)

PleasingFungus:
How would you like to be credited in `CREDITS.txt`?

SentientSupper:
"SentientSupper" would do.

PR #3519, opened 2024-02-06
Let players win the game with Rising Flame

grumposus:
Ignis's Rising Flame ability allows the player to travel rapidly to the floor above them, if that floor is a valid destination (next higher floor of the current branch, branch entry if they're at the top floor of a connected branch).  It does not, at present, work on D1...but if they player has the orb, they SHOULD be able to Rising Flame their way out of the dungeon to glorious victory.  This pull request implements this EXTREMELY IMPORTANT feature, the absence of which has no doubt caused untold suffering.

I have tested and confirmed this works in local tiles on Windows.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -1485,6 +1485,10 @@ static bool _can_rising_flame(bool quiet)
             mpr("You're already rising!");
         return false;
     }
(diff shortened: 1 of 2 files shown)


grumposus:
dangit I realized I overflowed the commit message length on this.  If you want this pull request and that's annoying to resolve, let me know and I'll redo this with a shorter commit message.

semitonal:
Merged, thanks for the patch!

PR #3523, opened 2024-02-07
Mini Sif Muna Rework

SentientSupper:
Going through a bunch of Sif Muna wins, a majority of them ended with 0 uses of her amnesia ability. I can't help but feel Sif could use a bit more than just channeling, spell gifts (which other gods have) and the rarely used divine exegesis, so I made a bunch of changes which could make going Sif feel better.

Divine exegesis's spell power bonus has been reworked to give a flat spell power bonus of Invo * 150 / 27 while having a minimum spell power of Invo 200 / 27. This is intended to make divine exegesis better when using it on spells that you don't have much training in as well as spells than you can currently already cast. For an ability with a heavy piety cost, its effects can be underwhelming. If you divine exegesis a spell you didn't train for, it will likely be at spellpower so low you're better off casting what you already can cast, while a spell you can already cast will often become ~33% better in terms of damage which is not worth using the ability for. This will also make training Invocations past the 10-15 range better.

Divine exegesis's piety cost has also been reduced when used with lower level spells. The cost is 4-6 with level 1 spells which scales linearly to 12-18 (same as current) with level 9 spells. This is intended to make divine exegesis less bad if you haven't found level 9 spells yet while maintaining the ability's main use with high level spells.

Sif also grants friendly fire protection to summons and spells at 2*. This is intended to allow Sif worshippers to use spells more flexibly in situations where self damage may occur and to grant a more subtle power boost which contrasts with Vehumet's raw destructive power. This also allows summons to be used with spells like Starburst or Refrigeration, allowing worshippers to pursue a summoner/conjurer hybrid build without needing to go Servitor every time. While it is possible to do that with current Sif using Ignition/MCC, this forces you to skip multiple lower level spells which doesn't feel good.

Currently this is implemented using other gods' protection code, which means it also affects things like immolation scrolls, ranged attacks and evocables. This could be improved but I don't feel like putting in that much effort for something that might not be merged in the first place.

Sif's piety decay has also been adjusted to match slow decay gods like Chei and Fedhas since Sif is notoriously the hardest god to reach 6* of piety with with most worshippers hovering between 4* and 5* piety. After testing this change, 5* piety is still where you're going to end up most of the time so it couldn't hurt slowing down her piety loss a little.

These changes aren't final and are intended to be refined or discarded after more testing.


randomdevpete:
There was some recent dev discussion around Sif. It is not generally felt that Exegesis is not powerful enough, but a different approach is planned for making the ability more useful rather than simply boosting spell power.

There is reasonable support for the idea of removing piety decay altogether across ALL gods, but this requires a rather more significant look at balancing piety gain. Slowing down the decay doesn't do as much as you think here and doesn't solve the core problems with the gifting system.

So I'm going to close this but if anyone wants to improve sif please enquire in #crawl-dev about specifics of existing plans.

PR #3524, opened 2024-02-07
Give deep elves a quickcasting passive

SentientSupper:
After humans inherited the meteorans' explore regeneration, deep elves are now one of the two races to be told of their mediocrity when opening their abilities/mutation screen remaining with the other race (hill orcs) soon to receive changes and could use a bit of spice. They also lose part of their advantage in extended when any other species get to max out a few skills to level 27. Although their current status is fine, they are also one of the lowest winrate races so it wouldn't hurt to make them a bit more exciting to play.

This pr gives them the quickcast mutation at level 1 which increases their spellcasting speed by 25%, allowing them to cast spells at 8 auts rather than 10 like every other race. This works with channeled spells like searing ray and flame wave but not MCC. 

Regarding balance, this is lowest I can make it while still having its effect remain noticeable. While it may seem strong, they are still casting from the same pool of mp and it benefits summoning spells, necromancy buffs and dot spells like frozen ramparts, freezing cloud and poison spells less. Could be tweaked after more testing.


elliptic:
As I said in discord when this suggestion came up, I am against unnecessary use of unusual delays for spellcasting. This is one of the few ways in which spellcasting is simpler than melee (mostly it is far more complicated and intimidating to use) and I'd like to keep it that way.

My proposal for a simple spellcasting-related buff for DE was to increase their base MP regen by some amount, though I don't personally think any buff is necessary.

SentientSupper:
I think it should be fine to take a mechanic (attack delays) from melee, simplify it by taking skills and weapon types out of the equation, and confine it to a single species without adding a significant amount of complexity to spellcassting.

I suppose you could slap on the DS mutation that doubles natural mp regen instead, but DS already have that and in my opinion it's not as exciting to have. Balance-wise they are indeed fine, so this is intended to make DE more interesting hopefully without significantly disrupting the balance.

PleasingFungus:
Hi, thanks for submitting this PR! I had a related design a few years ago.

A few notes:

> Although their current status is fine, they are also one of the lowest winrate races so it wouldn't hurt to make them a bit more exciting to play.

I'm quite confused by this. Is the implication that their winrate is low because they're boring to play, or that they're boring because they have a low winrate? They are, and always have been, one of the top species by frequency of play, and are especially popular with newer players. Surely those are better statistical proxies for 'excitement' than winrate?

> I think it should be fine to take a mechanic (attack delays) from melee, simplify it by taking skills and weapon types out of the equation, and confine it to a single species without adding a significant amount of complexity to spellcassting.

The thing with action speed is that it's quite hard to actually notice. The time that the *player* takes is unchanged, and the *impact of their* actions is also unchanged - their spells do the same damage and affect the same radius, etc. It can certainly be very powerful, but it often doesn't *feel* powerful.

Deep Elves gain their complexity from the entirety of Crawl's spell system. Their aptitudes allow them to very easily access a wide range of spells from quite early on - this is the core appeal of the species. As long as we keep making exciting spells, we won't need to add more complexity to Deep Elves.

Elegance is not a crime. :)

SentientSupper:
I don't mean current deep elves are boring in any way. I mean it's ok to not touch them at all, but also that there is nothing to stop you from experimenting with giving them something more to play with.

That being said, I'm not particularly attached to this concept so it's fine.

PR #3529, opened 2024-02-10
Vampire Revamp

SentientSupper:
Vampire form changing is problematic because there is an optimal form to assume for each branch, and switching is actively discouraged because of the penalties. Most players end up picking a form and sticking with it which is uninteresting gameplay.

Instead of being able to switch at any time, vampires now have a blood meter that is charged by gaining xp and storing up to 100 blood. Fully charging your blood meter takes xp equal to half the amount required to level up. This is capped at 3 for each kill to prevent it from being too easily charged in the early game. 

When fully charged, you can revivify over 3 turns to enter a bloodcraze. Successful transformations heal you for ~25% of your max health. This is intended to make revivify useable mid-combat while also distinguishing it from other buffs and panic buttons by giving different considerations (can you survive standing still for 3 turns? would losing your resistances be fatal? does draining work against what you're fighting?).

While bloodcrazed:

- You become alive, losing your undead resistances.
- You cannot regenerate. Half of the lost regeneration becomes MP regen.
- Melee and ranged attacks will drain life from susceptible foes, dealing bonus damage which scales with xl and healing you for half the damage dealt. You also drain on kills through other means like spells and summons.
- You gain long-range enemy detection.
- Your move delay is reduced to 7 auts. (suppressed by chei)

The maximum amount of healing you can receive per turn has a cap which scales with xl to make axes/quickblades not as overpowered. The fast movement speed is meant to help you catch up with and stab enemies who are unaware of you if you trained stealth for stabbing and close the distance with ranged foes instead of having to lure them around corners. It can also be used to escape but that is generally discouraged because bloodcraze is also an incredible offensive ability and charges slowly.

The bonus MP regen is intended to make the ability more worthwhile for casters outside of being an escape while also making guardian spirit and +regen less useless while bloodcrazed.

Each turn, you lose blood unless you drained life from an enemy that turn. Killing a susceptible enemy increases your blood by some amount, extending its duration. When bloodcraze ends, you gain weak for some turns. You can still exsanguinate to end a bloodcraze early (this still gives weakness and sets blood to 0, taking only 1 turn) in case you really need to. 

Bloodcraze is intended to support the stabber playstyle by providing a powerful buff with limited usage which you can use if a stabbing attempt fails. It works well with fast attacks by giving a flat damage boost and healing to each attack. Fighters will also appreciate having the buff and mages too if they encounter an enemy which resists their spells or if they run out of mp. Necromancers will appreciate having a way to kickstart an undead deathball and make the most out of their necromancy spells.

While bloodless:
- No longer have inhibited regeneration.
- rC++ reduced to rC+. (they don't need that much and to be consistent with other undead)
- Stealth++ to stealth+. (they already have high stealth apt and this is now their default form)
- No longer receive 2 levels of frail. 
- No more batform. (we don't need two different transformations that both receive speed boosts)

Other changes:
- Vampires now default to being bloodless. 
- Hp aptitude is now -2.
- Bite is always vampiric regardless of their blood state.
- They lose 1 axe apt. (now -2 because cleaving is very strong with bloodcraze)
- Can shapeshift and be polymorphed normally in both states. (they can mutate already so it wouldn't be weird for their bodies to be a more malleable than other undead)
- Vampiric bite no longer requires target to have blood, reflavored to draw life force instead. (so that the player doesn't need to remember what has blood and what doesn't)

They also gain a new passive ability that lets their vampiric effects ignore rN+ and affect demonic enemies so that they don't lose their entire gimmick in extended. A lot of the most dangerous enemies in extended are undead/nonliving so I don't expect this to be unbalanced while also being flavorful.

Added new tiles for bloodcraze vampires and blood bolt tile (recolored damnation) for vfx. Perhaps there could be a better tile but I'm not a pixel artist.


SentientSupper:
For some reason it's not passing the unbrace test even though I ran the unbrace utility on my side and looked at the code myself.

DracoOmega:
I've recently discussed this PR with other members of the devteam at some length, and while there's certainly broad agreement that vampires could use some changes, the bloodcraze mechanic as proposed is substantially more complex that feels appropriate for a baseline species gimmick, doing at least 8 different things at once (some of which debatably don't intuitively follow from the others).

That being said, I do think that the basic idea of 'toggable super mode that you shift in and out of on a semi-regular basis' has promise as a species gimmick, if it were streamlined and simplified in some ways. Given that there have been serious alternate proposals for what to do with vampires that appear to already have some support, I think it might be more fruitful if you considered what you could do with this mechanic in the context of an *entirely new species*. If anything, I think trying to make it fit the vampire theming is responsible for several layers of its complexity (such as the holiness changes, or needing a solution to how there's so much less blood available in some parts of the game than others).

You've obviously put a fair bit of work into this, and I hope this sort of response doesn't feel demoralizing. When undertaking something of this scale, it may be more useful in future to try to solicit specific feedback *before* so much effort is put into implementation. However, I am serious that a new species proposal incorporating some of the basic ideas here (don't let yourself get too fixated on it being life-steal based!) has the chance of being genuinely looked at.

SentientSupper:
I plan on taking some complexity out of bloodcraze by removing some of the more superfluous bonuses from it like the mpregen. This is a second draft after all and many things can be tweaked after testing if it ever gets tested.

I'm not sure what you mean by the "holiness change" because this pr doesn't touch anything regarding that. Blood gain is entirely based on xp gain so your ability charges no matter where you are even if it's a bit unintuitive. I think it's possible to retheme this rework as a new species but to be frank I think it suits the vampire theme very well and I very much like the lifesteal theme.

I actually did try to solicit feedback by posting on tavern, reddit and discord even though I ended up with no dev feedback at all. Perhaps I could be a bit more annoying about it next time.

DracoOmega:
>I'm not sure what you mean by the "holiness change"

I mean, the semi-living/undead resistances change during bloodcraze mode, sorry.

regret-index:
The holiness change this proposal uses is, strictly speaking, something inherited from the previous alive-versus-dead state of vampires kept up this whole time. It is also, frankly speaking, a very large chunk of acceptable complexity for a given species to have as a base concept alone.

There's four different components of how Bloodcraze is overwhelmingly complicated, really:

1. To make the species count as undead, and so there's a tangible upside to playing a low-apts species outside of a sparingly-available buff, it has to gain and lose rPois+++, rN+++, and rC+ (and Stlth+ for vampires only!) across the two states, a varied set of resistances that shifting regularly makes many monsters across the whole game fluctuate notably in threat. Mummies and ghouls don't change, while turning off Death Form is mostly a matter of preference for Tartarus or for the inconsistent appearance of Holy Pan.
2. To make the buff not just something to blindly spam (c.f. Charms removal), it has a very deliberate duration-gain-and-loss system: a separate meter to gain, slow activation with instant healing, duration going down and up over time with killing. This is far more complicated than comparable god abilities of just hitting or being in sight of monsters (c.f. Heavenly Storm and Blood For Blood).
3. To make the reckless behaviour of maintaining it worthwhile and tangible beyond forcing manual exploration and fighting, it focuses on a no-regen + lifesteal theme that goes against the rules of quite a few other forms of lifestealing already available in the game (including _divine_ lifestealing), despite the fact we don't try to make lots of other species gimmicks equally strong all game long (e.g. formicid digging versus rune vaults, several draconian breath resists, tengu acrobat versus hellpan lords, etc),
4. Then on top of that, there's a lot of extraneous buffs to help the buff status last longer (mp regen, enemy detection, fast movement) entirely.

Points `2` and `4` are relatively easy enough to adjust, it's true, but `2` will still stay something we've otherwise assigned to capstone god abilities. `1` and `3` are specifically tied to being a vampire, and add a ton of complexity burden to what a no-autoplay-buff already requires. The entirety of it may thematically fit together, and we may need something rather dramatic to sell both relatively unpopular undead species overall and vampires in particular as troubled monsters as they are. The proposal is still inherently quite volatile and complex even with more trimming, and it'd be much less trouble to make differently-flavoured undead species with other core gimmicks and an xp-gated-megabuff species as two separate design fronts instead, a track multiple devs have expressed interest in.

(I'd prefer the flavour of vampires over a lot of options on a completely non-mechanical basis. Even so, I'd say some sort of non-lupine lycanthrope would easily pick up the concept and be easily popular for a playerbase pleased by playable dwarves and goblins. It'd also easily focus on defensive benefits and melee upgrades, to contrast reasonably against how Draconians have an offensive / debuffing conjuration focus in the same xp-gated-abilities space.)

Several devs haven't talked about this much because actually extracting a workable concept from the proposal as stands unfortunately requires completely changing its scope and application to get anywhere near acceptable. Out of various options, the notion of splitting a megabuff species mechanic away from undead species entirely is most likely the best path forward.

SentientSupper:
Thanks for the feedback. I think even if it seems complex when reading about it, it doesn't feel that way in an actual game although making it less complicated is definitely preferable.

Seems to me that splitting this off into two races might be what ends up happening in the end. That being said I don't feel like throwing it out just like that given the amount of work I've put into this. I would like if an experimental happened so I that could at least see how players feel about it before sending it off into pr hell (which could help with seeing what direction the next concept of a shifter should go).

There are also a bunch of other things like shapeshifting and polymorphing normally in both states or vampiric bite not requiring blood that I feel are suited for vampires excluding the whole bloodcraze mechanic and I wonder how devs feel about those.

regret-index:
While rough concepts for a defensive super-mode species are good, Crawl's playable vampires specifically have been handled through the mix of 6737559, f08bb6f, and 60819be, and as such, I am closing this old PR.

PR #3530, opened 2024-02-11
EXPERIMENTAL: Reenable glowing players; let them burn monsters.

pooka109:
This uses the same formula as the cloak of Starlight does for blinding. Also, the burning is accomplished via the formula for Scorch (still too strong at power 5).

Warning: The burning is still too strong, even at power 5, and it currently crashes if it actually kills a monster.


pooka109:
After overhauling some things massively, the glowing mutation is now 3 levels instead of 1; there is no corona effect on monsters from it anymore; it now works when the player is damaged in melee (a la foul stench), blasting enemies with befouling fire; befouling fire is effectively cleansing flame in reverse, giving undead and demons half damage and holy beings/good god worshippers 1.5x damage.

The current damage formula is that of Scorch at power 50 at level 1, power 100 at level 2, and power 150 at level 3, except that the damage is unresistable except in the aforementioned cases, and goes through armor.

pooka109:
Turned the damage back to roughly that of Scorch, since at the 3rd level of the mutation it wasn't doing much otherwise. It can always be adjusted.

More importantly, cleaned things up so that it has 5 patches instead of as many as 19.

pooka109:
Added exactly one real weapon that uses the foul flame brand: a revived version of the eveningstar "Brilliance" with all of its other old qualities dropped except the +1 enchantment, and a new description of its corruption and foul flame lore.

PR #3537, opened 2024-02-14
Update ZH translation for 0.31

zhangjoke:
Update Chinese translation for Dungeon Crawl: Stone Soup 0.31. By the way, Modify some expressions.

Hoping it could be used in the future revised version 0.31.1.


PleasingFungus:
Thanks, as always :)

PR #3539, opened 2024-02-14
Save the Djinn!

HansSchultz:
Instead of complete random spells from all sets that make the Djinn nearly completely unplayable, give the Djinn random spells from a reasonable set, where this is a suggestion.

Consider characters based on Fire, Ice, Air, Earth and Alchemy as elementalists and they get random skills from within those sets, as you need more than one element to defeat various enemies that are immune to attacks from one or more elements.

Conjurers get their skills from the Conjurer set alone that is quite large.

Hexslingers get spells from the Hexes and Ice schools, with added spells Infestation and Shatter to patch the set.

Summoners gets spells from Summoning set and Earth set as otherwise the character will not be viable, and the Earth set is a powerful supplement. Moreover, Summon Ice Beast is used for a level 2 spell as there is no additional level 2 spells besides the initial spells from the Summoning spell set.

Necromancers are based on the school of Necromancy spell set only, but since there are several spell levels that might not have an option, it is heavily patched. The patched spells used are:
Mephitic Cloud, Iskanderun's Mystic Blast, Orb of Destruction and Polar Vortex.

Warpers use spells from the school of Translocation, but as with the Necromancer there is a need for many patched spells. The patched spells used are:
Irradiate, Spellforged Servitor, Maxwell's Capacitive Coupling, Chain Lightning and Dragon's Call.

Enchanters were difficult to define, and as they are not elementalists, the decision was to use spells from the Hexes, Necromancy, Translocation and Conjuration spell schools, thus giving them a wide range of spells.

The characters have been tested (not to the highest skill) and are fine playable, with the Warper being the toughest as the Warper is partly melee, but the Djinn is not the best melee species.

### DIFF
--- a/crawl-ref/source/ng-setup.cc
+++ b/crawl-ref/source/ng-setup.cc
@@ -430,8 +430,7 @@ static void _setup_innate_spells()
             case JOB_ICE_ELEMENTALIST:
             case JOB_AIR_ELEMENTALIST:
             case JOB_EARTH_ELEMENTALIST:
-                if (!(typeflags & spschool::conjuration)
-                    && !(typeflags & spschool::alchemy)
+                if (!(typeflags & spschool::alchemy)
                     && !(typeflags & spschool::fire)
                     && !(typeflags & spschool::ice)
                     && !(typeflags & spschool::earth)
@@ -441,7 +440,7 @@ static void _setup_innate_spells()
                 }
                 break;
             case JOB_HEXSLINGER:
-		if (!(typeflags & spschool::hexes)
+                if (!(typeflags & spschool::hexes)
                     && !(typeflags & spschool::ice))
                 {
                     continue;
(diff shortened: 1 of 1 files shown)


semitonal:
Hi, and thanks for the pull request! Unfortunately this isn't a change that would be merged - the premise here (that Djinn are "nearly completely unplayable" due to the random nature of their spell sets) is a bit of a flawed one. Djinn spell sets being fully randomised is very much an intentional feature of the species, distinguishing them from other spellcasters and encouraging them to use spells that other characters might not choose to memorise. Design-wise, Crawl's backgrounds are also intended to be a starting kit only, with the player adapting to what they find in the dungeon in terms of items, spells and gods, so while starting spells will obviously have a big impact, they don't and shouldn't dictate your entire character (in the same way that, for example, Demonspawn mutations and Draconian colour aren't in any way affected by your starting background).

The online stats also indicate that Djinn have a slightly above-average winrate so they're certainly very much playable to a good degree of success.

In general if you're interested in contributing to Crawl, it's often worth raising ideas in the development IRC channel (#crawl-dev on Libera.Chat) - folks there will be able to give feedback and can let you know if a proposed change is likely to be accepted.

PR #3566, opened 2024-02-18
Remove the "Eos" unrand, along with its halo.

pooka109:
The halo overlaps with the recently revived "Brilliance", and, without it, "Eos" is just a +10 morningstar of electrocution. It's also not placed by any vaults.


pooka109:
Not as much of an issue since Brilliance was changed to have an umbra instead of a halo.

PR #3568, opened 2024-02-18
Make Krakens scary again

Monkooky:
Old Krakens functioned as a quasi-stationary enemy, being easy but nontrivial to run from, difficult to kill safely, and dangerous enough that having run from a kraken warped the rest of the floor.

However, modern Krakens are easily trivialized. Scrolls of poison will deplete most of a kraken's health before it reaches you, aoe spells or axes enable you to cut down tentacles as fast as they spawn, and if all else fails you can stand at the edge of LOS with a ranged weapon and repeatedly step out of LOS if the tentacles reach far enough.

This branch seeks to change anti-Kraken tactics from solely applying AoE to instead encourage hitting the head- ideally, a more interesting risk-reward decision. Additionally, it seeks to mitigate or remove some of the degenerate tactics for fighting krakens.

The main changes in this commit are:
1: Quarter the damage the head takes when tentacles are hurt. This both makes AoE do less to trivialize Krakens, and helps disincentivize sitting at the edge of tentacle range and killing one tentacle at a time

2: Make tentacles not require LoS, and reach out to 2 past it This strongly disincentivizes strategies where the player repeatedly leaves and re-enters LoS.

3: Make the head more frail and less dangerous
This seems necessary to make krakens feasible to fight for melee characters.

4: Give Krakens fast regen
This is necessary for much the same reason why it's needed on stationary monsters- to make repeated disengagement a losing proposition

Less central to the goal is the change to tentacle stats, making them more frequent, fast moving, and frail. Frequent and fast moving were chosen to lower the window in which a character with a bow can shoot a kraken head unopposed and increase the danger of repeated attempts; frail both limits the effectiveness of single high aoe hits, and allows melee characters to more effectively fight their way to the kraken's head with tentacles in the way.

### DIFF
--- a/crawl-ref/source/dat/mons/kraken-tentacle-segment.yaml
+++ b/crawl-ref/source/dat/mons/kraken-tentacle-segment.yaml
@@ -6,7 +6,7 @@ genus: kraken
 will: invuln
 attacks:
 hd: 12
-hp_10x: 480
+hp_10x: 240
 ac: 5
 ev: 7
 intelligence: animal
--- a/crawl-ref/source/dat/mons/kraken-tentacle.yaml
+++ b/crawl-ref/source/dat/mons/kraken-tentacle.yaml
@@ -7,12 +7,13 @@ will: invuln
 attacks:
  - {type: tentacle_slap, damage: 29}
 hd: 12
-hp_10x: 480
+hp_10x: 240
 ac: 5
 ev: 7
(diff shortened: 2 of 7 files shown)


PleasingFungus:
Quick comment: seems probably reasonable but would require quite a bit of review. I'm told DracoOmega is doing that review, so will defer. :)

PR #3587, opened 2024-02-24
remove redundant "unblindable" tag from elementals

yrdzrfxndfvh:
nonliving monsters have been immune to blindness since commit b856850


PleasingFungus:
Nice fix, thanks! :)

PR #3589, opened 2024-02-24
Rework Ghouls

SentientSupper:
Ghouls are one of the least popular races and they could use a touch-up to make them more interesting to play. This rework turns them into blood-thirsty melee/caster hybrids who regain a temporary moment of lucidity when feasting upon the life energy of slain foes.

Ghouls now temporarily gain stacks of lucidity up to 3 when they slay an enemy in melee. When a spell is cast, it will consume 1 stack of lucidity to gain a spell power boost (double before stepdown, ~40% bonus damage in practice for damage spells), cost no mp and +wiz.

To push players towards making use of this gimmick more, they gain 3 ranks of low MP, giving them 30% less MP.

Their heal-on-kill now works against all enemies including the undead and non-living. Ghouls and vampires are the only species in crawl which have innate traits that get completely denied by the types of enemies they're facing. Normally it's fine to have enemies who are immune to certain things because it forces the player to come up with a new strategy, but in this case having an enemy say "nope you're worse than a human now" just feels unfun. Demonspawn powered by death works against non-living and undead too so it should be fine here for the sake of gameplay. Kills by allies now also trigger heal-on-kill to remove antisynergy with ally gods and summons. It's a buff to their trait, but I think it's healthier to have -Regen exist to balance out heal-on-kill than to have heal-on-kill be a trait that makes -Regen hurt a little less.

Some apts have been shifted around slightly.
- Ranged and throwing from -1 to -2. Since ranged attacks don't give lucidity, make it more obvious you aren't good with them. 
- Summoning from -1 to -2. Summons don't work well with lucidity, so make it more obvious you aren't good with them.
- Conj from -2 to -1 to push players towards using conjurations more. 
- Fire and Air from from -2 to -1 to make using non-ice/earth schools slightly better for better variety in choosing spell schools.

TODO:

- Further balancing if necessary
- Maybe have lucid spell haste return if the idea is popular
- Maybe replace lucid +wiz with something else. While I understand the motive behind not having temporary sources of spell failure changes and agree with the commit that removed wiz from brilliance, I personally want it to stay because I think it's crucial to letting player ghouls make use of their trait in the early game by allowing them to use their spells while focusing on training up their melee skills. It avoids some of the issues with temporary wiz by being available often enough. Without it, there's a high likelyhood of pure melee players staring at the lucid status and sighing because they can't cast any spells and training both magic and melee skills equally would probably just leave you without a suitable melee or magic killdudes.


elliptic:
I really don't want to have a spellcasting delay change - we avoid that and only use it for specific spells where it is really essential to the design. Always taking 1.0 is one of the few simple things about casting.

I'm also generally skeptical that this sort of buff is enough to make melee ghouls pick up spells as well, and I'm concerned than the -MP will just make casty backgrounds on ghoul less appealing.

SentientSupper:
I hope it's fine because it's actually just haste+brilliance for one spellcast. The hasting is important because melee/caster hybrids often run into the turn economy problem of "why should I cast this low-mid level spell for less damage than my axe?". This also gives them more reason to use a lucid spell when they've already hit the spell power cap for a spell.

I've found playtesting ghouls as hybrids after this change to be pretty fun. A pure casting ghoul before changes is just the same as casting spells on other species but you're worse at it and evidently this isn't very popular so removing this playstyle should be fine.

Hellmonk:
Closing because ghoul got reworked into revenant which has some similar features. Feel free to continue experimenting with the gimmick and resubmit as another species, though. There may still be some room for a "melee, then cast" species as opposed to the "cast and get to melee" gimmick of revenant.

PR #3593, opened 2024-02-26
Ban -Cast gear for wizard orc apostles.

pooka109:
There's currently nothing preventing wizard orc apostles, when they get artefact gear, from getting the -Cast property on that gear. This should fix that, albeit by crudely rerolling such artefacts via do/while loops until they don't get the -Cast property.

The orb of energy that orc wizard apostles sometimes get is the one bit of gear that doesn't need this, since -Cast is a conflicting property for orbs of energy.


pooka109:
-Cast doesn't actually do anything to monsters, and it could be flavored as Beogh blocking the puny -Cast attribute on his wizard followers, so this is likely not worth the effort.

PR #3594, opened 2024-02-26
refactor handle_monster_move

Monkooky:
handle_monster_move uses a global variable 'mmov' which is manipulated by several functions. This obfuscates how mmov is set, and makes it difficult to be confident about the current state of mmov at any given time.

This refactor makes several functions return a new value for 'mmov' instead of directly manipulating it, and renames such functions to better document their function. Other functions are directly passed a reference to mmov with minimal change to functionality.


DracoOmega:
Merged into trunk, thanks!

PR #3601, opened 2024-02-28
Add option to remove dropped items from autopickup

SentientSupper:
Adds a new option: drop_disables_autopickup which removes all dropped items from autopickup automatically. A QoL thing that allows you to drop items (especially from local tiles inventory panel) without needing additional keypresses to enter the autopickup menu and disable autopickup manually. Defaults to off.


PleasingFungus:
Not sure how many people will use this, but we can try it.

PR #3602, opened 2024-02-28
Added improved behavior for allies

SentientSupper:
This commit adds BEH_STICK, a behavior that causes allies to stick to the player while still attacking enemies that they can reach.

Managing suicidal allies is often an exercise in tedium as they rush forward to give up their lives in the name of the chosen one while blocking your line of fire to the enemy and waking up enemies you intend to stab. Faced with this, the player has three choices.

1. Constantly spam t-f to remind their allies that their life is precious. This is however annoying and also stops allies from attacking. They will also resume attacking the moment that you do so you can't use any ranged attacks without them leeroying.

2. Charge forward together. This means melee brute builds are heavily favored and ranged/magic/stabber builds are disfavored. This also puts the player in a bad position when their allies die and they are now surrounded in the open.

3. Just let them die :(.

With the "stick together" command accessed from the yell menu, allies will now stay as close as they can to you while attacking any enemies within reach, holding your formation until you give the order to charge. Recalled and permanent allies like ancestors and apostles will default to this behavior while temp summons will still use the old behavior.


PleasingFungus:
Interesting idea. Code and design will both take some time to review.

Elliptic suggested that ancestors should keep the current reckless behavior, but I know other people have requested more caution. I don't feel strongly.

Hellmonk:
There appears to be no interest in this at the moment, so I am closing it.

PR #3607, opened 2024-02-29
make regen/regenMP gizmos attune

Monkooky:
previously, if tinkered up while not at full hp/mp, would never attune.


pooka109:
It's in. Thank you.

PR #3608, opened 2024-02-29
Prevent crash on equip

### LINKED ISSUE #3606: Equipping a hand cannon or amulet of reflection crashes the game
windows local tiles

0.32-a0-562-g199a9b2549

**DESCRIPTION:**
Equipping a hand cannon or amulet of reflection as any species crashes the game.

Monkooky:
prevents several crashes introduced  in 4e4bab2fd2a8b753dc4cfd1b8ffeccd21831bb09

Fixes #3606

### DIFF
--- a/crawl-ref/source/player-equip.cc
+++ b/crawl-ref/source/player-equip.cc
@@ -1059,7 +1059,8 @@ static void _handle_regen_item_equip(const item_def& item)
     const bool regen_hp = (is_artefact(item)
                            && artefact_property(item, ARTP_REGENERATION))
                           || item.is_type(OBJ_JEWELLERY, AMU_REGENERATION)
-                          || armour_type_prop(item.sub_type, ARMF_REGENERATION);
+                          || (item.base_type == OBJ_ARMOUR
+                              && (item.sub_type, ARMF_REGENERATION));
 
     const bool regen_mp = (is_artefact(item)
                             && artefact_property(item, ARTP_MANA_REGENERATION))


crawl-ref/source/player-equip.cc
```diff
@@ -1059,7 +1059,8 @@ static void _handle_regen_item_equip(const item_def& item)
     const bool regen_hp = (is_artefact(item)
                            && artefact_property(item, ARTP_REGENERATION))
                           || item.is_type(OBJ_JEWELLERY, AMU_REGENERATION)
-                          || armour_type_prop(item.sub_type, ARMF_REGENERATION);
+                          || (item.base_type == OBJ_ARMOUR
+                              && (item.sub_type, ARMF_REGENERATION));
```
matta, line 1063: Thanks, I ran into this crash too.

FYI don't understand what the fix is trying to do. I think this might compile, but the comma operator evaluates the first arg, discards it, then evaluates to the result of the second, so this new expression equivalent to:

```c++
  || (item.base_type == OBJ_ARMOUR)
         && ARMF_REGENERATION
```

...which does not look right.

matta: Ahh, maybe this is a typo and you want:

```
    || (item.base_type == OBJ_ARMOUR
              && armour_type_prop(item.sub_type, ARMF_REGENERATION))
```

to match.

But maybe even better to share these expressions with the new `_is_regen_item()` function in `player-reacts.cc`, also added in https://github.com/crawl/crawl/commit/4e4bab2fd2a8b753dc4cfd1b8ffeccd21831bb09

pooka109: And right below it, the regen_mp bool looks like it should use the _is_mana_regen_item() function from player-reacts.cc.

Monkooky: agh, it was a typo

pooka109:
Should be fixed now.

Also used the functions to check for regen/mana regen items, as matta brought up.

PR #3617, opened 2024-03-01
Switch wizmode item creation to use letters.

### LINKED ISSUE #3610: Wizard mode item creation keyboard shortcuts for scrolls/potions do nothing
local tiles, Ubuntu 22.04.4 LTS

0.32-a0-569-g667c705b62

**DESCRIPTION:**
This can be reproduced as follows (assuming that prompt_menu = true in your rc file):

1. Activate wizard mode and try to create an item, via "&o".
2. At the popup window, try to create a scroll by pressing "?". or a potion by pressing "!". Instead of prompting you for which scroll or potion type you want to create, the game will act as though no key was pressed. (All other key shortcuts, when you press them, prompt you as expected.)

pooka109:
Instead of item symbols, since "!" and "!" currently get misinterpreted as commands and don't work as expected.

(This should close #3610, but the menu will be less flavorful with this applied.)

### DIFF
--- a/crawl-ref/source/wiz-item.cc
+++ b/crawl-ref/source/wiz-item.cc
@@ -96,18 +96,21 @@ void wizard_create_spec_object_by_name()
 
 void wizard_create_spec_object()
 {
-    string title = "Which item class (ESC to exit)?";
-    vector<WizardEntry> options =
+    vector<WizardEntry> options;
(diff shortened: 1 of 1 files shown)


pooka109:
Obsoleted by a better solution.

PR #3624, opened 2024-03-05
New tier 2 form: conduit form

SentientSupper:
This commit adds the prototype talisman, which transforms one of the user's arms into a mass of magic cables which discharge spells uncontrollably. 

Each melee attack has a chance to cast a chosen spell without costing magic once per turn, with the chance rising with Shapeshifting skill but falling with the level of the spell. The cables can discharge spells even when silenced, but the artificially-implanted magic circuits drain the user's own magic capacity and prevent the regular use of spells. Additionally, the exposed magic circuits can absorb the user's own spells, preventing damage and dissipate magical contamination at a faster rate.

This is a very fun form that opens up a lot of new build possibilities.

While transformed, you gain:

- 5 int
- Immunity to your own spells
- Increased decontamination rate
- 2/3 less max mp which doesn't stack with antimagic

While transformed, equipped shields and weapons in the off-hand are melded. The messy cables are hard to coordinate, and cannot be used to wield a weapon. You also lose the ability to cast spells the regular way outside of Sif divine exegesis.

This is a tier 2 form because I want it to be usable earlier while low-level spells aren't useless yet.

Other things this pr adds:

- Int modifier for forms
- Refactored simulacrum spell to allow querying its summon cap and current summon amount
- Fixed a melded jewellery slot message for forms

TODOs
Add form tile (current tile is a placeholder. I'm not a pixel artist so I'm fine with someone else doing this)
Add talisman tile
Balancing


Paperbell:
Maybe this could be tier 5 but alternate with death form.

SentientSupper:
The current plan is to retheme this form as Conduit form.

Hellmonk:
closing pending forms additions later in 0.33.

PR #3625, opened 2024-03-05
Fix jinxbite doing extra attacks

### LINKED ISSUE #3581: Jinxbite double trigger and inconsistency
windows local tiles

0.31

**DESCRIPTION:**
Jinxbite can trigger twice on the same target if your target blocks your ranged attack.
It can also trigger with darts and throwing nets if your target blocks or dodges it when it usually doesn't on hits.

**EXTRA FILES:**
image
image

WizardIke:
Fix jinxbite doing two attacks when shooting a ranged weapon that is blocked. Also, remove jinxbite attacks when blocking and dodging darts and nets (they never did attacks on hit).

Should fix #3581

### DIFF
--- a/crawl-ref/source/attack.cc
+++ b/crawl-ref/source/attack.cc
@@ -84,8 +84,6 @@ bool attack::handle_phase_blocked()
     if (attacker->is_player())
         behaviour_event(defender->as_monster(), ME_WHACK, attacker);
 
-    maybe_trigger_jinxbite();
-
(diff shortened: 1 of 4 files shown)


PleasingFungus:
Thanks for this! This change isn't quite trivial so it'll take a *little* time to review, but we should be able to merge it sooner than later.

DracoOmega:
Merged

Issue #3626, opened 2024-03-05
Hepliaklqana ancestor proper nouns miscapitalized and given articles


naddleman:
**PLATFORM and OS or BROWSER:**
Local console, online console

**DCSS VERSION AND GAME INFORMATION:**
` 0.32-a0-613-g40f424ee2b`

**DESCRIPTION:**
Hepliaklqana ancestor names can begin with lowercase letters but these will be capitalized when beginning a sentence:   ![Screenshot from 2024-03-05 09-38-48](https://github.com/crawl/crawl/assets/9171245/c7875b8a-1115-448d-bc21-d4b17ea2c3e3)

Swinging at an ancestor named "foo" gives the message "Really attack the foo?"
![Screenshot from 2024-03-05 09-39-29](https://github.com/crawl/crawl/assets/9171245/a8bce4ea-442e-490f-b509-2e1e6138685f)

I think this is `stop_attack_prompt` in [fight.cc](https://github.com/crawl/crawl/blob/master/crawl-ref/source/fight.cc#L1206) 

**EXTRA FILES:**

wheals:
Partly fixed. The assumption that the first word of a line is capitalized is pretty baked into crawl's message system, but I agree it's not ideal.

naddleman:
Extra capitalization also happens in `xv`

PR #3627, opened 2024-03-05
Remove "the" when swinging at Hepliaklqana ancestor

naddleman:
Fixes part of (#3626)


wheals:
Thank you for the contribution! I went with a more general fix in 9e401d1

PR #3633, opened 2024-03-07
Gozag shouldn't offer jewellery shops to Coglin

chucksellick:
Jewellery shops have a 10% chance per item of generating a talisman, but it's highly unlikely you'd buy one for that reason when general stores can include them anyway.

### DIFF
--- a/crawl-ref/source/god-abil.cc
+++ b/crawl-ref/source/god-abil.cc
@@ -3295,6 +3295,10 @@ bool gozag_call_merchant()
         {
             continue;
         }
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Thanks!

Minor:
- Don't forget to run `util/checkwhite` and `util/unbrace` to make our linter happy :)
- Commit messages should wrap at 80 characters. (And the first line should be no more than 56 characters, though we break that rule regularly.

PleasingFungus:
Oh, third minor point: commit messages should be imperative. So, `Forbid jewellery shops from Gozag for Coglins`, or similar.

PR #3641, opened 2024-03-10
Make slugs not look like stones

SentientSupper:
Makes hand cannon projectiles use a new tile. Feels weird for the description to say it shoots metal slugs while the actual tiles look like sling stones.


crawl-ref/source/tilepick.cc
```diff
@@ -2481,6 +2481,7 @@ static tileidx_t _tileidx_missile_base(const item_def &item)
         }
 
     case MI_SLUG:
+        return TILE_BOLT_IRON_SHOT;
```
PleasingFungus, line 2484: I don't think this matters (these aren't real items), but I'd be dubious about using iron shot as an item tile if it did matter.

(This whole section should probably be cleaned up.)

PleasingFungus (approved):
Thanks! Looks good other than the line I commented on, which I don't believe is used. I can merge this later.

T-Minutiae:
I much prefer the way it is now... "Slug" is simply another word for any solid projectile/bullet. The iron shot animation looks gigantic and hyper-impact in comparison and not like something that would come out of the hand cannon. Plus it would be slightly confusing sharing the same animation as the spell (kobold blast miner with hand cannon comes to mind).

PleasingFungus:
Very good point. I'll think about it more.

SentientSupper:
It makes the mule unrand attacks look more appropriate because that one actually emulates the iron shot knockback effect. Most blastminers spawn with arbalests so this would make one with a hand cannon easier to spot. Plus if you're a coglin with a sling and a handcannon, I think it would be better if the projectiles from your two different weapons looked different. 

Could perhaps use a new tile that's not iron shot's though. I saw some tiles for branded stones and those are suitably magical for some enchanted shots but I don't like how round they are.

SentientSupper:
I've added a new tile that's based on iron shot but with less outline (to reduce the impact) and colors taken from the enchanted boomerang tile (because hand cannons enchant their projectiles).

PR #3645, opened 2024-03-10
Improve colours of spells in spellbooks

NormalPerson7:
Currently, in WebTiles, the colours of spells in spellbooks are far from useful. The most important question when x-v-ing a book (e.g. in a gauntlet, or a ghost vault) and seeing its spells is "do I already have this spell in my library?". There was previously no way to check this without leaving the x-v menu and manually checking.

With this PR, all spells will be greyed out in book descriptions if you already have them in your library (or if they are useless, or if you have memorised them via Vehumet), lightred if forbidden, lightgrey if eventually useful, and lightblue if you could memorise it immediately, in that order of priority.

Better suggestions for colouring are welcome. I had wondered whether it would be better to change WebTiles to use the local tiles (and I presume console) formatting for spellbook descriptions, but I figure there's a reason why WebTiles and local tiles have different formatting in the first place. On the other hand, doing so would bring spell tiles to local tiles' spellbook descriptions, and it would bring the "already known" column to WebTiles, which would also solve this problem.

### DIFF
--- a/crawl-ref/source/describe-spells.cc
+++ b/crawl-ref/source/describe-spells.cc
@@ -264,13 +264,10 @@ static int _spell_colour(spell_type spell, const item_def* const source_item)
     if (!source_item)
         return spell_highlight_by_utility(spell, COL_UNKNOWN);
 
-    if (you.has_spell(spell))
-        return COL_MEMORIZED;
-
-    // this is kind of ugly.
-    if (!you_can_memorise(spell)
-        || you.experience_level < spell_difficulty(spell)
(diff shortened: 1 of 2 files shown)


PleasingFungus:
This looks neat! Doesn't look very hard to review, so I'd hope we could get this in sooner than later.

> I had wondered whether it would be better to change WebTiles to use the local tiles (and I presume console) formatting for spellbook descriptions, but I figure there's a reason why WebTiles and local tiles have different formatting in the first place...

Dangerous assumption. :) I think you're *probably* right, but...

PR #3646, opened 2024-03-10
Add a nostalgic hats-only shop

mainiacjoe:
This is a themed shop that has only hats.  The nostalgia is that the shop owner is Wiglaf.  He has collected a few quite nice hats over the years but now that he isn't adventuring anymore he wants to sell these to fund his retirement.  The Lua code intends that there be two ego hats, two randart hats, and either a single unrand hat or third randart hat.  These hats use the rare hat tiles without duplication.  DEPTH is set deep because these are nice items, though a shallower placement and high greed would work too and maybe be more thematic.

### DIFF
--- a/crawl-ref/source/dat/des/builder/shops.des
+++ b/crawl-ref/source/dat/des/builder/shops.des
@@ -1982,38 +1982,43 @@ ENDMAP
 NAME:   mainiacjoe_wiglaf_hat_shop
 TAGS:   transparent float no_monster_gen no_item_gen no_trap_gen
 KFEAT:  O = altar_okawaru
-FTILE:  .@ = floor_cobble_blood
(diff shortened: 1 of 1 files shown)


pooka109:
If you want to place this on the final floors of the S-branches (Swamp, Shoals, and Snake, I assume?), the DEPTH: syntax for those places appears to be:

Swamp:$, Shoals:$, Snake:$

mainiacjoe:
I didn't have the colon between the branch name and $ duh

pooka109:
No problem.

PleasingFungus:
Fun theme! Unfortunately, I'm not a vault dev, so will have to defer review.

Don't forget to run util/checkwhite. :)

PR #3648, opened 2024-03-11
Rework and bring back sword of the doom knight

SentientSupper:
Replace pain brand with the doom brand which causes it to inflict bonus damage up to 10% of its victim's maximum health on hit. This does the opposite of the sword of power, dealing more damage the tougher the foe you face and avoids making you train necromancy for bonus damage when the weapon has -cast.

Give it ^Drain to discourage swapping it in and out to use spells because of -cast or only when you face a tough enemy like a unique panlord.

Also make it evil because doom sounds evil and it was used by a yred worshipper.

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -549,17 +549,18 @@ BOOL:    evil
 DBRAND:  Damnation: Any ammunition fired from it explodes in a burst of
  damnation on striking a foe, dealing damage which bypasses armour.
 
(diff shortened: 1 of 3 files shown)


pooka109:
Interesting.

(Although the melee effect messaging seems to assume that the defender will never be the player. If that's not intended, maybe genericize the melee effect messaging for that case, as Wyrmbane does?)

SentientSupper:
Should be fixed now.

pooka109:
Thank you. It's in now.

PR #3650, opened 2024-03-11
Add SH preview to item descriptions

NormalPerson7:
Similarly to how item descriptions show the effect equipping that item would have on your AC/EV, make it also show the effect it would have on your SH, if relevant.

For consistency, make AC/EV/SH previews in descriptions not include temporary effects (previously these were included for AC and excluded for EV). If you prefer the alternative, we can do that too; the most important thing is that it's consistent.

Also make AC/EV/SH previews use the same function to compute them, instead of using entirely separate functions with entirely separate logic.

AC/EV/SH preview will now show on all items that have a non-zero effect on them, except non-artefact rings of protection/evasion and amulets of reflection (as for these items, it is not useful). AC preview will always show on body armour and aux armour; EV preview will always show on body armour, aux armour and true shields; and SH preview will always show on true shields, even if there is no effect. Debatably EV shouldn't always appear on aux armour, but I note this was something considered in https://github.com/crawl/crawl/commit/64bcc3171f6afe03e50909bde2252f2c2c0d74ea just last week so I haven't touched this behaviour.

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -1878,18 +1878,37 @@ static string _equipment_switchto_string(const item_def &item)
         return "wearing";
 }
 
-static string _equipment_ac_ev_change_description(const item_def &item, bool remove = false)
+/**
+ * Describe how (un)equipping a piece of equipment might change the player's
+ * AC/EV/SH stats. We don't include temporary buffs in this calculation.
+ *
+ * @param item    The item whose description we are writing.
+ * @param remove  Whether the item is already equipped, and thus whether to
+                  show the change solely from unequipping the item.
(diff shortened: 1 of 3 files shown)


NormalPerson7:
I note commit https://github.com/crawl/crawl/commit/5fa0bccb637462735179f4649d3c13a62652af78 and will update this PR accordingly later.

NormalPerson7:
This PR should now be up to date. I haven't been able to check it is working properly, because crawl/master is failing to compile for me at the moment (Windows 10, msys2) because I'm getting a `windres: preprocessing failed` error, but hopefully the automatic checks will be able to verify that it's ok.

PR #3652, opened 2024-03-12
Rework and bring back Crown of Eternal Torment

SentientSupper:
Replace its mhp malus with *Torm which gives a 5% chance to torment you
and surrounding living creatures when you take non-torment damage. This
serves to give it a more interesting effect and drawback that also
works before extended.

Replace torment immunity with half torment damage so *Torm is an actual
drawback unless you're already undead.

Remove SInv because it doesn't quite fit thematically and we already
have a good number of hat unrands with that.

This commit also adds KILLED_BY_TORMENT as a kill type which is used to
prevent the crown from recursively triggering. It should never actually
kill anything.

### DIFF
--- a/crawl-ref/source/dat/descript/unrand.txt
+++ b/crawl-ref/source/dat/descript/unrand.txt
@@ -634,7 +634,6 @@ contamination almost as effectively. Anyone reckless enough to place it on
 their head would find their magic regeneration dramatically increased, at the
 cost of multiplying any magical contamination received.
 %%%%
-
 crown of Eternal Torment
 
 A gold crown set with jet-black gems. It frequently afflicts its wearer and
--- a/crawl-ref/source/dat/descript/unrand.txt
+++ b/crawl-ref/source/dat/descript/unrand.txt
@@ -634,12 +634,12 @@ contamination almost as effectively. Anyone reckless enough to place it on
 their head would find their magic regeneration dramatically increased, at the
 cost of multiplying any magical contamination received.
 %%%%
-# TAG_MAJOR_VERSION == 34
+
 crown of Eternal Torment
 
-A gold crown set with jet-black gems. It permanently afflicts its wearer with
-agonising pain, substantially reducing their resilience but rendering them
-immune to further torments.
+A gold crown set with jet-black gems. It frequently afflicts its wearer and
+surrounding living creatures with agonising pain. Over time, the wearer becomes
+numb to the pain, greatly reducing the effects of torment on them.
 %%%%
 robe of Vines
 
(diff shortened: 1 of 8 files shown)


PleasingFungus:
What's your vision of this artefact? Pretend that crown of eternal torment never existed before, and you're adding it fresh - what's the design goal that you're working toward? What experience do you want players to have? What interesting choices do you want them to make?

SentientSupper:
The new vision is a double-edged artifact that has a chance of damaging both you and your enemies when you take damage before extended, and remains useful in extended when enemies become immune to torment by reducing the damage you take from torment while still being risky to use.

pooka109:
I'd note that, since it torments monsters around you without regard to whether they're allies or not, you'd either have to have no allies or only allies with torment resistance. So for your god, you'd be pigeonholed into Okawaru, Ru (leaning toward the "hated by all" sacrifice), or Fedhas; and for any allies you summoned, you'd be limited to undead, demons, or plants.

crawl-ref/source/dat/descript/unrand.txt
```diff
@@ -634,12 +634,12 @@ contamination almost as effectively. Anyone reckless enough to place it on
 their head would find their magic regeneration dramatically increased, at the
 cost of multiplying any magical contamination received.
 %%%%
-# TAG_MAJOR_VERSION == 34
+
```
pooka109, line 637: Just removing the TAG_MAJOR_VERSION comment is enough here; the comment shouldn't be replaced with this blank line.

crawl-ref/source/player.cc
```diff
@@ -6839,11 +6839,7 @@ bool player::res_torment() const
     // doesn't currently make you a plant, and I suspect changing that
     // would cause other bugs. (For example, being able to wield holy
     // weapons as a demonspawn & keep them while untransformed?)
-           || you.form == transformation::tree
-#if TAG_MAJOR_VERSION == 34
-           || player_equip_unrand(UNRAND_ETERNAL_TORMENT)
-#endif
-           ;
+           || you.form == transformation::tree ;
```
pooka109, line 6842: The space between transformation::tree and the semicolon is unnecessary.

pooka109:
To paraphrase DracoOmega's commentary on Discord:

* the random chance of torment is not really an interesting effect if you're not undead and immune to it
* the random chance of torment occurs rarely enough that it won't do much
* in extended, pretty much every monster will be immune to torment, so you'll be the only one taking damage
* the Kiku invocation that lets you choose to torment everything is rarely used, so this would be used even more rarely

If it could be reflavored to use a gimmick other than torment, it would have a much better chance of being deemed acceptable.

SentientSupper:
I have an idea for a "crown of empathy" that would increase damage taken and also give a passive aoe pain mirror effect but I'll be closing this because I don't see devs accepting it or me having the time to make the changes necessary in the near future.

PR #3653, opened 2024-03-12
Make Jiyva protect jellies from harm

SentientSupper:
Jiyva's mpregen and mp/wild magic/int mutations support casters well, but the random jelly spawns are especially annoying for casters. You can't firestorm when a jelly decides to hug your target. You can't shatter if there are jellies on the floor because there's a chance you damage a jelly behind a rock wall and incur penance. You can't even immolate at any time because a jelly could spawn and throw itself into the flames. This is no fun and I'm changing that.

Jiyva now protects slimes from harm similarly to Fedhas with the difference that you can't fire through slimes and they'll absorb projectiles if you try to do so.

Jiyva also prevents immolation scrolls from affecting jellies because using jellies as bombs is not cool (at least from Jiyva's perspective).

Additionally, all god protection will protect from torment. Previously this wouldn't have had an effect because plants and ancestors already resist torment.

Based on 75f1ca80bcda43d3fbc33714f3299d0209dd9cc1 with changes.

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -1234,11 +1234,14 @@ void bolt::do_fire()
                 && flavour != BEAM_DIGGING && flavour <= BEAM_LAST_REAL
                 && !cell_is_solid(target)
             // or visible firewood with a non-penetrating beam...
-                || !pierce
-                   && monster_at(pos())
+                || monster_at(pos())
                    && you.can_see(*monster_at(pos()))
+                   && (!pierce
                    && !ignores_monster(monster_at(pos()))
-                   && mons_is_firewood(*monster_at(pos())))
+                   && mons_is_firewood(*monster_at(pos()))
+            // or visible jelly with Jiyva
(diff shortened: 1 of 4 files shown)


crawl-ref/source/beam.cc
```diff
@@ -5202,6 +5204,23 @@ void bolt::affect_monster(monster* mon)
 
     hit_count[mon->mid]++;
 
+    // Jiyva absorbs attacks on slimes.
+    if (!is_tracer && agent()->is_player()
```
pooka109, line 5209: In the arena, agent() can return NULL, so checking agent()->is_player() here will segfault if you don't check agent() first (which is why some of the tests are failing).

SentientSupper: Thanks. Should be fixed now.

PleasingFungus:
Awesome, thanks! Merged with a few changes.

Possible that more work should be done in Jiyva's god description to communicate this. I added one word; unsure if more is needed.

PR #3655, opened 2024-03-13
Don't display monster to hit with 100x the chance

SentientSupper:
When examining a monster when your sh > 0. Hopefully the math isn't wrong. Fixes an issue introduced in 5fa0bccb63.


PleasingFungus:
Fixed by 6677626c5906d5863a90756d0b875000ce6fb659 (not sure if @elliptic saw this PR?)

PR #3658, opened 2024-03-14
Improve Gell's Gravitas + show collision damage for spells

SentientSupper:
It's weird that we have a telekinesis spell that doesn't allow the common demonstration of telekinetic might by slamming dudes into walls, so allow casting Gell's Gravitas on walls.

Give Gell's IMB's ministun. This is intended to make it feel more effective by making monsters not always immediately move away from where you're chucking them at.

Rework Gell's attraction distance. As far as I can tell with the old formula, the attraction distance for Gell's would be always be 1 at radius 1 and 3, 1 at radius 2 below 60 spellpower and 2 otherwise which is pretty unintuitive and has a weird breakpoint. I don't care if gravity realistically weakens with distance, it isn't well communicated to players that the radius 3 pull which only happens when your power is above 80% would be weaker and it doesn't feel good in practice. Change the distance to 1.25 scaling to 2.5 at max spellpower with random rounding and remove distance from the center of attraction from the formula. This does not affect collision damage.

Increase Gell's radius to always be 3 to remove another weird breakpoint. No random radius because we already have collision damage and pull distance scale with power and don't need a third factor to complicate things.

Sort gravitas targets by distance from attraction center rather than distance from you which I assume was a mistake. This makes attracting targets towards a point more reliable.

Using Gell's now counts as an attack for penance. This fixes being able to attack under sanctuary with Gell's.

Show collision damage for Gell's and IMB in description. Describe spellpower effects for these spells.

### DIFF
--- a/crawl-ref/source/actor.cc
+++ b/crawl-ref/source/actor.cc
@@ -22,6 +22,7 @@
 #include "mon-behv.h"
 #include "mon-death.h"
 #include "religion.h"
+#include "spl-damage.h"
 #include "stepdown.h"
 #include "stringutil.h"
 #include "terrain.h"
@@ -1003,7 +1004,7 @@ void actor::collide(coord_def newpos, const actor *agent, int pow)
     if (is_monster() && !god_prot)
         behaviour_event(as_monster(), ME_WHACK, agent);
 
-    dice_def damage(2, 1 + div_rand_round(pow, 10));
+    const dice_def damage = collision_damage(pow, true);
     const int dam = apply_ac(damage.roll());
 
     if (other && other->alive())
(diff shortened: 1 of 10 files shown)


PleasingFungus:
I believe @DracoOmega also has some ideas for changing Gell's Gravitas. Could be good to discuss them and see where we land.

> Sort gravitas targets by distance from attraction center rather than distance from you which I assume was a mistake. This makes attracting targets towards a point more reliable.
> 
> Using Gell's now counts as an attack for penance. This fixes being able to attack under sanctuary with Gell's.
> 
> Show collision damage for Gell's and IMB in description. Describe spellpower effects for these spells.

If it's simple, pulling out these changes into their own PR would make them more likely to land, since they sound like straightforward bugfixes and improvements. If it'd be a lot of work, don't worry about it.

(Ha at the sanctuary thing - good secret tech!)

regret-index:
The collision part was merged in- see 4c2c1c8. The rework to Gravitas has been made obsolete with e52afbe. As such, I am closing this PR.

PR #3659, opened 2024-03-14
Give wjc lunges a bonus with polearms, edit text

SentientSupper:
Polearms gain no special bonus with WJC martial attacks but lose their ability to do reach attacks when lunging. The optimal action when using polearms with WJC is to reach attack when there is a space between you and your enemy for an extra turn and that means you're not making use of lunge, which makes WJC polearm users a little sadder (emotionally) than others.

Fix this by giving lunge a 140% damage multiplier instead of 120% when used with reaching attacks.

Edited some descriptions to give more information about what the different martial attacks do.

### DIFF
--- a/crawl-ref/source/dat/descript/gods.txt
+++ b/crawl-ref/source/dat/descript/gods.txt
@@ -687,12 +687,10 @@ Wu Jian extra
   ..<lightred>@</lightred>.O..  -->  ...<lightred>@</lightred>O..  140% damage if you're using a polearm, and 120%
   .......       .......  otherwise.
 
-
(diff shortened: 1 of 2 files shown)


pooka109:
Interesting.

(Although the ".......  Multiple adjacent foes" portion of the edited text is setting off checkwhite, because all checkwhite sees is a "." followed by two spaces and then a capital letter, which it interprets as French spacing that needs to be "fixed". I'm not sure how to work around this.)

PleasingFungus:
It seems better for some weapon types to be worse in some contexts than others than to try to add extra complexity to keep everything balanced. As long as polearms are good in some contexts, they don't need to be made good with every god, etc.

PR #3662, opened 2024-03-15
Show wand spells in monster description

SentientSupper:
Allow inspecting the damage and success rates of wands possessed by a monster.

Adds placeholder tiles for bolt of light and some new descriptions.

image


crawl-ref/source/dat/descript/spells.txt
```diff
@@ -179,7 +179,8 @@ Fires a penetrating bolt of flames.
 %%%%
 Bolt of Light spell
 
-Fires a ray of searing light, injuring and blinding those struck.
+Fires a ray of searing light, injuring and blinding those struck. Confuses
+instead of blinding if the player is the victim.
```
NormalPerson7, line 183: I don't think we ever want to use 'the player' in a description as it breaks immersion. Perhaps a separate paragraph saying something like "If the bolt hits you, you would be confused' would better?

Otherwise I'm excited for this PR as currently monster wand damage is very obscure.

SentientSupper: Sounds reasonable as "the player" isn't mentioned in any other spell descriptions.

pooka109:
Nice.

Although using the teleport icon for the wand of warping spell seems odd. I've put together a placeholder tile for it (it's the unused Lugonu Bend Space tile, turned sideways and edited to use the same colors as the wand of warping's icon, so it looks something like a bigger version of that icon).

warp_space

SentientSupper:
It is done.

crawl-ref/source/rltiles/dc-spells.txt
```diff
@@ -262,6 +263,7 @@ summon_ufetubus SUMMON_UFETUBUS
 summon_tzitzimitl SUMMON_TZITZIMITL
 summon_undead SUMMON_UNDEAD
 summon_vermin SUMMON_VERMIN
+teleport TELEPORT
```
pooka109, line 267: This line should be removed now that the wand of warping has its own spell tile, and doesn't use the teleport tile anymore.

pooka109:
Applied. Thank you.

PR #3663, opened 2024-03-15
Cheapen of rings of slaying

### LINKED ISSUE #3279: Improper price set?

chrome, webtiles, kelbi

0.30

**DESCRIPTION:**

A mundane ring of slaying costs 1092g in a ring shop. However, in the same shop, a artefact ring {rPois rCorr Slay+4} costs a cheaper 672g.

**EXTRA FILES:**

seed: 4017248900655775592
on D:7
Screenshot 2023-09-02 205901

SentientSupper:
Rings of slaying are more than twice as expensive as other variable strength rings despite being arguably not that much better. Tone down its price to match other rings so it isn't often more expensive than good randart rings that generate in stores.

Slay rings are great in the early game and with short blades in the late game, but can end up being outclassed by stat rings which can provide more damage + ev. Using them also means forgoing potentially important defenses.

Closes #3279

### DIFF
--- a/crawl-ref/source/makeitem.cc
+++ b/crawl-ref/source/makeitem.cc
@@ -2070,7 +2070,7 @@ jewellery_type get_random_amulet_type()
     return *random_iterator(valid_types);
 }
 
-static jewellery_type _get_raw_random_ring_type()
+jewellery_type get_random_ring_type()
 {
     static vector<jewellery_type> valid_types;
     if (valid_types.empty())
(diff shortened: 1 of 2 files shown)


elliptic:
Ring of slaying is rare for a reason - it's universally great on a wide range of characters, especially early on where it is far better than +6 str or +6 dex (which only become better at some point in midgame if ever and are more character-dependent). Making them 3 times as common would be a large early game player buff, and I see nothing wrong with the status quo.

SentientSupper:
Suppose that's true. I think the price could still go down if they are to remain rare.

PleasingFungus:
Seems reasonable. Let's try it. :)

PR #3664, opened 2024-03-15
Give lesser beckoning a ministun

SentientSupper:
This makes using lesser beckoning on melee monsters less bad compared to just waiting for them to walk next to as you no longer spend 2 MP just to give them a free turn.

This also makes it consistent with IMB and buffed Gell's ministunning monsters.

It's fine for spells to be situational and this doesn't change that.

### DIFF
--- a/crawl-ref/source/spl-transloc.cc
+++ b/crawl-ref/source/spl-transloc.cc
@@ -1830,7 +1830,11 @@ bool beckon(actor &beckoned, const bolt &path)
 
     beckoned.apply_location_effects(old_pos); // traps, etc.
     if (beckoned.is_monster())
+    {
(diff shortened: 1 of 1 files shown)


elliptic:
We've moved away from use of sub-turn ministuns as much as possible - see e.g. Freeze not having it any longer. The main reason why knockback effects have it is to increase the chance that knockback will visibly push the monster back (without it immediately moving to its previous location), which isn't an issue here.

I'm also quite concerned that this buff would make lesser beckoning too strong against spellcasters.

PleasingFungus:
To elaborate on what elliptic said, ministuns are a bad combination of power (making monsters not do anything) with unexcitement. They don't *feel* good to players, since there's no clear feedback that anything is happening.

Lesser Beckoning is already good and useful, if situational, and doesn't need further buffs.

PR #3665, opened 2024-03-15
Buff Norris cast rates

SentientSupper:
Even after his previous buffs, Norris's killratio remains below 1%. In an actual game, he shows up at a time when you're already steamrolling most enemies, and then gets steamrolled.

Raise his draining gaze frequency dramatically so it can actually outdrain your mp regen and also increase primal wave's cast frequency slightly.


PleasingFungus:
Sure, let's try it. :)

PR #3666, opened 2024-03-15
Rework gadgeteer gloves and revmp gizmo, improve gadgeteer gizmo

SentientSupper:
Gadgeteer is fairly unimpactful given that most games you have enough wands to never run out and the faster charge gets wasted because misc evocables are not used the moment they come off cooldown, so give it the bonus wand power the gloves have too.

Fix a bug with wand power surge message being incorrect at low mp and let gadgeteer gizmo print the message too.

Rework gadgeteer gloves to give all misc evocables 1 extra max charge. This is a more impactful and exciting effect that lets you use them multiple times in a row and also reduces wasted charge by letting you charge when you already have 1 charge which lets you use them more frequently. Set base armor to 0 to compensate.

RevMP is an awkward gizmo effect because in order for it to be impactful, you need to use up your MP reserves and melee for a few turns before you get fully revved which then takes a while to replenish enough MP for you to do casty things.

Add a new effect to RevMP that lets you rev up by casting spells. This allows you to benefit from it earlier in a fight and also fully rev up from hitting enemies with spells at range so you can start dealing full melee damage when they get close.

This does not work with spells that already cause melee attacks and is reduced for low level spells to discourage spamming magic dart out of los just to rev up.


elliptic:
Gadgeteer gloves is an unrand, so quite rare - I don't think there's any concern about the name overlap causing confusion, though it might be nice if they did stack.

Your Overclock sounds far stronger than the intended target for (part of) a Coglin gizmo. Maybe it would be good on the unrand, but I think the coglin gizo is fine as it is. Even as someone who doesn't like misc evokers much in general or train evocations super high, I felt I was getting decent value from it in my games.

Although I agree that RevMP is quite underwhelming as is, I am concerned about the complexity of making it modify how the already complicated Rev mechanic works. This is just a tiny piece of a species (one piece of one gizmo), we don't want it to use up lots of complexity points. I would far rather give RevMP an unrelated buff or rework it altogether.

SentientSupper:
I'm fine with swapping the effects of overclock and gadgeteer gloves. I would have to handle equipping and unequipping the gloves better though.

Hellmonk:
Would be interested in taking the evocable charge increase part of this pr on its own, but it does not seem to function properly for me. Maybe in the future?

PR #3668, opened 2024-03-15
Rework antimagic brand

SentientSupper:
Simplify antimagic's chance to prevent a spellcast. It is now a flat 2/3 to fizzle a spell rather than 4/(4 + duration) to fail. Having its effect scale with duration means it is hard to evaluate its usefulness as a monster with antimagic could either have only 1 turn left and thus have an 80% chance to succeed casting or enough to never succeed. Since duration is tied to damage dealt, it also means severely antimagicked monsters are those who are 1-2 hits away from dying and will die before antimagic does much and those with high health have so little antimagic that they will succeed at casting anyway despite the little icon.

This number is equivalent to always having 8 turns of antimagic and could be adjusted if the brand ends up being too powerful, though this could affect other methods of applying antimagic (ru, enfeeble, etc).

Antimagic duration will be capped when applied via direct attacks so that it quickly wears off if not constantly reapplied in order to discourage swapping to another weapon after applying antimagic. It will also have a minimum of 10 auts when applied after the cap. This prevents situations where the "magic leaks into the air" message is printed but no antimagic is actually applied and slightly improves its performance with low damage weapons.

Antimagic halves your mp rather than reducing it by 2/3. This is still debilitating enough to casters that they would never want to use it, but allows non-casters a bit more mp for using god abilities.

Antimagic-branded weapons also gain a huge bonus to accuracy. This is flavored as antimagic being attracted to latent magic present in all beings the same way antimatter is attracted to matter.

Since antimagic only works against a small proportion of monsters you face, and even then likely not a monster with a spell you would care to disable or one that would survive being hit for more than a couple of turns, it is only valuable as a swap brand. Players with an antimagic weapon would either need to swap it frequently with a main weapon to get value out of it or get so annoyed that they just stick to their main weapon or rebrand the antimagic weapon. It is also arguable if disabling a single spellcaster is better than simply killing them faster when facing multiple enemies is common and antimagic only disables the enemy you are currently damaging.

Frequent weapon swapping is definitely something we need to move away from encouraging, so I'm giving antimagic an effect that enhances its overall effectiveness. This also makes applying antimagic more consistent and reduces situations where antimagic wears off despite you making constant attacks against an enemy.

The numbers could use some tweaking but this represents a 5-20% damage increase in most situations depending on your target's ev.

### DIFF
--- a/crawl-ref/source/attack.cc
+++ b/crawl-ref/source/attack.cc
@@ -209,6 +209,9 @@ int attack::calc_pre_roll_to_hit(bool random)
             }
             else if (weapon->base_type == OBJ_STAVES)
                 mhit += property(*weapon, PWPN_HIT);
+
+            if (get_weapon_brand(*weapon) == SPWPN_ANTIMAGIC)
+                mhit += 27;
         }
 
         // slaying bonus
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -6299,7 +6299,7 @@ mon_resist_type bolt::apply_enchantment_to_monster(monster* mon)
             break;
 
         const int dur =
-            random2(div_rand_round(ench_power, mon->get_hit_dice()) + 1)
+            (1 + random2(div_rand_round(ench_power, mon->get_hit_dice()) + 1))
                     * BASELINE_DELAY;
         mon->add_ench(mon_enchant(ENCH_ANTIMAGIC, 0,
                                   agent(), // doesn't matter
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -6298,9 +6298,16 @@ mon_resist_type bolt::apply_enchantment_to_monster(monster* mon)
         if (!mon->antimagic_susceptible())
             break;
 
-        const int dur =
-            (1 + random2(div_rand_round(ench_power, mon->get_hit_dice()) + 1))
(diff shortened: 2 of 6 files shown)


elliptic:
I think this pull request fails to take into consideration how dangerous design-wise an effect like antimagic is. Causing monsters to just do nothing instead of casting their scary spells is potentially both very powerful and something that can make fights much less interesting. Antimagic's current design tries to keep this in check - it is a rare (outside Trog) brand that is quite painful for most non-Trog characters to use (due to the 1/3 MP) and does normal damage, and requires constantly hitting and doing good damage to a monster to keep the duration stacked (which makes swapping it mid-fight bad, which is a major concern I would have with the changes in this pull request).

Antimagic is intended to be a fairly unreliable brand that rewards keeping hitting the same monster while doing a bunch of damage (a classic Trog thing).

Although it is true that antimagic encourages swapping to some extent, I don't think it is significantly worse about this than other brands checking resists etc, and I'm not convinced an overall buff helps this at all - it just changes which other weapons you might swap to over it when not interested in the antimagic.

SentientSupper:
Antimagic potentially being too good at simplifying fights could be addressed by tweaking antimagic's numbers. The current numbers are meant to be tweaked after more testing.

Capping the duration of antimagic applied via the brand should help with discouraging mid-fight swaps.

PleasingFungus:
Closing per elliptic's comments. Thank you for the submission.

PR #3669, opened 2024-03-15
Give deep elves bonus mp regeneration

SentientSupper:
Give DE the mp regeneration mutation from DS that doubles their base mp regen. This is intended to be a tiny boost to DE in the early game that also makes their mutation screen look less bare without adding complexity or shaking up species balance.

Add a new debug man species for use in the mutation test.

### DIFF
--- a/crawl-ref/source/dat/descript/species.txt
+++ b/crawl-ref/source/dat/descript/species.txt
@@ -17,6 +17,7 @@ wrapping slow-revving metal around their puny goblinoid frames.
 Deep Elf
 
 Deep Elves have a strong affinity for all kinds of magic, but are very fragile.
+They regenerate magic rapidly.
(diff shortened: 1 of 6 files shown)


elliptic:
Giving DE MUT_MANA_REGENERATION sounds good to me. I'm wary of adding a new fake species just to fix a test though (and it should definitely not be named "Debug Man" if so).

SentientSupper:
I don't feel like plumbing the depths of crawlcode to figure out a good bugless way of suppressing species innate mutations just for the test, so I'm going to do the simple thing of adding a dummy species for that purpose.

PleasingFungus:
Thank you for the submission. However, deep elves are currently an elegant and widely loved species. We don't need to complicate them, and we shouldn't.

PR #3670, opened 2024-03-15
Allow self targeting with nullifying breath

SentientSupper:
Nullifying breath feels weaker than other breaths because of its low damage and situational use compared to just being able to kill whatever's troubling you.

The reasons why self targeting with purple draconian breath were that it provided you with too many sources of self cancellation, to make it consistent with wands of quicksilver and that it was unintuitive.

Now that the breath has been reworked into an aoe, hitting yourself with it is no longer as unintuitive as before since you no longer have to explicitly target yourself with the bolt. Its limited charges also balances it better than before. It's also completely different from quicksilver wands, removing confusion between the two.

So allow dispelling yourself again to give it more usages.

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -3097,11 +3097,8 @@ bool bolt::harmless_to_player() const
     if (you.cloud_immune() && is_big_cloud())
         return true;
 
-    if (origin_spell == SPELL_COMBUSTION_BREATH
-        || origin_spell == SPELL_NULLIFYING_BREATH)
-    {
+    if (origin_spell == SPELL_COMBUSTION_BREATH)
         return true;
-    }
 
     switch (flavour)
     {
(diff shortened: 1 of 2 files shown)


PleasingFungus:
Quick comment: I approve of effects that have multiple uses, and I agree this is less bad than when it was essentially unlimited cancel uses. However, I'm still not sure if we want this to be usable as a cancel - cancellation is a very strong effect, and one that can take a lot of tension out of otherwise exciting moments.

It might be fine. @DracoOmega or @elliptic might have more opinions about this.

Code looks good, at a glance.

SentientSupper:
It also comes at the cost of self damage + antimagic now.

DracoOmega:
Regardless of the utility or design value of being able to use a purple breath charge as a worse !cancellation in a pinch, this seems like a significant nerf to the breath as-is. It is a radius 2 fireball-style AoE and it is deliberate that the player is unaffected by it or it would be impractical to use it in many, many combat situations without injuring (and antimagicing!) yourself. Changing this just so that sometimes you can get marginal additional value out of it seems bad for almost every other situation.

PleasingFungus:
Closing per DO's comment.

PR #3671, opened 2024-03-15
Remove failure chances from racial abilities

SentientSupper:
Ability failure chance for innate racial abilities fails to improve gameplay in any meaningful way compared to failure chances for spells, evocations and invocations.

For other abilities, you can affect your chances with a mixture of stats, skill, equipment and other factors. Only your XL affects the failure chance of racial abilities so it improves on its own without needing you to make any different decisions and usually the chances are so low that they rarely factor into your decision making and feel bad on the off chance they fail.

Many of these racial abilities already have a chance to "fail" by missing or low rolling without making you repeat the same action you just did again on the next turn. Having failure simply waste a turn without consuming other resources also isn't interesting and merely adds friction to gameplay.

We already have a bunch of racial abilities like barachi hop and formicid shaft that don't make use of failure chances to generate interesting decisions. There is no reason why it can't be the same for other species.

This change affects naga poison spit, draconian breath, vampire bat form, demonspawn word of chaos and damnation.


PleasingFungus:
> Having failure simply waste a turn without consuming other resources also isn't interesting and merely adds friction to gameplay.

Losing a turn can be very relevant for powerful, combat-shifting abilities, like most of those here. If you're in a situation where you need to pull out your big powerful breath attack, or need to quickly turn into a bat and fly away, losing a turn can be the difference between life and death.

> We already have a bunch of racial abilities like barachi hop and formicid shaft that don't make use of failure chances to generate interesting decisions. There is no reason why it can't be the same for other species.

Potentially, but it's a nontrivial balance shift, and it's not obviously better. There's a tension that's lost.

I'm especially leery of removing this from bat form. I feel least strongly about spit poison.

SentientSupper:
I also don't feel that strongly about spit poison but I'm including that for the sake of consistency. Same goes for bat form but I have a pr that removes that from vampires.

Hellmonk:
Going to close this as won't do for now, there doesn't seem to be consensus that this is a necessary change.

PR #3673, opened 2024-03-15
Allow damnation to work with dimensional bullseye

### LINKED ISSUE #3532: Damnation Bolts are not copied by Dimensional Bullseye
**PLATFORM and OS or BROWSER:** Local tiles

**DCSS VERSION AND GAME INFORMATION:** 0.32-a0-136-gc02482bde3

**DESCRIPTION:** Casting Dimensional Bullseye does not allow Damnation bolts to hit another target. I guess this makes sense since the damnation bolts explode before they "hit", but it's a strange special case. I tested with a random arbalest picked off the ground to show that it works for other ranged weapons.

**EXTRA FILES:**
image
image

SentientSupper:
The only ranged weapon dimensional bullseye doesn't work with is damnation which tends to surprise players. Given that damnation is a rare unrand, it wouldn't hurt to remove this special case.

Closes #3532


PleasingFungus:
Quick comment: goal seems good but the code is tricky, so I (or someone else) will need to read for a while to be confident about this.

Hellmonk:
crashes if the explosion from the first damnation bolt kills the monster targeted by dimensional bullseye. I will try to fix this later if nobody else gets around to it.

PR #3675, opened 2024-03-15
Allow spellvamp from majin bo to work with vile clutch

### LINKED ISSUE #3453: spellvamp of Majin-Bo doesn't work for BVC
webtiles

0.31-a0-1589-g8f5e8aebd6 (latest trunk)

**DESCRIPTION:**
spellvamp of Majin-Bo doesn't work for BVC. BVC is a direct damage spell, so it should work.

SentientSupper:
Closes #3453

### DIFF
--- a/crawl-ref/source/actor.cc
+++ b/crawl-ref/source/actor.cc
@@ -22,6 +22,7 @@
 #include "mon-behv.h"
 #include "mon-death.h"
 #include "religion.h"
+#include "spl-goditem.h"
 #include "stepdown.h"
 #include "stringutil.h"
 #include "terrain.h"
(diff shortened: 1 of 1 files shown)


PleasingFungus:
I'm not fully sold on this one. It's not clear to me that BVC is a direct damage spell - it seems indirect, as with cloud spells, sticky flame damage-over-time, poison damage-over-time, etc. Also, the code seems like a self-admitted hack. :)

What do you think?

SentientSupper:
It feels like a direct damage spell even if it technically isn't one. I also remember being disappointed by the time I discovered that it didn't work with BVC. I suppose it's fine to have it remain as is but I lean towards making rare unrands more functional.

Hellmonk:
Closing. BVC isn't intended to be a direct damage spell, imo, and we'd want a more robust implementation if we allowed it.

PR #3677, opened 2024-03-15
Don't evolve while undead

### LINKED ISSUE #3535: Evolution Mutation can produce [a different number than 2] Good Mutations when Swapping between Living and Undead
Offline, tiles version

0.31

**DESCRIPTION:**

**Discovery / Reproduction**

I was playing as a Lich, with rank one of Evolution: [You have hidden genetic potential.]
When this triggered, my Lich mutation immunity went off, draining my stats instead (So much for a good mutation!)

In order to benefit from my mutation (and stop it from draining my stats) I switched to living form on floors of the Ziggurat which don't have enemies that malmutate. Then on floors which can malmutate you, I switched back to Lich form.

After doing this for a while, I found that I had three new good mutations, and still had the hidden genetic potential

**Code Links / Mechanics of Bug**

This is the code that evolves you:

https://github.com/crawl/crawl/blob/598725c5f22ab752ec310488f62d41206f82b7da/crawl-ref/source/timed-effects.cc#L217-L228

Essentially it follows this logic:
(1) Attempt to mutate; if successful go to step (2), else return
(2) Increment mutation success counter; If counter reaches (2), go to step (3), else return
(3) Decrement mutation success counter by 2; Attempt to remove evolution mutation

The problem with this logic as it currently stands, is mostly in step (1):

If you are a Lich, the initial mutation attempt will return successful, even though it doesn't mutate you,
allowing the logic above to proceed to step (2). We can see that in the code here:

https://github.com/crawl/crawl/blob/598725c5f22ab752ec310488f62d41206f82b7da/crawl-ref/source/mutation.cc#L2052

Which returns true, even though the mutation didn't really occur.

Similarly, the removal of the evolution mutation in step (3) evaluates to a No-Op here:

https://github.com/crawl/crawl/blob/598725c5f22ab752ec310488f62d41206f82b7da/crawl-ref/source/mutation.cc#L2530-L2531

This means we can have the following sequence:

(1) You are living; evolve once (muts counter is now 1)
(2) You are undead; evolve once (no mutation occurs, muts counter incremented to 2, then reset to 0, evolution isn't removed)
(3) change back to living; GOTO (1)

**Potential Solutions**

We should think about what the desired behavior is for this. I can think of a few possibilities:

(1) Perhaps we should force remove the evolution mutation. This means undead players will simply miss out on the good mutation if it occurs while they are undead

(2) We could force BOTH the good mutation and the mutation removal, effectively allowing this entire mechanic to bypass undead mutation invulnerability (note that Jivya's withdrawal of slime mutations wrath, but not his malmutate wrath, currently bypasses Zin's 100% mutation immunity, so mutation immunity is already a bit messy)

(3) We could fail the mutations while the player is undead, meaning only once they are living will they be able to get 2 good mutations from their evolution, and before that will simply get their stats continuously drained. This is actually the behavior I thought existed before discovering this bug.

If we do (3), we should think about why the mutation attempt returned true when the player is undead. Do we use that somewhere else? If that needs to return true for another reason, we could add special casing logic to the evolution mutation code; otherwise if undead attempted mutation returns false, it would implement (3) here.

(4) We could implement one of the above, but remove the stat drain effect from your evolution mutation while undead.

**Additional Potential Bug**

I haven't doubled checked everything but it looks like we might have the same bug if you alternate between maxed and unmaxed piety with Zin.

Whatever solution we use for this should work for Zin as well, as well as other potential ways of alternating between possessing and not possessing mutation resistance / immunity.

**EXTRA FILES:**
I have saves, but reproducing the bug is tricky and time consuming, so please consult my code references above instead.

SentientSupper:
Closes #3535

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -2634,7 +2634,7 @@ void apply_exp()
     _handle_god_wrath(exp_gained);
 
     // evolution mutation timer
-    if (you.attribute[ATTR_EVOL_XP] > 0)
+    if (you.attribute[ATTR_EVOL_XP] > 0 && you.can_safely_mutate(true))
         you.attribute[ATTR_EVOL_XP] -= exp_gained;
 
     // modified experience due to sprint inflation
--- a/crawl-ref/source/timed-effects.cc
+++ b/crawl-ref/source/timed-effects.cc
@@ -203,6 +203,9 @@ static void _jiyva_effects(int /*time_delta*/)
 
 static void _evolve(int /*time_delta*/)
 {
+    if (!you.can_safely_mutate(true))
+        return;
+
     const bool malignant = you.has_mutation(MUT_DEVOLUTION);
     if (!malignant && !you.has_mutation(MUT_EVOLUTION))
         return;


pooka109:
Applied. Thank you.

PR #3681, opened 2024-03-16
Display PE center targets with more opacity

SentientSupper:
Display permafrost center targets with more opacity. This includes a nonstandard use of AFF_YES but I don't think any players would complain.


PleasingFungus:
Wow, these branch names...

Thanks!

PR #3684, opened 2024-03-16
Fix an issue with randart staves in local tiles

### LINKED ISSUE #3518: Randart staff tile overlaps with tile of its base type in local tiles inventotry
local tiles, windows 10

0.31

**DESCRIPTION:**
In the inventory, the tile of randart staves overlaps with the tile of its base type. This does not happen in the wield menu.

**EXTRA FILES:**
image
image

SentientSupper:
When an item that has a randomized appearance (potions, staves) on the floor is rendered, the item's main tile is actually the "brand". The randomized tile gets added on later by calling add_main_tile which calls tileidx_known_base_item on the main tile when. This is a hack, but it lets you see the brands of identified items on the ground.

However with randart staves, the main tile gets put in the special tile slot instead. Normally this won't cause any issues, but the local tiles inventory renders items as if they were on the ground. It calls add_main_tile on the special slot which then renders the randomized staff appearance on top of the randart staff tile.

Fixed by not doing this for special tile slots.

Closes #3518

### DIFF
--- a/crawl-ref/source/tiledgnbuf.cc
+++ b/crawl-ref/source/tiledgnbuf.cc
@@ -165,20 +165,26 @@ void DungeonCellBuffer::add_dngn_tile(int tileidx, int x, int y,
         m_buf_feat.add(tileidx, x, y);
 }
 
-void DungeonCellBuffer::add_main_tile(int tileidx, int x, int y)
+void DungeonCellBuffer::add_main_tile(int tileidx, int x, int y, bool no_base)
 {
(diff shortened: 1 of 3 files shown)


PleasingFungus:
What was the outcome of your discussion about this with Ge0ff in #dcss?

SentientSupper:
No further comment from Ge0ff but I disapproved of his version of the fix because it is bad practice to have an overloaded function work differently depending on the number of variables passed into it. I suppose the a whole new function could be used instead though.

pooka109:
Applied. Thank you.

PR #3674, opened 2024-03-15
Let flying undead aquatic monsters traverse land

SentientSupper:
This allows bound krakens to fly across land.
Closes #3461

PR #3685, opened 2024-03-16
Fix randart staff names

SentientSupper:
Fixes an issue introduced in 55136d7

image

### DIFF
--- a/crawl-ref/source/artefact.cc
+++ b/crawl-ref/source/artefact.cc
@@ -1432,7 +1432,7 @@ string make_artefact_name(const item_def &item, bool appearance)
     // get base type
     lookup += _get_artefact_type(item, appearance);
 
-    const string base_name = _base_name(item);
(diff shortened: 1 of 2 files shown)


PleasingFungus:
Thanks for taking a look at this issue!

The root problem is that 55136d7841389a13275f47e37acfe5727eb7597f made `item_base_name` incoherent. For most items, the base name was a reasonable string that could appear on its own - for example, `dagger`, or `ring`. After 55136d7841389a13275f47e37acfe5727eb7597f, that was not true for staves.

Unfortunately, this PR didn't fix that underlying issue, rather patching over it for some use cases. This would have caused further pain in future.

Apologies for causing the bug in the first place. For future PRs, it'd be helpful to move a little slower and try to understand the underlying logic and purpose of functions, and to try to leave the code more coherent and easier to comprehend for the next person who comes along. That'll save you the time spent writing PRs like this and #3674.

Issue #1966, opened 2021-05-30
Autofire vamp drain will target an immune monster if theres a not immune monster also adjacent


ardl:
**PLATFORM and OS or BROWSER:**
webtiles trunk version 0.27-a0-1200-g94dc6ad4cf

**DESCRIPTION:**
If there are multiple adjacent monsters, one that is immune to vamp drain and one that is not immune to vamp drain autofiring vamp drain will occasionally target the immune monster. If you press f to manually fire quivered vamp drain the targeter does highlight a valid target.
**EXTRA FILES:**
![image](https://user-images.githubusercontent.com/70080435/120122743-a39ace80-c178-11eb-86dd-9dc644238b4e.png)
![image](https://user-images.githubusercontent.com/70080435/120122752-bc0ae900-c178-11eb-9cf9-4b9581fc407a.png)

rawlins:
Hopefully fixed by d5a6684c9 and associated commits, but I haven't verified.

ebering:
Sadly still replicates in 0.28

DracoOmega:
I am fairly sure this shouldn't happen since the spell auto-targeting rewrite of 0.33

PR #3687, opened 2024-03-16
Improve autoaim behavior and make ranged targeting avoid actors in between

### LINKED ISSUE #3232: Fast Roots aiming tries to avoid allies and yourself

webtiles, kelbi, chrome

tested in 0.30 and 0.31(trunk)
**DESCRIPTION:**

I evoked a wand of roots and tried to kill gnolls in front of me. As I know Fast Roots does not friendly fire, but the auto-aiming mechanics still tried to avoid hitting me by aiming the space left-behind the gnoll.
**EXTRA FILES:**

SentientSupper:
Don't avoid aiming at self with harmless spells. This is relevant for wand of roots.

When autofiring, don't attempt to hit immune targets. This is relevant for any spell that has immune monsters.

Also attempt to choose a target who isn't blocked by allies / plants / slimes you can't shoot through.

Closes #3232 #1966

Make most ranged attacks from you and monsters avoid hitting actors on the way to the target. This change is intended to make ranged attacks and spells much less annoying to use when you have firewood / allies / slimes blocking the path by making it choose a slightly different path with the same rules as permissive los.

TODO: Don't leak invisible monster locations. #3017 would solve this in a more elegant fashion.

Before:
image
After:
image

Before:
image
After:
image

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -818,7 +818,9 @@ void bolt::apply_beam_conducts()
 void bolt::choose_ray()
 {
     if ((!chose_ray || reflections > 0)
-        && (no_actor_perm_lof || !find_ray(source, target, ray, opc_no_actor))
+        && (no_actor_perm_lof || !(agent() && agent()->is_player()
+            ? find_ray(source, target, ray, opc_shoot_through)
+            : find_ray(source, target, ray, opc_no_actor)))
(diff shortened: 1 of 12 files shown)


PleasingFungus:
Quick comment: the goal seems good but the code is scary, since directn.cc is a nightmare. I (or someone else) will need to spend a while reading to be confident about this one.

PR #3688, opened 2024-03-16
Fix max HP value in description of your ancestor

bjiorn:
Previously the value was not up-to-date after gaining a level until your ancestor died and respawned.


PleasingFungus:
Looks good, thanks! How would you like to be credited?

bjiorn:
I'm already listed in the credits as 'bjobae', but I would appreciate it if you could change that to "Björn Bähre".

PleasingFungus:
Done!

PR #3689, opened 2024-03-16
Fix pressing f twice to aim quivered Dr breath

SentientSupper:
Previously pressing f with a quivered breath would not let you start aiming your breath until you pressed it again.


crawl-ref/source/spl-zap.cc
```diff
@@ -159,7 +159,6 @@ static pair<ability_type, zap_type> _abil_zaps[] =
     { ABIL_SPIT_POISON, ZAP_SPIT_POISON },
     { ABIL_BREATHE_FIRE, ZAP_BREATHE_FIRE },
     { ABIL_BREATHE_POISON, ZAP_BREATHE_POISON },
-    { ABIL_NOXIOUS_BREATH, ZAP_NOXIOUS_BREATH },
```
PleasingFungus, line 162: ?

SentientSupper: Noxious breath uses a spell like any other drac breath instead of zapping directly so I think this was an oversight. Having this here also causes the double f issue.

DracoOmega:
Merged

PR #3693, opened 2024-03-17
Allow spell vamp with vamp draining and necrotise

SentientSupper:
These didn't work before and feel very "direct damage"-ish.

### DIFF
--- a/crawl-ref/source/monster.cc
+++ b/crawl-ref/source/monster.cc
@@ -73,6 +73,7 @@
 #include "tilepick.h"
 #include "tileview.h"
 #endif
+#include "spl-goditem.h"
 #include "traps.h"
 #include "view.h"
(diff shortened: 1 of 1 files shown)


PleasingFungus:
Quick comment: the goal of this change seems reasonable but I'm leery of putting this change in monster::hurt. Suspect it should go somewhere in the respective spells' logic, or maybe in beam.cc. But I haven't done a full review so I'm not sure how viable alternatives are.

SentientSupper:
I think it's viable to put the code in apply_enchantment_to_monster and _vampiric_draining_effect, but that would make extending this to newer spells require a bit more work. I didn't do this because I felt having stuff in one place would be better and we already have a nice comment "// Gigantic-switches-R-Us" in there.

PR #3694, opened 2024-03-17
Show player as immune to polymorph/charming when immune

SentientSupper:
Show player as immune to charming with clarity and polymorph when undead.


PleasingFungus:
Thanks for this. Would be nice to have a solution that involved less duplication, but that could be tricky.

PR #3695, opened 2024-03-17
Allow aux attacks to receive bonuses from lash

### LINKED ISSUE #3298: Auxiliary attacks can miss when performing a martial attack with Serpent's Lash
(Webtiles trunk, 0.30-something)
According to description, martial attacks when Serpent Lashing can't miss.
I lunge at Mnoleg while in Dragon Shape and my auxiliary offhand punch (claw) attack misses. I checked it on other guy with the bite attack and it also missed.
I don't know if it's a bug or intended behaviour.
mnolegmiss

SentientSupper:
WJC's lash claims to work on melee attacks and doesn't specifically mention not benefiting aux attacks. Also the issue has been open for months and nobody's claimed that it's working as intended so it probably isn't.

Closes #3298


PleasingFungus:
Quick comment: seems vaguely reasonable but I don't feel super strongly about it. Curious about others' opinions.

Code looks maybe fine but I'd need to review more thoroughly to be confident (look for other uses of `wu_jian_has_momentum` and read `player_aux_test_hit` more thoroughly).

pooka109:
Opinion-wise, I figure if Serpent's Lash melee attacks aren't supposed to miss, that should apply to auxiliary melee attacks as well.

PR #3697, opened 2024-03-17
Correct how the display_char option handles gems and gold.

Aliscans:
DCHAR_ITEM_GOLD and DCHAR_ITEM_GEM were in a different order in dchar_names[] (used by the "display_char" option) to the one in dungeon_char_type, meaning that the option had unexpected effects.

Fix this, and add item_gem to the docs for this option.


PleasingFungus:
Thanks!

PR #3701, opened 2024-03-18
Updated version of Monkooky's Charlatan's Orb PR.

pooka109:
The original is at #3302.

The only changes made in this version are to update it for current trunk, make the fake orb glow the same way as the real one, and add an inscription indicating its Evocations boost.


pooka109:
Closing, since it's been merged into the original pull request at #3302.

PR #3704, opened 2024-03-19
Fix coglin weirdness with cursed weapons

### LINKED ISSUE #3562: Attempting to equip weapon as coglin will destroy cursed weapon in slot
windows local tiles

0.32-a0-393-g38ff01cdb3

**DESCRIPTION:**
1. Curse a weapon as a coglin with Ashenzari.
2. Click on another weapon to equip it.
3. The cursed weapon will be swapped off and destroyed without asking for confirmation.

image

SentientSupper:
It would probably be better to split can_unwield from can_wield, but I don't want to touch this more than I need to.

The code about trying to equip a two handed weapon as a coglin with cursed weapons is unnecessary because you are already asked to remove your off-hand weapon first even if neither is cursed and I don't think anyone actually wants to use a two handed weapon as a coglin.

Fixes #3562.

### DIFF
--- a/crawl-ref/source/item-use.cc
+++ b/crawl-ref/source/item-use.cc
@@ -1276,10 +1276,10 @@ static bool _safe_to_remove_or_wear(const item_def &item,
 // Rather messy - we've gathered all the can't-wield logic from wield_weapon()
 // here.
 bool can_wield(const item_def *weapon, bool say_reason,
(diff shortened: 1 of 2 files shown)


DracoOmega:
This was fixed (in a different manner) by 8529b04c429318c8a

PR #3705, opened 2024-03-19
Make permissive lof treat actors as obstacles

SentientSupper:
This change is intended to make ranged attacks and spells much less annoying to use when you have firewood / allies / slimes blocking the path by making it choose a slightly different path with the same rules as permissive los.

TODO: Don't leak invisible monster locations.

Before:
image
After:
image

Before:
image
After:
image


SentientSupper:
Going to close this because I'm merging it with another pr which has a lot of overlap in code.

PR #3708, opened 2024-03-19
Improve drop_disables_autopickup behavior

SentientSupper:
Don't disable autopickup when dropping missiles because missiles with different brands share the same base type.

### DIFF
--- a/crawl-ref/source/items.cc
+++ b/crawl-ref/source/items.cc
@@ -2645,7 +2645,7 @@ bool drop_item(int item_dropped, int quant_drop)
 
     ASSERT(item.defined());
 
-    if (Options.drop_disables_autopickup)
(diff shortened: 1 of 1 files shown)


NormalPerson7:
How feasible would it be to differentiate the autopickup settings for different brands? For example most characters want curare on autopickup at the start of the game, but probably not so for say datura.

SentientSupper:
Currently autopickup settings are handled as a 2d array of item base type and sub type. It should be possible to add brands on top of that but I don't have a desire to that right now.

PleasingFungus:
Thanks for the follow-up! We'll also need to update the docs - I've done so in 0a3110eaddf12d9b83930124f97cb9dfe22b6811.

PR #3709, opened 2024-03-19
Describe mindburst and pain damage

SentientSupper:
In monster spell description.
image

### DIFF
--- a/crawl-ref/source/describe-spells.cc
+++ b/crawl-ref/source/describe-spells.cc
@@ -499,6 +499,15 @@ static string _effect_string(spell_type spell, const monster_info *mon_owner)
         }
         if (you.immune_to_hex(spell))
             return "(immune)";
+
+        if (spell == SPELL_MINDBURST
+            || spell == SPELL_PAIN)
+        {
+            const dice_def dam = _spell_damage(spell, hd);
+            return make_stringf("(%dd%dx%d%%)", dam.num, dam.size,
+                hex_chance(spell, mon_owner));
+        }
+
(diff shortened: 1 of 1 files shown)


crawl-ref/source/describe-spells.cc
```diff
@@ -499,6 +499,15 @@ static string _effect_string(spell_type spell, const monster_info *mon_owner)
         }
         if (you.immune_to_hex(spell))
             return "(immune)";
+
+        if (spell == SPELL_MINDBURST
+            || spell == SPELL_PAIN)
```
PleasingFungus, line 504: When possible, it's better to avoid hardcoding specific constants. Instead, think about the pattern that you're trying to describe ("hex spells with associated damage zaps") and write your code to target that.

crawl-ref/source/describe-spells.cc
```diff
@@ -499,6 +499,15 @@ static string _effect_string(spell_type spell, const monster_info *mon_owner)
...
+        {
+            const dice_def dam = _spell_damage(spell, hd);
+            return make_stringf("(%dd%dx%d%%)", dam.num, dam.size,
```
PleasingFungus, line 507: This seems a bit confusing. "It does `1d15x20%` damage? So, 1d3 damage?"

I changed it to `20%,1d15` (in the example.) Probably still confusing, but maybe less so?

`20%->1d15` might also be reasonable. Open to more suggestions.

PleasingFungus (commented):
Thanks for the improvement! :)

PR #3710, opened 2024-03-19
Fix blink targeter revealing invisible enemies + Bind Soul targeter revealing the true Mara

RypoFalem:
Blink targeter only shows invis enemies when they are known
image
image

Bind soul targeter allows targets that are summons that you don't know for sure are summons. (Actually casting bind soul on these will reveal that the monster is a fake, but that costs a turn, MP, and piety). Maybe it shouldn't cost piety when it fails? 
image


PleasingFungus:
Thanks for looking at these fixes! This code is a bit concerning, so it'll take a little review before one feels comfortable merging it.

DracoOmega:
Merged, thanks!

PR #3711, opened 2024-03-19
Propose unrand guarded vault for Rift.

orjb1:
So the flavour I'm going for here is based on this description on the Wiki: "A small wrinkle in the very fabric of reality, magically bound to a shaft by some mad acolyte of Lugonu."  Is not abyss/leakage themed, did not want a portal (in my headcannon the acolyte got stuck here experimenting).  I stole evilmikes acolyte npc, no idea if that's ok!

Anyway, my idea was that the acolyte has been busy "forging/blessing/animating" demon tridents repeatedly eventually created "Rift".  Elf/Vaults seemed reasonable?  I think the mobs are pretty terrifying for any melee who might reasonably want to use the weapon.

Things I'm unsure about:
 - Is the sheer volume of distortion tridents too much?
 - Is my dream of flying rift unreasonable?
 - Is there enough decoration/visualisation to support the theme?  I was struggling to come up with more supporting elements.  
 
Anyway, I'd love some thoughts/feedback!

### DIFF
--- a/crawl-ref/source/dat/des/variable/mini_monsters.des
+++ b/crawl-ref/source/dat/des/variable/mini_monsters.des
@@ -9728,8 +9728,8 @@ MONS:   dancing weapon ; demon trident ego:distortion
 KMONS:  B = pile of debris
 KMONS:  D = dancing weapon ; rift
 KMONS:  E = generate_awake occultist god:Lugonu name:mad_acolyte_of_Lugonu n_rpl n_des \
-		col:lightgreen tile:mons_mad_acolyte_of_lugonu; \
-		robe good_item . demon trident ego:distortion | demon trident ego:none \
+        col:lightgreen tile:mons_mad_acolyte_of_lugonu; \
+        robe good_item . demon trident ego:distortion | demon trident ego:none \
         spells:smiting.20.priest;banishment.10.priest;invisibility.10.priest;
 MAP
       c===c
--- a/crawl-ref/source/dat/des/variable/mini_monsters.des
+++ b/crawl-ref/source/dat/des/variable/mini_monsters.des
@@ -9727,8 +9727,8 @@ NSUBST: ! = 3:e / *:.
 MONS:   dancing weapon ; demon trident ego:distortion
 KMONS:  B = pile of debris
 KMONS:  D = dancing weapon ; rift
-KMONS:  E = generate_awake wizard name:mad_acolyte_of_Lugonu n_rpl n_des col:lightgreen \
-        tile:mons_mad_acolyte_of_lugonu; \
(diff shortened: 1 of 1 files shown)


pooka109:
Quick note: If you're going to use the mad acolyte of Lugonu, you'll need to copy the fix from b37f33011ee95e779136f67a220b14905bebfcfb to make sure they *actually* worship Lugonu.

pooka109:
Regarding the mad acolyte of Lugonu, you'll also need to copy the fix from d272cc43a007ec2f50bf4a85fb5ace1859173685 to make sure it leaves a "human corpse", and not a "human corpse of mad acolyte of Lugonu".

regret-index:
Sorry I didn't get around to this until near feature freeze and thus don't have much time to wait on the feedback I'll provide here. I was divided for a while on how to deal with there _already_ being a guarded_unrand_rift vault by hellmonk, until I decided recently that it's fine as long as they further split the rare weight and share a `uniq_guarded_rift` tag.

What I adjusted in the commit merging this in, none of which is all that pressing, and all of which you can keep in mind for future submissions to get more quick attention:
- The order of lines in most vaults should always go all `MONS` and `KMONS` lines, then all `ITEM` and `KITEM` lines, then all `KFEAT` lines, then all randomization lines- it makes it much easier to parse absolutely everything a vault exactly contains overall. The only real exceptions are for vaults defining different colours / tiles / rewards alongside monsters in differing lua branch / set configurations, which will probably need to put some randomization before everything else.
- While most punctuation marks are fine for using for vault randomization purposes, I'd prefer @ to be almost never used as such with in its fixed already-extant purpose of connecting to the rest of the level- it's confusing on first glance otherwise. Only use the glyph outside of this purpose when a vault's so full of randomization it actually needs over 60 different glyphs in it.
- Since there's already an artefact dancing demon trident of distortion, I'm nerfing the other dancing demon tridents by giving them the "damaged" tag. It's probably still a bit excessive, but banishment's fun.
- While vaults don't have nearly as much pressure as elsewhere to keep to an 80 character line limit, it's usually for the better if it doesn't add that much more total lines anyway, as the three branding states of demon tridents fits for this vault's other items.
- With 1bdf802 having added abyss-unique stone that still looks reasonably deliberately built, I'm using it to finish up the Abyss-cultist-outside-of-the-Abyss theme trying to avoid feeling too much like the Abyss itself leaking.

Thank you for the submission.

orjb1:
I cannot believe I missed that there was already a vault for rift!  Thanks for the help, and apologies to hellmonk for the duplication.

I don't mean to make cleaning work for devs to merge, but I am hoping for feedback on "is this a reasonable idea" before trying to get approval on implementation.  A PR feels like a reasonable channel for this (and I really am ok with being rejected) - but if there's an better avenue do let me know :)

PR #3712, opened 2024-03-20
more monster-based skill titles

yrdzrfxndfvh:
this adds some skill titles based off of monsters you can find in game (or used to be able to). Cool! i like most of them but some are admittedly questionable, also i wanted to add "Shadow Fiend" for Ds dith worshippers but i couldnt figure out how to make it work

also un-includes misc.h because there are no more time dependent titles

### DIFF
--- a/crawl-ref/source/skills.cc
+++ b/crawl-ref/source/skills.cc
@@ -26,7 +26,6 @@
 #include "item-prop.h"
 #include "libutil.h"
 #include "message.h"
-#include "misc.h" // now_is_morning
 #include "notes.h"
 #include "output.h"
 #include "random.h"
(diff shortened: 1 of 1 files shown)


regret-index:
I like the rough idea of this, but not a fair number of the examples. Most prior non-Draconian examples are for particularly notably dangerous, unique, or even just extended enemies (Demonologist, Hellbinder, Blademaster, Master Archer, Rockslime, Iron Golem Searing Wretch, and so on). Several proposed titles either previously were or currently are generic fodder post-Lair (e.g. Salamanders, Ravens, Cane Toads), which makes for a passable _joke_ but a very undignified title otherwise, and also a deeply strange species shifting implication on the regular. There's also a duplicate (Sharpshooter's already a ranged title in general).

I'd probably keep without concern Iron Troll, Crusher, Defender, War Gargoyle (sigh), and Pearl Dragon. Alongside that:
- A title of just "Molten" would probably be better for 27 Fire Magic Gargoyles, keeping the spirit without the lengthiness and direct naming.
- I'd make Hell Knight exclusive to Makhleb, as Makhleb is specifically the monster hell knight god.
- Some other devs can quibble over Death Drake, Shard Shrike, and Shadow Dragon being too far or not. I'm on the fence for the dragons, as Draconian 

The rest I think are far too silly to actually use.

yrdzrfxndfvh:
thank you for the response i just wanted to throw some ideas out there (also i thought sharpshooter was already a title, damn you badwiki)

pooka109:
I did put in your un-include of misc.h.

PR #3715, opened 2024-03-20
Preview attack delay in equipment description

SentientSupper:
Include attack delay when previewing equipment. This is important for rangers especially those who are wearing cursed or ^drain/contam/fragile armor.

Currently wondering if I should add this for damage rating too but I'd like feedback before proceeding with it.

image

### DIFF
--- a/crawl-ref/source/player-act.cc
+++ b/crawl-ref/source/player-act.cc
@@ -271,15 +271,16 @@ brand_type player::damage_brand(int)
  *
  * @param projectile  The projectile to be thrown, if any.
  * @param rescale         Whether to re-scale the time to account for the fact that
- *                   finesse doesn't stack with haste.
+*                   finesse doesn't stack with haste.
+ * @param ignore_temporary  Whether to ignore temp modifiers like haste.
  * @return           A random_var representing the range of possible values of
  *                   attack delay. It can be casted to an int, in which case
  *                   its value is determined by the appropriate rolls.
  */
-random_var player::attack_delay(const item_def *projectile, bool rescale) const
+random_var player::attack_delay(const item_def *projectile, bool rescale, bool ignore_temporary) const
 {
     const item_def *primary = weapon();
-    const random_var primary_delay = attack_delay_with(projectile, rescale, primary);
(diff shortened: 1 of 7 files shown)


PleasingFungus:
Cool change! Will take a bit of review.

> Currently wondering if I should add this for damage rating too but I'd like feedback before proceeding with it.

Hm, seems complex. Weakly think it'd be better to hold off on that, at least for now.

crawl-ref/source/player-act.cc
```diff
@@ -379,8 +380,8 @@ random_var player::attack_delay_with(const item_def *projectile, bool rescale,
         attk_delay = div_rand_round(attk_delay, 2);
     }
 
-    return rv::max(div_rand_round(attk_delay * player_speed(), BASELINE_DELAY),
-                   random_var(1));
+    return rv::max(ignore_temporary ? attk_delay : div_rand_round(attk_delay * player_speed(),
```
Cgettys, line 383: This seems to be getting a bit wordy, might be clearer as a if /else statement?

crawl-ref/source/player.cc
```diff
@@ -6597,9 +6597,10 @@ int player::evasion_scaled(int scale, bool ignore_temporary, const actor* act) c
  * @param ac        The player's AC if this item were equipped.
  * @param ev        The player's EV if this item were equipped.
  * @param sh        The player's SH if this item were equipped.
+ * @param ad        The player's attack delay if this item were equipped.
  */
-void player::ac_ev_sh_with_specific_item(int scale, const item_def& new_item,
-                                         int *ac, int *ev, int *sh)
+void player::status_with_specific_item(int scale, const item_def& new_item,
+                                         int *ac, int *ev, int *sh, int *ad)
```
Cgettys, line 6603: Feels like this is getting towards meriting a struct for these  4 parameters?

Cgettys:
Overall, mostly just a few nits, nothing I see as horrible.

CrawlOdds:
An updated version of this was eventually merged in 2f29fc7

Issue #2447, opened 2022-02-25
Problem with autotargeting with Tab/p in ranged


ASilver:
**PLATFORM and OS or BROWSER:**

**DCSS VERSION AND GAME INFORMATION:**

Chrome - version 0.29-a0-219-g0a02b1cf4f

**DESCRIPTION:**

There is an issue with the Tab/p autotargeting. I enter a room and press tab and it insists on choosing an enemy that is not directly in line of fire, despite having a wide choice.

**EXTRA FILES:**

![image](https://user-images.githubusercontent.com/3589847/155808328-e4798f11-067a-461e-a68b-1dc1a751c83c.png)

CrawlOdds:
I think this is working as intended - autofight is intentionally very basic in its target selection, to be predictable over being clever.

PR #3716, opened 2024-03-20
Improve autoaim behavior and make ranged targeting avoid actors in between

### LINKED ISSUE #3232: Fast Roots aiming tries to avoid allies and yourself

SentientSupper:
(In case you're wondering why this is a new pr, I didn't know renaming a branch would close the pr. Original: #3687)

Don't avoid aiming at self with harmless spells. This is relevant for wand of roots.

When autofiring, don't attempt to hit immune targets. This is relevant for any spell that has immune monsters.

Also attempt to choose a target who isn't blocked by allies / plants / slimes you can't shoot through.

Should possibly put the code for the above inside the autofight lua somehow?

Closes #3232 #1966 #2447

Make most ranged attacks from you and monsters avoid hitting actors on the way to the target. This change is intended to make ranged attacks and spells much less annoying to use when you have firewood / allies / slimes blocking the path by making it choose a slightly different path with the same rules as permissive los.

TODO: Don't leak invisible monster locations. #3017 would solve this in a more elegant fashion.

Before:
image
After:
image

Before:
image
After:
image

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -2450,9 +2450,9 @@ bool activate_talent(const talent& tal, dist *target)
         args.hitfunc = hitfunc.get();
         args.restricts = testbits(abil.flags, abflag::target) ? DIR_TARGET
                                                               : DIR_NONE;
-        args.mode = TARG_HOSTILE;
+        args.mode = is_targeted ? TARG_HOSTILE : TARG_NONE;
         args.range = range;
-        args.needs_path = !testbits(abil.flags, abflag::target);
+        args.needs_path = testbits(abil.flags, abflag::dir_or_target);
         args.top_prompt = make_stringf("%s: <w>%s</w>",
                                        is_targeted ? "Aiming" : "Activating",
                                        ability_name(abil.ability).c_str());
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -818,7 +818,9 @@ void bolt::apply_beam_conducts()
 void bolt::choose_ray()
 {
     if ((!chose_ray || reflections > 0)
-        && (no_actor_perm_lof || !find_ray(source, target, ray, opc_no_actor))
+        && (no_actor_perm_lof || !(agent() && agent()->is_player()
+            ? find_ray(source, target, ray, opc_shoot_through)
+            : find_ray(source, target, ray, opc_no_actor)))
         && !find_ray(source, target, ray, opc_solid_see)
         // If fire is blocked, at least try a visible path so the
         // error message is better.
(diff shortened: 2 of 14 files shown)


DracoOmega:
While I agree that the targeter automatically providing unobstructed paths that are nonetheless possible for the player to manually select could be good, I feel like some of these alternate paths you are showing up there are ones that are *not* currently possible to realistically attain manually at present, with spell range being as it is. Am I incorrect?

Also, I feel you've glossed over 'A side effect is that ranged monsters become more dangerous because they are blocked a lot less frequently by their friends now.' far too quickly. That is a significant mechanical change and not one that I personally think is a good idea. Regardless of the degree to which it makes specific individual enemies more dangerous, deliberately manipulating enemy lines of fire via player positioning is an interesting element of player agency at present, and making that significantly harder for the player to do feels like a direct gameplay loss. (I imagine it will also be a lot less clear to the player when an enemy is or isn't blocking another's shots, given how flexible your proposed paths are).

To some degree, the player being actively smarter than monsters with how they can aim things is probably also good gamefeel, regardless? Players like to feel they're outsmarting things.

SentientSupper:
The alternate paths I am showing use the same rules as permissive los so it just achieves them by treating actors as if they were walls.

The impact on balance is something I'm currently testing. The current feeling I get is that there's no noticeable change in difficulty. I think the loss in player agency is suitably compensated for by reducing gameplay friction.

I think it actually makes it more clear if a monster's shots would be blocked since before there's a lot of ambiguity with diagonals unless you check with your own tracer. After the change you can assume a monster can hit you by default unless it's blocked by more than one adjacent blockers.

I'm not sure about making the player better at aiming than monsters. That could lead to situations where you take cover behind plants to shoot an enemy who can't hit you (maybe good?) but also rat value.

elliptic:
This is a massive buff to ranged (for both player and monster). A fundamental change like this needs a much stronger argument than "less annoying to use when you have firewood / allies / slimes blocking the path".

SentientSupper:
I reckon the current balance of ranged isn't great enough to warrant rejecting major usability improvements that would affect it.

Implojin:
Personally I wouldn't want to merge this because it represents a massive balance shift to players being able to positionally block monster ranged attacks, and it would need a hell of a lot more justification for that than wanting to make the UI slightly smoother when tabbing.

Because of this, and the age of the PR, and the above comments by DO and elliptic, I'm closing this one.

PR #3717, opened 2024-03-20
Fix Jiyva attack absorption failing to find a name

SentientSupper:
For zap types constructed from hex beams.


PleasingFungus (commented):
Thanks for spotting the issue and submitting a fix! These changes seem potentially pretty wide-reaching, and involve adding new null pointers, so someone will need to spend some time reviewing and being confident that this doesn't add new bugs/crashes before merging it.

PR #3718, opened 2024-03-20
Teach dogs to stay

SentientSupper:
Refresh dog duration every time you cast it to avoid situations where the dog times out immediately after you cast the spell to heal it.

### DIFF
--- a/crawl-ref/source/mon-ench.cc
+++ b/crawl-ref/source/mon-ench.cc
@@ -197,7 +197,7 @@ bool monster::add_ench(const mon_enchant &ench)
     // If the duration is not set, we must calculate it (depending on the
     // enchantment).
     if (!ench.duration)
-        added->set_duration(this, new_enchantment ? nullptr : &ench);
+        added->add_duration(this, new_enchantment ? nullptr : &ench);
 
     if (new_enchantment)
         add_enchantment_effect(ench);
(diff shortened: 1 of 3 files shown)


DracoOmega:
It was originally intentional that recasting to heal did not give additional duration, at least partially to avoid encouraging the player to 'top up' their dog periodically to prevent it from disappearing. I could see maybe see an argument for giving it a very brief bit of additional duration on recast so that the bonus attack the player expects to get from recasting actually happens, but am unsure about any more than that.

However, I'm additionally leery about how you are adjusting duration. Not only have you pulled out a private method that isn't otherwise used externally, but you are resetting the duration to its initial baseline it in a way that has significant randomization afterward, and so it is very possible that recasting will now *decrease* remaining duration sometimes.

SentientSupper:
I suppose there could be a cleaner way of coding this once I take another look at it.

The reason I want this is because of situations where you cast the dog for a fight, finish that fight then head to the next one, heal up your dog when the fight begins, but then have to resummon the dog because it decided to time out right at that time. This encourages you to wait for your dog to disappear at the end of each fight just so the next one you summon has a longer duration.

Possibly decreasing the duration when it's reset is intentional, as we don't want players casting it multiple times just to get a dog that lasts longer.

Hellmonk:
closing as won't do for now; may get handled in another way in future

PR #3722, opened 2024-03-22
fix: Solve the Mara Hall problem

RypoFalem:
(DracoOmega named this the "Mara Hall problem" and I love that so I'm using it here.)

Before this commit, when Mara summoned two illusions of himself, he would have a 1/2 chance of swapping positions with each illusion consecutively. This led to a spoilery outcome where you had better odds of attacking the real Mara if you attacked one of the Maras at the new positions as demonstrated below:

Call Mara's original position 'a', the position of the first illusion 'b', and the position of the second illusion 'c'.

```
1st action 2nd action Result position
swap(1/2)  swap(1/2)  c(1/4)
dont(1/2)  swap(1/2)  c(1/4)
swap(1/2)  dont(1/2)  b(1/4)
dont(1/2)  dont(1/2)  a(1/4)
```

You can see here that the probabilities of the true Mara being at position a and b are both 1/4 while the sum of the probabilities of Mara being at position c are 1/2. I don't know of any way position c and b can be distinguished, but you can usually distinguish position a because that's where Mara starts. Knowing this, you know that you have equal or greater odds of attacking the true Mara if you switch from attacking the Mara at the original position to one of the ones at the new positions. Like with the Monty Hall problem, you should always switch!

This commit keeps the 1/2 swap chance for the first illusion but uses 1/3 chance for the next potential swap. (and 1/4, 1/5, 1/6, etc for subsequent swaps but Mara only makes two clones so that never happens in practice) This evens out the odds as demonstrated by the new table:

```
1st action 2nd action Result position
swap(1/2)  swap(1/3)  c(1/6)
dont(1/2)  swap(1/3)  c(1/6)
swap(1/2)  dont(2/3)  b(1/3)
dont(1/2)  dont(2/3)  a(1/3)
```

Now the two c's sum to 1/3 and all three positions have an equal chance of containing the true Mara!

I don't actually have a proof for the general case of any number of swaps but simulations suggest that it is true. It seems like something relevant enough to computer science that there's probably an existing proof out there. Let me know if you know it!

### DIFF
--- a/crawl-ref/source/mon-cast.cc
+++ b/crawl-ref/source/mon-cast.cc
@@ -6192,8 +6192,10 @@ void mons_cast(monster* mons, bolt pbolt, spell_type spell_cast,
         // Swap chance is 1/2, then 1/3, 1/4, etc. This gives Mara an equal
         // chance of ending up at any of the original or illusion positions.
         for (sumcount = 0; sumcount < sumcount2; sumcount++)
+        {
             cast_phantom_mirror(mons, mons, 50, SPELL_FAKE_MARA_SUMMON,
                                 sumcount + 2);
+        }
 
         if (you.can_see(*mons))
         {
--- a/crawl-ref/source/mon-cast.cc
+++ b/crawl-ref/source/mon-cast.cc
@@ -3464,7 +3464,7 @@ static void _corrupting_pulse(monster *mons)
 
 // Returns the clone just created (null otherwise)
 monster* cast_phantom_mirror(monster* mons, monster* targ, int hp_perc,
-                             int summ_type)
+                             int summ_type, int swap_chance_reciprocal)
 {
     // Create clone.
     monster *mirror = clone_mons(targ, true);
@@ -3503,7 +3503,7 @@ monster* cast_phantom_mirror(monster* mons, monster* targ, int hp_perc,
 
     // Sometimes swap the two monsters, so as to disguise the original and the
     // copy.
-    if (coinflip())
+    if (x_chance_in_y(1, swap_chance_reciprocal))
         targ->swap_with(mirror);
 
     return mirror;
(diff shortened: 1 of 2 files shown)


NormalPerson7:
The proof is simple. Suppose Mara makes n copies (including the real one). The kth copy is real precisely if the kth swap succeeded and all subsequent swaps failed, with probability
$$ \frac{1}{k}\frac{k}{k+1}\cdots\frac{n-2}{n-1}\frac{n-1}{n} = \frac{1}{n} $$

crawl-ref/source/mon-cast.cc
```diff
@@ -6189,8 +6189,13 @@ void mons_cast(monster* mons, bolt pbolt, spell_type spell_cast,
         if (sumcount2 <= 0)
             return;
 
+        // Swap chance is 1/2, then 1/3, 1/4, etc. This gives Mara an equal
+        // chance of ending up at any of the original or illusion positions.
```
PleasingFungus, line 6193: The comment about the odds of swapping belong with the logic, inside `cast_phantom_mirror`. If you have a comment distant from the corresponding logic, then people who change the logic are unlikely to also update the comment.

crawl-ref/source/mon-cast.cc
```diff
@@ -3503,7 +3503,7 @@ monster* cast_phantom_mirror(monster* mons, monster* targ, int hp_perc,
 
     // Sometimes swap the two monsters, so as to disguise the original and the
     // copy.
-    if (coinflip())
+    if (x_chance_in_y(1, swap_chance_reciprocal))
```
PleasingFungus, line 3506: You can use `one_chance_in` here. (Same thing, just a little shorter and nicer to read.)

crawl-ref/source/mon-cast.h
```diff
@@ -57,6 +57,7 @@ spell_type living_spell_type_for(monster_type mtyp);
 
 monster* cast_phantom_mirror(monster* mons, monster* targ,
                              int hp_perc = 35,
-                             int summ_type = SPELL_PHANTOM_MIRROR);
+                             int summ_type = SPELL_PHANTOM_MIRROR,
+                             int swap_chance_reciprocal = 2);
```
PleasingFungus, line 61: `swap_chance_reciprocal` is pretty verbose, and requires the caller to have big opinions on what's going on inside the function. Passing in which summon this is (first? second?) is a nicer API to explain.

PleasingFungus (commented):
Thanks for this! Merged with some small changes. :)

PR #3724, opened 2024-03-22
Convert most remaining species fake mutations to real mutations

NormalPerson7:
The objective of this PR is to add detailed descriptions for most species fake mutations that currently exist, by converting them to real mutations and adding descriptions to `/dat/descript/mutations.txt`. I am not touching bloodless vampires or forms.

After this PR, the only remaining species fake mutations for non-deprecated species are octopodes' eight rings, and the effects of size. The former should be reasonably simple to deal with.. I am wary of adding mutations for size, because it affects so much and there are many separate cases (spriggan, felid, kobold, ogre/troll, armataur/naga all have slightly different behaviour), but ideally we would have descriptions that indicate the effect size has on EV, attack penalty in water, item use, etc.

This PR is broken down into several commits, designed to be able to be cherry-picked if only some of these changes are favoured, or if some are approved more quickly than others. The mutations added are:
* stasis (Fo)
* digging (Fo)
* almost no armour (Op)
* no body armour (Dr)
* poison immunity (also includes miasma and sickness immunity) (Gh, Mu, Gr, Dj)
* tough serpentine skin (Na)
* resilient stone body (Gr)
* draconian scales (one for each colour of Dr)
* draconian breaths (one for each colour of Dr except base)
* amphibious (Op, Ba, grey Dr)

Save compatibility has been added. The first five of the above mutations I expect to be relatively uncontroversial. The biggest change is with the scales mutations, where I have refactored the racial AC code. I have added two new mutflags to deal with Dr dragon form and calculating racial AC; I hope this approach is acceptable but if not I can change things to use a more awkward setup. Individual commits go into more detail.

I would of course like feedback on the mutations.txt descriptions and the gain/loss flavour in mutation-data.h. Any other help is also welcome.


NormalPerson7:
The recent changes to allow examining fakemuts accomplish the objective of this PR in a cleaner and more pragmatic way, and thus, this PR is now obsolete,

PR #3727, opened 2024-03-24
Show coglins' offhand weapon in the WebTiles interface

NormalPerson7:
Add a second weapon line for coglins to the WebTiles interface, just below the normal weapon line and above the quiver, which is hidden in all cases where you don't have an offhand weapon equipped.

This works differently from console/offline tiles, but I personally think this is how those formats should work too. (Currently, they have both weapons listed on the same line, which gives each weapon awfully little space for its properties, or in the case of artefacts, even just the name can be too long.)

[Important note: as this code relates to WebTiles I am unable to test it! I'm hoping the checks will reveal any mistakes.]

### DIFF
--- a/crawl-ref/source/tileweb.cc
+++ b/crawl-ref/source/tileweb.cc
@@ -1250,6 +1250,9 @@ void TilesFramework::_send_player(bool force_full)
     }
     json_close_object(true);
 
+    _update_int(force_full, c.offhand_weapon, (bool) you.offhand_weapon(),
+                "offhand_weapon");
+
     _update_int(force_full, c.quiver_item,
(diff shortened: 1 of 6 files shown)


PleasingFungus:
My hero! Thank you for diving into the webtiles mines to implement this. :)

> This works differently from console/offline tiles, but I personally think this is how those formats should work too. (Currently, they have both weapons listed on the same line, which gives each weapon awfully little space for its properties, or in the case of artefacts, even just the name can be too long.)

I'm not sure if it's possible to add more lines to the sidebar in local UIs. I'd be especially worried about trying to do so in console. I strongly suspect that doing so would break some deep assumptions in the current ui (for local tiles and/or console), though someone would need to test or code dive to be sure.

PR #3732, opened 2024-03-26
Revamp qaz

sdynet:
Qaz is a god with interesting mechanics, but he's been around for a long time without any major changes, so he has some stale parts. I wanted to do this without making any major changes, but to improve things internally that weren't up to date.

 * the dust cloud changes to a sandstorm, dealing frag type damage to enemies.
 * Increases the number of clouds maintained each turn from 7 to 9 (based on piety 160) (i.e., increases the rate at which clouds are generated).
 * I relaxed the generation criteria so that clouds would still not be generated in corridors, but clouds could be generated in ambiguous terrain that would not be considered a corridor.
 * Slightly lowered the cost of Upheaval.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -645,7 +645,7 @@ static vector<ability_def> &_get_ability_list()
 
         // Qazlal
         { ABIL_QAZLAL_UPHEAVAL, "Upheaval",
-            3, 0, generic_cost::range(3, 4), LOS_MAX_RANGE,
+            3, 0, generic_cost::range(2, 3), LOS_MAX_RANGE,
             {fail_basis::invo, 40, 5, 20},
             abflag::none },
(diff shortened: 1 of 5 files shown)


regret-index:
Whether or not the sandstorm change is the best of ideas (I'm a little wary of the message spam it'll include later in the game, as it's unlikely to do damage to many enemies post-first-rune; we also have enough cloud tile parsing issues as is so we could definitely do with a proper new tile readied for this), the general community consensus is definitely that Upheaval losing friendly fire for the invoker in e6bea51 has significantly increased Qazlal's overall power level to the point of leaving them off as an overall stronger god. As such, I'm doubtful making such cost less will pass muster.

Hellmonk:
Closing this as won't do for now, however I am interested in adding sandstorm clouds for another purpose later.

PR #3738, opened 2024-03-27
Add damage to ability descriptions

NormalPerson7:
For the following abilities:
* all breath weapons
* storm form: blinkbolt
* demonspawn: hurl damnation
* orb of Dispater: evoke damnation
* Beogh: smiting
* Cheibriados: slouch (damage against a normal-speed monster)
* Ignis: foxfire swarm
* Jiyva: oozemancy
* Qazlal: upheaval, disaster area
* Ru: power leap, apocalypse
* Uskayaw: stomp, grand finale
* Yredelemnul: hurl torchlight
* the Shining One: cleansing flame

Makhleb's minor destruction and major destruction are too random to describe in a simple formula in any meaningful way, except potentially we could make it state the average damage in the future. The same applies to Nemelex' cards.

In many cases, abstract the power and damage formulae for these abilities to ensure the output damage is the same as listed. This often involves listing the rounded-down damage formula, where abilities use div_rand_round on skill/XL.

There are many more abilities that it would be nice to quantify in their descriptions based on the player's current skill but don't fit neatly into the damage category. It would also be nice to list ability noise, and damage at max skill, but since abilities often don't have a clearly defined power-cap and noise is often dealt with elsewhere, this is challenging.

Nonetheless, this will hopefully be useful in allowing players to see how their skill/XL affects damage output. Players should remember that there are usually no actual breakpoints and that the damage listed has usually been rounded down.

I have checked that this all seems to be working correctly.

### DIFF
--- a/crawl-ref/source/ability.cc
+++ b/crawl-ref/source/ability.cc
@@ -724,6 +724,18 @@ static vector<ability_def> &_get_ability_list()
     return Ability_List;
 }
 
+static map<ability_type, spell_type> breath_to_spell =
+    {
+        { ABIL_COMBUSTION_BREATH, SPELL_COMBUSTION_BREATH },
+        { ABIL_GLACIAL_BREATH, SPELL_GLACIAL_BREATH },
+        { ABIL_NULLIFYING_BREATH, SPELL_NULLIFYING_BREATH },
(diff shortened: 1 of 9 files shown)


DracoOmega:
Merged, thanks!

PR #3741, opened 2024-03-28
Draft: Alistair's Ungoldify (new level 8 Alchemy spell)

chucksellick:
A reversal of classic lead-to-gold alchemy: converts a handful of gold coins into base metals (randomly lead, silver or iron). Fires them in a 90 degree cone of spray but only after a turn of movement to choose the direction of the cone - not moving will cause a smaller weaker explosion in a 360 degree spray.

Costs actual gold but deals silver damage vs chaos, and corrosion and poison. Encourages movement and use of space, and synergises interestingly with other effects such as translocations, rampage, acrobat.

A few things need to be done before it's mergable (damage display, possibly monster version, fitting it into books and vaults, more tweaking and balance, make beam tiles better).


PleasingFungus:
I don't know that I'll have enough dev time to review this properly anytime soon, so I'll defer to DracoOmega here.

PR #3745, opened 2024-03-30
Fix three subtle issues with stats panel weapon display in WebTiles

NormalPerson7:
Having fiddled with my menu_colour settings extensively in my rc file I have noticed that the weapon display in the stats panel (top-right corner) does not work as expected.

Three related issues exist:
  1. Due to these colours being set on the JS side, which only has access to the inventory set of menu_colour options, the setting `menu_colour += stats:<colour>:<weapon description>` was essentially useless, as the inventory setting was always used.
  2. The item colour for the stats-weapon panel was not updated when a weapon was equipped, meaning it only displayed in the correct (equipped, typically lightgreen) colour when reloading the game, or when equipping the weapon directly from the ground.
  3. Most concerningly, this also applied to when a weapon was adjusted with the `=i` command if the slots of two weapons with the same base type and enchantment were swapped. For randart weapons this could mean the weapon display could become completely decoupled from reality.

Since the expansion of quiver functionality, these issues have not existed for the quiver slot, as the entire quiver description including the colour is computed on the C++ side. Hence (particularly with launchers) you could end up with the same item in your weapon and quiver slots written in different colours.

This PR fixes both issues by passing the weapon colour(s) directly from the C++ side as a separate piece of information. Incorporate the old unarmed_attack_colour information into this. The separate checks for corrosion on the JS side have also been removed, as they are handled by this new information.

[Note: I've done this by duplicating some code. It would be nice to unduplicate it, I just have absolutely no idea which file the resulting `weapon_colour` function should go in, so if anyone has any suggestions, I'd be happy to improve that. Alternatively, it's probably not a very big deal to leave it as it is here.

Tested on a local WebTiles server.]

### DIFF
--- a/crawl-ref/source/tileweb.cc
+++ b/crawl-ref/source/tileweb.cc
@@ -1386,7 +1386,7 @@ void TilesFramework::_send_item(item_def& current, const item_def& next,
     changed |= _update_string(force_full, current.inscription,
                               next.inscription, "inscription", false);
 
-    // TODO: props?
+    // TODO: props? Hopefully covered by the name check below
 
     changed |= (current.special != next.special);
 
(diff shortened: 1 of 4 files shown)


NormalPerson7:
Ge0ff has indicated that there is even more to this issue, which this PR won't fix in its current state. I will have another look later.

CrawlOdds:
Thanks for the fix! Issue 3 was fixed in ab51d6d; the other commit is merged in 0bed8fc, with a little tidying and especially the code deduplication the PR description hints at.

PR #3750, opened 2024-04-02
Fix red draconians in dragon form having two breath abilities

NormalPerson7:
Prevent them from having the usual Breathe Fire. Now they just get their innate breath.


wheals:
Thanks for catching this, I fixed it with a slightly different approach.

NormalPerson7:
With your approach, base draconians using dragon form at XL < 7 (which I admit is never likely to happen) won't get a breath weapon. Just in case this hadn't been considered, I thought I'd mention it.

wheals:
Good point. Let's call it an easter egg :P

PR #3751, opened 2024-04-02
Allow demigod monks

NormalPerson7:
This reverts commit df79c29.

Demigod monks were banned way back in 0.14 after the piety bonus was added and monks were allowed to use weapons, as they were strictly inferior to fighter, gladiator, skald, warper, and later reaver, without being suitably distinct.

Now, monks start with an orb of light and a potion of ambrosia to help them on their way. They also start with high dex, compared to every other background with a melee weapon choice. They feel unique enough even without the 2* piety bonus. Thus, let's allow demigods to be monks again.

[Also, the image of a demigod beginning as a monk worshipping itself and reflecting on its own divine ancestry seems perfectly acceptable.]


Hellmonk:
Tentatively closing this. "enough" devs feel that the piety bonus is the main draw and losing that is sufficient to disallow the combo. This is not a definitive won't do, but it's probably not happening in the near future.

PR #3755, opened 2024-04-05
Areaprop cleanup

Monkooky:
removes some vestigial areaprops


Monkooky:
It appears sanctuary_1 and sanctuary_2 were added in 
https://github.com/crawl/crawl/commit/6ee6a6e5288df02352cc728d51670a84f6fe8a41
along with a 
    // TODO: update sanctuary here.
It does not appear that these variables have ever been in use throughout crawl's history

it does

pooka109:
And it's applied now. Thank you.

PR #3760, opened 2024-04-11
Fix wanderer evoker starts from allowing duplicate evokers

NormalPerson7:
Previously, if your wanderer rolled a misc evoker, you could still get the same misc evoker generate in your game. Since they're meant to essentially be unrands, fix this.

[Sidenote: misc evokers are generally pretty unbalanced for starting equipment. The lightning rod in particular is insanely strong, with 4 charges of (2+)d15 unavoidable lightning damage. Some rebalancing may be desirable in future.]

### DIFF
--- a/crawl-ref/source/makeitem.cc
+++ b/crawl-ref/source/makeitem.cc
@@ -1636,16 +1636,9 @@ misc_item_type get_misc_item_type(int force_type, bool exclude)
     return NUM_MISCELLANY;
 }
 
-static void _generate_misc_item(item_def& item, int force_type, int item_level)
+// May also be called when a wanderer gets assigned a misc evoker at start
+void handle_generated_misc(misc_item_type typ)
 {
-    const auto typ = get_misc_item_type(force_type);
-    if (typ == NUM_MISCELLANY)
-    {
-        item.base_type = OBJ_WANDS;
-        generate_wand_item(item, OBJ_RANDOM, item_level);
-        return;
-    }
-    item.sub_type = typ;
     switch (typ)
     {
     case MISC_SACK_OF_SPIDERS:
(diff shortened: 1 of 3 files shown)


crawl-ref/source/makeitem.cc
```diff
@@ -1656,13 +1649,26 @@ static void _generate_misc_item(item_def& item, int force_type, int item_level)
     case MISC_TIN_OF_TREMORSTONES:
     case MISC_CONDENSER_VANE:
     case MISC_GRAVITAMBOURINE:
-        you.generated_misc.insert(typ);
+        you.generated_misc.insert(item.sub_type);
```
pooka109, line 1652: Shouldn't `item.sub_type` be `static_cast<misc_item_type>(item.sub_type)` here?

NormalPerson7: It should be, yes. Hopefully fixed now (I decided it would be better if this function just took `misc_item_type` instead of the full `item_def` reference.

crawl-ref/source/ng-wanderer.cc
```diff
@@ -6,6 +6,7 @@
 #include "item-use.h"
 #include "items.h"
 #include "jobs.h"
+#include "makeitem.h"
```
pooka109, line 9: Is there a reason this include is here? The code seems to build properly with it removed.

NormalPerson7: Huh, I thought it was necessary to include the definition of `handle_generated_misc`. Is it not always necessary to include header files for all functions called within the file? I can certainly remove the include.

NormalPerson7: Oh nevermind I just added it to the wrong file, oops.

PR #3764, opened 2024-04-14
Reuse monster_iterator::advance in operator++

bernardosulzbach:
Their implementations were only different in syntax, and we have already done this for actor_near_iterator and monster_near_iterator.


wheals:
Thank you! How would you like to credited in CREDITS.txt?

bernardosulzbach:
I thought this PR was never going to be picked up. Full name (Bernardo Sulzbach) is best, thanks.

PR #3766, opened 2024-04-14
Add Permafrost Eruption as a monster spell for servitors/ghosts

NormalPerson7:
Modelled off the code for Plasma Beam, as Permafrost Eruption is similar (lvl 6, two blasts, automatic targeting). I have set the damage to be about 70% of that of monster Plasma Beam, accounting for the fact Permafrost Eruption doesn't miss, and noting that a similar ratio exists for the player spells. We can always tweak this later.

Servitors will cast this spell, with priority set just below Plasma Beam and just above Fireball (I figure it should be a bit stronger than monster fireball, but the close targeting restriction will make it slightly more awkward for servitors to have than Plasma Beam). Player ghosts will also have it; perhaps at some point a unique will get it too.

[I've tested that this behaves well with player servitors (e.g. won't hit the player). I can't figure out how to make a monster that casts it in wizmode, so I haven't been able to test it works okay with monsters.]


DracoOmega:
Merged, thanks!

(For the future, you can spawn monsters in wizmode with arbitrary spells by using something like "rat spells:permafrost_eruption" or "rat spells:permafrost_eruption.200" if you want them to cast it every action, which is helpful for testing.)

PR #3767, opened 2024-04-15
Fix a few things about wanderer

NormalPerson7:
- Troll, oni and spriggan wanderers now actually get armour when they roll armour skill (acid dragon scales for good quality, otherwise an aux slot) as originally intended. They'll also get armour skill if they start with acid dragon scales.
  - Ghoul wanderers can now get potions of attraction like every other species
  - Formicid wanderers can now start with a quarterstaff or a ranged weapon and a shield
  - Starting wands of warping now get 2-4 charges like iceblast/roots instead of getting 15

Also, to incentivise using starting unarmed combat skill when wanderers get it, make wanderers that start with unarmed skill not get the default club/dagger if they weren't assigned another weapon.

This is courtesy of me generating hundreds of wanderers for testing their stats, and a couple of other changes I intend to put in separate PRs (namely: reducing the strength of staves skill, as +0 quarterstaves are too good for "decent" quality, some adjustments to throwing skill, and adjustments to starting stats).

### DIFF
--- a/crawl-ref/source/ng-wanderer.cc
+++ b/crawl-ref/source/ng-wanderer.cc
@@ -465,7 +465,7 @@ static void _decent_potion_or_scroll()
         { { OBJ_POTIONS, POT_LIGNIFY },
             you.is_lifeless_undead(false) ? 0 : 5 },
         { { OBJ_POTIONS, POT_ATTRACTION },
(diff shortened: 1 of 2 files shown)


NormalPerson7:
There is one more thing to do here: make Tr/On/Sp starting with acid scales actually get armour skill. Currently, it's prevented.

PR #3775, opened 2024-04-20
New Alchemy spell: Gastronomic Expanse

Monkooky:
A draft implementation of a new Alchemy spell, Gastronomic Expanse.
This spell creates an area which expands while you remain in it, but irreversibly begins retracting once you leave. Monsters in the area take acid damage over time and are corroded; players in the area suffer ramping corrosion.

At present, the following remains to do:
-All descriptions need writing
-Targetter needs restriction to cardinal directions
-Gut status needs to provide rCorr
-Gut status needs a less dumb name
-Spell level/schools need thought
-expansion rate/cap need changing
-Damage needs consideration


crawl-ref/source/dat/descript/spells.txt
```diff
@@ -678,7 +678,13 @@ Galvanic Breath spell
 %%%%
 Gastronomic Expanse spell
 
-summon stomach
+Transposes part of the dungeon with the stomach of a great extraplanar
+beast, which will rapidly digest any other creatures within. The transposed
+region will expand as long as the caster remains in it, but they are only 
+partially protected from it's corrosive effects.
```
cfcfcfcfcfcfcfcfcfcf, line 684:
```suggestion
partially protected from its corrosive effects.
```

crawl-ref/source/duration-data.h
```diff
@@ -640,6 +640,9 @@ static const duration_def duration_data[] =
     { DUR_BEOGH_CAN_RECRUIT, LIGHTBLUE, "Recruit", "", "can recruit",
       "You may recruit a defeated apostle into your service", D_EXPIRES,
        {{ "", end_beogh_recruit_window}}},
+    //TODO this needs rewrite
+    { DUR_GASTRONOMIC, YELLOW, "Gut", "transposing a stomach", "gastronomic expanse",
```
chucksellick, line 644: As much as I could infer any consistent logic for different status light colours, YELLOW is generally used for maluses of a less serious nature. Various colours are used for magical effects but I would go with MAGENTA here (same as toxic radiance).

crawl-ref/source/target.cc
```diff
@@ -2024,6 +2024,18 @@ targeter_bog::targeter_bog(const actor *a, int pow)
         affected_positions.insert(c);
 }
 
+targeter_gastronomic::targeter_gastronomic(const actor *a, int pow)
```
chucksellick, line 2027: Doesn't seem to do anything with power, maybe remove it from constructor?

crawl-ref/source/target.cc
```diff
@@ -2024,6 +2024,18 @@ targeter_bog::targeter_bog(const actor *a, int pow)
...
+    : targeter_smite(a, 1, 1, 1, {})
+{
+    
+}
+
+bool targeter_gastronomic::valid_aim(coord_def loc)
+{
+    coord_def delta = loc - you.pos();
```
chucksellick, line 2035: I would try allow aiming anywhere in the 90 degree cone and use a method similar to targeter_cone to preview where the expansion is going to go? Maybe using AFF_YES vs AFF_MAYBE to show the difference between short term and long term expansion

crawl-ref/source/duration-data.h
Monkooky: Thanks, corrected

Monkooky:
drafting this for a bit, as I think it needs at least one major change

regret-index:
With the heavily buggy implementation concerns provided in the dev channel at the original time this was posted, with notable refactorings done since causing several merge conflicts, and with #4554 having provided an example of pulling this forward, I am closing this PR. Monkooky, of course, may open up a new PR with a cleaned up version when he is ready to present such.

PR #3782, opened 2024-04-26
fix corrosion chances for high degree corrodes

Monkooky:
High degree corrodes currently compute the odds of avoiding the second corrode as if corrosion is 1 higher instead of 4 higher.

Likely broken in 7d2b172

not likely relevant since only ever applies to wu jian wrath

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -7211,7 +7211,7 @@ bool player::corrode_equipment(const char* corrosion_source, int degree)
         if (!x_chance_in_y(prev_corr, prev_corr + 28))
         {
             props[CORROSION_KEY].get_int() += res_corr() ? 2 : 4;
-            prev_corr++;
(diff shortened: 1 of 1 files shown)


crawl-ref/source/player.cc
```diff
@@ -7211,7 +7211,7 @@ bool player::corrode_equipment(const char* corrosion_source, int degree)
         if (!x_chance_in_y(prev_corr, prev_corr + 28))
         {
             props[CORROSION_KEY].get_int() += res_corr() ? 2 : 4;
```
chucksellick, line 7213: Would be a little simpler this way round:

```
prev_corr += res_corr() ? 2 : 4;
props[CORROSION_KEY] = prev_corr;
```

(Maybe rename `prev_corr` to just `corrosion`.  It doesn't do anything outside of this loop anyway.)

pooka109:
Applied (both this and mumra's simplification).

PR #3785, opened 2024-04-27
Show weapons, monster weapon brand/magical staff/ranged/throwing damage properly in x-v

NormalPerson7:
**Summary (see individual commits for details):**
In the x-v screen for monsters
* Add terse weapon descriptions for monsters using weapons in the Attacks column of the attacks table
* Show weapon brand max damage for most relevant brands in the Max Damage column of the attacks table, indicating clearly the bonus damage from the brand
* Show magical staff bonus damage in the Max Damage column in the same way
* Make the attack table's columns have flexible width, so that weapons, brand damage etc don't get cut off unless line length would otherwise exceed the 80 character maximum
* Show the max damage of AF_DRAIN
* Clarify AF_PURE_FIRE
* Fix an off-by-one error in x-v for all calculations involving monster weapons
* Let monsters use staves of conjuration
* Include monster slaying equipment in the damage in the Max Damage column
* Display ranged weapon damage as "Shoot" instead of "Hit" in the attack table
* Include archer bonus damage in the main damage number in the Max Damage column of the attack table, and include Nessos's special effect in the Bonus column; remove the corresponding special text notes
* Include throwing in the attack table, showing monster throwing damage properly for the first time
* Remove the now-redundant "It is quivering ..." text from x-v.

**Further todo list**
* Work out which damage bonuses do or don't apply to ranged/throwing and describe this correctly

**Some examples:**
| Monster | Before | After |
| ---- | ---- | ---- |
| tengu reaver | image | image |
| two-headed ogre |  image| image |
| Sonja | image | image |
| Nessos |  image | image |

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -5442,23 +5442,6 @@ static string _monster_attacks_description(const monster_info& mi)
     return result.str();
 }
 
-static string _monster_missiles_description(const monster_info& mi)
-{
-    item_def *missile = mi.inv[MSLOT_MISSILE].get();
-    if (!missile)
-        return "";
-
-    string desc;
-    desc += uppercase_first(mi.pronoun(PRONOUN_SUBJECTIVE));
-    desc += mi.pronoun_plurality() ? " are quivering " : " is quivering ";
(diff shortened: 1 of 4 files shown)


DracoOmega:
Very thorough and extremely helpful. Thanks a ton!

PR #3792, opened 2024-05-02
Fix formicids able to shaft while in non-mutaion-preserving forms

### LINKED ISSUE #3776: formicid shafting in a tree (as well as pig etc) form possible via f, but not via a menu
console (various linux, shouldn't matter)

reproducible both in 0.31 and fresh master

**DESCRIPTION:**
it seems like tree (*) form is supposed to prevent formicids from shafting (at least it disappears from ability menu), but quivering this ability and using `f` lets you do it anyway

(*): encountered in actual game; later on i confirmed same behaviour in master for tree, pig, fungus etc)

caryoscelus:
Previously, if formicid character was in a form that disables mutations, ability menu would remove both Dig and Shaft self, but the latter was still possible by using ability quiver. Now this workaround is removed.

fixes #3776

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -7962,6 +7962,13 @@ bool player::do_shaft()
 
 bool player::can_do_shaft_ability(bool quiet) const
 {
+    if (!form_keeps_mutations())
+    {
+        if (!quiet)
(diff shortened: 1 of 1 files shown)


DracoOmega:
Merged, thanks!

PR #3797, opened 2024-05-04
Mark transporter destinations as known on entering a fully_mapped level

Aliscans:
This lets the player use travel commands to reach an altar in hellmonk_temple_divided_pantheon (where there are transporters between the stairs and most of the altars) without first visiting the transporters manually.

brflag::fully_map now also prevents transporters from appearing in the level overview. On other levels this happens until the transporter is used.


DracoOmega:
Merged, thanks!

PR #3803, opened 2024-05-05
Reaver background rebalancing + Adjust fugue

sdynet:
There are two reasons why I want to turn the reaver into a de facto necro-warrior.
1. reaver is a warrior-mage class, but there is no synergy between weapons and magic, they simply mechanically use weapons and magic together.
 * Cinder Acolyte: I start with a flame-branded weapon. Scorch spells can be used to create synergy.
 * HexSlinger: Training with weapons makes it easier to hit enemies with Inner Flame debuffs for the intended effect; training with magic, on the other hand, makes it easier to inflict debuffs on enemies. Either way, there are synergies.
 * Enchanter: Practicing weapons increases your stab damage. Practicing magic makes it easier to assassinate enemies. There is a synergy between practicing either.
However, the reaver has no such synergy, so I believe the reaver's spell list should be changed to allow weapons and magic to synergize with each other.

2. I think that the backgrounds of necromancers and necro-warriors should be separated from each other. necromancers are mages, not warrior-mages. however, their spell composition is a mix of both types. therefore, i would like to see the necromancer's spell list changed to be more mage-like and less warrior-mage-like.

The Reaver needs to change, and the Warrior-mage spells that the Necromancer has need to find a new job, so I decide that it would be a good idea to rebalance the Reaver to Necro-Warrior.


Grimdank-Godhead:
I have been looking for a necro-warrior type background for along time, i hope this is merged.

PleasingFungus:
Thanks for submitting this!

The point of Reaver is to be a 'melee fighter' who also has powerful conjurations as supplemental and emergency attacks, rather than as their primary means of killing enemies. This works.

As others pointed out on Discord, Necromancer is already a necro-warrior. Adding another one would feel a bit redundant.

Cheers!

sdynet:
T_T cheers!
No thoughts on this one?
https://github.com/crawl/crawl/pull/3803/commits/fe9b2f4bef29965080bc05f2caa6e36262f0cae7

PleasingFungus:
> While FUGUE's max stack effect (negative energy attack) is fun and useful, it's not easy to build up to 7 stacks in a real-world combat setting, as the average player won't put themselves in a situation where they're fighting 7 creatures in a row unless they have to. Also, it's not uncommon to have no or very little left at the end of 7 stacks. Therefore, I'm lowering the max stack of this spell to 4 to make it easier to see the final effect. I'm also lowering the level to 2 to reflect the weakened effect.

Sadly, I haven't had a chance to play with Fugue yet, so my opinion isn't very useful. :) @DracoOmega might have some thoughts.

PR #3807, opened 2024-05-08
Tune down quicksilver bolt by 3d5 to bring more in line with other breaths

Rytisgit:
Tune down quicksilver bolt by 3d5 to bring more in line with other breaths

Quicksilver bolt is powerful, accurate, irresistible, and has a strong effect. The damage of quicksilver matches other breaths which don't have a special effect. To bring it more in line with acid breath, which is tuned to a lower damage level to account for the corrosion, reduce the damage by 3d5, bringing purple draconian from 3d19 to 3d14, and quicksilver dragon from 3d20 to 3d15.


regret-index:
> The damage of quicksilver matches other breaths which don't have a special effect. To bring it more in line with acid breath, which is tuned to a lower damage level to account for the corrosion,

This kind of feels patently untrue. In terms of dragons: acid dragons are pre-Lair spawns, and literally every other dragon besides Steam dragons (also pre-Lair spawns), swamp dragons (who inflict more damage after poison), Storm dragons (which have half-ac checks and double-zaps) deal more damage when un-resisted. While a resist does put them lower than Quicksilver's average damage, the resistible ones frequently have positions for this context: fire and ice dragons are mostly threats in early portals or the end of Lair where people aren't guaranteed to resist them, storm dragons are mostly resisted through randart properties or a consumable, golden dragons check a conflicting pair of resists, and there's heavy debate over ever swapping to rN in general throughout the game as is, much less for shadow dragons. A monster mostly meant for Vaults and Depths being more dangerous than average hardly stands out as needing adjustment.

In terms of draconians: yellow draconians have twice the accuracy of purple draconians, which make them much more likely to wreck EV-focused characters, while black draconians also have the half-ac and double-zap properties storm dragons do. Purple draconians are indeed some of the most dangerous draconians overall, but if anything it's good that they stand out because unclassed / base draconian bands in Zot are mostly in a fodder position anyway, so having specific enemies in them be notably dangerous is fine. Using current-version and trunk Zot death statistics to include proteans at a reasonable sample size of over 900 deaths, quicksilvers are 15th, shifters are 16th, and base purple draconians are tied for 22nd well below said yellow and black draconians. Shifters got nerfed in 9efa865 as part of constriction adjustments, also.

I'm rather disinclined to weaken either of their breath over such terms. I might look at further nerfing the one purple draconian spamming vault in Zot entirely incidentally outside of this (cheibrodos_zot_enchanting_chambers), and the quicksilver dragon Snake end is still a little overtuned regardless, but those are vault affairs disconnected from the monsters themselves.

DracoOmega:
Yes, I see no reason to nerf this breath in particular, and in the average case it is not even the most dangerous draconian breath from a damage perspective and hardly seems out of line.

(Also, yellow drac breath has something closer to triple the accuracy of purple breath, not double)

Rytisgit:
Thanks for your thoughts. Just a couple of points I wanted to add - the indirect kill count of the quicksilver stripping buffs like resistance/haste/teleport should be quite impactful in helping the oofs/orb guardians to make the kills zot. I don't think there is any other real adjustment which could be made to take some bite out of quicksilver, instead of just a damage decrease(now thinking about it, perhaps quicksilver could only remove 1 or some number of enchantments instead of all of them, Similar to how corrosion stacks up.). Also quicksilver dragons are speed 15 so they get more mileage than the other dragons with their breaths.

I would say I am mostly thinking about Zot in terms of balancing the quicksilver not the rest of the game, since in most cases dealing with any enemy 1v1 is possible in crawl.

PR #3830, opened 2024-05-23
Replace mprf("string") and mprf("%s", string) with mpr(string).

Aliscans:
Also wrap some long lines.

### DIFF
--- a/crawl-ref/source/traps.cc
+++ b/crawl-ref/source/traps.cc
@@ -537,7 +537,7 @@ void trap_def::trigger(actor& triggerer)
             break;
         trap_destroyed = true;
         if (you_trigger)
-            mprf("You set off the alarm!");
+            mpr("You set off the alarm!");
         else
             mprf("%s %s the alarm!", triggerer.name(DESC_THE).c_str(),
                  mons_intel(*m) >= I_HUMAN ? "pulls" : "sets off");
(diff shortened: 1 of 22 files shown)


crawl-ref/source/traps.cc
```diff
@@ -1162,18 +1162,19 @@ void do_trap_effects()
...
             }
-            mprf("With a horrendous wail, an alarm goes off!");
+            mpr("With a horrendous wail, an alarm goes off!");
             fake_noisy(40, you.pos());
             you.sentinel_mark(true);
             break;
 
         case TRAP_TELEPORT:
         {
-            string msg = make_stringf("%s and a teleportation trap spontaneously manifests!",
```
pooka109, line 1172: The commit message states that a make_stringf() is replaced with a "std::string +", but this is the only make_stringf() in the commit, and, while its line is wrapped, it's still a make_stringf() after the change. I'm guessing this is an oversight?

Aliscans: Sorry. I thought about changing that line, but must have decided against it.

Replacing that line with a + would have saved a line, but I can imagine someone finding the make_stringf() format easier to read. I was trying to avoid "personal preference" stuff here.

I've removed the reference from the PR description. I've left the commit message as I don't know how well it would work if I tried to change it.

pooka109: No problem. I've applied it, and edited the commit message. Thanks.

NormalPerson7:
Aliscans, just for future reference, you can use `git commit --amend` to adjust a commit message (and this also incorporates any additional staged changes into the commit) and then push your changes with `git push --force`

PR #3832, opened 2024-05-24
Smooth out stat modifiers for weapon/unarmed damage

beeceropteryx:
Currently, Crawl uses integer division when applying stat modifiers to a weapon's base damage, which results in sharp breakpoints: using a dagger, the player does exactly the same amount of damage with Dex 29 as they do with Dex 20. This is counterintuitive, misleading, and inconsistent with Crawl's design philosophy. This implements the simple fix of replacing integer division with random rounding.

### DIFF
--- a/crawl-ref/source/fight.cc
+++ b/crawl-ref/source/fight.cc
@@ -1494,9 +1494,8 @@ int stat_modify_damage(int damage, skill_type wpn_skill, bool using_weapon)
     const bool use_str = weapon_uses_strength(wpn_skill, using_weapon);
     const int attr = use_str ? you.strength() : you.dex();
     damage *= max(1.0, 75 + 2.5 * attr);
-    damage /= 100;
 
-    return damage;
+    return div_rand_round(damage, 100);
 }
 
 int apply_weapon_skill(int damage, skill_type wpn_skill, bool random)


chucksellick:
On a quick scan of the codebase, I see a couple of problems with making this change in the way you've done it here (although the intent sounds very good, as such breakpoints are definitely undesirable).

Firstly, stat_modify_damage is called from damage_rating in describe.cc, meaning the displayed weapon description will randomly vary the amount of damage it reports. We should be showing a max damage here, I assume, (and you have effectively increased the max damage by 1 for most weapons most of the time, even if the chance of that highest damage is very low). 

Secondly, in the places where stat_modify_damage is called in combat (in attack.cc and melee-attack.cc), it is always immediately followed by a random2 roll to determine the actual damage. So it's adding an extra random roll where really only one should be needed. (I don't *think* this affects the overall probability of hitting the new max damage, but it's confusing enough that it makes me unsure)

How I would change this: make stat_modify_damage return the *undivided* number instead, and leave it up to the callers to handle this. For combat you can then pass this value straight into random2, and do integer division *after* that, basically achieving what div_round_round does but without that extra roll. In damage_rating you can just divide with rounding wherever appropriate (not sure if round up or round to nearest is better; stat_modify_damage is used twice and i think "max" and "average" are both useful; maybe here we could actually show decimals to show the stats making a continous difference?)

The final thing that needs some consideration is: this is an overall buff to combat, by my reckoning this averages 0.5 extra damage per attack over the entire game. With the number of attacks the player makes over the course of a game, this is actually a huge amount of free damage added up, and I wonder if a small nerf is required in the equation to compensate this?

chucksellick:
(Sorry it took so long to reply, I only just noticed!)

I just meant to use div_round_up after the random2 roll i think, so you get that extra damage point if your roll is over by whatever size chance you have.

So the stat function returns 150:

random2(150) -- if it's below 100, 1 damage; 100 or above, 2 damage; there's a 1/3 chance of 2 damage which is correct because i was halfway between breakpoints. This works out the same as doing a div_rand_round elsewhere, but using one less rng call.

PR #3834, opened 2024-05-28
Abyss decor: colourful horror statue.

pooka109:
The tile in question is made from colour-shifted combinations of UNUSED/monsters/colourful_horror.png and UNUSED/features/statue_bases.png, and in-game is a redefined metal statue. I've added a small Abyss vault that puts it next to a Lugonu altar.

(I considered this for chaotic gods, but Xom and Nemelex have scintillating statues already, and it didn't seem demonic enough for Makhleb.)


regret-index:
While I'm fond of additional statues types, and would like to consider in the future partitioning the current huge breadth of aimlessly varied statue tiles into being tied to specific branch flavours, I'm worried that this particular PR's statues look too much like a regular monster with the distinguishing dais being easily missed. (Xom's chess pieces being made into a statue at least inherently benefited from being an already-abstracted object.) Abyss _is_ somewhat suiting as a place where one could confuse features for monsters and vice-versa, but the untidiness of terrain at all times makes such a notion rather concerning from a user interface front.

pooka109:
Okay. I wasn't entirely certain about a color-shifting statue in the already-shifting Abyss (another case of fitting thematically versus fitting in gameplay).

I am curious about how a desaturated version would turn out, but if it's really not worth the effort, don't bother.

regret-index:
I ended up trying out the desaturation anyway, as visible below:
vivid_horror_statues
Unfortunately, a quick survey said they all looked too much like monsters still. The tile might be reusable in the future, and rainbow abyssal statues without nearly so much detail might also be drawn and added eventually, but for now repurposing it for entirely decorative purposes is a bust, alas.

PR #3843, opened 2024-06-01
Make various small improvements to ability summary on ^ screen

NormalPerson7:
- Beogh: add lines for Beogh sending challenges as you gain piety, and explain that they can be recruited if you defeat them; improve various other inconsistencies around ^ for new Beogh
 - Ignis: explain that Rising Flame is one-use-only
 - Lugonu: show passive banishment on other gods' wrath
 - Yredelemnul: explain that light the torch gives allies at 1*

Also move a number of passive abilities with hardcoded descriptions describe-god.cc to the table of god powers in religion.cc. The only abilities left in describe-god.cc are those that have unusual criteria (e.g. Hep frailty), or those whose text changes as you gain piety (e.g. lifesaving, umbra).

There remain several minor passives (e.g. Trog's ally protection and berserk extension, Jiyva's jelly eating and jelly protection, Elyvilon's ally lifesaving) which are too minor to appear on ^.

### DIFF
--- a/crawl-ref/source/describe-god.cc
+++ b/crawl-ref/source/describe-god.cc
@@ -773,9 +773,9 @@ static formatted_string _describe_god_powers(god_type which_god)
             desc.textcolour(god_colour(which_god));
 
         if (piety >= piety_breakpoint(5))
-            desc.cprintf("Orcs frequently recognize you as Beogh's chosen one.\n");
+            desc.cprintf("Orcs frequently recognise you as Beogh's chosen one.\n");
         else
-            desc.cprintf("Orcs sometimes recognize you as one of their own.\n");
+            desc.cprintf("Orcs sometimes recognise you as one of their own.\n");
     }
     break;
 
--- a/crawl-ref/source/describe-god.cc
+++ b/crawl-ref/source/describe-god.cc
@@ -766,13 +766,16 @@ static formatted_string _describe_god_powers(god_type which_god)
     {
     case GOD_BEOGH:
     {
-        if (piety >= piety_breakpoint(5))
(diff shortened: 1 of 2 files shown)


NormalPerson7:
I have no idea why this branch said it had conflicts (I built it on top of `master` earlier today), but hopefully that's fixed now. I've also added Jiyva's passive jelly eating and jelly protection, and unlocking the Slime Pits, to the commit.

pooka109:
For the record, the conflicts were probably caused by the "recognize" -> "recognise" spelling change in the Beogh-related messages, which the current version of the PR undoes (it shouldn't).

NormalPerson7:
Thanks, I hadn't noticed that. I'm away at the moment so can't edit the original commit/squash the fix, but otherwise, this is ready.

pooka109:
Applied (squashed and with a few minor fixes to Beogh ability logic). Thank you.

NormalPerson7:
Thanks for the fixes.

I put Beogh apostle recruitment at 3* because you can't get challenges to recruit an apostle below 3*, so it seemed wrong to say this is possible at 0*. I totally see your argument though: you don't lose the ability to recruit just because you drop below 3*. It's a weird one, it's hard to find a perfect solution, but it needs to say something there, else it looks like Beogh's main thing is "fight extra monsters for no reward", not what we want.

Perhaps it would be best to turn it on at 3* and then never turn it off again, or something, if that's even possible. 

In any case, any threshold for this ability is better than the old behaviour, which was that the line about recruiting apostles appeared only immediately after winning a challenge and before actually recruiting the apostle.

pooka109:
Indeed. The recruitment message stuff is tricky, but it's a definite improvement relative to what was there before.

PR #3844, opened 2024-06-01
Tweak wanderer good-throwing and good-weapon starting equipment

NormalPerson7:
Previously, the good set of throwing items for wanderer was 1-4 curare-tipped darts and a +2 dagger. Neither of these scale well with throwing skill, which is likely to be your highest skill. This could also be extremely disappointing if you got 1 curare dart and it mulched immediately. To encourage actually training throwing skill, replace the enchanted dagger with large rocks/javelins/boomerangs (depending on size), up to a total of 7 pieces of ammunition. (Note: this uses the same code as the gladiator background, so small species will get double the number of boomerangs.)

Also, the good set of weapon items for all weapon skills had a 4/5 chance of giving a +2 enchantment to a decent-quality weapon (falchion, mace, etc), and a 1/5 chance of upgrading to a good-quality weapon (long sword, flail, etc). In practice this upgrade was rather rare, and starting with a +2 decent weapon is a little too frequent. So let's bump the upgrade chance to 1/3, for the sake of variety.

[Notes: quarterstaves are still way too good compared to all of the other decent-quality weapons. I'd like to replace it with a magical staff, but a magical staff felt just slightly too weak in melee without a matching magic school in my testing to be viable as a starting weapon. Any thoughts on this?]

### DIFF
--- a/crawl-ref/source/ng-setup.cc
+++ b/crawl-ref/source/ng-setup.cc
@@ -191,14 +191,14 @@ item_def* newgame_make_item(object_class_type base,
     return &item;
 }
 
-static void _give_throwing_ammo()
+void give_throwing_ammo(int n)
 {
     if (species::can_throw_large_rocks(you.species))
-        newgame_make_item(OBJ_MISSILES, MI_LARGE_ROCK, 1);
+        newgame_make_item(OBJ_MISSILES, MI_LARGE_ROCK, n);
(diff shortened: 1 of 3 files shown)


pooka109:
Looks good to me; it's in.

(As for quarterstaves, I'm not certain what to do about those.)

PR #3852, opened 2024-06-03
Mumra/monster wands revised

chucksellick:
* Make wands available for monsters later in the game with better power scaling (and hopefully without breaking the early game!)
* Fixed damage and "chance to defeat your will" when displaying wand spells in xv

### DIFF
--- a/crawl-ref/source/dat/mons/grinder.yaml
+++ b/crawl-ref/source/dat/mons/grinder.yaml
@@ -1,6 +1,6 @@
 name: "Grinder"
 glyph: {char: "5", colour: blue}
-flags: [no_ht_wand, speaks, unique, female]
+flags: [speaks, unique, female]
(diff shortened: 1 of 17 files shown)


crawl-ref/source/artefact.cc
```diff
@@ -1062,7 +1058,7 @@ static void _get_randart_properties(const item_def &item,
     // enhance one good property.
     int good = max(quality + fixed_bad + bad - fixed_good, 0);
 
-    // We want avoid generating more then 4-ish properties properties or things
+    // We want avoid to generating more then 4-ish properties or things
```
pooka109, line 1061: This should start with "We want to avoid generating...".

pooka109: It happens :)

PR #3853, opened 2024-06-04
new yred altar

fbwer:
I changed it to be more traditionally altar. 

To respect the original skull design, I modified the existing tile.

Since yred was once a good god, I made the altar a bright purple to show traces of they old appearance.

Due to yred fall, the top of the altar is broken.

yred spilled black blood on the ground.

The black blood covered the altar, revealing the bright purple parts as skulls.

yredelemnul

but it isn't basalt. its hard basalt good god altar. i think it is neither holy nor unholy.
maybe its vandalized mauve marble altar.

fat yredelemnul
and this is fat ver


regret-index:
The mixture of black blood and neon pink for the altar base mostly looks like an image error rather than any actual notion of Yredelemnul.... having all of their altars spontaneously explode with black blood when the good gods rejected them?.... It's a cool metal notion, but the current appearance rejecting any notion of even having an altar is a far better look for the tile, and I'd rather just adjust the description to mention such instead of shrinking it.

PR #3860, opened 2024-06-08
Fix a couple more things about the new Webtiles colours

NormalPerson7:
Previously, using black as a menu_colour option was never useful, because black text is always invisible on the black game background. However, we can now redefine the colour black, as of d719f66, opening up a 16th possible text colour to use with menu_colour.

Upon trying to do exactly this myself, I discovered the Webtiles inventory UI implicitly, and apparently unintentionally, replaces black text with lightgray, regardless of what black may have been rebound to. Let's fix this, so we really can have a 16th colour.

Also fix the poison HP-bar colour being `yellow` instead of redefinable yellow.


DracoOmega:
Merged, thanks!

PR #3861, opened 2024-06-09
Improve picking up part of a stack of items with the mouse in local tiles

WizardIke:
Previously cntrl-clicking an item on the ground in the local tiles onscreen inventory would bring up the item pickup menu where you could type in a number that couldn't be seen and then press a to select the item again and then press enter. Now, cntrl-clicking the item prompts for the number you want to pick up and then picks up that number when you press enter. The new way is similar to how dropping part of a stack works.

Before:
Screenshot (1661)

After:
Screenshot (1662)

### DIFF
--- a/crawl-ref/source/item-use.cc
+++ b/crawl-ref/source/item-use.cc
@@ -4489,12 +4489,19 @@ void tile_item_pickup(int idx, bool part)
         return;
     }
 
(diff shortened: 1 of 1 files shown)


pooka109:
It's in. Thank you.

PR #3862, opened 2024-06-09
Update README.md

Korachof:
Added a link right below # How to Play to help navigate users to the installation help documentation (https://github.com/crawl/crawl/blob/master/crawl-ref/INSTALL.md)

This documentation is nested under several folders. For new users or those who aren't technically-savvy, having these installation instructions directly in the README, under How to Play, makes sense. That way they don't have to go searching if they run into any confusion, which will cause frustration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ Dungeon Crawl Stone Soup is a game of dungeon exploration, combat and magic, inv
 
 ## How to Play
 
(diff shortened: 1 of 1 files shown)


pooka109:
Applied. Thank you.

pooka109:
Since this is your first contribution here, what name do you want to be listed under in the credits?

(Also, sorry it took so long to get to this.)

pooka109:
Good to know, and you're welcome.

PR #3864, opened 2024-06-13
A few more tweaks/fixes to the x-v attacks table

NormalPerson7:
1. Fix ranged weapon damage including might/berserk/weak effects.
2. Treat ghost/pan lord melee brands like any other weapon brand (showing the additional damage information in the damage column rather than the bonus column).
3. Stop pretending that monster ranged weapon attacks apply melee attack flavours.
4. Add an extra row to the attacks table for monsters with ranged weapons to show what their base melee damage and flavour would be if they weren't wielding their ranged weapon. This is so that players can still always see what a monster's attack flavours are, even if it's wielding a ranged weapon.

I have additionally split up `_monster_attacks_description` into several different functions, to allow me to dramatically simplify the code for point 4.

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -1700,29 +1700,6 @@ static string _category_string(const item_def &item, bool monster)
     return description;
 }
 
-static string _ghost_brand_extra_info(brand_type brand)
-{
-    switch (brand)
-    {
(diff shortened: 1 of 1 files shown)


DracoOmega:
Greatly appreciated work, as usual. Thanks!

PR #3873, opened 2024-06-18
Don't let the player take actions when their turn is over

### LINKED ISSUE #3563: Game crashes after repeatedly equipping and unequipping weapon in off hand
windows local tiles

0.32-a0-393-g38ff01cdb3

**DESCRIPTION:**
1. Hold down ctrl and left click a weapon in inventory repeatedly.
2. Clicks will become unresponsive shortly before crashing.

**EXTRA FILES:**
[crash-Vucighu-20240218-212040.txt](https://github.com/crawl/crawl/files/14322753/crash-Vucighu-20240218-212040.txt)

### LINKED ISSUE #3799: Rebinding '<' to 'r' causes mouse clicking on an upstairs you are on to open the read scrolls menu

local tiles

Version: 0.32-a0-1074-g4fd57f6e19

**DESCRIPTION:**

I used the in game Ctrl-d menu to rebind '<' to 'r' and then left mouse clicked on an up staircase I was on. Instead of going up the staircase, the read scrolls menu opened up.

**EXTRA FILES:**

Screenshot (1588)

### LINKED ISSUE #3801: Control left mouse clicking an adjacent open door says "Unknown command"

local tiles

Version: 0.32-a0-1074-g4fd57f6e19

**DESCRIPTION:**

If you control left mouse click an adjacent open door with no known creatures in it, you get the message "Unknown command". Looking at the code, it looks like it is meant to close the door. I'm not sure weather it's a good idea to make it close the door though as that would interfere with control clicking to attack unseen creatures.

The function with an error is _adjacent_cmd in travel.cc where is has `return CMD_CLOSE_DOOR_LEFT - CMD_MOVE_LEFT` instead of `return cmd + CMD_CLOSE_DOOR_LEFT - CMD_MOVE_LEFT`.

**EXTRA FILES:**

### LINKED ISSUE #3802: Shift left mouse clicking the map in map view mode results in auto explore starting after you exit map view

local tiles

Version: 0.32-a0-1074-g4fd57f6e19

**DESCRIPTION:**

Pressing X to go into map view followed by shift left mouse clicking on the map won't no anything immediately, but will start auto explore when you press escape to exit map view.

**EXTRA FILES:**

### LINKED ISSUE #3804: Right mouse clicking the map and dragging in map view does nothing

windows local tiles

Versions where clicking and dragging does nothing: 0.32-a0-1074-g4fd57f6e19, 0.31.0, 0.30.2, 0.29.1, 0.28.0
Versions where dragging does nothing, but the initial click works: 0.27.1, 0.26.1

**DESCRIPTION:**

While in map view mode, you should be able to right mouse click the map to view that part of the map. You should then be able to drag the mouse around the map while holding the right mouse button to view different parts of the map. However, nothing happens when I do this.

**EXTRA FILES:**

WizardIke:
If the player entered multiple mouse inputs in a single frame in local tiles, they would all be processed and the time taken for the player's action would be set to the time for the last action. The monsters and environment would then get to react.

Equipping a weapon as a coglin doesn't happen in a single turn but rather adds a "delay" to a queue that makes you skip your turn until it finishes equipping or gets interrupted. So, if you start equipping the same weapon twice in the same turn with this bug, it would skip your turns until the weapon equipped and then skip your turns until the weapon equipped again which would crash as you were already wielding the weapon.

The was another bug were the function to check if there was more input keycodes (a keycode being e.g. a, D, ctrl-b) was incorrect and would often return true when ctrl, shift or alt were held even though they don't give a keycode on their own. This would sometimes cause the game to freeze when reading in all the available keycodes when one of these keys were held making it easy to input multiple mouse inputs in one frame.

The previous two bugs were the cause of issue #3563 so this change should fix #3563.

I also fixed #3804, fixed #3802, fixed #3801, and fixed #3799 when I was rewriting the UI code.


crawl-ref/source/macro.cc
```diff
@@ -56,6 +56,10 @@
 #include "unicode.h"
 #include "version.h"
 
+// XXX: we might not want to use an std::deque here as they can waste a
```
chucksellick, line 59: 4096 bytes seems overkill certainly, but also probably not an issue for a one-off global.  Maybe take a note or create an issue and leave this for another PR?

chucksellick:
This seems overall like it's making some really useful changes and fixes. I particularly like that it changes a good number of incomprehensible magic number comparisons into more meaningful enum checks.

However it's rather a large PR and touches a lot of fairly old and not well understood areas of crawl code, making it very difficult to assess the total impact and whether it might introduce any new (and worse) bugs. Especially as crawl runs on so many different systems and environments.

I'm wondering if it's possible to break any parts of the PR off into some smaller more granular commits, as it might be easier to review the bigger changes if they weren't mixed in with so many small changes? And anything irrelevant to the main change of the PR deferred for a separate one.

Also, some of the XX and TODO comments added should be cleaned up and either dealt with or also deferred for later.

Definitely some nice work here, it's just a slightly scary PR currently for a dev to merge!

crawl-ref/source/macro.cc
WizardIke: Each entry in the keymaps contains two deques so it's about 8 kB of wasted space per keymap entry with Clang (1 kB with GCC). It looks like keymaps only holds user defined keymaps and not any of the default key mappings so this isn't as bad as I initially thought.

I will move this to another issue or PR as it doesn't belong here

WizardIke:
I've pulled the smaller fixes out into their own issues. The rest of the code is too outdated to be useful (and probably should be two PRs anyway).

PR #3875, opened 2024-06-18
Don't leak invisible monsters when casting Gell's Gavotte

NormalPerson7:
Fix the targeter revealing the locations of invisible monsters in line of sight when they would be hurt by the spell.


DracoOmega:
Merged (with an addition). Thanks!

PR #3879, opened 2024-06-21
Non mindless TRJ

aaronegolden:
It doesn't make sense for TRJ to be "mindless". With this change TRJ can be pacified by Ely worshipers (with a very heavy investment in invocations). Jiyva dies with the usual effects, not on pacification, but when TRJ goes up the stairs.

### DIFF
--- a/crawl-ref/source/dat/mons/royal-jelly.yaml
+++ b/crawl-ref/source/dat/mons/royal-jelly.yaml
@@ -13,7 +13,7 @@ hd: 21
 hp_10x: 2310
 ac: 8
 ev: 4
-intelligence: brainless
+intelligence: human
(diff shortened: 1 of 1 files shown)


chucksellick:
I'm curious, why doesn't it make sense? All other J monsters are the same intelligence (actually brainless - literally they don't have brains, they're just blobs of jelly).
The only exception is Dissolution, an ex-human who converted to Jiyva and retained their intelligence. So we have an explanation for them.
The intelligence change has a fair few other mechanical implications, not just pacification ... e.g. avoiding dangerous clouds, susceptible to more debuffs / attacks, and it would be hard to justify such big changes to a very distinct and memorable late game boss.

PR #3886, opened 2024-06-25
display quiet

fbwer:
The dampen noise effect is now output from the display.

This is expressed in a format that replaces Display Noise: with Quiet:.

This can only be seen in the cloak of Thief and Dithmenos.

### DIFF
--- a/crawl-ref/source/output.cc
+++ b/crawl-ref/source/output.cc
@@ -753,7 +753,7 @@ static void _print_stats_noise(int x, int y)
     if (have_passive(passive_t::dampen_noise) || player_equip_unrand(UNRAND_THIEF))
     {
         textcolour(LIGHTMAGENTA);
-        CPRINTF("Quiet: ");
+        CPRINTF("Noise/2: ");
     }
     else 
     {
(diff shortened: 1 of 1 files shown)


crawl-ref/source/output.cc
```diff
@@ -750,8 +750,16 @@ static void _print_stats_noise(int x, int y)
 
     bool silence = silenced(you.pos());
     int level = silence ? 0 : you.get_noise_perception(true);
-    textcolour(HUD_CAPTION_COLOUR);
-    CPRINTF("Noise: ");
+    if (have_passive(passive_t::dampen_noise) || player_equip_unrand(UNRAND_THIEF))
```
Cgettys, line 753: Very minor nit, but I'm pretty sure it should be
```
    if (have_passive(passive_t::dampen_noise) || player_equip_unrand(UNRAND_THIEF))
    {
        textcolour(LIGHTMAGENTA);
        if (have_passive(passive_t::dampen_noise) && player_equip_unrand(UNRAND_THIEF))
             CPRINTF("Noi/4: ");
        else
             CPRINTF("Noi/2: ");

    }
    else 
    {
        textcolour(HUD_CAPTION_COLOUR);
        CPRINTF("Noise: ");
    }
```

Hellmonk:
Discussed this with DracoOmega, and we'd like to display noise dampening somewhere else if necessary. Specifically, there were concerns that it is misleading because the noise bar is still displaying the actual noise made, not noise/2 or quiet.

PR #3888, opened 2024-06-27
Don't waste a turn when summoning with no space

### LINKED ISSUE #3837: Summon Lightning Spire wastes a turn when terrain clearly prevents spire placement
Webtiles

Latest trunk

**DESCRIPTION:**
Casting Summon Lightning Spire while in/over deep water and with the entire 5x5 casting area blocked by deep water causes it to waste a turn and 4 MP with the message "Nothing appears to happen." It should return `spret::abort` instead.

WizardIke:
When trying to summon a creature where you can't see any available space to place it, don't spend a turn or any resources.

Fixes #3837


crawl-ref/source/mon-util.h
```diff
@@ -290,8 +290,13 @@ mon_intel_type mons_intel(const monster& mon);
 // since the class variants do not handle zombies correctly.
 habitat_type mons_habitat_type(monster_type t, monster_type base_t,
                                bool real_amphibious = false);
+// The monster type should already be adjusted for zombie types
+habitat_type mons_habitat_type(monster_type t, bool real_amphibious = false);
```
AlejandorLazaro, line 294: Functions in `mon-util.cc` whose first parameter is `monster_type` *mostly* start with the prefix `mons_class_`. For consistency it might be good to update this function name to `mon_class_habitat_type`.

That said, `mons_zombie_size`, `mons_skeleton`, and even the current `mons_habitat_type` right above this new function don't have that prefix and have been in the codebase for years... so maybe we hand-wave that away. 😅

crawl-ref/source/mon-place.cc
```diff
@@ -160,51 +156,78 @@ bool monster_habitable_grid(monster_type mt,
...
+/**
+ * Can monsters of this class survive on actual_grid?
+ *
+ * @param mt the monster class to check against
+ * @param actual_grid the terrain feature being checked
+ * @param wanted_grid if == DNGN_UNSEEN, or if the monster can't survive on it,
+ *                    ignored. Otherwise, return false even if actual_grid is
+ *                    survivable, if actual_grid isn't similar to wanted_grid.
+ */
+bool monster_habitable_grid(monster_type mt,
```
AlejandorLazaro, line 209: Personally, I like this new function a lot and feel like it'd be better to use universally than the function right after, since in the header file you even mention how it has the limitation of `// Can't handle krakens and eldritch tentacles as their habitats are special`.

That said, reading through your changes to `spl-summoning.cc`, I understand the want for the new function and like the concept of determining proper habitable tiles for to-be-summoned creatures. However, I think even the starting spell Summon Mammals shows a potential issue with it—specifically, the fact that flying bats can be summoned from the spell, but quokkas and rats do not fly, but would still be affected by the `flying=true` being used here.

```c++
spret cast_summon_small_mammal(int pow, god_type god, bool fail)
{
    if (!_can_summon_monster(you, HT_LAND, true))  // <-- Only one of the small mammals picks can actually fly
    {
        mpr("There is no available space!");
        return spret::abort;
    }
```

That said, I think this might be the best approach we could take as of now if we want to be able to generically find out if ANY of the possibly summoned creatures would be able to be summoned in open spaces near the caster.

Two things could occur as a result of these checks as I see it:
1. Mass deaths of any summoned HT_LAND creature with no flying on a drowning or lava cell
2. If other checks force only habitable creatures to be summoned, we could force nothing but bats and other flying creatures to be summoned from otherwise random spells

That said, I'm *very* new to the codebase and don't have expertise here, so feel free to ignore this comment if there are other checks or conditions that address those concerns already. 😄

crawl-ref/source/mon-util.h
WizardIke: I've renamed the function to `mons_class_habitat` as it is just the class version of `mons_habitat`.

crawl-ref/source/mon-place.cc
WizardIke: I'm honestly thinking of replacing the `_can_summon_monster` function with a function that takes an array of monster types and returns true if we can summon any of them seeing that's want `_can_summon_monster` is effectively being used for anyway. This probably wouldn't be as efficient, but it would mean that I wouldn't have to expose the hacks used to handle krakens and eldritch tentacles. Same with replacing the `monster_habitable_grid` function that takes a habitat.

Regarding casting summon small mammals with only space available for bats, I haven't changed the behavior here, you will have a chance of getting a bat and a chance of nothing happening. Although it would be nice to force only bats or prompt the player in this situation, that feels to me like a separate change to not letting the player waste a turn and magic points summoning when it would have no effect.

AlejandorLazaro: Yeah, a uselessness check with a prompt *may* make sense, but prompts do feel (personally) so jarring it might be better for the player to just learn under what circumstances a spell makes sense to cast.

If we wanted to get real technical or feature heavy, I can imagine a HUD element and flavor text warning of the spell’s limited usefulness in certain circumstances.

Agreed in that this change is more singular in its purpose, so we can worry about those cases later.

PR #3909, opened 2024-07-05
block entry of > 27 skill target

elliottbernstein:
Skill target gets set to 27 silently when user entered a number greater than 27. This change blocks setting the skill level to 27 and prompts the user to enter a number of 27 or lower in webtiles.


pooka109:
Applied. Thank you.

PR #3914, opened 2024-07-09
Make fair_adjacent_iterator fair

WizardIke:
The old fair_adjacent_iterator would visit the tiles around it in the following order
1 4 6
2 x 7
3 5 8
starting from a random tile and wrapping around when it reached the last tile e.g. if the random start was 4 then 4,5,6,7,8,1,2,3. So imagine for example a yellow ugly thing surrounded by walls and other ugly things trying to pick a random ugly thing around it when changing colour as shown below

w=wall
r=red ugly thing
y=yellow ugly thing
b=brown ugly thing
c=cyan ugly thing
g=green ugly thing

w w r
w y b
w c g

if the random start was an adjacent ugly thing that ugly thing would be selected, however, if the random start was one of the walls the iterator would be advanced until it reached the cyan ugly thing and it would be selected givng the cyan ugly thing a 5/8 chance of being selected and the red, brown and green ugly things 1/8 chance each.

This change makes fair_adjacent_iterator visit the tiles around it in a random order (without repeats)

### DIFF
--- a/crawl-ref/source/coordit.cc
+++ b/crawl-ref/source/coordit.cc
@@ -469,6 +469,83 @@ int distance_iterator::radius() const
     return r;
 }
 
+static void _set_number_at_index(std::uint_fast32_t& numbers,
+    std::uint_fast8_t index, std::uint_fast8_t value) noexcept
+{
+    std::uint_fast32_t mask = 15 << (index << 2);
+    mask = ~mask;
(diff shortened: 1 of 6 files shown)


Hellmonk:
this looks kind of scary to me but I'm going to merge and then revert later if it's messed up

PR #3926, opened 2024-07-18
Don't say that an undead kraken has died.

Aliscans:
A kraken's "death" message referred to "The dead kraken's tentacles" even if it had been undead.

It now omits "dead" in this case, and says that simulacrum tentacles melt away (not "evaporate", like the head, as there's no icy vapour).


pooka109:
Fixed a bit differently.

Killed krakens are now unconditionally no longer referred to as dead, and their tentacles are now described as "disappearing", since that's technically what they do in-game. (References to water and melting don't apply for soui-bound krakens produced by Yred's power, since those fly.)

PR #3929, opened 2024-07-21
Add a note to distinguish explore mode deaths from Felid deaths

DevJac:
I've sometimes practices in explore mode and have wanted to know how many times I've literally *cheated* death. There's no consistent string to search for to identify this. Especially with Felids it is hard to tell which deaths are cheats and which are legit revivals.


pooka109:
It's in. Thank you.

PR #3946, opened 2024-07-31
Update Charlatan's Orb

Monkooky:
The Charlatan's Orb can currently break the 27 skill limit for evocations, and this does not behave correctly for some misc evokers.
The orb was already probably an incorrect choice for a significant majority of characters, and simply capping the boost at 27 would have made this worse.

Instead, we give the orb a more unusual effect of directly boosting skills based on unmodified evocations.

This effect is hopefully desirable enough that a reasonable portion of characters would consider picking up the orb and training more evocations to take advantage of it despite the drawbacks; but not desirable enough that a character would pick it up and train evocations without a single wand or misc evokable to their name.


pooka109:
Applied. Thank you.

PR #3948, opened 2024-08-01
Mosters-spellcasters and -Cast artefacts.

### LINKED ISSUE #3936: monsters with -cast item can cast spells
PLATFORM and OS or BROWSER:
local tiles

DCSS VERSION AND GAME INFORMATION:
0.32-20-1786-g0c0b035568 (latest trunk)

DESCRIPTION:
monsters who are wearing -cast equipment can cast spells.

CipHuK:
Spellcasters drop -Cast artefacts.

Fixes #3936

### DIFF
--- a/crawl-ref/source/mon-cast.cc
+++ b/crawl-ref/source/mon-cast.cc
@@ -4554,7 +4554,8 @@ bool handle_mon_spell(monster* mons)
                 && artefact_property(item, ARTP_PREVENT_SPELLCASTING);
         });
 
-        simple_monster_message(*mons, " takes off the burden that blocked casting.");
+        simple_monster_message(*mons,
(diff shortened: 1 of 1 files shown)


chucksellick:
Would perhaps be better to prevent them receiving these items at all in the first place.

Additionally -- handle_mon_spell is not just used for "magical" spells but also natural and religious abilities which shouldn't be blocked by -Cast anyway. So you need to also check whether they have any spells that *should* be affected by -Cast before doing anything about this.

Hellmonk:
replaced almost all -cast with *silence and made it work for monsters. If someone gets a spellcasting monster with mad mage's mauler's feel free to open a new pr and I'll fix it.

PR #3952, opened 2024-08-02
remove auto-hit flag

Monkooky:
this flag is entirely redundant; it is true only when aimed_at_feet is and checked only when aimed_at_feet is


Implojin:
Would you mind force pushing this with an updated commit title before I windmill slam it?

Implojin:
Thanks! :)

PR #3954, opened 2024-08-04
Remove 'attack previous target' shout

Monkooky:
'Attack previous target!' appears in the shout menu if you have a target, and they are alive in line of sight. This could theoretically make ordering allies smoother if e.g. you want to order your allies to attack, then order a retreat, then repeat the attack order. However the current implementation has flaws that ruin even this narrow use case, and buggy enough behaviour that it would be better removed than present. 

-The 'Attack previous target!' shout can be issued even when not displayed on the menu; if you have a previous target alive but out of LOS it will command allies to target them. If it's not displayed for other reasons (your previous target is a tile or nothing) then it is treated as an attack shout. This leaks information, but is mostly just strange and buggy looking.

-What your previous target is is obtuse. Your target is changed when you perform any action which places a targetter, along with some other conditions. For example, firing past a monster, casting apport, or ordering a retreat will clear your previous target.

This commit removes the 'previous target' shout entirely instead of trying to fix it.


NormalPerson7:
Sounds good to me.

Given the number of extra commits on this branch, including some other feature commits and fixes that aren't the subject of this PR, you may want to make a new branch off `master`, cherry-pick 667f46e, and make a fresh clean PR. I'm suggesting this because other causes of action could risk losing some of the other work you've done here.

DracoOmega:
I pulled out the relevant commit myself and merged it. Thanks (and good riddance to this thing).

PR #3955, opened 2024-08-04
Remove undocumented 'wait here' command

Monkooky:
The behaviour of the undocumented 'wait here' shout is clearly similar to 'guard this area'; but the different parameter for _set_friendly_foes obfuscates that they are identical.

This makes the code harder to parse with no real gain to gameplay.


pooka109:
Applied. Thank you.

PR #3968, opened 2024-08-09
Improve malign gateways targeter

### LINKED ISSUE #3910: FR: Malign Gateway targetter shows you where it might appear
webtiles

trunk

**DESCRIPTION:**
The Malign Gateway targetter highlights every non-adjacent square in LoS. However, the actual squares where the gateway may appear are chosen by different logic (non-adjacent, on cardinal directions, surrounded by empty space). It would be nice to display this so the player can tell where the gateway might appear.

WizardIke:
Instead of highlighting all spaces that are more than one space away from you, it now only highlights spaces that the gateway might be summoned at.

It currently doesn't take into account magic mapped terrain (this is the same as airstrike).

should close #3910

Screenshot (1887)
Screenshot (1888)
Screenshot (1889)

### DIFF
--- a/crawl-ref/source/spl-cast.cc
+++ b/crawl-ref/source/spl-cast.cc
@@ -1392,9 +1392,12 @@ unique_ptr<targeter> find_spell_targeter(spell_type spell, int pow, int range)
     }
     case SPELL_FOXFIRE:
         return make_unique<targeter_maybe_radius>(&you, LOS_NO_TRANS, 1, 0, 1);
-    // TODO: these two actually have pretty wtf positioning that uses compass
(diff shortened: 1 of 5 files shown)


regret-index:
The actual gateway placement behaviour of solely looking for the eight directions itself probably should be changed to place in potentially all squares in sight not adjacent to walls, rather than secretly relying on such a strange eight-directions targeting in the first place.

WizardIke:
@regret-index I created a new PR where malign gateway can target non-compass directions

Hellmonk:
Merged the newer of the pull requests, so closing this one. Thanks.

PR #3980, opened 2024-08-15
Add a whole new generator for shop names

chucksellick:
Since the beginning of time, shop names have been one thing that stayed pretty uninspiring in a game otherwise chock full of very colourful and varied procedural text generators for other random game elements.

The old generator consisted of:
"<name>'s <type> [Shoppe|Boutique|Emporium|Shop]"

And that was it.

This commit fixes all that with a brand new and far more colourful generator largely consisting of a nearly 3000-line textdb file.

Now many synonyms are used for different shop contents, words for "shop", as well as adding a couple of entirely new paths for shopkeeper names that are rather more memorable than the classic make_name() which is used everywhere (e.g. Pan Lords, artifacts, ghosts...) There are a few grammatical variations on the name structure (e.g. Dave's Weapon Shop vs Weapons By Dave vs Ye Olde Weapon Shop).

Names start off fairly straightforward in the early levels but increase in variety and eccentricity based on dungeon level (or player XL for Gozag shops) by varying the length of the generated names, both upwards and downwards. So you end up with very baroque names as well as very short ones.

There are enough permutations that no two names are ever likely to be similar over a single game, although if by chance they occasionally are it's enough of a funny coincidence and just makes it look like the shops are competitors or perhaps a chain.

This is also something of a flavour enhancement for Gozag as they get some special handling and rather more extravangant shopkeeper names. (Shopkeepers are now considered always neutral gender for the purposes of welcoming you to their shop: with some of the shop owners now being *multiple* persons it didn't make sense to try and keep track of this...)

Vault-defined shop names are left alone and will resort to the old generator where they are partially defined.

Support for Gadget shops is included in expectation of them being reinstated in a separate commit.

A "shopnames.lua" script is also included to generate a large number of shop names for different levels and different shop types to test the generator. A sample dump is attached to showcase the huge variety of names.

[shopnames_test.txt](https://github.com/user-attachments/files/16629334/shopnames_test.txt)


regret-index:
I do like the spirit and direction of this PR, to be quite clear- it is indeed true that regular shops were always quite plain relative to quite a few other things in Crawl. I would have liked to get this handled pretty quickly even as a purely flavour-based reworking, Unfortunately, after having built the branch and looked through hundreds of, I'm rather worried about trying to cram this in this close to version release. To explicitly list off my compunctions:

* Shop names here can roll a few too many descriptors, enough that it loses individual effective meaning, and especially when an individual shop is hardly guaranteed to actually be of interest. I'd prefer to trim out some of the longest word choices in the total selection (or make them be chosen in separate conglomerations themselves) plus the name arrangements to only choose one or two of shopkeeper-adjectives, [& Daughter / & Family], shop adjectives, antique-shop adjectives. It's otherwise a mess to see even e.g. four shops at once in Orc when they can have names like "Weighty Uhyuqodd's Vintage Waivered Bits & Pieces Museum", "Monosyllabic Allface's Artisan Organs of Battle", "Ugasha's Firsthand Back Alley Pawn House O' Home Wares", and "Dick the Fluffy's Brilliantly Crafty Weaponry for Sale". (Coglin gizmo names have gotten unwieldy, but nowhere near this kind of length.)

* While Crawl is inherently in a generic medieval-fantasy setting, nothing in Crawl besides Wiglaf's deep Scottish accent uses "Ye" or "Yon", and while this is hardly unknown for current-day establishments, it feels weird to use more obviously and prominently. 

* Choice-of-words-wise, there's a lot of confusable pieces and questionable innuendos among the actual parts use to composed these names. "Milk" as a descriptor for potions is strange, calling an entire book store an "encyclopedia" reads weirdly when that could read easily as a singular book, jewellery shouldn't be called "gems" when that's an extremely specific system prominent elsewhere in the game, and for all the other innuendos present in the game I have to draw a line at being able to produce "Icky Dicky" as a name.

Early in 0.33, if the first two points are handled, I can reasonably do the editing for the third point myself. I'm sorry to say I'm reluctant to try and quite fit it in otherwise.

chucksellick:
@regret-index Your feedback largely aligns with aspects of the generator I wasn't quite sure about (I wasn't quite sure how far was "too far" with the general tone and some of the more eccentric stylings). In some cases I was just trying to find as many words that could fit as possible and/or cover as much of the alphabet as I could for starting letters, but yeah this resulted in some odd fits and certainly was planning to cut/improve some stuff further.

Personally I felt "Ye Olde Shoppe" is such a classic fantasy joke I wanted to have a chance to generate it (and we do already have "Shoppe" leaning into such archaic spellings) and maybe this form just needs to be much rarer and used in less places?

So I'll do another couple of passes and trim things back a lot and see where we're at, this feedback helps a lot. Thanks

PR #3981, opened 2024-08-15
Don't let non-damaging walls or clouds stop running.

Aliscans:
Let the player run (shift+dir) through safe clouds.

Let the player run up to slimy walls if immune to their effects.

In both cases, the player could already travel through an affected square.


DracoOmega:
Merged, thanks.

PR #3984, opened 2024-08-17
Add vaults; three arrival, three ecumenical, one ice cave portal

lexi-h:
More vaults! Learning syntax and style guidelines. 

Feedback is much appreciated! I'd like to ensure the basics are sane before I start cooking up more complex vaults. Specifically I'm most curious on my handling on the muddy ecumenical altar; unsure if I did that in the "best"/cleanest way possible.

### DIFF
--- a/crawl-ref/source/dat/des/altar/ecumenical.des
+++ b/crawl-ref/source/dat/des/altar/ecumenical.des
@@ -1272,7 +1272,7 @@ ENDMAP
 
 NAME:     angelslayer_ecumenical_altar_muddy
 TAGS:     transparent ruin
-NSUBST:   D = 1:t / 1:_ / 3:F / *:.
+NSUBST:   D = 1:_ / 1:t / 3:F / *:.
 NSUBST:   E = 1:t / 5:F / *:.
 KMONS:    F = plant
 FTILE:    t_.DEF = floor_mud
--- a/crawl-ref/source/dat/des/altar/ecumenical.des
+++ b/crawl-ref/source/dat/des/altar/ecumenical.des
@@ -1242,3 +1242,16 @@ xxxGx._.xGxxx
 xxx..xxx..xxx
  x    x    x
 ENDMAP
+
+NAME:   angelslayer_ecumenical_altar_simple_liquid
+TAGS:   transparent no_monster_gen no_pool_fixup
+SUBST:  W : W:10 w:5 l:1 
+: ecumenical_altar_setup(_G)
+MAP
+.WWW.
+W...W
+W._.W
+W...W
+.WWW.
+ENDMAP
+
--- a/crawl-ref/source/dat/des/altar/ecumenical.des
+++ b/crawl-ref/source/dat/des/altar/ecumenical.des
@@ -1255,3 +1255,17 @@ W...W
 .WWW.
 ENDMAP
 
+NAME:     angelslayer_ecumenical_altar_broken_walls_simple
+TAGS:     ruin transparent
+SHUFFLE:  _E
+FTILE:    ._Ec = floor_crypt
+TILE:     c = wall_catacombs
+SUBST:    E = G:10 V:5 A:3 .:3
+: ecumenical_altar_setup(_G)
+MAP
+ccccccccc
+c.......c
+c._...E.c
+c.......c
+cc.cccccc
+ENDMAP
--- a/crawl-ref/source/dat/des/altar/ecumenical.des
+++ b/crawl-ref/source/dat/des/altar/ecumenical.des
@@ -1245,7 +1245,7 @@ ENDMAP
 
 NAME:   angelslayer_ecumenical_altar_simple_liquid
 TAGS:   transparent no_monster_gen no_pool_fixup
-SUBST:  W : W:10 w:5 l:1 
+SUBST:  W : W:10 w:5 l:1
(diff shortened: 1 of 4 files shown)


regret-index:
Hi! Thank you for your submission. Sorry this took a while to get to- this submission was in the middle of feature freeze, so actually attending to it was going to inevitably take a while since it would have to wait until after that plus the tournament were over.

There's a couple of minor tweaks that should be done before this vault submission goes in, all mostly for the sake of consistent house style (which I really ought to get around to writing into the docs themselves) before anything else:
* The header order I've been trying to standardize vaults across the game towards usually goes NAME -> TAGS -> DEPTH / PLACE -> ORIENT -> MONS / KMONS -> ITEM / KITEM -> KFEAT -> TILE / FTILE -> shared Lua functions. Alongside adjustments in each of the vaults, this'll also mean the FTILE listing for `angelslayer_ecumenical_altar_broken_walls_simple` will need to list the arch / statue / dry fountain AGV glyphs.
* For `angelslayer_ecumenical_altar_muddy`, it's easier to parse vaults if they don't use capital letter glyphs for sections of vaults that primarily exist as potentially floor features (unless they're part of randomization that could). Since `.`s are floor, this usually goes towards other non-door punctuation marks- here, I'd use probably `?`s or `"`s for the outside glyphs, and `-`s for the inside glyphs. Also, the FTILE line doesn't need the pre-randomization glyphs in its definition.
* In `angelslayer_arrival_pockets_within_ring`, SUBST lines can and should squash together glyphs all doing the same redefintions- the lines could thus be compressed down into `SUBST: ( = >, [H = ., DEF = x, T = TV`.
* `angelslayer_arrival_single_sparkle` eats up a lot more lines in the .des file than it should by being extremelly tall and relatively thin- it's not a super high priority, but Crawl will already rotate and mirror vaults by default, so the vault definition in the code should probably be wide instead of tall. This is most easily done by using a text-art rotation utility, like REXPaint, or [some other old tools](http://xom.montres.org.uk:3333/).

Also, for future reference, vault submissions usually should squash their commits together into one for the sake of a submission. Individual vault additions getting their own individual commits is mostly for large and obvious features, like new branch ends or portal layouts. I can handle that part myself, of course.

(None of this is _entirely_ a blocker to having a vault submitted- if a given vault creator gives their consent, I can handle this sort of thing myself, as incredibly little of it has to do with the actual contents of a vault. It does, however, inevitably make it take a bit longer for me to get around to given vault PRs, as I keep writing up these sorts of things.)

lexi-h:
Wonderful, thank you for the feedback! I figured the delay was feature-freeze related. I'll work on this once the tournament is over ^^

I have been wanting to build some large branch-end vaults (I'll probably start with Lair), so it's good to learn the proper way to do things on the small stuff. Are there other text-art utilities like REXpaint?

lexi-h:
>     * The header order I've been trying to standardize vaults across the game towards usually goes NAME -> TAGS -> DEPTH / PLACE -> ORIENT -> MONS / KMONS -> ITEM / KITEM -> KFEAT -> TILE / FTILE -> shared Lua functions. Alongside adjustments in each of the vaults, this'll also mean the FTILE listing for `angelslayer_ecumenical_altar_broken_walls_simple` will need to list the arch / statue / dry fountain AGV glyphs.

Fixed to my knowledge, but I'll let you shuffle around the header order. I'm unsure of the priority (since no exhaustive list was provided).

>     * Also for `angelslayer_ecumenical_altar_broken_walls_simple`, `wall_catacombs` is used as for rock walls in three vaults and as stone for only one vault, so I'd prefer it to only be used for rock walls (and to fix the one other use of such to fit)- you can just use `x`s instead of `c`s in that vault.

Fixed!

>     * For `angelslayer_ecumenical_altar_muddy`, it's easier to parse vaults if they don't use capital letter glyphs for sections of vaults that primarily exist as potentially floor features (unless they're part of randomization that could). Since `.`s are floor, this usually goes towards other non-door punctuation marks- here, I'd use probably `?`s or `"`s for the outside glyphs, and `-`s for the inside glyphs. Also, the FTILE line doesn't need the pre-randomization glyphs in its definition.

Fixed!

>     * In `angelslayer_arrival_pockets_within_ring`, SUBST lines can and should squash together glyphs all doing the same redefintions- the lines could thus be compressed down into `SUBST: ( = >, [H = ., DEF = x, T = TV`.

Fixed!

>     * `angelslayer_arrival_single_sparkle` eats up a lot more lines in the .des file than it should by being extremelly tall and relatively thin- it's not a super high priority, but Crawl will already rotate and mirror vaults by default, so the vault definition in the code should probably be wide instead of tall. This is most easily done by using a text-art rotation utility, like REXPaint, or [some other old tools](http://xom.montres.org.uk:3333/).

Fixed!

> Also, for future reference, vault submissions usually should squash their commits together into one for the sake of a submission. Individual vault additions getting their own individual commits is mostly for large and obvious features, like new branch ends or portal layouts. I can handle that part myself, of course.

Noted!

> (None of this is _entirely_ a blocker to having a vault submitted- if a given vault creator gives their consent, I can handle this sort of thing myself, as incredibly little of it has to do with the actual contents of a vault. It does, however, inevitably make it take a bit longer for me to get around to given vault PRs, as I keep writing up these sorts of things.)

Thank you for your time! o7 

Let me know if there's anything else.

PR #3986, opened 2024-08-19
Fix stash search prefixes for golden dragon scales

seandewar:
Presumably broke after the rename from gold -> golden.

I also noticed test/stress/qw.rc has the old name still, but it already seems to have been fixed in the qw repo, so I left that alone. (The file seems rather out-of-date anyway; it generates other errors as a result)


PleasingFungus:
Looks good, thanks!

PR #3988, opened 2024-08-19
Android: Add a full screen option to the launcher

robertxgray:
I got this message from an Android user:
    
> can't enable full screen mode by setting "tile_full_screen = true",
> still shows a black bar on top with device status.
    
I agree a full screen option for Android can improve the user
experience. Mostly when playing on landscape mode with modern devices.
    
You can read this in the docs:
    
> Where you set the UI flags makes a difference. If you hide the system
> bars in your activity's onCreate() method and the user presses Home,
> the system bars will reappear. When the user reopens the activity,
> onCreate() won't get called, so the system bars will remain visible.
> If you want system UI changes to persist as the user navigates in and
> out of your activity, set UI flags in onResume() or
> onWindowFocusChanged().
    
Tried both options and the behaviour when using onWindowsFocusChanged
is more consistent. When using onResume, full screen is lost whenever
you open the top menu.

Screenshot_20240820_102151

### DIFF
--- a/crawl-ref/source/android-project/app/src/main/java/org/develz/crawl/DCSSLauncher.java
+++ b/crawl-ref/source/android-project/app/src/main/java/org/develz/crawl/DCSSLauncher.java
@@ -10,6 +10,7 @@ import android.util.Log;
 import android.view.View;
 import android.widget.AdapterView;
 import android.widget.ArrayAdapter;
+import android.widget.CompoundButton;
 import android.widget.EditText;
 import android.widget.Spinner;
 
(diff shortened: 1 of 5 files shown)


robertxgray:
After some testing, the option has been renamed to "Full screen" and it's enabled by default.

Also, the flags are now set on onWindowsFocusChanged, where the behaviour is more consistent. When using onResume, full screen is lost whenever you open the top menu.

robertxgray:
The full screen option is already available on Bloatcrawl 2.

https://play.google.com/store/apps/details?id=org.develz.bloatcrawl2

PR #3991, opened 2024-08-23
Update player doll to visually lose wings when big wings mutation is lost (Resolves #3973)

### LINKED ISSUE #3973: Draconian wings cosmetic bug
Version: trunk 
Dungeon Crawl Stone Soup version 0.32-a0-2080-gfe854bf2fb (webtiles) 

Problem: gained wings mutation from potion on pale draconian. Removed it, but the "wings" on the paperdoll were still on. After saving and reloading, the paperdoll correctly doesn't have wings anymore.

AlejandorLazaro:
## Summary

Draconian wings would remain on player dolls until saving and reloading if they had lost their big wings mutation. This change has the loss of the big wing mutation initialize the player doll so we properly lose the wings when this occurs.

Reported by [kuniqsX](https://github.com/kuniqsX) and resolves [#3973](https://github.com/crawl/crawl/issues/3973)

## Changes

* Added `init_player_doll();` with a `USE_TILE` def compiler flag under the `MUT_BIG_WINGS` switch case for the following code:
  * `crawl-ref/source/mutation.cc`
    * `static bool _delete_single_mutation_level(mutation_type mutat, ...`

## Investigation

Tested with the committed code change alongside the *NOT* committed code changes to `potion.cc` for easy reproduction:
```c++
    bool effect(bool = true, int = 40, bool = true) const override
    {
        if (have_passive(passive_t::cleanse_mut_potions))
            simple_god_message(" cleanses your potion of mutation!");
        else
            mpr("You feel extremely strange.");
        bool mutated = false;

        static bool true_to_add = true;

        // Remove mutations.
        if (!true_to_add)
        {
            mprf(MSGCH_DIAGNOSTICS, "Removing MUT_BIG_WINGS");
            mutated |= delete_mutation(MUT_BIG_WINGS, "potion of mutation", false);
        }
        // Add mutations.
        else
        {
            mprf(MSGCH_DIAGNOSTICS, "Adding MUT_BIG_WINGS");
            mutated |= mutate(MUT_BIG_WINGS, "potion of mutation", false);
            learned_something_new(HINT_YOU_MUTATED);
        }
        true_to_add = !true_to_add;
        return mutated;
    }
```
1. Starting out with no wings on a draconian
Screenshot 2024-08-23 at 6 40 09 PM

2. Wings appear on the player doll after mutating and gaining the big wings mutation
Screenshot 2024-08-23 at 6 40 38 PM

3. Wings properly disappear after removing the big wings mutation
Screenshot 2024-08-23 at 6 40 50 PM

### DIFF
--- a/crawl-ref/source/mutation.cc
+++ b/crawl-ref/source/mutation.cc
@@ -2570,6 +2570,9 @@ static bool _delete_single_mutation_level(mutation_type mutat,
 
     case MUT_BIG_WINGS:
         land_player();
+#ifdef USE_TILE
+        init_player_doll();
+#endif
         break;
 
     case MUT_HORNS:


chucksellick:
This seems like a fairly uncontroversial bugfix (and it mirrors the same switch case in `mutate(...)`)

Digging into what init_player_doll() actually does, it's slightly odd that it reads dolls.txt every time it's called; otherwise it would be a very light function and we could just call it at the start of every player turn and catch any anomalies like this. But that would just be a good refactor, I don't think it should stop this fix going in as is.

PR #4016, opened 2024-09-08
don't crash if shadow can't be placed (ChongLi)

geekosaur:
If a player is surrounded by monsters to depth 3+, `_get_shadow_spots` produces an empty vector and the game crashes trying to dereference the first entry. Check for an empty vector first.
(https://cbro.berotato.org/morgue/ChongLi/crash-ChongLi-20240908-132253.txt)


Implojin:
Thanks for the fix!

PR #4032, opened 2024-09-16
Fix HP/MP not updating when (un)equipping the Charlatan's Orb

### LINKED ISSUE #4015: Charlatan's Orb fails to adjust max HP correctly, and skill boost has missing text
0.32.0 CXC (sidenote: CXC has not been updated to the latest version of 0.32, meaning the shield bug remains in place at the very least on that server).

Charlatan's orb bonus displays "Shields enhanced by  are displayed in green" - needs text there

image

NormalPerson7:
Make the Charlatan's Orb explicitly update the player's HP and MP in its equip/unequip functions.

Also add the Charlatan's Orb as a cause of increased skills on the skills screen to fix the previous text
`Skills enhanced by  are in green.`

Resolves #4015. (Forgot to mention in commit message, sorry)


Implojin:
Looks good to me, thanks for the fix!

PR #4034, opened 2024-09-16
Don't auto-hide currently unmemorisable spells when picking up books

### LINKED ISSUE #3994: Spellbooks marked as useless while worshipping Okawaru do not give you spells after you abandon him.
Dungeon Crawl Stone Soup version 0.33-a0-5-g34d56c8fa2 (webtiles)
- Playing as formicid gladiator of Okawaru.
- Abandon Okawaru for TSO
- Abandon TSO for Cheibriados
Problem: spellbooks which had spells forbidden by Okawaru (i.e. Animate Dead, Death Channel) did not give me those spells when I was no longer a follower of Okawaru. They give the "Unfortunately, you learn nothing new" message as with books containing spells you already know.
Those spells are effectively forbidden to me in that game.

NormalPerson7:
Previously, if you worshipped Okawaru (for example), spells which could not be memorised would be silently added to your library and also hidden in your library.

Spells that can't be memorised do not appear on the spell library interface, so this had no actual effect as long as the spell could not be memorised. But, say, if you then abandoned Okawaru, you would find all your previously unmemorisable spells hidden in the library (so that, to unhide and memorise them, you would have to visit the "Show" screen).

In practice, this meant many players would just assume the spell was lost forever.

Resolves #3994.

### DIFF
--- a/crawl-ref/source/spl-book.cc
+++ b/crawl-ref/source/spl-book.cc
@@ -427,10 +427,9 @@ bool library_add_spells(vector<spell_type> spells, bool quiet)
         if (!you.spell_library[st])
         {
             you.spell_library.set(st, true);
-            bool memorise = you_can_memorise(st);
-            if (memorise)
+            if (you_can_memorise(st))
                 new_spells.push_back(st);
-            if (!memorise || Options.auto_hide_spells)
+            if (Options.auto_hide_spells)
                 you.hidden_spells.set(st, true);
         }
     }


Implojin:
IMO this seems fine; it defeats the intent of be7b7cd180 but the player confusion here is probably reason enough to just default to not-hiding spells unless the player is specifically requesting we do so through the rc option.

Gonna leave this PR open for just a little bit in case anyone else has comments but otherwise I'll merge later tonight.

Implojin:
Thanks for the patch!

PR #4035, opened 2024-09-16
Forbid Cheibriados coglins from making rampaging gizmos

### LINKED ISSUE #4025: Coglins can invent unusable gizmos
Webtiles

0.32

**DESCRIPTION:**
Coglins can invent Rampage artefacts while worshipping Cheibriados. This makes the gizmo unusable. This behaviour should be made consistent with acquirement mechanics, where the item is guaranteed to be usable.

**EXTRA FILES:**
https://media.discordapp.net/attachments/317641873709596693/1284219680373866629/image.png?ex=66e5d647&is=66e484c7&hm=444f8476258582e9628fb4f93b1d51a12a9ea9c35acf103f36838b2d1309860b&=&format=webp&quality=lossless

NormalPerson7:
It feels a bit lame to get a gizmo that is forbidden by your god, and we should assume thematically that coglins have some control over the gizmos they create.

Chei worshippers will now just get Acrobat on gizmos that would otherwise have Rampage/Acrobat. These gizmos will be a bit weak, but I figure we probably shouldn't tweak anything else about gizmo properties for Chei worshippers.

[I'd love a better way than just explicitly referencing Cheibriados here, but I am not aware of a suitable function.]

Resolves #4025.

### DIFF
--- a/crawl-ref/source/artefact.cc
+++ b/crawl-ref/source/artefact.cc
@@ -2035,7 +2035,9 @@ static void _apply_gizmo_prop(item_def& gizmo, gizmo_prop_type prop)
             break;
 
         case GIZMO_RAMPAGE:
-            artefact_set_property(gizmo, ARTP_RAMPAGING, 1);
+            // Don't give rampage to Chei worshippers
+            if (!you.religion == GOD_CHEIBRIADOS)
+                artefact_set_property(gizmo, ARTP_RAMPAGING, 1);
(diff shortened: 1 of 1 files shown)


DracoOmega:
I opted not to give Chei worshippers 'half' of a gizmo prop (which I thought could feel kind of sad) and instead not generate this major property for Chei worshippers altogether. But thanks for reminding me about this!

PR #4037, opened 2024-09-17
Fix crash in Xom's wave of despair effect

seandewar:
If `place_monster_corpse` returns nullptr, the game crashes when trying to call `turn_corpse_into_skeleton`.

Possible when placing a skeleton on a terrain feature that destroys items, like lava. (`move_item_to_grid` destroys the would-be corpse item)

Example crash log: https://underhound.eu/crawl/morgue/seandewar5/crash-seandewar5-20240902-225318.txt

### DIFF
--- a/crawl-ref/source/xom.cc
+++ b/crawl-ref/source/xom.cc
@@ -2878,9 +2878,11 @@ static void _xom_wave_of_despair(int sever)
             dummy.position = *di;
 
             item_def* corpse = place_monster_corpse(dummy, true);
-            turn_corpse_into_skeleton(*corpse);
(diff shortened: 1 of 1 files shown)


regret-index:
(...Sorry your name isn't on the commit- I was trying out Github's PR merging options and didn't realize that it'd override the author after auto-wrapping the crash log url unless I specified otherwise. Thank you for submitting this PR and fixing the bug.)

seandewar:
Thanks and no problem! Seems it was added via a merge commit, which references the commit from my branch as a parent anyway. :+1:

PR #4039, opened 2024-09-19
Take another pass at the mage background descriptions

NotAJumbleOfNumbers:
I'm gonna be honest, some of the current descriptions say basically nothing about the background itself.  "Confront problems with damaging spells"?  "Provides them with the flexibility to survive"?  I thought I could do better, so here's a first pass at trying to do so.

### DIFF
--- a/crawl-ref/source/dat/descript/backgrounds.txt
+++ b/crawl-ref/source/dat/descript/backgrounds.txt
@@ -1,8 +1,8 @@
 %%%%
 Air Elementalist
 
-Air Elementalists have learned the Shock spell. Later on, their selection of
-Air Magic spells provides them with the flexibility to survive.
+Air Elementalists have learned the Shock spell. Their library of spells are 
+jittery and sometimes uncontrollable, but very effective.
 %%%%
 Artificer
 
@@ -30,7 +30,8 @@ there is nothing left.
 %%%%
 Conjurer
 
-Conjurers confront problems with damaging spells.
+Conjurers confront problems with reliable spells. Their starter Magic Dart spell 
+is an attack that always hits.
 %%%%
 Delver
 
@@ -55,7 +56,7 @@ choice and a potion of might.
 Fire Elementalist
 
 Fire Elementalists have learned the Foxfire spell. Their other starting spells
-are likewise focused on fiery destruction.
+are likewise focused on fiery, explosive destruction.
 %%%%
 Gladiator
 
@@ -99,8 +100,8 @@ when in need. They start with the Kiss of Death spell.
 %%%%
 Summoner
 
-Summoners are able to cast Summon Small Mammal from the start, and begin with
-other Summonings spells in their library.
+Summoners are able to cast Summon Small Mammal from the start, and can
+summon deadlier beings further down the line.
 %%%%
 Shapeshifter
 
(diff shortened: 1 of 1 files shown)


crawl-ref/source/dat/descript/backgrounds.txt
```diff
@@ -109,8 +110,8 @@ fighters (especially when unarmed) at the cost of armour.
 %%%%
 Alchemist
 
-Alchemists start with the Sting spell, and their magic is very effective against
-foes susceptible to poison.
+Alchemists start with the Sting spell, and their wide range of effects
+can harm a variety of foes.
```
regret-index, line 114: While Alchemists do have very different endspells for a longer while than most starting books, "harm a variety of foes" reads very weirdly when nearly all starting damage spells can hurt most early game foes. I'm not sure what uses of words like "condition", "circumstances", and other such words can properly emphasize their specific limitations to both rPois and the risks of Sticky Flame, but it'd probably be better than going from mentioning they're one of the most conditional ones to vaguely implying they can hurt many things.

NotAJumbleOfNumbers:
Okay, I rewrote those two.  Does this sound better?

chucksellick (commented):
My take on reading this is that these new descriptions mostly don't really tell me anything more than the old ones and in some cases even a bit less (or at least leave me with more questions).

Adding Magic Dart to the Conj description looks like a good idea as it makes it consistent with all the other backgrounds, but I would follow their template and mention it as the very first thing.

In general however, is knowing the starting spell *actually* that useful? We could assume the new player doesn't know what any of the spells do anyway. Maybe it'd be a good policy to not even mention it and use those words to write a bit more about what *gameplay* to expect with a background. I feel like Hexslinger and Enchanter are both pretty good descriptions without actually having to mention a specific spell. I should note as well there is some space to play with; Monk is the longest description and fits exactly on two lines in console, so these can all be a bit longer.

Looking through more descriptions, several more all suffer from the same problems: Hedge Wizard, Warper, Ice Elementalist, Necromancer: these all tell me very little.

Probably the best written existing description is Earth Elementalist:

```
Earth Elementalists know the Sandblast spell, and can later learn more earthen
spells suitable for subtle stabbing or brute-force blasting.
```

This gets across a lot of information very succinctly, and nicely lets me know what to expect.

chucksellick:
Just suggesting some further things that could be worked into some of the backgrounds to give more of a flavour of gameplay. (Writing this i realised just how hard to is to distill things into such short sentences, as most of the schools do a lot of different things!)

Air Magic:
"jittery and uncontrollable" isn't particularly accurate when one half of the spellbook is Swiftness and Airstrike and it's also very open to interpretation.
How about something like "Air Elementalists attempt to control the untamed forces of weather, utilising electrical magic to harm groups of foes, and harnessing the wind to assist them."

Alchemist:
Unusual forms of damage such as time-based poison.

Hedge Wizard:
Hedge Wizards are magic generalists, good at using a wide variety of tools to frustrate their foes and gain an edge in the battlefield.

Ice Elementalist:
Another one that is just completely vague: "The spells in their starting library are quite versatile."
"ice magic is good at hindering foes as well as offering protection, but can restrict the movement of the caster"

Necromancer:
I think everyone knows what necromancy *means* but this spends a whole sentence explaining it anyway without saying anything about gameplay.
"Necromancers can turn the dead into allies to fight alongside them, and use the life force of the living to regenerate themselves."

Hellmonk:
Merged some changes based on pr and feedback, closing.

PR #4051, opened 2024-09-23
Let the user inscribe nearby items on the floor. Fix some bugs.

Aliscans:
These are separate changes (apart from the last two), but I've put them all together here as some of them touch the same functions.

1. (fix) Don't let the player use off-level items with CMD_DISPLAY_MAP.
2. (fix) Don't use item.index() for items outside env.items (there's a warning about this in item-def.h).
3. (fix) If two waypoints are in the same place, searching for * no longer repeats items there.
4. (fix) Don't look at inventory if the player searches for 0-9 or *. This means that the search terms just look at waypoints.
5. (feat) Let the examine commands (Xv and xv) be used to inscribe items.
6. (feat) Let the find command (Ctrl-F) be used to inscribe items (or for any other interaction xv would normally permit).


DracoOmega:
Thanks for finding that *incredible* bug about manipulating things on other levels. I've merged the fix for that one directly, since it seemed high priority.

I'll take a closer look at the rest of the PR later, but for the moment I want to note that is_floor_item() in 45a2b570796f77e30dd9a will also return true for many things that aren't actually on the floor. It's probably unexpected to most people, but the way Crawl handles items means that the inventory of all shops on the floor and also all items in the inventories of all monsters on that floor  are also stored in env.items, so at the very least the function should be named something else. (But also, while being able to use ::index() in an invalid way isn't great, I'm a little hesitant on making the code perform a linear search of the entire floor's inventory every time you want to check an item's index.)

Aliscans:
The name `is_floor_item()` is probably more a reflection of how I was using it than anything. I don't think there's a way to inspect monsters' inventory, and it doesn't catch shop items as they're in `shop_struct::stock` rather than `env.items`. I'm thinking that `floor_item()` may be better off inside `item_def` as it operates on a specific `item_def` and `item_def::index()` calls it. Some other name would be useful if it did move, but I don't know what wuold work best.

I'm not sure what linear search you are referring to. I know `is_floor_item()` just calls `FixedVector::begin()` and `FixedVector::end()`, neither of which should need to do extra work for large arrays.

PR #4060, opened 2024-09-27
Fix compilation against modern vendor-neutral OpenGL implementation.

vanfanel:
Adds the GLVND makefile parameter and corresponding defines so the game can be built on GNU/Linux against modern vendor-neutral OpenGL implementation supported by MESA, instead of having to rely on deprecated GLX which is X11 only.

Pass `GLVND=y` to build against modern OpenGL, default is for now using legacy GLX implementation.

### DIFF
--- a/crawl-ref/source/Makefile
+++ b/crawl-ref/source/Makefile
@@ -788,7 +788,14 @@ ifneq ($(uname_S),Darwin)
 ifeq ($(msys),Yes)
 LIBS += -lopengl32 -lglu32
 else
-LIBS += -lGL -lGLU
+
+ifdef GLVND
+LIBS += -lOpenGL   # Use modern vendor-neutral OpenGL
+DEFINES_L += -DUSE_GLVND
+else
+LIBS += -lGL -lGLU # Use legacy GLX
+endif
+
 endif
 endif
 endif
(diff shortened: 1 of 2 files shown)


Cgettys:
@vanfanel  - we removed the glu dependency recently. I think that makes the crawl-ref/source/glwrapper-ogl.cc part of this change unnecessary, but do you want to rebase the Makefile change?

vanfanel:
> @vanfanel - we removed the glu dependency recently. I think that makes the crawl-ref/source/glwrapper-ogl.cc part of this change unnecessary, but do you want to rebase the Makefile change?

I'm trying to do so, but now on my system `libOpenGL.so` lives in `/usr/local/lib64`, and the Makefile is not looking there unless I manually add `-L/usr/local/lib64` which is not a proper solution.

If I build with `make V=1` I see these are the libs and paths passed to the linker:

`-lz -lSDL2_image -L/usr/local/lib  -lSDL2   -lfreetype  -lOpenGL    -Lcontrib/install/x86_64-linux-gnu/lib contrib/install/x86_64-linux-gnu/lib/liblua.a contrib/install/x86_64-linux-gnu/lib/libsqlite3.a -pthread`

...as you can see there, `-L/usr/local/lib` is passed, but not `-L/usr/local/lib64`.
Where does this  `-L/usr/local/lib` come from? I've been trying to read the Makefile and find out but it's SO big...


Normally, on my system I can do:

`gcc example.c -lOpenGL` and the linker finds OpenGL as required.

Cgettys:
Hmm, I'm not seeing that behavior on my Pop_Os 24.04 machine... I'm fairly baffled.

PR #4069, opened 2024-10-02
Allow Okawaru gift reroll, at immense piety cost

anderspapitto:
Mechanically: add an ability that clears the existing set of gift offerings, at a cost in piety that is guaranteed to drop standing below 6*. If neither gift has been accepted yet, both will be rerolled. If one gift has been accepted, only the other is rerolled, and if both have been accepted then this ability does not exist.

Motivation: This is a hybrid of old and new gifting behavior. The bad playfeel of "unique, once-per-game capstone options are all garbage" is removed. The player can choose to wait for something better, like in the old system. However the gifts are still unique, and there is no long-term piety-draining trickle of ignorable gifts once a capstone has been selected.

Thematically: Rejecting Okawaru's offerings is a touchy topic to navigate, and will cause offense, hence the large piety cost.

This is submitted as a POC conversation-starter. It is entirely untested.


DracoOmega:
While variations on ways to reroll Okawaru gifts comes up periodically, the odds in a real game of Okawaru giving you nothing of meaningful value are very slim (it's 8 acquirement-grade equipment to choose from already!) and offering players the option to 'hold out for a better roll' at such an immense cost feels like encouraging poor behavior. An okay upgrade now is almost certainly more valuable than a perfect upgrade quite a while later, especially considering how many Finesses/Duels you're holding back on using to get it.

The gifts are already an overvalued part of Okawaru's strength by many players, despite Oka having some of the most powerful invocations in the game. I don't think we should be actively encouraging many players to use those abilities *even less*.

So, while we appreciate the initiative, I'm closing this PR now.

Issue #4066, opened 2024-09-30
Felids can escape Collapse by dying


caryoscelus:
**PLATFORM and OS or BROWSER:**
local console / nixos

**DCSS VERSION AND GAME INFORMATION:**
0.32, 0.33-a0-151-g9ec8a70d

**DESCRIPTION:**
get felid character to collapse from 0 strength, die and collapse status is gone despite strength still lacking. haven't checked other zero stat states, but i suppose they will have the same issue

kleiner-fisch:
I can reproduce this, and will look into this.

Seems stat-zero is only checked on stat changes. 

I think on reviving stat-zero should be checked.

PR #4076, opened 2024-10-05
Fixed 0_stat status for revived characters.

kleiner-fisch:
This fixes issue #4066.

Description from the issue:
> Get Felid character to collapse from 0 strength, die and collapse status is gone despite strength still lacking.

### Implementation

There are duration counters to track how long Collapse should hold. If strength > 0 the counter is decreased, and Collapse is removed when the counter is 0. On reviving the counter was mistakenly set to zero. Same for Brainless and Clumsy.

Now revive does not set those counters to 0, to retain the 0_stat status across being revived.

Sidenote: The location of the method `revive` inside `nearby-danger.cc` seems weird.

### DIFF
--- a/crawl-ref/source/nearby-danger.cc
+++ b/crawl-ref/source/nearby-danger.cc
@@ -505,7 +505,10 @@ void revive()
             && dur != DUR_TRANSFORMATION
             && dur != DUR_BEOGH_SEEKING_VENGEANCE
             && dur != DUR_BEOGH_DIVINE_CHALLENGE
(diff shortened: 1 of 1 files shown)


caryoscelus:
looks good, thanks!

DracoOmega:
I elected to fix this in a slightly different way. Felid revival already restores all drained stats, which usually puts the player above 0 again, so I think it is consistent with other revival healing to remove those status effects in that case. Instead, it maintains the status only if their stat is still somehow 0 or less.

But nonetheless, thanks for the submission pointing out where the fix needed to go! It is appreciated.

PR #4077, opened 2024-10-06
work around crash in actor::knockback

geekosaur:
https://cbro.berotato.org/morgue/sjaakdevlaming/crash-sjaakdevlaming-20241006-014359.txt

Passive retaliation killed the caster of a force lance before knockback could occur; since the caster's position was long gone by the time actor::knockback finally got called, there is no position to knockback away from any more, so just return false.


geekosaur:
Clearly I'm a C programmer….

geekosaur:
Arguably the knockback should happen before the retaliation, but I presume that caused other problems or it wouldn't be a fineff now.

Implojin:
Thank you for fixing this, I've merged it and backported it to 0.32.

Issue #4075, opened 2024-10-04
Yredelemnul bind soul succeeds even when monster leaves sight


kleiner-fisch:
**PLATFORM and OS or BROWSER:**

Linux Offline

**DCSS VERSION AND GAME INFORMATION:**

 Dungeon Crawl Stone Soup version 0.33-a0-214-g42c94764d4 (console) character file.

**DESCRIPTION:**

I used bind soul from Yredelemnul on a monster. Binding should only succeed if in between using the ability and killing the monster, the monster does not leave my sight.

However, I went up and down some stairs without the monster following me. Still, when I killed it, it became my slave, even though it was not within my sight without interruption. 

When I use the ability, and the monster leaves my sight by walking away (I do not go down stairs) it works as expected.

kleiner-fisch:
Fixed in #4078

PR #4078, opened 2024-10-07
Fixed Yred bind soul ability

kleiner-fisch:
Before, bind_soul had infinite duration. Whenever we lost sight of the monster, the duration was changed to finite
duration, and changed back to infinite duration again, when we saw the monster again.

However, in the case that we went up some stairs the code to make the duration finite is not executed.

Now, bind_soul has finite duration (I randomly picked 30 auts), but the
duration is only decreased when the monster is not in our sight.

If we change floors without the monster, the enchantment is cancelled on
our return to the floor. 

Sidenote: I think having for every enchantment an extra case in `timed-effects.update_monster` seems strange, but I am not sure how to do it better. Happy for suggestions.


kleiner-fisch:
I forgot, this fixes #4075

PR #4094, opened 2024-10-17
Changed resting with inhibited regeneration mutation

### LINKED ISSUE #4018: inhibited regeneration weirdness near invisible monsters

Webtiles

0.32-a0-2226-g9043e79ccf

**DESCRIPTION:**

I have the Inhibited Regeneration mutation. No monsters appeared to be around, but the -regen status was active. Trying to rest resulted in the message "You start waiting." and no time passing. It turned out that there was an invisible monster present.

I reproduced this in trunk wizmode:
- Reduce HP or MP to less than full (it still happens at full HP)
- Give self Inhibited Regeneration
- Spawn a shadow wraith
- Check that -regen status appears, and that there is no sensed monster tile
- Press 5
- Observe the "You start resting." message and lack of turn count change

Autoexplore does not get interrupted.

I'm not sure what the intended behavior would be. Should the -regen status get applied? Should there be a more descriptive failure message?

Without the mutation, resting will proceed until the monster hurts you or gets a "sensed monster" tile.

**EXTRA FILES:**

Screenshot from 2024-09-10 16-54-53

kleiner-fisch:
### Problem 
This fixes #4018 

So far, resting with the mutation inhibited regeneration while an invisible monster is nearby that the player is not aware off resulted in the normal `You start resting` message, without letting any time pass.

Why we get a zero delay is not clear to me. As the control flow is 
* `main._do_rest()` -->
*  `_start_running(...)` --> 
* `you.running.initialise(...)` and here I would expect to get a non-zero delay. But that is not what happens.

Perhaps someone can explain why that does not happen?

### Suggested Change
Now, a message is generated that indicates that there is a monster nearby, which inhibits regeneration, and still we do not let time pass.

Additionally, changed the description of the mutation to indicate that monsters need not be visible to prevent regeneration.

A different possible way to handle things might be to let time pass, e.g. 1 aut. However, in my view that is usually not what the player intends when pressing 5 with some monster nearby. If the player intends to let time pass, they can still let it pass in small increments using `.` or `s`.

### DIFF
--- a/crawl-ref/source/mutation-data.h
+++ b/crawl-ref/source/mutation-data.h
@@ -248,7 +248,7 @@ static const mutation_def mut_data[] =
 { MUT_INHIBITED_REGENERATION, 3, 1, mutflag::bad, false,
   "inhibited regeneration",
 
-  {"You do not regenerate when monsters are visible.", "", ""},
+  {"You do not regenerate when monsters are nearby.", "", ""},
 
   {"Your regeneration stops near monsters.", "", ""},
 
--- a/crawl-ref/source/mutation.cc
+++ b/crawl-ref/source/mutation.cc
@@ -1091,7 +1091,7 @@ static vector<string> _get_fakemuts(bool terse)
         else
         {
             result.push_back(
-                _formmut("You do not regenerate when monsters are visible."));
+                _formmut("You do not regenerate when monsters are nearby."));
             result.push_back(
                 _formmut("You are frail without blood (-20% HP)."));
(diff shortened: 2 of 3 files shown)


Implojin:
Thanks for the PR! I've fixed this in a slightly different way in b65ea80194.

Re: Why time wasn't passing: The rundelay being issued was immediately being stopped by `regeneration_is_inhibited` (in delay.cc). As far as I can tell, that behavior seems correct, and we wouldn't want to change it.

PR #4101, opened 2024-10-24
Apply location effects to flank and swoop

adelrune:
Before this commit, flank and swoop attack flavours would not trigger traps and trap-like effects like sigil of bindings when moving the monster.

I haven't found an issue for this but I stumbled upon that bug while feeling clever about a ufetubus getting sigile'd. The ufetubus didn't get sigiled as I expected.

### DIFF
--- a/crawl-ref/source/mon-act.cc
+++ b/crawl-ref/source/mon-act.cc
@@ -886,6 +886,8 @@ static bool _handle_swoop_or_flank(monster& mons)
             }
         }
         mons.move_to_pos(tracer.path_taken[j+1]);
+        // Apply traps and other location effects to the monster.
+        mons.apply_location_effects(mons.pos());
         fight_melee(&mons, defender);
         mons.props[SWOOP_COOLDOWN_KEY].get_int() = you.elapsed_time
                                                   + 40 + random2(51);


crawl-ref/source/mon-act.cc
```diff
@@ -886,6 +886,8 @@ static bool _handle_swoop_or_flank(monster& mons)
             }
         }
         mons.move_to_pos(tracer.path_taken[j+1]);
+        // Apply traps and other location effects to the monster.
+        mons.apply_location_effects(mons.pos());
```
Implojin, line 890: It looks like `apply_location_effects` wants the monster's old position, not its new one. Could you fix this?

adelrune:
Thanks, I'll take a look at that later today or tomorrow !

adelrune:
I have addressed your two comments, thanks for the feedback !

Implojin:
Thank you for the bugfix :)

As this was your first contribution, what name would you like to be listed as in the credits?

adelrune:
Thanks for merging! adelrune is good for the credit name (even though my computer has another name for the commit ...)

Implojin:
Done. Thanks again!

PR #4108, opened 2024-11-04
reinstate yak island

yrdzrfxndfvh:
re-adds the old yak island vault removed in 530620f - but greatly increases the threat level

- 50% of the time it will be enclosed with the island inaccessible, and have a one way hatch in the middle if so
- lair (2-5) gets yaks, death yaks, dream sheep, elephants and catoblepae
- swamp (1-3) gets spectral yak and yak-adjacent creatures, weighted towards spawning death yaks but can also create some weaker (dream sheep) and scarier (dire elephants) monsters, these will fly off of the island when they hear you. will also be mostly filled with shallow water
- shoals 1-3 gets the standard yak creatures and centaurs, and can spawn a yaktaur with a decent quality arbalest or a cyclops
- shoals 4 gets the scariest set, spawning potentially multiple centaur warriors, yaktaurs and catoblepae, as well as placing either an apis, stone giant or yaktaur captain (ouch), but also has the most loot

i have a slight feeling that the lair set might be a bit undertuned and the shoals:4 set might be overtuned (is using an apis too much? i was inspired to make shoals:4 its own set because apis can already spawn there and they are a yak type monster) but i would need feedback from other people to make any solid decisions

### DIFF
--- a/crawl-ref/source/dat/des/variable/mini_monsters.des
+++ b/crawl-ref/source/dat/des/variable/mini_monsters.des
@@ -1423,6 +1423,51 @@ MAP
      x3
 ENDMAP
 
+# Original by Lemuel - Stranded yaks
+NAME:   lemuel_tekkud_yak_island
+TAGS:   no_pool_fixup no_monster_gen
+DEPTH:  Lair:2-, Swamp, !Swamp:$, Shoals
+WEIGHT: 5
+: if crawl.coinflip() then
+SUBST:  - = W , > = .
+: else
+SUBST:  - = w
+: end
+NSUBST: . = 6:1 / 1:2 / 1:% / *:.
+: if you.branch() == "Lair" then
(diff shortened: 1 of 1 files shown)


regret-index:
I haven't gotten around to this one for a while because it'd need substantial reworks, though the author of this PR has assented to such as I plan to get around to today. Immediately comparable reinsating of vaults is in 2bbf05e, which had significant monster set reshufflings dealing with perfectly fine layouts. lemuel_yak_island was removed, however for being melee-only enemies. This PR.... seems to assume catoblepae stuck over water actual do much meaningful damage by themselves (and that it's fun gameplay to plink that down with a 0-ranged-skill shortbow after clearing the floor), or that vanilla spectrals of out-of-branch monsters suit a branch with many discrete themes because of one spectral-summoning monster. There's not even much for rewards beyond free experience after glacially doing the former!

I'm preserving a version of the Shoals vault, which can contain yaks and dream sheep and ranged monsters alike, and tossing out the rest.

PR #4118, opened 2024-11-10
Use more nouns when a boulder pushes something into the unknown.

Aliscans:
Brom's Barrelling Boulder can push monsters into terrain you haven't seen. This created a message such as " Your boulder crushes the orc against  and falls apart!" as feature_description_at() returns an empty string for unknown terrain.

Replace the empty string with "something" for that message.


nlavsky:
Thanks!

PR #4119, opened 2024-11-10
Fix grow oklob not placing under the player

### LINKED ISSUE #4048: Using Grow Oklob ability on the player costs MP (and I assume piety) and takes a tun, but summons nothing.
0.33-a

**DESCRIPTION:**
Using Grow Oklob ability on the player costs MP (and I assume piety) and takes a tun, but summons nothing.  Using Grow Ballistomycene on the player works properly.

WizardIke:
There was a bug that made plants hostile while being placed and Fedhas worshippers cannot be on top of hostile oklob plants (but can be on top of other hostile plants). This allowed growing ballistomycetes under the player but not oklobs.

Fixes #4048


pooka109:
Applied; thanks.

PR #4126, opened 2024-11-13
Don't blame the player when a confused ally kills

### LINKED ISSUE #3304: Confused TSO summons can incur penance on the player
**PLATFORM and OS or BROWSER:** webtiles

**DCSS VERSION AND GAME INFORMATION:** 0.31-a0-771-g505c3d1

**DESCRIPTION:** TSO worshippers will incur penance if one of their TSO angel summons attacks/kills itself when affected by confusion.

An angel I'd summoned via TSO's god powers became confused by a Tarantella.  It proceeded to hit itself, die from it's own damage at which point I incurred penance.  I tried to find an instance of an angel attacking itself without dying but I couldn't find one in my game's log, so I'm not sure if the angel has to die to trigger penance.

Also, I found a comment on Discord from 3 years ago where someone had a similar issue, except it was friendly fire between two of TSO's summons while the summon was confused.

**EXTRA FILES:**

WizardIke:
Angels summoned by The Shining One worshippers could become confused (e.g. by being hit by a tarentella) and then kill themselves or other angels putting the player in penance. Fix this by not blaming the player for kills by confused allies unless they (or an ally) confused the ally.

Fixes #3304

### DIFF
--- a/crawl-ref/source/mon-death.cc
+++ b/crawl-ref/source/mon-death.cc
@@ -1802,6 +1802,29 @@ static bool _god_will_bless_follower(monster* victim)
            && random2(you.piety) >= piety_breakpoint(0);
 }
 
+static bool should_blame_you_for_kill(int killer_index, bool pet_kill) noexcept
+{
+    if (killer_index == YOU_FAULTLESS)
+        return false;
+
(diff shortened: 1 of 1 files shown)


DracoOmega:
Merged, thanks.

Issue #3822, opened 2024-05-17
gozag shop shows up in search


cfcfcfcfcfcfcfcfcfcf:
**PLATFORM and OS or BROWSER:**
webtiles

**DCSS VERSION AND GAME INFORMATION:**
0.31.0-32-g396128a36d

**DESCRIPTION:**
searching for rpois returns an amulet in a shop closed by gozag wrath

**EXTRA FILES:**
![image](https://github.com/crawl/crawl/assets/78773087/115eb463-c9f8-4e21-bd6b-ad70bf4a2029)
![image](https://github.com/crawl/crawl/assets/78773087/324787c8-11f3-4234-a599-11a105b8855c)

PR #4127, opened 2024-11-13
Add method to remove abandoned shops from StashTracker

kleiner-fisch:
#### Problem 
When leaving Gozag his shops close. However, in #3822 it is mentioned that the closed shops still appear in the StashTracker.

#### Implementation
To fix this I use a similar approach to what is done to remove these
kind of shops from the shopping list in
`shopping_list.remove_dead_shops()`.

Additionally, `shopping_list.remove_dead_shops()` assumed that when calling 
`level_excursion::go_to(level_id)` dactions are performed. 
However, calling that method will not trigger dactions. Hence, I added a call to execute dactions to the method.

#### Potential Issues
I am unsure whether the similar code of removing dead shops from
the StashTracker and the shopping list can/should be combined in a
single method. I would be happy for feedback.

### DIFF
--- a/crawl-ref/source/religion.cc
+++ b/crawl-ref/source/religion.cc
@@ -68,6 +68,7 @@
 #include "skills.h"
 #include "spl-book.h"
 #include "sprint.h"
+#include "stash.h"
 #include "state.h"
 #include "stringutil.h"
 #include "tag-version.h"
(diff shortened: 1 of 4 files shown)


Hellmonk:
I believe this was fixed in 931650e. Please reopen if it's still a problem.

PR #4129, opened 2024-11-14
Fix slime not showing around slime walls after reload

WizardIke:
After reloading and sometimes after changing levels, the slimey acidic
floors around the walls in the Slime Pits (and oozemancy walls)
wouldn't appear.

After reloading before this change:
Screenshot (1951)

After reloading after this change:
Screenshot (1952)


DracoOmega:
Merged, thanks.

PR #4130, opened 2024-11-14
Fix autofight not moving over web with slick slippers

### LINKED ISSUE #3125: Autofight doesn't walk over web traps when wearing slick slippers
online console, CBR2, Mac OS X Terminal
0.30.0-5-g1a98f26d00
https://cbro.berotato.org/saves/jejorda2-crawl-0.30-230508-1818.tar.bz2

**DESCRIPTION:**
I am wearing the slick slippers, so I cannot be webbed.
My melee weapon does not have reaching.
There is an enemy (steelbarb worm) two spaces from me, and a web trap in the space between us:
```
^w.
#^.
#@^
```
No other enemies are visible.
I press tab, expecting to walk onto the trap to close the distance. Or maybe I walk to the empty space diagonal from me that is also adjacent to the worm.
Instead, I get the error: "Failed to move towards target."
Does the tab auto-fight know that it is safe to walk on web traps while wearing the slick slippers?

WizardIke:
When autofight was trying to move towards an enemy to attack it and the
only way you could see to reach the enemies was through a web, autofight
would give a message saying there was no safe path to the enemy and do
nothing. It should have moved through the web as the slick spippers give
immunity to them.

Fixes #3125

### DIFF
--- a/crawl-ref/source/dat/clua/autofight.lua
+++ b/crawl-ref/source/dat/clua/autofight.lua
@@ -116,7 +116,7 @@ local function have_quiver_action(no_move)
 end
 
 local function is_safe_square(dx, dy)
-    if view.feature_at(dx, dy) == "trap_web" then
+    if view.feature_at(dx, dy) == "trap_web" and not you.is_web_immune() then
         return false
     end
     return view.is_safe_square(dx, dy)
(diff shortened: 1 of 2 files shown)


DracoOmega:
Merged, thanks.

PR #4132, opened 2024-11-14
Fix display of GDR in wizmod

WizardIke:
Use the rounded down instead of randomly rounded version of the players AC when calculating guaranteed damage reduction for display.


DracoOmega:
Merged, thanks.

PR #4135, opened 2024-11-16
Don't crash if a wizard flees the Vaults without a rune.

Aliscans:
Entering the Vaults via the stairs and then leaving with &u triggered an ASSERT in _rune_effect().

Fix this by briefly "unlocking" the Vaults as the wizard escapes.


DracoOmega:
Merged, thanks!

PR #4136, opened 2024-11-16
Describe fewer things which have been excluded as stairs.

Aliscans:
Only describe an excluded square as being up- or downstairs in the dungeon overview if it goes to another level, and if its exact destination can be predicted after using it once.

This means that it no longer describes things such as one-way gates, portals, shops or transporters as being stairs.

### DIFF
--- a/crawl-ref/source/exclude.cc
+++ b/crawl-ref/source/exclude.cc
@@ -521,7 +521,7 @@ void set_exclude(const coord_def &p, int radius, bool autoexcl, bool vaultexcl,
                 const dungeon_feature_type feat = cell.feat();
                 if (feat_is_door(feat))
                     desc = "door";
-                else
+                else if (feat_is_travelable_stair(feat))
                 {
                     const command_type dir = feat_stair_direction(feat);
                     if (dir == CMD_GO_UPSTAIRS)


Implojin:
Why?

Aliscans:
> Why?

Are you asking about the principle or the implementation?

I don't find it helpful when the dungeon overview refers to stairs if I put an exclusion on a square which happens to contain an altar. And doing it like this only involves one line of code.

Implojin:
I'm asking about the context behind why you want to make this commit: Your PR description and the commit message have stated what you're doing, but it's more important for future maintainers to have a record of why the commit was made.

You've mentioned that placing an exclusion marks an altar as a staircase. Okay. If the game is doing that, that sounds buggy.

However, maybe we can do better than just choosing to not mark things as staircases? On a brief look at the code, we're building a generic exclusion desc and we also have `feat_type_name()` as a thing we could call. We could build the desc using `feat_type_name` in fewer lines with more functionality while deduplicating.

I don't have a lot of time left tonight to look at this; I'll try to get to it tomorrow -- please let me know what you think.

Aliscans:
You can always do more. A lot of it comes down to knowing how you want the game to behave.

Personally, I make notes about what's happening in each game outside Crawl, so don't rely on annotations for more than a short time. This was just something which looked wrong, and looked like something which could be easily improved.

Implojin:
Sorry it took me so long to get back to this, I've fixed this in cd0d8a32d with `feat_type_name()`.

PR #4145, opened 2024-11-23
disable installation of pkg-config and libpng

geekosaur:
Both are already installed on current GHA images; the former is actually redirected to install a different variant (there are two: the original, and a fast but buggy C version), which then conflicts with the already-installed version. So simply skip installing both.


Implojin:
And thanks again.

PR #4146, opened 2024-11-23
silence flash_tile warning on console builds

geekosaur:


geekosaur:
…which wants the other PR or it fails on MacOS, naturally. Oh well.

crawl-ref/source/view.cc
```diff
@@ -1519,8 +1519,14 @@ void view_add_glyph_overlay(const coord_def &gc, cglyph_t glyph)
     glyph_overlays.push_back({gc, glyph});
 }
 
+#ifdef USE_TILE
+#define CONS_UNUSED(x) x
+#else
+#define CONS_UNUSED(x)
+#endif
```
WizardIke, line 1526: Instead of making a new macro, I would changed the part of the function body that looks like
```
#ifdef USE_TILE
    tiles stuff
#endif
```
into
```
#ifdef USE_TILE
    tiles stuff
#else
    UNUSED(tile);
#endif
```
as this is what I've seen used elsewhere in the codebase.

WizardIke, line 1526: Instead of making a new macro, I would changed the part of the function body that looks like
```
#ifdef USE_TILE
    tiles stuff
#endif
```
into
```
#ifdef USE_TILE
    tiles stuff
#else
    UNUSED(tile);
#endif
```
as this is what I've seen used elsewhere in the codebase.

WizardIke, line 1526: Instead of making a new macro, I would changed the part of the function body that looks like
```
#ifdef USE_TILE
    tiles stuff
#endif
```
into
```
#ifdef USE_TILE
    tiles stuff
#else
    UNUSED(tile);
#endif
```
as this is what I've seen used elsewhere in the codebase.

WizardIke, line 1526: Instead of making a new macro, I would changed the part of the function body that looks like
```
#ifdef USE_TILE
    tiles stuff
#endif
```
into
```
#ifdef USE_TILE
    tiles stuff
#else
    UNUSED(tile);
#endif
```
as this is what I've seen used elsewhere in the codebase.

geekosaur: Fine with me, I just did the first thing that popped into mind as I'm not fully familiar with all the weirdnesses in this codebase. (I would actually have moved that macro into a header and added the other two variants, but specifically wanted comments like yours first.)

Implojin:
Thanks!

PR #4147, opened 2024-11-26
Cast OTR with uppercase Z when all enemies resist poison (CrawlOdds)

### LINKED ISSUE #3957: OTR behaviour inconsistent when all enemies rpois
webtiles

0.32-a0-2006-gc3173615b2

**DESCRIPTION:**
When casting OTR:

- With no enemies on the screen, you get a confirm message "You can't see any susceptible monsters within range! (Use Z to cast anyway."
- With only rpois enemies on the screen, you cast the spell

This is inconsistent - either both should cast the spell (with the idea being that pre-casting is "expected"), or neither. My vote is neither - this is nice behaviour when you aren't sure if something has rpois.

WizardIke:
When casting Olgreb's toxic radiance with no enemies in view, it would give the message "You can't see any susceptible monsters within range! (Use Z to cast anyway)". However, when casting with only poison resistant and immune enemies in view, it would cast normally even though it couldn't effect anything. Fix this inconsistency by also requiring uppercase Z to cast with only poison reistant and immune enemies in view.

This change also fixes a bug where monsters would always act as if they could see invisibly even when they couldn't when deciding whether to cast OTR.

Fixes #3957

### DIFF
--- a/crawl-ref/source/spl-util.cc
+++ b/crawl-ref/source/spl-util.cc
@@ -1648,7 +1648,6 @@ bool spell_no_hostile_in_range(spell_type spell)
     case SPELL_FIRE_STORM:
         return false;
 
-    case SPELL_OLGREBS_TOXIC_RADIANCE:
     case SPELL_IGNITION:
     case SPELL_FROZEN_RAMPARTS:
     case SPELL_FULSOME_FUSILLADE:
@@ -1681,6 +1680,9 @@ bool spell_no_hostile_in_range(spell_type spell)
 
         return true;
     }
+
+    case SPELL_OLGREBS_TOXIC_RADIANCE:
+        return cast_toxic_radiance(&you, pow, false, true) == spret::abort;
 
     case SPELL_IGNITE_POISON:
         return cast_ignite_poison(&you, -1, false, true) == spret::abort;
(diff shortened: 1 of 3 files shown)


crawl-ref/source/spl-util.cc
```diff
@@ -1625,7 +1625,6 @@ bool spell_no_hostile_in_range(spell_type spell)
         return true;
 
     const int range = calc_spell_range(spell, 0);
-    const int minRange = get_dist_to_nearest_monster();
```
Implojin, line 1628: Why are you moving this, then defining it twice (as const)? I'd suggest leaving it where it is, and not repeating ourselves.

crawl-ref/source/spl-util.cc
```diff
@@ -1649,11 +1648,16 @@ bool spell_no_hostile_in_range(spell_type spell)
         return false;
 
     case SPELL_OLGREBS_TOXIC_RADIANCE:
+        return cast_toxic_radiance(&you, pow, false, true) == spret::abort;
```
Implojin, line 1651: This OTR case could perhaps be moved lower in the function, for a visual grouping with the other spells returning a `cast_spell_name` handler?

WizardIke: Good idea, done

crawl-ref/source/spl-util.cc
WizardIke: `get_dist_to_nearest_monster` is a fairly expensive function to call and most spells don't use its result so I moved it to the two places that use it. I've undone this change as it had nothing to do with OTR's inconsistency.

Implojin:
Thanks for the fix :)

PR #4148, opened 2024-11-26
Fix merfolk equipping barding when in beast form (EscapistDaydreamer)

### LINKED ISSUE #4142: dcss bug report
PC Windows 10

0.32.1

**DESCRIPTION:**
Im playing Naga shapeshifter, when i go into beast form my barding melts into my body. But if i unequip the barding
before transforming, i can equip it while in beast form and beneficiate from the AC bonus. Idk whats intended, if you can
or if you cannot have a barding equipped while in beast form, but both are possible in current version it seems.

**EXTRA FILES:**
bardingunmelded
bardingmelded

WizardIke:
Beast form should meld barding and it was melding them correctly if you put on the barding and then activated the form. However, it was letting you put on and benefit from barding if you activated the form first.

fixes #4142

### DIFF
--- a/crawl-ref/source/item-use.cc
+++ b/crawl-ref/source/item-use.cc
@@ -1952,7 +1952,7 @@ bool can_wear_armour(const item_def &item, bool verbose, bool ignore_temporary)
 
     if (sub_type == ARM_BARDING)
     {
-        if (you.can_wear_barding(ignore_temporary))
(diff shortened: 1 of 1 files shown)


Implojin:
I've fixed this slightly differently in 7c3642a because the pre-existing code here was buggy, but thank you for the PR.

PR #4153, opened 2024-12-01
adjust meatsprint

yrdzrfxndfvh:
i am not very good at using github desktop


regret-index:
The permarock is a good idea to avoid further Hellfire Mortar shenanigans and Formicid abuse, but I have to ask: why bother changing the numbers on any of the weapons? I guess the sprint is already innately screwing with the maximum numbers that can be acquired from gear, but considering all the massive visual noise and message noise the sprint relies upon, it doesn't feel like enticing more possible ways for future enchant-proofing to run into issues is particularly worth it.

(Also, why not leave even the minimum of a git comment on those two things being what this commit actually _does_?)

yrdzrfxndfvh:
the enchantment thing was just because i thought it was cool

PR #4158, opened 2024-12-04
fix: Yred keeps bound hydra heads

### LINKED ISSUE #4115: Using Yred's Bind Soul ability doesn't replace the current soul by the new one if they're the same species.
Local tiles on Windows

Version 0.32
Seed: 1610517359015032543
Floor: Lair 5

**DESCRIPTION:**
I was worshipping Yred, and used Bind Soul to capture a 5-headed hydra on Lair 5. Later on, I met another hydra with more heads that I decided to capture. Using Finisher, I managed to cut more heads until it had 12. Then I blinked away and used Hurl Torchlight to kill it. The backlog said 'The 12-headed hydra's soul is now yours'. After checking my ally, it turns out it still has 5 heads.

**EXTRA FILES:**
[yred hydra.zip](https://github.com/user-attachments/files/17682712/yred.hydra.zip)

bobross419:
Yred bound soul was not consistent for hydra.  Bound hydra would revert to the original number of heads.  Other transformed or polymorphed monsters would retain their final form when bound.  Zombie and spectral hydra also maintain their final number of heads.

This change allows Yred to maintain the final number of hydra heads; additionally, Yred will restore any heads that were permanently removed with a flaming weapon.

No change in heads
```
You wrap your dark will around the six-headed hydra's soul!
The six-headed hydra vaporises in an electric haze!
The six-headed hydra's soul is now yours.
You can see a friendly bound six-headed hydra.
```

Cutting off heads with fire
```
You wrap your dark will around the five-headed hydra's soul!
You hack one of the five-headed hydra's heads off!
You hack one of the four-headed hydra's heads off!
The three-headed hydra vaporises in an electric haze!
Yredelemnul restores the hydra's soul to it's natural glory!
The three-headed hydra's soul is now yours.
You can see a friendly bound five-headed hydra.
```

Growing additional heads
```
You wrap your dark will around the seven-headed hydra's soul!
You lop one of the seven-headed hydra's heads off!
You lop one of the eight-headed hydra's heads off!
The nine-headed hydra vaporises in an electric haze!
The nine-headed hydra's soul is now yours.
You can see  a friendly bound nine-headed hydra.
```

Resolves #4115

### DIFF
--- a/crawl-ref/source/god-abil.cc
+++ b/crawl-ref/source/god-abil.cc
@@ -1747,7 +1747,7 @@ void yred_make_bound_soul(monster* mon, bool force_hostile)
     {
         if (orig.heads() < mon->props[OLD_HEADS_KEY].get_int())
         {
-            mprf("Yredelemnul restores the hydra's soul to its natural glory!");
+            mpr("Yredelemnul restores the hydra's soul to its natural glory!");
             mon->num_heads = mon->props[OLD_HEADS_KEY].get_int();
         }
         else
--- a/crawl-ref/source/god-abil.cc
+++ b/crawl-ref/source/god-abil.cc
@@ -1702,8 +1702,8 @@ void yred_make_bound_soul(monster* mon, bool force_hostile)
     remove_bound_soul_companion();
     add_daction(DACT_OLD_CHARMD_SOULS_POOF);
 
-    const string whose = you.can_see(*mon) ? apostrophise(mon->name(DESC_THE))
-                                           : mon->pronoun(PRONOUN_POSSESSIVE);
+    string whose = you.can_see(*mon) ? apostrophise(mon->name(DESC_THE))
+                                     : mon->pronoun(PRONOUN_POSSESSIVE);
 
     // Heal the health we paid wrestling for this soul
     you.heal(mon->get_ench(ENCH_SOUL_RIPE).degree);
@@ -1742,6 +1742,20 @@ void yred_make_bound_soul(monster* mon, bool force_hostile)
     // Use the original monster type as the zombified type here, to get
     // the proper stats from it.
     define_zombie(mon, mon->type, MONS_BOUND_SOUL);
+
+    if (orig.mons_species() == MONS_HYDRA)
+    {
+        if (orig.heads() < mon->props[OLD_HEADS_KEY].get_int())
+        {
+            mprf("Yredelemnul restores the hydra's soul to it's natural glory!");
+            mon->num_heads = mon->props[OLD_HEADS_KEY].get_int();
+        }
+        else
+            mon->num_heads = orig.heads();
+
+        whose = you.can_see(*mon) ? apostrophise(orig.name(DESC_THE))
+                                           : orig.pronoun(PRONOUN_POSSESSIVE);
+    }
 
     // Modify health based on invocations skill
     mon->max_hit_points = yred_get_bound_soul_hp(orig.type);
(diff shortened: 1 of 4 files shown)


bobross419:
This is ready for automated testing or lmk if there are any changes needed.

crawl-ref/source/god-abil.cc
```diff
@@ -1743,6 +1743,20 @@ void yred_make_bound_soul(monster* mon, bool force_hostile)
     // the proper stats from it.
     define_zombie(mon, mon->type, MONS_BOUND_SOUL);
 
+    if (orig.mons_species() == MONS_HYDRA)
+    {
+        if (orig.heads() < mon->props[OLD_HEADS_KEY].get_int())
+        {
+            mprf("Yredelemnul restores the hydra's soul to it's natural glory!");
```
bstrie, line 1750: Should be "its natural glory".

crawl-ref/source/god-abil.cc
```diff
@@ -1747,7 +1747,7 @@ void yred_make_bound_soul(monster* mon, bool force_hostile)
     {
         if (orig.heads() < mon->props[OLD_HEADS_KEY].get_int())
         {
-            mprf("Yredelemnul restores the hydra's soul to it's natural glory!");
+            mprf("Yredelemnul restores the hydra's soul to its natural glory!");
```
pooka109, line 1750: Since this message is a string without any parameters, it should use mpr() instead of mprf().

NormalPerson7:
My concern is that this would introduce unwanted behaviour: namely, you could be encouraged to keep chopping heads to get a 20-headed hydra, e. g. with polearms + BVC/other methods to keep the hydra away from you. I can definitely see the other argument too - that this would be a risky strategy.

Implojin:
From a gameplay perspective, this sounds like the kind of tedium we'd prefer to avoid: Giving any player who wants to bind a hydra incentive to sit around chopping heads to make their minion more powerful does not sound like fun gameplay.

I won't close this immediately in case any other devs want to make an argument in favor of this, but I'm not a fan.

pooka109:
What if, instead of preserving the number of heads a hydra has when it dies, we preserve the number of heads the hydra had when it was initially generated? That would avoid the unintuitiveness of randomly rerolling the number of heads, but it would also not encourage degenerate play like giving hydras as many heads as possible before binding them.

Implojin:
Sure, that sounds fine.

bobross419:
I moved saving the ORIG_HEADS_KEY from when the bind ability is invoked to when the monster is placed; otherwise, I could still pre-cut heads before binding to artificially inflate the head count.   
```
You wrap your dark will around the three-headed hydra's soul!
{...}
Your grip on the 19-headed hydra's soul is slipping...
You lose your grip on the 19-headed hydra's soul.
You wrap your dark will around the 19-headed hydra's soul!
The 19-headed hydra vaporises in an electric haze!
The 19-headed hydra's soul is now yours.
You can see a friendly bound three-headed hydra.
```

I also went ahead and updated the original derived undead code to make it more consistent with binding's behavior.  Now, death channelled spectral hydra will return to their original number of heads, but zombies will still maintain the number of heads they died with.  I can fix zombies too, it just didn't make sense for physical heads to disappear. 

```
The nine-headed hydra vaporises in an electric haze!
You kill the nine-headed hydra!
The dead are walking! A glowing mist starts to gather...
You can see a friendly spectral three-headed hydra (minion) and a friendly nine-headed hydra
zombie (minion).
```

bobross419:
I will have time to work on this over the next couple weeks if there are changes that need to be made.

CrawlOdds:
Done similarly in 1795ca5 (the version in this PR had some small issues with polymorphing and a crash in some cases)

PR #4159, opened 2024-12-04
remove old g++ check

geekosaur:
The pattern matches Arch's 14.2.x, which leads to inconsistent optimization and crashes with their build flags.

It was later noted that g++ 4.2 didn't support C++11, which we require these days, so the test no longer serves a purpose anyway. And Macs have shipped clang instead of gcc for years.

### DIFF
--- a/crawl-ref/source/Makefile
+++ b/crawl-ref/source/Makefile
@@ -558,14 +558,7 @@ ifdef USE_ICC
 # Some very good optimization flags.
   CFOPTIMIZE := -O2 -parallel
 else
-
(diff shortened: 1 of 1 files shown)


draeath:
If this is accepted, please pull merge/cherrypick/whatever into `stone_soup-0.32` branch as well.

Implojin:
Thanks for the patch!

Re: Backporting this: We'll be backporting it to 0.31 and issuing point releases for those branches soon-ish, to hopefully signal to any other downstream packages that may have had affected builds that they should pick this fix.

PR #4164, opened 2024-12-09
Fix barding being incorrectly marked permanently useless

WizardIke:
Barding for races that can wear it should be marked temporarily useless in a form that melds it, not permanently useless. This was causing it to be impossible to get from acquirement and show as the wrong colour in the inventory while in a form that melds it.

### DIFF
--- a/crawl-ref/source/item-use.cc
+++ b/crawl-ref/source/item-use.cc
@@ -1952,15 +1952,10 @@ bool can_wear_armour(const item_def &item, bool verbose, bool ignore_temporary)
 
     if (sub_type == ARM_BARDING)
     {
-        if (!you.can_wear_barding(ignore_temporary))
+        if (!you.can_wear_barding())
         {
             if (verbose)
-            {
-                if (ignore_temporary)
-                    mprf(MSGCH_PROMPT, "You can't wear that!");
-                else
-                    mprf(MSGCH_PROMPT, "You can wear that only in your normal form.");
-            }
+                mprf(MSGCH_PROMPT, "You can't wear that!");
             return false;
         }
     }


Implojin:
Thank you for bringing this up again, because yes, you're right, there are still bugs here, but looking more deeply at the callers of `can_wear_armour`, I suspect some of them are incorrectly passing the desired temp bool, probably because this function is flipping the convention for temp bools used elsewhere in the code.

Implojin:
Also, I'm inclined to take the inventory highlighting as correct behavior(?), because it's being highlighted in grey while the player can't equip it.

Implojin:
Re: Your latest force push, I think we don't want to get rid of the temp check completely, because we have at least one check in invent.cc `_temp_unwearable_armour` that would be broken for bardings if we did.

WizardIke:
> Also, I'm inclined to take the inventory highlighting as correct behavior(?), because it's being highlighted in grey while the player can't equip it.

It does look right, but this isn't what happens with other items like boots as only permanently useless items are greyed out in that menu.

WizardIke:
> Re: Your latest force push, I think we don't want to get rid of the temp check completely, because we have at least one check in invent.cc `_has_temp_unwearable_armour` that would be broken for bardings if we did.
> 
> (It's possible some behavior of these callers may not need to be preserved, but it does seem like it needs to be reviewed.)

We check for temporary things that disable wearing boots later in `can_wear_armour` so this check wasn't needed. It was also wrong as it would print "You can wear that only in your normal form." for boots using races. We could correct the check if that would be better, something like
```
if (!you.can_wear_barding(false))
{
    if (verbose)
        mprf(MSGCH_PROMPT, "You can't wear that!");
    return false;
}
else if (!ignore_temporary && you.can_wear_barding(true)
{
    if (verbose)
        mprf(MSGCH_PROMPT, "You can wear that only in your normal form.");
    return false;
}
```

WizardIke:
The reworked `can_equip_item` function doesn't have this problem

PR #4178, opened 2024-12-16
Let the user look at more "nogen" items in the help menu.

Aliscans:
Replace the "nogen" tag with a "deleted" one for "TAG_MAJOR_VERSION = 34" sections of art-data.txt. Both tags mean the item isn't created in the normal way. The "deleted" tag alone stops the help menu from giving details of the item.

Remove the code in extant_unrandart_by_exact_name() which let some (but not all) of the nogen items be looked up.

My hope here is that, if the data is all in art-data.txt, it's more likely to be updated correctly than it would be with two lists in completely different bits of code.

For what it's worth, the artefacts which have nogen but weren't on the list are:

faerie dragon scales (created with the Enchantress)
dreamdust necklace (created by using the dreamshard necklace)
Axe of Woe (Arena of Blood)
amulet of invisibility (Fedhas' Mad Dash)
scarf of invisibility (Fedhas' Mad Dash)

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -72,6 +72,7 @@
 # * chaotic:  Is treated as a chaotic item.
 # * clarity:  Provides clarity.
 # * corrode:  Randomly causes corrosion when taking damage.
+# * deleted:  Never randomly generated. Should only appear in old games.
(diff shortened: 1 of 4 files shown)


Cgettys:
Very nice, less places to update seems sane to me.

PR #4179, opened 2024-12-19
Say something after "You have ." when wielding a non-weapon item (fix #4172)

### LINKED ISSUE #4172: "You have" literally nothing
Chrome/CXC
Trunk (0.33-a0-566-gaf1b155deb)
**DESCRIPTION:**
The text where to-hit data used to be becomes "You have ." when wielding a non-weapon. 
**EXTRA FILES:**
image

code2828:
Method `describe_to_hit()` returns immediately if the item in hand is not a weapon, resulting in printing "You have ." in the monster desc page. These commits attempts to solve that by telling the player they're not wielding a valid weapon. Also wielding a non-weapon item should be undesirable so an exclamation mark is used instead of a period.

Should resolve #4172.

Didn't talk about this in the irc chatroom although the guide told me to do so because my IP is blocked (idk why). Hope it fixes the issue well and does not violate any conventions. (12/25: sry for the merge commit)

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -5772,7 +5772,7 @@ bool describe_to_hit(const monster_info &mi, ostringstream &result,
     if (weapon != nullptr
         && !(is_weapon(*weapon) || is_throwable(&you, *weapon)))
     {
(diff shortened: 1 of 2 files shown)


code2828:
I guess this issue is 'fixed' in 2dbd7c1 because you can no longer breadwield. Dunno if good to close.

PR #4181, opened 2024-12-20
Make the `base_type` of classed draconians their corresponding colors instead of randomly chosen.

### LINKED ISSUE #4001: Monster lookup shows random draconian colours for classed draconians
Version 0.33-a0-41-gb3de7b944d (though even my 0.28 stable game has this issue)

**DESCRIPTION:**
`?/mdraconian stormcaller` should always show a yellow draconian but instead it shows a randomly coloured draconian. This is evident in the tile, description, breath ability, and sometimes resistances (purple draconians will show more will+ but red/white/green ones won't show fire/cold/poison resistance)

code2828:
In #4001, when you lookup information about draconian stormcaller, the color is randomly generated instead of white. This is because of lines lookup-help.cc:1061-1062. This commit attempts to fix that by assigning colors to them using a switch-case statement. Info about which color should be which is grabbed from lines mon-util.cc:2723-2728.

Should resolve #4001.

### DIFF
--- a/crawl-ref/source/lookup-help.cc
+++ b/crawl-ref/source/lookup-help.cc
@@ -1059,7 +1059,32 @@ static int _describe_monster(const string &key, const string &suffix,
     // Might be better to show all possible combinations rather than picking
     // one at random as this does?
     if (mons_is_draconian_job(mon_num))
-        base_type = random_draconian_monster_species();
+    {
+        // Classed draconians have fixed colours since 0.28
+        switch (mon_num)
+        {
+        case MONS_DRACONIAN_STORMCALLER:
+            base_type = MONS_WHITE_DRACONIAN;
+            break;
+        case MONS_DRACONIAN_MONK:
+            base_type = MONS_GREEN_DRACONIAN;
+            break;
+        case MONS_DRACONIAN_SHIFTER:
+            base_type = MONS_PURPLE_DRACONIAN;
+            break;
+        case MONS_DRACONIAN_ANNIHILATOR:
+            base_type = MONS_YELLOW_DRACONIAN;
+            break;
+        case MONS_DRACONIAN_KNIGHT:
+            base_type = MONS_BLACK_DRACONIAN;
+            break;
+        case MONS_DRACONIAN_SCORCHER:
+            base_type = MONS_RED_DRACONIAN;
+            break;
+        default:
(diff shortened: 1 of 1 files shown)


crawl-ref/source/lookup-help.cc
```diff
@@ -1059,7 +1059,32 @@ static int _describe_monster(const string &key, const string &suffix,
     // Might be better to show all possible combinations rather than picking
     // one at random as this does?
     if (mons_is_draconian_job(mon_num))
-        base_type = random_draconian_monster_species();
+    {
+        // Classed draconians have fixed colours since 0.28
```
Cgettys, line 1063: Might be nice to have a dictionary instead of a case statement. 
And the bit in the comment about 0.28 doesn't quite make sense to me?

But overall, looks reasonable to me.

code2828: Thanks for your review!

About the comment, [The 0.28 release notes](https://github.com/crawl/crawl/releases/tag/0.28.0) said that "Classed draconians have a fixed colour per job." so I figured that it might be good to note here since this is where the issue rose.

code2828: I wonder if it looks good now.

Cgettys: Better, but this is now assigning a base_type to non-draconians by mistake. It also does a double lookup in the success case.

I think it should probably be something like this:
```
    monster_type base_type = MONS_NO_MONSTER;
    if (mons_is_draconian_job(mon_num))
    {
        // Classed draconians have fixed colours since 0.28
        const auto colour_it = draconian_job_to_color.find(mon_num);
        if (job_colour != draconian_job_to_colour.end())
            base_type = job_colour->second;
       else 
            // Might be better to show all possible combinations rather than
            // picking one at random as this does?
            base_type = random_draconian_monster_species();
    }
```
nit: this codebase afaik uses the British spelling of colour

code2828: Oof, that was a really dumb mistake. Sorry for late response, couldn't access github from time to time.

Cgettys: To err is to be human, or something like that.

pooka109:
Applied. Sorry it took so long to get to this.

PR #4191, opened 2024-12-29
Stop mortars from colliding with "a floor" and then being "destroyed!".

Aliscans:
If the lava stream for Hellfire Mortar is shortened by an obstacle which is no longer there when the mortar reaches the spot, and the player sees it, this could give the message "The hellfire mortar collides with a floor and sinks back into the magma."

Omit the "collides with a floor " part unless the mortar hits a monster or "solid" terrain.

Remove the death message which followed this (which seemed redundant) and stop blaming a monster for the death. That is, change the killer_index in monster_die() to NON_MONSTER and the silent flag to true.

...

I don't have an opinion on the mechanics of Hellfire Mortar, but the messages seemed a bit off.

A mortar could collide with "a floor" (not even "the floor") if it ran out of magma. It called monster_die() with a killer_index of "true". As killer_index isn't a boolean, this was just some monster on the level. I've changed this to "no-one" and silenced the death message.

### DIFF
--- a/crawl-ref/source/mon-act.cc
+++ b/crawl-ref/source/mon-act.cc
@@ -1148,7 +1148,7 @@ static void _handle_hellfire_mortar(monster& mortar)
             if (i + 1 == path.size())
             {
                 simple_monster_message(mortar, " sinks back into the magma.");
-                monster_die(mortar, KILL_NON_ACTOR, true);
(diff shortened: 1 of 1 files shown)


gammafunk:
Merged, thanks for the PR!

PR #4192, opened 2024-12-29
Fix the way the scroller works when a menu grows.

Aliscans:
Previously, if a player had a spell library containing every spell, searching it for "hat" led to Shatter being selected. Searching for "a" after this led to Shatter still being selected, but only the top of the spell list being displayed.

This happened because UIMenu::update_items() resized item_info but didn't set the x, y, row or column fields, and UIMenu::get_item_region() relied on those being up to date.

Call UIMenu::do_layout() to fix it.


gammafunk:
Merged, thanks for the PR!

PR #4193, opened 2024-12-29
Use article_a() instead of "a article" in a few places.

Aliscans:
Replace "a %s" (and equivalent) with article_a() in some places where the noun can start with a vowel sound, or where the noun isn't clear from the context.

For instance, typing ?/i with the default setup now prints "Describe an item;" rather than "Describe a item;".


gammafunk:
Merged, thanks for the PR!

PR #4200, opened 2025-01-03
docs: Update INSTALL.md instructions to cover Ubuntu 24.04 package requirements

### LINKED ISSUE #3998: Build error: no GL/glu.h
When trying to compile the game in accordance with the [instructions](https://github.com/crawl/crawl/blob/master/crawl-ref/INSTALL.md#ubuntu--debian)  from source under Ubuntu 24.04, the following error occurs

> glwrapper-ogl.cc:25:13: fatal error: GL/glu.h: no such file or directory

which is solved by installing the package

> sudo apt-get install libglu1-mesa libglu1-mesa-dev

Perhaps one should add installation of missing packages to the instructions?

Cgettys:
The trunk branch of the apt repository is about a decade out of date and I prefer local to webtiles, so I went to build crawl from source because I wanted to try Forgecraft out.
I am running Pop!_OS 24.04 (which is a Ubuntu 24.04 derived distro), and I stumbled over #3998 

I was very glad that issue existed, since it saved me a lot of sleuthing.

But it would be better still to add it to INSTALL.md, like that issue suggested.
I'm not sure if these packages are auto-installed as transitive dependencies of some package, or commonly installed by default, or what on older distros (in which case I could safely add them to the existing list for crawl_tiles instead).

So instead, I'm just adding another few  lines saying "you need this for Ubuntu 24.04+", and adding an example error message to the troubleshooting message.

This is my first contribution to crawl - if you want to credit me, you can credit me as Charlie Gettys, but I'm really not fussed over credit for a few lines in a readme :smile: .

Closes #3998

### DIFF
--- a/crawl-ref/INSTALL.md
+++ b/crawl-ref/INSTALL.md
@@ -88,6 +88,11 @@ libfreetype6-dev libpng-dev fonts-dejavu-core advancecomp pngcrush
 
 Then follow [the above compilation steps](#compiling).
 
+Note: Ubuntu 24.04+ and related distros (such as Pop!_OS 24.04) may require a few more packages: 
+```sh
+sudo apt install libglu1-mesa libglu1-mesa-dev
+```
+
 ### Fedora
 
 These instructions may work for other RPM-based distros.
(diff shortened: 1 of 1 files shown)


gammafunk:
These requirements should apply to recent debian versions, so I just updated the main list of debian packages rather than putting more distro-specific instructions. Thanks for the PR!

PR #4203, opened 2025-01-03
fix: Make spells that never miss unaffected by blindness

Flugkiller:
When a player is blinded, they can miss with spells that are supposed to never miss, such as Magic Dart, LRD, and Chain Lightning. This only applies to spells that solely rely on having AUTOMATIC_HIT as their to hit for this purpose (e.g. Fireball doesn't currently miss even when blinded, since it is an explosion). This commit makes sure that the aforementioned spells will be unaffected by blindness.


Implojin:
This PR seems reasonable based on the blindness change in 7051e9d3e0, thanks for the fix. :)

PR #4208, opened 2025-01-05
Let beam targeter take whether it's harmless into account.

ahyangyi:
The situation I was trying to fix is to stop Mephitic Cloud targeter from trying to avoid the Qazlal-worshipping player. The harmlessness check is already there, but not called by the beam targeter.

`bolt::harmless_to_player()` was private. However, it makes sense for the beam targeter to call this, and making it public would be the straightforward way to do this.

### DIFF
--- a/crawl-ref/source/beam.h
+++ b/crawl-ref/source/beam.h
@@ -210,6 +210,7 @@ public:
 
     bool can_affect_actor(const actor *act) const;
     bool can_affect_wall(const coord_def& p, bool map_knowledge = false) const;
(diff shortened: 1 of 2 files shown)


ahyangyi:
[qazmeph.zip](https://github.com/user-attachments/files/18314416/qazmeph.zip)

Test savegame in wizmode.

Cast Mephitic Cloud.

Without patch: targeter attempts to avoid the player.
With patch: targeter no longer attempts to avoid the player.

PR #4210, opened 2025-01-06
Work towards removing glu dependency

Cgettys:
* Instead of load_textures accepting an opaque bool, have it take MipmapOptions directly
* Add a maybe_bool option to force enable or force disable mipmapping to make testing easier.
* Encapsulate 6 fields that always are passed together into a little class that enforces only valid combinations of arguments can be created, allow passing the args straight through in a few functions that do that, and deduplicate some logic to be consistent in how we check for a special case
* Give a magic number a name (NO_OFFSET).
* Try to create useful static methods to clarify what each call to load_texture does. Really we may want more than 1 load_texture method, like create_empty_texture and so forth, instead of this method that does everything depending on hard to follow conditions.
* Avoid a pointless reallocation when a resize (that no-ops if unnecessary) and memset will do.
* Fix a bug (that somehow doesn't matter?) in the glyph atlas handling of the special fully white glyph 0 - it wasn't filling in the box it clearly meant to be. Instead, simplify further and just fill in the whole temporary buffer with white.
* Also fix a loop to render glyphs to go row by row instead of column by column - should improve memory locality. Not sure if it really will matter at all for e.g. startup perf, given that the whole working buffer of pixels is often say, 2KB which should comfortably fit in L1.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -2390,6 +2390,11 @@ tile_filter_scaling = false
         Setting it to true filters the textures resulting in a smoother but
         blurrier image.
 
+tile_use_mipmaps = auto
+        Used in conjunction with tile_cell_pixels or tile_sidebar_pixels.
+        Setting it to true filters the textures resulting in a smoother but
+        blurrier image.
+
 tile_force_overlay = false
         Setting this option to true will force the message window to appear as
         an overlay on top of the screen. If you use this option, making your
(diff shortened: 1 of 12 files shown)


Cgettys:
This PR stands on its own, so publishing it to get feedback / avoid the PR getting too far behind

Cgettys:
I profiled the release configuration, it doesn't make a huge difference. May improve debug startup time though, measuring that now.

Cgettys:
This may help debug tile builds a bit because of the explicit memset instead of a loop that most modern compilers are usually smart enough to turn into a vectorized loop / memory ordering improvement
Maybe ~50ms in debug tiles startup.

Cgettys:
Moving to draft until #4217 is merged

chucksellick:
Is this a duplicate of #4240 ?

Cgettys:
Should be, between that and #4217
let me close this.

PR #4211, opened 2025-01-06
fix: some warnings produced by header_build_tests

Cgettys:
* Add missing include to conduct-type.h to ensure it is self-contained (e.g. not depending on other headers happening to include tag-version.h before including it)
* Remove unnecessary semicolon.
* Add missing cstdint includes


DracoOmega:
Merged, thanks!

PR #4217, opened 2025-01-10
Remove GLU dependency

Cgettys:
Avoid depending on deprecated GLU library by dynamically loading
OpenGL 3+ function glGenerateMipmap where available.
If it's not available, Where it's not, we just render without mipmapping.

Note that for a lot of resolution / rc setting / zoom level combos,
we don't mipmap anyway, and this PR doesn't change those cases at all.

Also note that OpenGL 3.0 was released in 2008.
glGenerateMipmap is available even on decade old GPUs and iGPUs as a result.
For players running truly ancient GPUs and systems that supported GLU,
but not OpenGL 3, and who happened to have just the right setup for us to
enable mipmapping, we will now render without it.

If anyone actually does turn out to be affected, they may be able to use
a version of llvmpipe that supports OpenGL 3+ as a software renderer.
llvmpipe is available on Linux and Unix-like systems.

Also note we could turn on mipmapping on GLES targets according to
https://registry.khronos.org/OpenGL-Refpages/es3.0/html/glGenerateMipmap.xhtml.

This may be quite useful for low resolution screens, like older android phones.
But I'll leave that for a future commit.

### DIFF
--- a/crawl-ref/INSTALL.md
+++ b/crawl-ref/INSTALL.md
@@ -83,7 +83,7 @@ libsqlite3-dev libz-dev pkg-config python3-yaml binutils-gold python-is-python3
 
 # Dependencies for tiles builds
 sudo apt install libsdl2-image-dev libsdl2-mixer-dev libsdl2-dev \
-libfreetype6-dev libglu1-mesa libglu1-mesa-dev libpng-dev fonts-dejavu-core \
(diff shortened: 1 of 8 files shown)


Cgettys:
I may need to rework #4210 to accommodate - I'll do that after this is merged..

Implojin:
This seemed fine to me so I've merged it in de7caed, thanks again for your work here. :)

PR #4221, opened 2025-01-11
Let wizards give items to shops.

Aliscans:
Add a wizard mode command to give an item to a shop, to give a straightforward way to test things which involve a shop with a specific inventory.

Allow an item prompt to select any item from inventory or the floor. Rename the existing OSEL_ANY item selector as OSEL_PORTABLE to make the distinction clear. Use the new selector for "give item to shop" so books can be donated. It may be useful elsewhere, but I haven't looked.

### DIFF
--- a/crawl-ref/source/wiz-item.h
+++ b/crawl-ref/source/wiz-item.h
@@ -20,3 +20,4 @@ void wizard_unidentify_all_items();
 void debug_item_statistics();
 void wizard_recharge_evokers();
 void wizard_unobtain_runes_and_orb();
+void wizard_give_shop_item(const coord_def& where);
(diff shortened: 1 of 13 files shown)


crawl-ref/source/wiz-item.h
```diff
@@ -20,3 +20,4 @@ void wizard_unidentify_all_items();
 void debug_item_statistics();
 void wizard_recharge_evokers();
 void wizard_unobtain_runes_and_orb();
+void wizard_give_shop_item(const coord_def& where);
```
Cgettys, line 23: This file needs to include coord-def.h now - that's why the header build is unhappy (or at least one reason why)

PR #4223, opened 2025-01-11
feat: penance title for players abandoning with an invo title (DracoOmega)

Flugkiller:
This commit would make it so that whenever a theist player would abandon their god while having an invocations title, they would gain the penance title for as long as their wrath lasts. 
For flavor reasons, this would be an appropriate way to make these titles available in game, as most of them currently are unobtainable.

(I'm very open to feedback if the implementation of this is anything less than ideal.)


crawl-ref/source/skills.cc
```diff
@@ -2092,6 +2092,15 @@ string skill_title_by_rank(skill_type best_skill, uint8_t skill_rank,
...
         }
+        // Make sure the traitor induced title overrides under penance
+        if (you.attribute[ATTR_TRAITOR] > 0)
+            {
+                for (god_iterator it; it; ++it)
+                {
+                    if (player_under_penance(*it))
+                        result = god_title(*it, species, piety);
+                }
+            }
```
Monkooky, line 2103: This will choose semi-arbitrarily between traitor titles if the player has multiple penances; in particular this can get you a traitor title by abandoning one god to get ATTR_TRAITOR to 1, then slightly upsetting your current god to pass the player_under_penance. See comment on 755-760 of religion.cc

crawl-ref/source/religion.cc
```diff
@@ -752,6 +752,12 @@ void dec_penance(god_type god, int val)
         return;
     }
 
+    if (you.attribute[ATTR_TRAITOR] > 0)
+    {
+        you.attribute[ATTR_TRAITOR]--;
+        you.redraw_title = true;
+    }
+
```
Monkooky, line 760: Instead of using an attribute for number of gods betrayed, consider having an attribute for which god should grant the traitor title. 
During this step, instead of decrementing you.attribute traitor, check if there are any gods which you have penance for but are not presently worshipping, and set the traitorizor attribute to one of them, likely the one with the most remaining penance. This allows you to implement a logical rule for which title to use when abandoning multiple gods, instead of picking the last god in the god iterator you have penance for.

DracoOmega:
Note that, at the moment, if you have betrayed multiple gods, you will have the traitor title of whichever god is first *in enum order*, which is pretty opaque. "The one you betrayed first" makes the most sense to me (ie: save that god in ATTR_TRAITOR upon excommunication, as opposed to figuring it out later)

(A more minor issue is that if you abandon a god in a way that shouldn't result in your title changing, you nonetheless are briefly shown reverting to the default invo title in the middle, before immediately returning to the title you had prior to this.)

DracoOmega:
Merged, thanks!

PR #4224, opened 2025-01-11
refactor: enum related cleanups

Cgettys:
* Split explore_stop_options into its own header.
 * Make initfile parsing use maps instead of brute force else-ifs
 * Make the layers enum a enum class
 * simplify base movespeed logic in form_base_movespeed
 * Fix return type of viewmap_flash_clour and fix compiler warning.
 * Fix incorrect enum bitfiddling in viewmap_flash_colour.
Only the last bullet point has any behavioral change.

The intent of a135081b66271c96c04bea8a314a81b1ae5e9791
was to display the flash colour if all 4 layers were
on, or at least that's what the comment said.
The actual behavior was to display it if any of the 4 were on.

### DIFF
--- a/crawl-ref/source/explore-stop-options.h
+++ b/crawl-ref/source/explore-stop-options.h
@@ -1,0 +1,51 @@
+#pragma once
+
+enum explore_stop_options
+{
+    ES_NONE                      = 0x00000,
+
+    // Explored into view of an item that is NOT eligible for autopickup.
+    ES_ITEM                      = 0x00001,
+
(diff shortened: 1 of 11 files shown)


Cgettys:
Happy to rebase/update commit messages /squash, but I'm guessing not all of it will go in together, so I'ml eaving it split for now.

Cgettys:
Think this is merge-ready from my perspective, unless there's feedback otherwise. Squashed and fixed comment m essage

DracoOmega:
Merged (with a few conflicts from new code resolved), thanks!

PR #4228, opened 2025-01-12
ci: Automatically check commit messages against guidelines

Cgettys:
Add a python script to check line lengths of all commits not in master.

Should work properly even in the case where it's ran on master, though it's a bit pointless there.


wheals:
Thanks! This gave me the idea to also create a local git hook for checking the commit message, which I did in 8bdf87329e4c56dbe30c3d27a7538ab963cd80be

PR #4238, opened 2025-01-14
Fix: white splash screen when OpenGL 3+ unavailable (Ge0ff)

Cgettys:
I thought GL_TEXTURE_MIN_FILTER had the same value for both branches.

I thought incorrectly.

Fix it in the simplest possible way for now.
I'll clean this up more in a future PR.

Bug was introduced in #4217.

Credit to Ge0ff for reporting.


Cgettys:
merged with changes

Implojin:
As mentioned, merged with minor changes in 6518f178e, thanks again.

PR #4240, opened 2025-01-15
feat: Mipmapping wherever available, rcfile option

Cgettys:
* Make MipmapOptions an enum class
* Add a struct to encapsulate a bunch of options that always go together
* Use MipmapOptions instead of an opaque bool
* Add rcfile option to opt/in out of trying to mipmap, to override a weird heuristic we have for some images.
* Support mipmapping if GL_ARB_framebuffer_object extension is available too (basically all HW)
* Support mipmapping on OpenGL ES 2+ as well.
* Simplify the LoadTexture code (even covering subtexture + offset case we don't exercise today)


Cgettys:
I'd like to see folks test this on:
[ ] Android
[ ] Pre-OpenGL 3 but GL_ARB_framebuffer_object supporting HW
[ ] Modern HW

PR #4243, opened 2025-01-15
feat: Improve developer experience by reminding them ccache exists

Cgettys:
Validate that ccache is installed and that the compiler is symlinked to use ccache.

Update docs to point to a helpful answer on setting up ccache.


Cgettys:
Shouldn't cause build failure on CDO, should just warn, unless I screwed up or am misremembering

PR #4244, opened 2025-01-15
Don't freeze corpses that have been turned to gold

### LINKED ISSUE #4185: Gozag-worshipping White Draconian gets no gold from enemies killed by ice blast
Webtiles, Firefox (version almost certainly irrelevant)

0.32.1-5-gba85492886 

**DESCRIPTION:**
1. Be a white draconian with at least one breath charge
2. Kill one or more enemies with the ice breath ability
3. Notice `The foo shatters and turns to gold!` in the message log
4. Destroy or wait for the ice statue to disappear, notice no gold left behind. No gold is added to the total automatically, either (like what happens when an enemy dies over deep water).

**EXTRA FILES:**
~~Game still in-progress, but~~ EDIT: Log [here](https://cbro.berotato.org/morgue/slitherrr/morgue-slitherrr-20241225-090135.txt), its key is `slitherrr:cbr2:20241121015841S`

[Backup Save Here](https://cbro.berotato.org/saves/slitherrr-crawl-0.32-241222-0645.tar.bz2)

WizardIke:
When you killed a monster with the white draconians breath ability while worshipping Gozag, you would get a message saying it shattered and turned to gold but you would get a block of ice and no gold. Fix this by not freezing a monsters corpse if it has been turned to gold.

Fixes #4185

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -5179,8 +5179,11 @@ void bolt::kill_monster(monster &mon)
 
     item_def *corpse = monster_die(mon, ref_killer, kindex);
 
-    if (origin_spell != SPELL_GLACIATE && origin_spell != SPELL_GLACIAL_BREATH)
+    if (origin_spell != SPELL_GLACIATE && origin_spell != SPELL_GLACIAL_BREATH
+        || goldify)
+    {
         return;
+    }
 
     if (corpse)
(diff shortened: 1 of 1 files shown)


Cgettys:
This seems like a no-brainer to me - why should white draconian have this unexpected downside in combination with Gozag, after all.

pooka109:
Applied; thanks.

PR #4245, opened 2025-01-15
Fix Hep ancestors sometimes spawning incorrectly

### LINKED ISSUE #3300: Ancestor Identity doesn't work properly if used before ancestor spawns
console, playing on underhound.eu via ssh

0.30.1-4-g462c640d31

**DESCRIPTION:**
I noticed a minor bug with Hepliaklqana's "Ancestor Identity".

I was using Passwall to go through a wall and got enough piety for an ancestor (_"Hepliaklqana forms a fragment of your life essence into the memory of your ancestor"_) immediately when I emerged on the other side. The ancestor didn't spawn right away, but appeared three turns later. I used "Ancestor Identity" before the ancestor spawned for the first time, and when it did it didn't have the name I chose, instead it had the default random name. I had to change the name to something else and then back to make it apply.

There is a ttyrec recording of this happening (frame 30266) at https://underhound.eu/crawl/ttyrec/chujev/2023-09-11.21%3A06%3A06.ttyrec.bz2
Morgue: https://underhound.eu/crawl/morgue/chujev/morgue-chujev-20230912-200136.txt

### LINKED ISSUE #3450: Two Hep ancestors appear instead of one.
<Webtiles, Firefox>

<https://crawl.kelbi.org/crawl/morgue/SpiritOfHugs/SpiritOfHugs.txt>

**DESCRIPTION:**
<Heplikquana spawned two ancestors instead of one. Abilities only affect one of them, though so far we only used Remember Identity. It happened after a random teleport revealed enough tiles for us to get piety to get the ancestor, in dungeon 4. >

--> backup: http://crawl.kelbi.org/saves/SpiritOfHugs-crawl-git-93c622d222-240103-1725.tar.bz2

WizardIke:
If the player gained enough piety to gain an ancestor on an enemy's turn or at the start of their turn (e.g. from an enemy digging or the player's passwal finishing), they will get a turn to act before their ancestor is summoned as the ancestor will be summoned at the end of the player's turn. Delaying the summoning of the ancestor to the end of the player's turn is done by storing a copy of the ancestor in a queue. This lead to a bug reported by krajj7 (aka chujev) where changing your ancestor's name on this turn would result in them spawning with the old name. Fix this by using the actual data for the ancestor instead of the data in the queue.

Queueing up spawning an ancestor also lead to a bug reported by NephilaWeaver where a teleport trap could teleport you into unexplored terrain giving you enough piety for an ancestor and then piety delay could bring you below the piety required for an ancestor but fail to remove your ancestor as it hadn't spawned yet. Then at the end of your next turn, your ancestor would spawn even though you didn't have enough piety. Then when your got enough piety for an ancestor, a second one would spawn. Fix this by checking if spawning an ancestor is needed before spawning one from the queue.

Fixes #3300
Fixes #3450

### DIFF
--- a/crawl-ref/source/religion.cc
+++ b/crawl-ref/source/religion.cc
@@ -4690,6 +4690,17 @@ void delayed_monster_done(string success, delayed_callback callback)
     _delayed_done_callbacks.push_back(callback);
 }
 
+static bool _needs_to_spawn_hepliaklqana_ancestor()
+{
+    return you_worship(GOD_HEPLIAKLQANA)
+            // You don't get an ancestor under penence
+           && !player_under_penance()
+           // You must have enough piety for an ancestor
+           && have_passive(passive_t::frail)
+           // An ancestor must not already exist
+           && hepliaklqana_ancestor() == MID_NOBODY;
+}
+
 static void _place_delayed_monsters()
 {
     // Last monster that was successfully placed (so far).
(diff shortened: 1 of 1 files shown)


DracoOmega:
Merged, thanks!

PR #4247, opened 2025-01-16
fix: Bring Asterion's ability names in line with overhauled Makhleb

ahyangyi:
Asterion's current ability name, "Greater Servant of Makhleb", comes from Makhleb's old ability names before the 0.32 overhaul.
    
Since the overhaul was mostly merging the lesser/greater pairs while leaving the actual mechanics untouched, I think we can keep the     existing monster spells (along with icons and descriptions), but use the new ability titles.
    
The other ability, "Major Destruction", is contrasted with another monster spell "Legendary Destruction". It is thus not changed in this PR.


ahyangyi:
BTW, please let me know if you think it's better to also rename `Major Destruction` here. I'm erring on the "safe" side for now but willing to change.

PR #4248, opened 2025-01-16
Automatically identify some obsolete items.

Aliscans:
Add food and rods to the "always known" list in item_type_known(). This restores some behaviour from before _full_ident_mask() was deleted, in that you are no longer asked if you want to identify "removed food" items.


Implojin:
Thanks!

PR #4249, opened 2025-01-17
When you use mon-tile as a player-tile, display their changes in water

tianmaster:
A small modification, when you use Merfolk's Mons Tile as your player tile, it correctly displays their differential in water


Cgettys:
Seems like a good fix.
Might be other call sites we should consider whether need in_water argument passed, but this is a step in the right direction to me.

PR #4252, opened 2025-01-18
Improve reaping brand messaging.

### LINKED ISSUE #4251: Spurious messages when killing monsters using the reaping brand
local tiles, Ubuntu 24.04

0.33-a0-763-g7bd0b9f747

**DESCRIPTION:**
Every visible monster killed with the reaping brand (whether a weapon or the Skull of Zonguldrok) gives the message that the monster's soul was torn from its body, even if the monster is unspectralizable (such as an obsidian statue).

pooka109:
_make_derived_undead() now allows overriding the default success and failure messages (the "mist gathering" ones). Do this for kills using the reaping brand, so that its messages are only displayed when monsters are successfully spectralised (and visible to the player).

The failure message only applies to zero-headed hydras, which should be impossible to get with a reaping brand, but it's in place just in case a bladed weapon somehow gets both the flaming and reapling brands, or otherwise non-reapable monsters of living holiness get added.

Fixes #4251.

### DIFF
--- a/crawl-ref/source/mon-death.cc
+++ b/crawl-ref/source/mon-death.cc
@@ -1323,14 +1323,14 @@ static string _killer_type_name(killer_type killer)
 }
 
 static string _derived_undead_message(const monster &mons, monster_type which_z,
-                                      const char* mist)
+                                      string msg)
 {
     switch (which_z)
     {
     case MONS_SPECTRAL_THING:
     case MONS_SIMULACRUM:
         // XXX: print immediately instead?
-        return make_stringf("A %s mist starts to gather...", mist);
+        return msg;
     case MONS_SKELETON:
     case MONS_ZOMBIE:
         break;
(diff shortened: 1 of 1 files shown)


crawl-ref/source/mon-death.cc
```diff
@@ -1423,17 +1425,23 @@ static void _make_derived_undead(monster* mons, bool quiet,
     if (god == GOD_KIKUBAAQUDGHA || spell == SPELL_BIND_SOULS)
         mg.extra_flags |= (MF_NO_REWARD | MF_HARD_RESET);
 
-    const char* mist = which_z == MONS_SIMULACRUM ? "freezing" :
-                       god == GOD_YREDELEMNUL ? "black" :
-                       "glowing";
+    const string mist = which_z == MONS_SIMULACRUM ? "freezing" :
```
Cgettys, line 1428: Curiosity question: if you're making this a string anyway, why not make it concatenative for these two conditions?
A freezing black mist would be cool :D

DracoOmega: Yred reaping can't make simulacra, though. That param isn't for the player's god while raising the dead, but if the god themselves is the source.

Cgettys: Ah, gotcha, mutually exclusive then.

PR #4254, opened 2025-01-19
fix: overflow of overlays with Qazlal's Disaster area (prawnwizard)

Cgettys:
In rare cases, you can exceed the 20 overlay limit.

This leads to a crash.

To fix this:
* Deduplicate all overlays, not just the most recent
* Reduce overlay cap to 16 instead of 20 for vectorization reasons
* Make packed_cell fit into 196 bytes, because why not + cache alignment
* Make add_overlay an instance method for convenience

Modern clang versions auto-vectorize and unroll this std::find nicely. GCC doesn't do quite as well, but at least clones/unrolls it 4 wide. We'll see if it's a win or not.

### DIFF
--- a/crawl-ref/source/directn.cc
+++ b/crawl-ref/source/directn.cc
@@ -1352,8 +1352,7 @@ static void _draw_ray_cell(screen_cell_t& cell, coord_def p, bool on_target,
 {
 #ifdef USE_TILE
     UNUSED(on_target, p);
-    cell.tile.dngn_overlay[cell.tile.num_dngn_overlay++] =
-        _tileidx_aff_type(aff);
+    cell.tile.add_overlay(_tileidx_aff_type(aff));
 #endif
     const auto bcol = _colour_aff_type(aff, on_target);
     const auto mbcol = on_target ? bcol : bcol | COLFLAG_REVERSE;
@@ -1419,8 +1418,7 @@ void direction_chooser::draw_beam(crawl_view_buffer &vbuf)
         const bool inrange = in_range(p);
         auto& cell = vbuf(grid2view(p) - 1);
 #ifdef USE_TILE
-        cell.tile.dngn_overlay[cell.tile.num_dngn_overlay++] =
-            inrange ? TILE_RAY : TILE_RAY_OUT_OF_RANGE;
+        cell.tile.add_overlay(inrange ? TILE_RAY : TILE_RAY_OUT_OF_RANGE);
 #endif
         const auto bcol = inrange ? MAGENTA : DARKGREY;
         const auto cglyph = _get_ray_glyph(p, bcol, '*', bcol| COLFLAG_REVERSE);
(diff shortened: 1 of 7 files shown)


Cgettys:
The crash in question appears to have been pretty much a degenerate case - electrocution branch, Disaster Area, FogMachine Lua script???

But even so, might as well prevent this.

crawl-ref/source/tilecell.cc
```diff
@@ -82,9 +83,33 @@ enum wave_type
...
+    if (not_present)
+    {
+        const int insert_pos = num_dngn_overlay++;
+        // the lower bound is redundant, but helps e.g. clang
+        // realize that it can combine comparisons.
+        if (0 <= insert_pos && insert_pos < MAX_DNGN_OVERLAY)
+            dngn_overlay[insert_pos] = tileidx;
+        else
+        {
+            mprf(MSGCH_DIAGNOSTICS,
```
Implojin, line 107: We don't really use MSGCH_DIAGNOSTICS like this; if you look at some of the other instances in the code we mostly have diagnostics wrapped in ifdefs for debug builds. IIRC it displays in darkgrey(?) and we probably won't get many, if any, reports from players who see this.

Might be preferable to just keep an assert here so that we're at least notified if this case gets hit, even if it's technically only a UI problem?

crawl-ref/source/tilecell.h
```diff
@@ -15,72 +15,85 @@ struct packed_cell
 {
     // For anything that requires multiple dungeon tiles (such as waves)
     // These tiles will be placed directly on top of the bg tile.
-    enum { MAX_DNGN_OVERLAY = 20 };
-    int num_dngn_overlay;
+    enum { MAX_DNGN_OVERLAY = 32 };
```
Implojin, line 18: Are we capping at 32 or 16 here? I'm assuming the commit message reflects an earlier squashed change?

Cgettys: It's 32 here, you're right - thought I had switched back.
Powers of 2 seem to produce better code.
I was hesistant to both A) make it smaller and B) keep it an assert.
I think 16 should be enough for all cases. I can make it 16 + an assert if you like

crawl-ref/source/tilecell.cc
Cgettys: Made it an assert

crawl-ref/source/directn.cc
```diff
@@ -1432,8 +1430,8 @@ void direction_chooser::draw_beam(crawl_view_buffer &vbuf)
     // Only draw the ray over the target on tiles.
 #ifdef USE_TILE
     auto& cell = vbuf(grid2view(target()) - 1);
-    cell.tile.dngn_overlay[cell.tile.num_dngn_overlay++] =
-        in_range(ray.pos()) ? TILE_RAY : TILE_RAY_OUT_OF_RANGE;
+    cell.tile.add_overlay(
```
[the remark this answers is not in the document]
Cgettys: Like this?

crawl-ref/source/tilecell.h
Cgettys: Made it 16, fixed commit message and compile assert

crawl-ref/source/directn.cc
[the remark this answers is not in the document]
Implojin: Yup, thank you :)

Implojin:
Thanks for the fixes, and for the tests!

PR #4257, opened 2025-01-22
Add Oni's titles

CipHuK:
After Ogre was removed there no titles for Oni.
(I'm not an enthusiast of Japanese folklore, so can not suggest cool titles)

Also there were no Forgecraft titles, two are added for Oni


pooka109:
Applied; thanks.

PR #4261, opened 2025-01-23
ci: Generate docs in CI and publish to GitHub Pages

Cgettys:
* Generate Lua docs with ldoc
* Generate FAQ.html with make docs
* When on master, publish generated artifact to GitHub Pages
* Add a very rudimentary index.html to serve as landing page
* Once merged, https://crawl.github.io/crawl/ should show the new index.html
* With working links to FAQ.html and lua/index.html :)

One less task to do manually :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -669,6 +669,46 @@ jobs:
       - run: make -j$(nproc) monster
         working-directory: crawl-ref/source
 
+  build_docs:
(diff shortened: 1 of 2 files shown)


Cgettys:
See https://cgettys.github.io/crawl/ for what it'll look like as is. I know the index is about as boring of a page as you can get, but without it, you just get a 404 style page

PR #4264, opened 2025-01-26
fix: remove undefined behavior from map_cell

Cgettys:
Clang-tidy warns about these memcopies:
* Undefined behavior, source object type 'const map_cell' is not TriviallyCopyable
* Undefined behavior, destination object type 'map_cell' is not TriviallyCopyable

Pointer provenance / type-based aliasing analysis are part of the C++ standard, so this analysis seems correct.

* Remove the UB by changing to proper field-wise copies; compiler should optimize it pretty well, especially for the move case.
*  Simplify the operator== and != to do pointer comparison, as at present, the memcmp was comparing pointers that are unique per instance anyway. Leave a TODO about it.
* Also leave some TODOs for future work I'll do in a follow-up PR.
* Some general cleanup (e.g. ternaries where they make sense)

### DIFF
--- a/crawl-ref/source/map-cell.h
+++ b/crawl-ref/source/map-cell.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <libutil.h>
+
 #include "enum.h"
 #include "mon-info.h"
 #include "tag-version.h"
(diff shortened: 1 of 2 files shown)


wheals:
Merged in 00f9be72f1e76a6d9332deea553ec1e3d3fce03f with some followup code cleanup in ec844b119431a4cabd5d559dffcbcc40fea5bcd4

PR #4265, opened 2025-01-29
Fix swapping two-handed weapon with fly to one without

### LINKED ISSUE #4255: Unwielding 2-handed weapon prints messages twice
Webtiles @ CAO

0.33-a0-776-g802b72d (Console/Webtiles)

Likely due to the changes to how 2-handed weapons are handled.
It's not really noticeable with most weapons since it condenses to print just `x2`, but for those that print an extra message on unwield (such as protection brand) it doesn't condense it and prints both twice.

Image

Not a major problem, just figured I'd note it. Thanks for all the hard work :)

### LINKED ISSUE #4258: Artefact flight prevents switching equipment over deep water even when other items with flight are equipped
0.33-a0-783-g76bfe29 Webtiles

As seen in this screenshot, I had two randarts with the Fly property equipped, a weapon and an amulet.

Image

Despite this, while over deep water, I cannot switch weapons, as the game appears to overzealously think that removing one item with Fly would cause me to drown. 

Image

This prevents me from switching weapons in battle, which is a relevant and important tactical decision in some situations.

WizardIke:
When trying to swap a two-handed weapon with the flight property on it with another two-handed weapon without flight while wearing another item of flight, you would get a message saying that you would drown and be unable to swap weapons if you were above deep water. This was caused by the item swapping code trying to unequip the two-handed weapon twice (once to free the primary hand and once to free the offhand) which would mess up the count of equipped flight items.

Fixes #4258
Fixes #4255

### DIFF
--- a/crawl-ref/source/item-use.cc
+++ b/crawl-ref/source/item-use.cc
@@ -1344,6 +1344,23 @@ bool try_equip_item(item_def& item)
         {
             vector<item_def*>& candidates = removal_candidates[i];
 
+            // If one of the candidates has already been removed to free
+            // another slot, we don't need to remove anything to free this slot
+            bool slot_freed_when_freeing_other_slot = false;
+            for (const item_def* removed_item : to_remove)
+            {
+                for (const item_def* item : candidates)
+                {
+                    if (item == removed_item)
+                    {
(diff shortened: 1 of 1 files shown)


DracoOmega:
Merged, thanks!

PR #4266, opened 2025-01-29
Fix foxfire swarm using the wrong tile and power

WizardIke:
The foxfire swarm ability was creating foxfire with a power of GOD_NO_GOD (0) instead of 20 and marshlight set to 20 instead of its default value of false.


pooka109:
Applied; thanks.

PR #4272, opened 2025-01-29
Fix always invalidating the area grid when a monster dies

WizardIke:
We were checking if the radius of any of the monster's halo, umbra or silence were not equal to zero. However, they have a radius of -1 when they don't exist.


pooka109:
Applied; thanks.

PR #4275, opened 2025-01-30
Allow manifold assault with the Arc Blade

### LINKED ISSUE #3876: Can manifold assault with Arc Blade in offhand but not primary hand

Version: 0.32-a0-1468-g753a682cf

**DESCRIPTION:**

As a coglin you can manifold assault with a weapon in your primary hand and the Arc Blade in your offhand but not the other way around.

**EXTRA FILES:**

Screenshot (1703)
Screenshot (1702)

WizardIke:
The Arc Blade sometimes shoots out a number of arcs at monsters adjacent to you when you attack with it. This doesn't work very well when hitting a monster that isn't adjacent to you with manifold assault, so manifold assault didn't allow using it. This change makes the Arc Blade shoot an arc at the monster instead if it isn't adjacent to you and allows it to be used with manifold assault.

Fixes #3876


Hellmonk:
Currently, this prevents using manifold assault at all if an ally is potentially hittable. I believe it needs to use a stop_attack_prompt inside player_unrand_bad_attempt instead of returning directly. Should be mergeable w/ minor cleanup after that.

PR #4280, opened 2025-01-31
Fix: Reorder Forgewright starting items

LiciTheCrawler:
Reorders the starting items to be consistent with other backgrounds
with a similar kit.


gammafunk:
Merged, thanks for the PR.

PR #4285, opened 2025-02-01
Don't hang on SIGHUP in wizard_mod_tide().

Aliscans:
Make Crawl shut down correctly in some situations where it didn't already.

Make wizard_mod_tide() respond to seen_hups, rather than ignore it.
Make ui::delay() treat seen_hups as if the user had pressed the Escape key.


Cgettys:
Should everywhere that key_is_escape is called also check crawl_state.seen_hups?

Aliscans:
I wouldn't. get_ch() returns CK_ESCAPE if it receives a SIGHUP (with libunix.cc &ndash; I haven't checked other versions). This means that (say) activate_ability() already handles SIGHUP correctly.

It might be worth checking what Crawl does in that sort of situation, but most of the code works correctly.

PR #4291, opened 2025-02-04
Portuguese: Added abilities.txt

nam-ak:


powerbf:
Opened in error. Should not have been raised against upstream repo.

PR #4294, opened 2025-02-06
Fix incorrect max health and magic after melding items (ritt)

### LINKED ISSUE #4284: Max mp value unchanged after some mp+ items are melded by transformation
webtiles

 Dungeon Crawl Stone Soup version 0.33-a0-803-gcb9a0bf236 (webtiles)

**DESCRIPTION:**
base 56 + hat 10 + ring 9 = 75
...and still 75 in treeform (with hat melded.)

Image

Image

WizardIke:
After melding items that gave max health or magic your max health and magic would be set to the values for having the items unmelded. Also unmelding the items would set your max health and magic to the values for having them melded.

Fixes #4284


DracoOmega:
Merged, thanks!

PR #4295, opened 2025-02-07
bugfix: Zong text was ambiguous as to the equipment slot

bobross419:
One of the players noticed that the Skull of Zonguldrok's artefact text was unclear as to the equip slot which caused some problems with the 2H they were using.  This adds a "When held" to the second sentence to clarify.

### DIFF
--- a/crawl-ref/source/dat/descript/unrand.txt
+++ b/crawl-ref/source/dat/descript/unrand.txt
@@ -834,10 +834,10 @@ power to carry them out, no matter the adversity.
 %%%%
 skull of Zonguldrok
 
-The skull of a legendary necromancer, seemingly parted from the rest of his
-body. When held, it seems disinclined to explain the precise circumstances of 
-its ending up in this state, but offers no objection to others siphoning what 
-considerable necromantic power still remains in his ancient bones.
+The skull of a legendary necromancer, seemingly parted from the rest of its
+body. When held, it seems disinclined to explain the precise circumstances of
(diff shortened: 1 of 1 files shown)


crawl-ref/source/dat/descript/unrand.txt
```diff
@@ -835,9 +835,10 @@ power to carry them out, no matter the adversity.
 skull of Zonguldrok
 
 The skull of a legendary necromancer, seemingly parted from the rest of his
-body. It seems disinclined to explain the precise circumstances of its ending up
-in this state, but offers no objection to others siphoning what considerable
-necromantic power still remains in his ancient bones.
+body. When held, it seems disinclined to explain the precise circumstances of 
+its ending up in this state, but offers no objection to others siphoning what 
+considerable necromantic power still remains in his ancient bones
```
chucksellick, line 840: You lost the full stop from the end of the paragraph

chucksellick:
Apologies as I was going to add another comment the other day and didn't get round to it. The change is fine as it is but I brought this up on the Discord as I wondered if it seems awkward that there's such a mixture of pronouns ("it" vs "his") in the description, and this is perhaps amplified by the addition to the sentence.

We already have a canonically male skull (Murray) so perhaps the best thing is to reword it a bit more so it's all it/its and just leave it ambiguous; there's no reference anywhere else to Zonguldrok's gender (could also go with they/their but whatever works best).

PR #4301, opened 2025-02-11
fix: unistd.h inclusion for macOS 10.14

### LINKED ISSUE #4299: macOS 10.14 needs unistd.h include in initfile.cc to build
local tiles

0.33-a0-853-g30d10493ae

**DESCRIPTION:**
macOS 10.14 build needs unistd.h include added to initfile.cc in order to build.  Without it, I get:

% /usr/bin/make TILES=1 prefix=/usr/local
...
    CXX initfile.o
initfile.cc:4715:9: error: use of undeclared identifier 'readlink'
        readlink("/proc/self/exe", tempPath, sizeof(tempPath) - 1);
        ^
initfile.cc:5103:9: error: use of undeclared identifier 'unlink'; did you mean 'unlink_u'?
        unlink(filename.c_str());
        ^~~~~~
        unlink_u
./syscalls.h:39:5: note: 'unlink_u' declared here
int unlink_u(const char *pathname);
    ^
initfile.cc:5128:9: error: use of undeclared identifier 'unlink'; did you mean 'unlink_u'?
        unlink(out_name.c_str());
        ^~~~~~
        unlink_u
./syscalls.h:39:5: note: 'unlink_u' declared here
int unlink_u(const char *pathname);
    ^

WIth unistd.h added as an include (I put it after string, on line 27) everything builds just file.

FWIW this is not a new issue and has been the case for a while, I just wasn't motivated to submit the bug until now.

**EXTRA FILES:**
n/a

Cgettys:
Hopefully fix #4299.

I'm not entirely convinced all usage of unlink should not be unlink_u. But that's a separate discussion.

### DIFF
--- a/crawl-ref/source/initfile.cc
+++ b/crawl-ref/source/initfile.cc
@@ -88,9 +88,7 @@
 #include <shlobj.h>
 #elif defined(TARGET_OS_MACOSX)
 extern char **NXArgv;
-#ifndef DATA_DIR_PATH
 #include <unistd.h>
-#endif
 #elif defined(UNIX) || defined(TARGET_COMPILER_MINGW)
 #include <unistd.h>
 #endif


hbent:
I closed 4299, but this can/should be committed IMO.  It fixes the build and appears not to cause any problems for other OSes.

PR #4303, opened 2025-02-13
Do not let Vainglory prevent entering transporters or PoG

radio-gra:


DracoOmega:
Thanks for pointing out this issue! (I elected to fix it by moving the Vainglory check *after* the code for transporters/PoG instead.)

PR #4304, opened 2025-02-14
only deploy docs and try to send IRC notification on the main repo

geekosaur:
As currently written, the `deploy_docs` action runs on forks, which means anyone with a fork of the crawl repo gets a build failure when updating the fork. This patch makes it only run on the main crawl repo (hopefully).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -693,7 +693,7 @@ jobs:
   deploy_docs:
     name: Deploy docs
     needs: build_docs
-    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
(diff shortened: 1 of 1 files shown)


geekosaur:
Hm, nope, wrong context item. Whoops.

geekosaur:
While I'm at it, make the same check in "notify IRC". I haven't seen that one fire, but I suspect it's failing on forks anyway.

Cgettys:
Good call - there's nothing stopping anyone who wants to from enabling GH Pages on their fork to try out this part of the pipeline (which is why I didn't add this check when I added GH Pages publishing), but probably better to leave it off by default.

PR #4305, opened 2025-02-15
style: a few cleanups in Beogh apostle code

Cgettys:
Add a named constant to make it clear why some of these loops start from 1.
Fix a few typos and add comments to help myself or anyone else in the future.
Make a few parameters const where they already could be.


DracoOmega:
Merged, thanks!

PR #4311, opened 2025-02-17
fix: Ash "Shatter the Chains" not giving unequip_item a chance to do its job

### LINKED ISSUE #4309: Persistent halo when orb of light is shattered
Webtiles, Safari 18.2, cbro

 Dungeon Crawl Stone Soup version 0.33-a0-932-g1afe1e7492 (webtiles)

**DESCRIPTION:**
When orb of light was shattered, the halo remained stuck to the location.  Saving and reloading cleared the visual bug.

I did not check the visual description of the tiles to see if they showed haloed.

**EXTRA FILES:**
Image

Cgettys:
The item was being destroyed before unequip_item was called.
As a result, the call to unequip_item did nothing at all.
Because the item was already OBJ_UNASSIGNED,
rather than being the specific item that unequip_item
needed to handle.

This resulted in several bugs:
* Visual artifacts (halo/umbra/etc)
* Likely various messages not being printed
* Possibly other persistent issues?
E.g. update_vision_range not being called.
as well as potentially other bugs.
For example, update_vision_range would not be called.

Closes https://github.com/crawl/crawl/issues/4309

### DIFF
--- a/crawl-ref/source/god-abil.cc
+++ b/crawl-ref/source/god-abil.cc
@@ -2410,12 +2410,13 @@ bool ashenzari_uncurse_item()
         canned_msg(MSG_OK);
         return false;
     }
-
     mprf("You shatter the curse binding %s!", item.name(DESC_THE).c_str());
     item_skills(item, you.skills_to_hide);
-
     vector<item_def*> to_remove = {&item};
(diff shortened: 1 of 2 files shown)


DracoOmega:
Merged (with edits). Thank again!

PR #4312, opened 2025-02-17
ci: fix checkconventionalcommit.py

Cgettys:
Python version in CI doesn't like line 36.
My local version of Python is fine with it.
Split the statement in 2 to satisfy more Python versions.

Also exclude certain branches from main repo.
We're not gonna fix past commit messages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,8 @@ jobs:
       - name: Unbrace
         run: util/unbrace -n
         working-directory: crawl-ref/source
-      - name: Checkconventionalcommit
+      - name: Checkconventionalcommit # If not in crawl/crawl, or if it's a branch other than master or the release branches, run
+        if: (github.repository != 'crawl/crawl' || !(github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/stone_soup-')))
         run: util/checkconventionalcommit.py
         working-directory: crawl-ref/source
         continue-on-error: true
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -25,7 +25,8 @@ jobs:
       - name: Unbrace
         run: util/unbrace -n
         working-directory: crawl-ref/source
-      - name: Checkconventionalcommit
+      - name: Checkconventionalcommit # Run on all branches outside of crawl/crawl, except master and release branches which have too many past violations
+        if: (github.repository != 'crawl/crawl' || github.ref != 'refs/heads/master' || !startsWith(github.ref, 'refs/heads/stone_soup-'))
```
chucksellick, line 29: Should this be && rather than || ?

crawl-ref/source/util/checkconventionalcommit.py
```diff
@@ -33,7 +33,8 @@ def main():
     print(f"Checking commits being merged from {branch_to_check} to {base_branch}")
 
     git_args = [ "git", "log", "--no-merges", pretty, branch_to_check, f"^{base_branch}" ]
-    print(f"Running: {" ".join(git_args)}")
+    command_to_print= " ".join(git_args)
```
chucksellick, line 36: Minor formatting nitpick, should have a space left of the =

.github/workflows/ci.yml
Cgettys: nope, it's
Run if not crawl/crawl
||
Run if not master
||
Run if not release branch

crawl-ref/source/util/checkconventionalcommit.py
Cgettys: Fixed

.github/workflows/ci.yml
Cgettys: oh, this can't be quite right, you're right.

Cgettys: Fixed to what I had been trying to write - non-crawl/crawl, or dev branch basically

PR #4313, opened 2025-02-18
Making stairs findable

Dzeddy:
## Summary
I seek to improve how stashes handle staircases and escape hatches, ensuring they are properly tracked and updated in the stash system for proper findability

## Changes
* Renamed parameter `feature` to `feat` in `is_boring_feature()` for consistency
* Fixed formatting in `is_boring_feature()` return statement for better readability
* Added explicit stash tracking for staircases and escape hatches
  * When updating stashes at a location, check if it's a staircase/hatch
  * Create or update stash entries for these features

### DIFF
--- a/crawl-ref/source/stash.cc
+++ b/crawl-ref/source/stash.cc
@@ -216,11 +216,15 @@ bool Stash::needs_stop() const
     return false;
 }
 
-bool Stash::is_boring_feature(dungeon_feature_type feature)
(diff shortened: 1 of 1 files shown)


gammafunk:
I think this modification is a good idea, so I've implemented it an easier way. Your implementation added a one-off stash creation in a method that's only supposed to update. The actual stash creation method simply looks at whether the feature is considered "boring", so making stairs/hatches not boring is a better way to go.

PR #4316, opened 2025-02-18
Use plural verbs in _xom_time_control().

Aliscans:
Give "you and everything else in sight" the plural verbs it deserves.

Split some long lines.


pooka109:
Applied; thanks.

PR #4320, opened 2025-02-20
Wall Monsters part 1

chucksellick:
The first couple of wall-dwelling monsters: Rock Fish and Wolf Lichen (and a Strolling Stone which sheds the latter, based on the Moss Beast from regret-index's notes).


regret-index:
Dev feedback from both DracoOmega and myself was provided for this previously, and while while we like the strolling stone concept and tile, their lichen summons are strange and ill-suited to deploy on a visual, flavour, and mechanical level. Wall-walking was fixed up from this branch and merged in dea4f60, rock fish were adjusted and merged in 21aa791, and additional monsters further away from early D deploy the new ability in both e37ab6d and 9f71ee0. As such, opening up a separate PR for another take on strolling stones would be a good idea in the future.

PR #4323, opened 2025-02-21
console: draw waypoints over other items and features

radio-gra:
Currently, waypoints are only displayed on otherwise empty squares on the level map view. In combination with them not being xv-able and not reporting any sort of coordinates other than what level they're on, this makes it so that if anything overrides them, they become impossible to find without actually traveling to their exact location.

For my use case, this is a very annoying problem - I use waypoints to track timed portals and mark encounters with monsters that indicate potential branch entrances (e.g. redback on L:3), and in general for annotations that are more location-sensitive where annotating the whole floor doesn't cut it. Especially when scouting portals, being able to see exactly where the waypoints are at all times is very important. I understand that x-marks-the-spot may not the intended use case for waypoints, but I suspect it to be by far the most common one.

### DIFF
--- a/crawl-ref/source/viewmap.cc
+++ b/crawl-ref/source/viewmap.cc
@@ -313,6 +313,13 @@ static void _draw_level_map(int start_x, int start_y, bool travel_mode,
                 if (travel_mode && travel_colour_override(c))
                     cell->colour = _get_travel_colour(c);
 
+                // If we've a waypoint on the current square,
+                // show the waypoint number.
+                if (uint8_t waypoint_char = is_waypoint(c))
+                {
+                    cell->glyph = waypoint_char;
+                }
+
                 if (c == you.pos() && !crawl_state.arena_suspended && on_level)
                 {
                     // [dshaligram] Draw the @ symbol on the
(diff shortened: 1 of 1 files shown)


crawl-ref/source/viewmap.cc
```diff
@@ -313,6 +313,13 @@ static void _draw_level_map(int start_x, int start_y, bool travel_mode,
                 if (travel_mode && travel_colour_override(c))
                     cell->colour = _get_travel_colour(c);
 
+                // If we've a waypoint on the current square,
+                // show the waypoint number.
+                if (uint8_t waypoint_char = is_waypoint(c))
```
Cgettys, line 318: Assignment inside if is generally considered bugprone.

And crawl uses a odd convention for braces: https://github.com/crawl/crawl/blob/3f0725400bbc8228084413c86b80aa763b9a73aa/crawl-ref/docs/develop/coding_conventions.md#d-braces
I think something like this might be clearer?
```cpp
uint8_t waypoint_char = is_waypoint(c)
// If we've a waypoint on the current square,
// show the waypoint number. 0 == not a waypoint
if (waypoint_char)
    cell->glyph = waypoint_char;
```

Cgettys:
Overall, I think this is a good change. If it proves to be controversial, a rcfile option to make it configurable depending on player taste is simple enough and would make everyone able to get what they want.

radio-gra:
amended

PR #4331, opened 2025-02-26
Fix the area grid not being updated on monster item unequip

WizardIke:
The `monster::unquip` function was checking if it changed the monster's halo and umbra radii and invalidateding the area grid if it did. However, because it only runs any effects that need to happen on unequip and doesn't actually remove the item from the monsters equipment slots, it doesn't change the halo or umbra radii. Instead we should check if the halo and umbra radii have changed in the places that actually remove the items.


pooka109:
Applied; thanks.

PR #4337, opened 2025-02-28
Allow coglins to benefit from two spectral weapons (jejorda2)

### LINKED ISSUE #3719: Coglin wields two spectral weapons, but only one processes.
online console CBRO
Morgue at https://cbro.berotato.org/morgue/jejorda2/morgue-jejorda2-20240320-125017.txt

0.32-a0-791-gfc315c9ecd

**DESCRIPTION:**
When my Coglin wields two spectral demon tridents and attacks an enemy in an open area, I expect to see two summoned floating spectral demon tridents beside my character attacking my enemy.

But I only see one.  Sometimes it is named for my primary weapon, and sometimes it is named for my off-hand weapon. But I never see both, and I think that I should see both.

WizardIke:
They used to only be able to bring out one spectral weapon when they were attacking with one spectral weapon in each hand.

Fixes #3719

### DIFF
--- a/crawl-ref/source/tags.cc
+++ b/crawl-ref/source/tags.cc
@@ -1315,6 +1315,110 @@ static void _shunt_monsters_out_of_walls()
     }
 }
 
+static bool _is_spectral_weapon(const item_def& weapon)
+{
+    return get_weapon_brand(weapon) == SPWPN_SPECTRAL
+           || is_unrandom_artefact(weapon, UNRAND_GUARD);
+}
+
+static void _fix_player_spectral_weapon()
+{
+    if (!you.props.exists(SPECTRAL_WEAPON_KEY))
+        return;
+
+    mid_t weapon_mid = you.props[SPECTRAL_WEAPON_KEY].get_int();
+    you.props.erase(SPECTRAL_WEAPON_KEY);
+
+    monster* spectral_weapon = monster_by_mid(weapon_mid);
+    if (!spectral_weapon)
+        return;
+
+    vector<item_def*> weapons = you.equipment.get_slot_items(SLOT_WEAPON);
+    weapons.erase(remove_if(weapons.begin(), weapons.end(),
+        [](const item_def* w) { return !_is_spectral_weapon(*w); }),
+        weapons.end());
+
+    item_def* spectral_item = spectral_weapon->mslot_item(MSLOT_WEAPON);
+
+    if (weapons.empty() || !spectral_item)
+    {
+        monster_die(*spectral_weapon, KILL_RESET, NON_MONSTER, true);
+        return;
+    }
+
+    // Because the spectral weapon monster holds a copy of the weapon and
+    // the weapon can be changed afterwards (e.g. by being inscribed), we may
+    // not be able to find an exact match.
+    item_def* best_match = weapons[0];
+    for (size_t i = 1; i < weapons.size(); ++i)
+    {
+        item_def* weapon = weapons[i];
+
+        if (weapon->sub_type == spectral_item->sub_type
+            && best_match->sub_type != spectral_item->sub_type)
+        {
+            best_match = weapon;
+            break;
+        }
+
+        if (weapon->plus == spectral_item->plus
+            && best_match->plus != spectral_item->plus)
+        {
+            best_match = weapon;
+            break;
+        }
+
+        std::string spectral_item_name = "";
+        if (spectral_item->props.exists(WEAPON_NAME_KEY))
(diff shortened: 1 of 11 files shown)


crawl-ref/source/tags.cc
```diff
@@ -1315,6 +1315,98 @@ static void _shunt_monsters_out_of_walls()
...
+    if (!spectral_weapon)
+        return;
+    item_def* spectral_item = spectral_weapon->mslot_item(MSLOT_WEAPON);
+    if (!spectral_item)
+        return;
+
+    // Because the spectral weapon monster holds a copy of the weapon and
+    // the weapon can be changed afterwards (e.g. by being inscribed), we may
+    // not be able to find an exact match.
+    item_def* best_match = weapons[0];
```
Cgettys, line 1346: Is there ever a case where weapons could be length zero, or would that be a bug?

crawl-ref/source/tags.cc
```diff
@@ -1315,6 +1315,98 @@ static void _shunt_monsters_out_of_walls()
...
+static void _fix_player_spectral_weapon()
+{
+    if (!you.props.exists(SPECTRAL_WEAPON_KEY))
+        return;
+
+    mid_t weapon_mid = you.props[SPECTRAL_WEAPON_KEY].get_int();
+    you.props.erase(SPECTRAL_WEAPON_KEY);
+
+    vector<item_def*> weapons = you.equipment.get_slot_items(SLOT_WEAPON);
+    for (size_t i = weapons.size(); i > 0;)
```
Cgettys, line 1327: Perhaps comment that this indexing and ordering is to avoid iterator invalidation issues and overflow?

crawl-ref/source/tags.cc
Cgettys: Ah, right, we'd return above.

crawl-ref/source/tags.cc
```diff
@@ -1315,6 +1315,98 @@ static void _shunt_monsters_out_of_walls()
...
+        if (weapon->plus == spectral_item->plus
+            && best_match->plus != spectral_item->plus)
+        {
+            best_match = weapon;
+            break;
+        }
+
+        std::string spectral_item_name = "";
+        if (spectral_item->props.exists(WEAPON_NAME_KEY))
+            spectral_item->props[WEAPON_NAME_KEY];
```
Cgettys, line 1367: Curiosity: are these setting props via default construction? I don't see what the weapon_name has to do with the props setting right below it, for example.

crawl-ref/source/tags.cc
```diff
@@ -1315,6 +1315,98 @@ static void _shunt_monsters_out_of_walls()
...
+            break;
+        }
+    }
+    best_match->props[SPECTRAL_WEAPON_KEY].get_int() = weapon_mid;
+}
+
+static void _fix_spectral_weapons()
+{
+    _fix_player_spectral_weapon();
+    for (monster_iterator mi; mi; ++mi)
```
Cgettys, line 1394: Curiosity: what does the second mi do here?

crawl-ref/source/tags.cc
WizardIke: This does look right at all. Its meant to assign to spectral_item_name.

crawl-ref/source/tags.cc
WizardIke: It checks if we still have monsters to iterate over (monster_iterator has a conversion to bool operator). This is the normal code for iterating over all the monsters.

Cgettys: Thanks for the clarification... should have realized that it has a conversion to bool... gotta love implicit conversions.

crawl-ref/source/tags.cc
WizardIke: I replaced the loop with remove_if.

crawl-ref/source/tags.cc
WizardIke: I think that you could end up with no spectral weapon if you had a unrand artifact that used to have spectral but doesn't anymore. Now that I think about this, we do need to clean up the spawned weapon in this case as it won't clean itself up if its not linked up to an equipped item.

WizardIke: The spawned spectral weapon should now be cleaned up when you aren't wielding a spectral weapon.

gammafunk:
This looks good, so merged with one follow-up commit. Thanks for the PR! We'll need to keep an eye out for crashes given the nature of these changes.

A future smallish project might be to make dual spectral weapons work for monsters as well. Not that useful, and it might be best to generalize some aspects of monster equipment first.

PR #4338, opened 2025-03-01
Fix a crash when initializing Nobody's memories

WizardIke:
We were accessing a vector out of bound for some game seeds.


DracoOmega:
I may be missing the obvious here (and your change is arguably clearer anyway), but can you explain *how* it can access the vector out of bounds. Is it not just copying the last member into the chosen slot and then removing the last member, each time?

WizardIke:
`random_choose_weighted` does not return an index, it returns the `first` member of the `std::pair` it chooses.  The `weights` vector starts with the `first` members of the `std::pair`s it contains set to the indices of the elements of `memories`. This means as we shrink the `weights` vector some of the `first` members of its `std::pair`s will no longer be valid indices into the `weights` vector.

gammafunk:
> `random_choose_weighted` does not return an index, it returns the `first` member of the `std::pair` it chooses. The `weights` vector starts with the `first` members of the `std::pair`s it contains set to the indices of the elements of `memories`. This means as we shrink the `weights` vector some of the `first` members of its `std::pair`s will no longer be valid indices into the `weights` vector.

The `weights` vector doesn't start with the `first` members of the `std::pair`s it contains set to the indices of the elements of `memories`, its `first` members are in fact indices of the elements of `_recollections`. To that end, since `_recollections` is never changed by this code (it's in fact `const`), those indices don't have their validity change regardless of what changes in `weights` (or `memories`). We in fact use the indices stored in `weights` to directly index `_recollections` and not `weights` itself, which is just a stack variable, so there shouldn't be a crash. At least, that's my understanding after reading this code.

Now, I'm a bit less sure of the idiom (common practice?) @DracoOmega used here to delete an element. Apparently something like this would work and hopefully be more clear:
```cpp
weights.erase(weights.begin() + index);
```
If that does work, that should probably be the change we make.

WizardIke:
Oh I meant index in `_recollections` not `memories`. The problem line is `weights[index] = weights.back()` as `index` is an index into `_recollections` and weights will be shorter after we remove values. 

When I had this crash the debugger told me that index was 16 and that the length of weights was 16.

DracoOmega's code for erasing an element looks like the "swap and pop" idiom and should be fine.

gammafunk:
Yep, you're right. The indexing later into the full data is not a problem, but using it to index back into `weights` during selection is. Merging this, thanks for the fix.

For anyone interested in the element deletion aspect, my proposed code above using `vector::erase()` does compile and would be shorter/preferred way to erase an element of a vector from a (valid) index.

PR #4340, opened 2025-03-01
Avoid crash when clockwork bee falls in deep water

chucksellick:
Prevents a segfault which would happen when the bee's target disappears from view at the same time as the bee is flying over a body of deep water (or lava). The case that the "inert" (and flightless) bee failed to place, due to no solid ground being near enough, wasn't handled and execution fell through to the next block where it was assumed a target did exist.

Rather than needing to figure out exactly why the bee won't place to have messages like splashing in the water, the bee can just self- destruct; it's the same result.


DracoOmega:
Merged (with small tweaks). Thanks for the fix!

PR #4341, opened 2025-03-01
Add some themed little-v vaults to big-V Vaults

nicolae-carpathia:
Most of them are using the themes regret-index whipped up a while back, but a few have more bespoke themes.

Edit: I'm realizing as I'm pushing this to github that we usually do one file per commit. Mea maxima culpa.


Hellmonk:
These seem good to me. Really like the library vault.

nicolae-carpathia:
Somehow missed that comment. Sounds good enough for me.

Issue #4344, opened 2025-03-02
Imp suck-up speech error


croikle:
**PLATFORM and OS or BROWSER:**
console, observed on cbro

**DCSS VERSION AND GAME INFORMATION:**
0.33-a0-1030-g170f05ed26

**DESCRIPTION:**
I summoned an imp and received this message:

` Your cerulean imp says, "I yield to your fealty, my grandiloquent and {{string
_"db_embedded_lua":7: 'then' expected near 'return'}} ban!"`

It looks like 1bb2c40a1794882406599fa7479d4ff789190ac6 lost a `then` in the elseif chain.

[/crawl-ref/source/dat/database/monspeak.txt:350](https://github.com/crawl/crawl/blob/master/crawl-ref/source/dat/database/monspeak.txt#L350)

PR #4345, opened 2025-03-03
Fix imp _suck_up_adj2_ syntax

croikle:
Imp speech `_suck_up_adj2_` throws an error because of a missing `then`. Restore correct syntax.

Should fix https://github.com/crawl/crawl/issues/4344.


pooka109:
Applied; thanks.

Good catch.

PR #4347, opened 2025-03-04
Rework robe of misfortune (again).

Hellmonk:
Based on a comment on a very old pr. Now provides triple wizardry and 15 MP, but has a flat 5% chance to give a miscast effect even if the spell succeeds. Bonuses were chosen to try and make it a casting item orthogonal to most of the existing casting armours, which tend to provide lots of spellpower. The other drawbacks of the robe are removed.

### DIFF
--- a/crawl-ref/source/art-data.txt
+++ b/crawl-ref/source/art-data.txt
@@ -859,11 +859,15 @@ OBJ:      OBJ_ARMOUR/ARM_ROBE
 # try to make it bad, this is the only bad SPARM:
 FB_BRAND: SPARM_PONDEROUSNESS
 PLUS:     5
+INSCRIP:  Wiz+++ *Miscast
 COLOUR:   LIGHTMAGENTA
 TILE:     urand_misfortune
 TILE_EQ:  robe_misfortune
-EV:       5
-BOOL:     harm, slow, corrode
+EV:       3
+MP:       15
+DESCRIP:  Wiz+++:    It greatly improves the wearer's spellcasting success.
+ *Miscast:    The wearer will sometimes suffer miscast effects even
+ when they successfully cast a spell.
 
 # start TAG_MAJOR_VERSION == 34
 NAME:     cloak of Flash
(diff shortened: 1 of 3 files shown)


crawl-ref/source/spl-cast.cc
```diff
@@ -2248,6 +2248,12 @@ spret your_spells(spell_type spell, int powc, bool actual_spell,
                            you.experience_level,
                            "the malice of Vehumet");
         }
+        else if (you.unrand_equipped(UNRAND_MISFORTUNE) && one_chance_in(20))
+        {
```
Cgettys, line 2252: nit: isn't there an extra brace here?

crawl-ref/source/spl-cast.cc
```diff
@@ -2248,6 +2248,12 @@ spret your_spells(spell_type spell, int powc, bool actual_spell,
...
+
+            mprf("You feel unlucky.");
+            miscast_effect(spell, 1 + random2(20) + random2(21));
```
Cgettys, line 2255: Shouldn't we have a miscast_source::unrand or something?

crawl-ref/source/spl-cast.cc
```diff
@@ -2248,6 +2248,12 @@ spret your_spells(spell_type spell, int powc, bool actual_spell,
                            you.experience_level,
                            "the malice of Vehumet");
         }
+        else if (you.unrand_equipped(UNRAND_MISFORTUNE) && one_chance_in(20))
+        {
+
+            mprf("You feel unlucky.");
```
pooka109, line 2254: mpr() will do here, rather than mprf(), since it's printing a constant string.

crawl-ref/source/art-data.txt
```diff
@@ -859,11 +859,15 @@ OBJ:      OBJ_ARMOUR/ARM_ROBE
 # try to make it bad, this is the only bad SPARM:
```
pooka109, line 859: Since the reworked robe has fewer drawbacks, maybe the comment here no longer applies, or no longer applies as much?

Hellmonk:
Closing, seems unlikely to be accepted.

PR #4349, opened 2025-03-06
Flaming/freezing warning trigger (alt #3868)

Hellmonk:
I think this version has neater messaging. It does require changing the current message for cold vuln, however. Long term that might be better for simpler pattern matching? Either way, one of these two prs should be merged and the other closed.


NormalPerson7:
I definitely agree that adding "terribly" to the damage message itself is neater. "The cold freezes you terribly" sounds a bit odd to me but probably it's a compromise worth making for consistency.

Hellmonk:
Dracoomega suggested "chills you terribly" as an alternative for the beam message (and then can pattern match "you terribly" to catch all 3 messages in rcfile)

PR #4350, opened 2025-03-06
Updated ghost crab claws unrand.

Hellmonk:
Revises #2745. Unrand gloves that prevent weapon or shield use, but give a chance for spectral mist clouds on melee or spell cast. Also enhances necromancy. Tried to incorporate some ideas and feedback from the original pr, and made the numbers scale with damage and spell level respectively. Still needs a tile.


crawl-ref/source/dat/descript/unrand.txt
```diff
@@ -859,3 +859,12 @@ crown of vainglory
 An extravagant crown of gold filigree and glittering jewels that even a king
 could feel immodest wearing. It grants its wearer a regal aura the match of any
 in heaven and also a great deal of unearned confidence.
+%%%%
+ghost crab claws
+
+A large pair of ghostly claws fashioned into gloves by a fenstrider coven for
+use in necromantic rites. The magically inclined and pugilistically favored
```
pooka109, line 866: This bit should use the Australian spelling of "favoured."

crawl-ref/source/art-data.txt
```diff
@@ -1901,6 +1901,21 @@ DESCRIP: Majesty:   You are bathed in light that prevents you from turning
  enemies and refuse to leave the floor again for some time.
 +Ring++:   You can equip two additional rings.
 
+ENUM:    CRAB_CLAWS
```
pooka109, line 1904: Shouldn't this enum section be at the very end of art-data.txt, just before DUMMY2?

Hellmonk: I believe we keep scarf of invisibility last because it's a "fake" unrand (only shows up in a sprint map). At least, recent unrands have been added just before it.

crawl-ref/source/dat/descript/unrand.txt
Hellmonk: fixed

crawl-ref/source/art-data.txt
pooka109: Okay. Keeping enum compatibility can get rather confusing.

PR #4359, opened 2025-03-10
Replace -Cast with *Silence

Hellmonk:
-Cast was very polarizing; it tended to either do nothing or make the item unusable. Try out a more dynamic negative effect along the lines of *Corrode; 3% chance for temporary silence when damaged. Silence duration is identical to blueberry klown pie. This effect works on monsters (who ignored -Cast entirely) in the rare case that one spawns with a *Silence artifact.

-Cast remains exclusively on mad mage's maulers, where it's necessary for the gimmick.


NormalPerson7:
Just a balance point: silence prevents the reading of scrolls and the use of god abilities, as well as spells, so this property goes from polarising (-Cast) to universally very dangerous. I wonder about reducing the rate this triggers from 3% down to 1% to match *Slow - what do you think?

PR #4363, opened 2025-03-10
fix plain revenants spawning in zig

yrdzrfxndfvh:
old monster revenants are now revenant soulmongers


pooka109:
Applied; thanks.

Issue #4352, opened 2025-03-07
Multi-attack monster continues attack even when target is replaced by friendly


croikle:
**PLATFORM and OS or BROWSER:**
console

**DCSS VERSION AND GAME INFORMATION:**
0.33-a0-1053-gd93d3284e2, further experiments on 0.32.1 wizmode

**DESCRIPTION:**
When a friendly monster with multiple attacks kills its target, causing a new friendly monster to spawn in the same location, the remaining attacks will strike this friendly target.

Observed for:

- Attackers: hydra, bunyip, juggernaut
- Friendly-target generators: Yred zombie-raising aura, Animate Dead, Death Channel, Infestation, Yred Bind Soul (any others to try?)

Note that this only occurs if the new friendly monster appears in the same tile.

Expected behavior:

The attacker will not attack the friendly target, or the friendly target will not be generated until all attacks are complete (perhaps like 11965a5a82f8fff7eeb6627bf469b57702e0d6c0?).

**EXTRA FILES:**
```
 You lift the black torch aloft and begin your conquest of Level 2 of the Orcish
 Mines in Yredelemnul's name!
 Yredelemnul sends servants to aid you!
 Your bound seven-headed hydra bites the orc!
 The orc dies!
 Its soul becomes fuel for the torch.
 The dead are walking!
 Your bound seven-headed hydra bites your orc zombie!
 Your orc zombie is destroyed!
```

PR #4365, opened 2025-03-11
Fix reusing monsters while still in use

WizardIke:
When a monster makes multiple attacks in `fight_melee` we need to fire final effects after each attack to handle blinking and banishment from distortion weapons properly (see mantis issue 6936). However, the memory for dead monsters is reused when firing final effects and there are a lot of places that call `fight_melee` (usually indirectly) that hold on to pointers to monsters and expect them not be be reused (`fight_melee` itself and `handle_monsters` also expect monsters not to be reused after final effects but are easy to fix). To fix this, don't reuse the memory of dead monsters when firing final effects, instead use a seperate function that is called to free up dead monsters and allow them to be reused that we call in places where we know its safe to do so.

Partial fix for #4352 although it doesn't handle the case where a monster just changes to friendly.


Cgettys:
This seems like a good thing to do to me - but why not go with reference counting (shared_ptr or otherwise) instead?

crawl-ref/source/mon-death.cc
```diff
@@ -4042,3 +4058,15 @@ bool mons_bennu_can_revive(const monster* mons)
     return !mons->props.exists(BENNU_REVIVES_KEY)
            || mons->props[BENNU_REVIVES_KEY].get_byte() < 1;
 }
+
+// Allows the memory used by dead monsters to be reused
+void free_dead_monsters()
```
Cgettys, line 4063: do we need to call this so many places? or could we do something like RCU, defer cleaning up dead monsters by like, a aut or something and have a new list every aut?

WizardIke:
> This seems like a good thing to do to me - but why not go with reference counting (shared_ptr or otherwise) instead? Edit: or something like RCU where we're not trying to clean up quite so promptly, but promptly enough (see comment elsewhere)

Reference counting would require changing everywhere to not use raw pointers/references or do it manually and if I was fine with changing everywhere I would probably just check for mid mismatches when checking if a monster is alive in the places that use `fight_melee`. RCU would have the same problem as you would have to say when you acquire or release a monster.

crawl-ref/source/mon-death.cc
WizardIke: The problem is debug commands that don't take turns (or in game time). We have some tests that run these in a loop allocating and killing a lot of monsters. Also, I presume you meant turn not aut, an aut is the in game unit of time where your average turn takes roughly 10 aut to complete.

Cgettys: Turn or aut, either could be viable for RCU - the point is that it's deferred by at least one "pass" such that nothing could still be accessing it
https://www.kernel.org/doc/html/latest/RCU/rcu.html

Cgettys:
Fair enough. Yes, moving to reference counting would be more work, just wondering if this is "enough" to largely fix these issues, or if it's not aggressive enough. Dunno, you may know better than I.

Cgettys:
We could also consider some sort of arena allocator generational approach I suppose, hmmmm.

crawl-ref/source/mon-death.cc
WizardIke: Yeah that actually makes sense for RCU. We could also just call it whenever we got low on free monsters to avoid problems with running out when no turns or auts pass.

WizardIke:
> Fair enough. Yes, moving to reference counting would be more work, just wondering if this is "enough" to largely fix these issues, or if it's not aggressive enough. Dunno, you may know better than I.

We use mids (which are basically weak references to monsters) when we want to reference monsters over multiple turns. This has the advantage that you can still reuse the memory for a monster even when someone is holding onto its mid. In fact, we could use mids to fix ours problems here, but that would require rewriting everywhere that uses `fight_melee` (and pretty much everywhere seems to use it indirectly).

PR #4366, opened 2025-03-11
Make monsters stop attacking when their target becomes friendly

WizardIke:
For example if you have an allied zombie hydra and it defeats an orc challenger sent after you by Beogh, it shouldn't use its remaining attacks to attack the now friendly orc.

Partial fix of #4352

### DIFF
--- a/crawl-ref/source/fight.cc
+++ b/crawl-ref/source/fight.cc
@@ -509,6 +509,8 @@ bool fight_melee(actor *attacker, actor *defender, bool *did_hit, bool simu)
     // Melee combat, tell attacker to wield its melee weapon.
     attacker->as_monster()->wield_melee_weapon();
 
+    bool was_hostile = !mons_aligned(attacker, defender);
+
     int effective_attack_number = 0;
     int attack_number;
     for (attack_number = 0; attack_number < nrounds && attacker->alive();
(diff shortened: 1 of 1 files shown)


crawl-ref/source/fight.cc
```diff
@@ -517,10 +519,11 @@ bool fight_melee(actor *attacker, actor *defender, bool *did_hit, bool simu)
         if (!attacker->alive())
             return false;
 
-        // Monster went away?
+        // Monster went away or become friendly?
         if (!defender->alive()
             || defender->pos() != pos
-            || defender->is_banished())
+            || defender->is_banished()
+            || was_hostile && mons_aligned(attacker, defender))
```
chucksellick, line 526: Wondering if there should be a check for if the monster is confused, berserk etc.? If they're not in control of their senses it wouldn't make sense for them to be aware to stop

WizardIke: That makes sense to me. I believe we only need to check for confused and frenzied as that's what friendly monsters attacking the player checks (but this code could be buggy). We don't need to check marionette as that changes the monster's attitude and I don't think we need to check for vexed as that doesn't call `fight_melee` (although maybe we should check it anyway in case it does in the future?).

PR #4367, opened 2025-03-12
Slimy Shroud Jiyva mutation (redux)

Hellmonk:
Loosely based on the removed shroud of golubria spell, but with a simplified design. It has a 1/4 chance to negate a melee attack, after which it goes on cooldown for a while depending on the damage prevented.


pooka109:
Since this adds a Shroud status, shouldn't there also be a description of the Shroud status in source/dat/descript/status.txt?

PR #4374, opened 2025-03-15
Fix a crash when mirrored Nobody uses pyrrhic recollection

### LINKED ISSUE #4369: Nobody Pyrrhic Recollection Crash

Played on berotato webtiles using Brave.

0.33-a0-1084-g778f541198
**DESCRIPTION:**

Encountered Nobody, mirrored Nobody and deployed box of beasts. Nobody tried to do the ignite memory thing and the game crashed. I have not played recently enough to know what Nobody does so could not tell what it was attempting to do.
**EXTRA FILES:**

https://cbro.berotato.org/morgue/shadowrider38/crash-shadowrider38-20250314-011059.txt

WizardIke:
During pyrrhic recollection we would end all enchantments on nobody which would end the summon timer on them, killing them and cleaning up their mid, we would then try and revive them using their mid which would crash. To fix this, set them as pending revival before ending their enchantments so they cannot be killed and then add back their summon timer if needed.

Fixes #4369


WizardIke:
Fixed in commit 85c0d3f

PR #4378, opened 2025-03-15
feat: New Unrand, Boots of Scraped Shins

Cgettys:
A stabber friendly boot unrand.

Intended to be a stab enabler.

* +50 Stealth
* 7 base damage - about what a kick does, or a base dagger
* Has a 1/3 chance to confuse, like Fungal Fisticloak (but minus the unbreathing limitation)
* Usual XL scaling - roughly 50% chance at XL1, nearly 100% chance at XL27

Currently steals the tiles of the old Boots of the Assassin.

Name needs bikeshedding.


WizardIke:
I really like the idea of dance shoes you can kick with. That chance of paralyses at 27 short blades does seem a bit high to me though (given how fast short blades attack and that its basically an insta kill with them), but maybe its ok for a rare artifact?

Cgettys:
Hellmonk suggested it as a random-ish starting point. Might be too strong, not sure yet.  

Spriggan's knife has a 50% chance of upgrading a low tier stab to a high tier stab for comparison - or a 50% chance per attack to make a low tier stab basically insta-kill.

Cgettys:
Lowered chance to 6 + SK_SHORT_BLADES - 20% chance at level 14 sbl (Quickblade mindelay), 33% chance at level 27 sbl

Monkooky:
It seems to me like there's two disjoint gimmicks going on here, they're thematically incoherent, and both have serious issues.

One, you have Stealth+ Gauntlets of War on your feet (highly enchanted boots you can kick people with)
Two, you have boots that paralyze when you make an attack.

Technically, you need to hit with the aux attack for the paralyze chance to come up, but hit chances get high enough that this is a minor concern, and the extra damage from the kicks isn't going to be noticeably related to the paralyze going off. 

So let's talk about each gimmick.

If you stripped the paralyze and just had +3 boots with a 5 damage aux attack, they would be near-equivalent to boot slot Gauntlets of War, and thus the default equip for most 3-rune chars. 
The big problem with this is that it's very good in a very boring way- you've got high enchant boots that make you do more damage! There's never a reason _not_ to wear +3 boots of ~+5 slay, and it takes a major highroll to find _better_ boots for any melee char.


If you made the boots less generically good, so you'd only wear them if you were interested in the on-hit para- the on-hit para is incredibly good for absolutely any melee char.
Para gives you a free turn, a guaranteed hit, and (at 1h mindelay with minimal stealth investment) double damage. Against any normal enemy, paralyze should be thought of as an instakill, shortblade or no. Against a unique, ood, or high tier enemy, it's probably only taking out 1/2-2/3 of their health unless you're using a shortblade.
If you're not using a shortblade, these boots will roughly translate to +10-15% total damage with trivial investment in stealth and shortblades- not even accounting for the free turns paralysis gives you. This is extremely strong.
With a shortblade, it's outlandishly strong. 16-25% instakill chance is not reasonable. _10%_ irresistable para on hit is probably too high.



So where to take this unrand:
First, the question of para-on-hit. 
I don't think a raw chance of para-on-hit can be low enough odds to be reasonable strength without it _feeling_ very bad and being a different kind of uninteresting. You almost certainly want one of the following:
A low tier stab condition instead (by far the simplest option, and probably the best for making this actually stabber focused as low-tier stabs need dex or stealth skill to stab)
A more restrictive activation condition (on-damage instead of on-hit _might_ be restrictive enough, but hard to say)
A severe drawback to wearing the item (requires designing a whole new gimmick to slap on the thing)
Regardless of how you nerf this, you probably don't want to scale the stab-enabling off of short blades skill- stabs themselves already scale off short blades skill, and it feels like an unintuitive (and broadly unsuccessful) kludge to try to force this to be stabber oriented.

Second, the question of taking Gauntlets of War off your feet.
Much easier. Do some subset of:
Dumpster the AC so you're making an actual (likely bad) tradeoff if you don't care about the stabby stuff
Make the aux attack no damage so you've just got decent boring boots if you don't care about the stabby stuff
Give the item a drawback so it's dubiously worth it if you don't care about the stabby stuff

PR #4382, opened 2025-03-18
Sign and notarize app bundle on macOS

aronson:
## Overview

These additional steps and modifications allow for developers to sign (with an Apple Developer certificate), notarize (server-side virus check from Apple), and staple (attach notarization result to .app bundle) the .app bundles that are shipped for macOS for the software.

## Benefits

Users will no longer face Gatekeeper lockout when launching the app. In the current state of the released application, one must go into System Settings and manually override the current, invalid signature to load the application. They must be an administrator of the system to do this. In the future, software may be even harder to distribute on macOS without a valid signature. 

Users will be able to immediately launch the application on any supported Mac after download.

Users will have confidence the software does not contain malware per Apple's required server-side notarization process.

## Drawbacks

Without a current certificate, the build will fail. Reverting this PR would be required to restore functionality. The error messages in this scenario would explicitly inform developers the certificate has expired.

Certificates cost $100/yr.

## Removals

I have removed the .zip creation in the `mac/Makefile.app-bundle` script. A .zip for release can only be created on macOS after the signature is applied, notarization passes, and the attestation ticket is available to staple to the bundle. These steps are post-build steps and I've found it easiest to perform them in the GitHub action instead of segmenting the build.

## Additions

A number of GitHub Actions secrets must be added for this to function. I am willing and able to provide values for these secrets on my personal development account with Apple, which I do for [another large, FOSS video game project](https://apotris.com), as I trust in the goodwill of the development organization behind this repo and wish to help support the greater Roguelikes community. I would prefer (as may you) for you to provide your own values. You must have or obtain an Apple Developer license for this to function.

- `APPLE_ACCOUNT_USERNAME`: your Apple account ID, which is also found as your iCloud account email
- `APPLE_ACCOUNT_PASSWORD`: an app-specific password if you use 2FA, or your iCloud account password otherwise
- `APPLE_ACCOUNT_TEAM_ID`: the Team ID associated with your apple account, this may be different from your personal Team ID if the organization forms a team through the Apple Developer web interface
- `APPLE_BUILD_CERTIFICATE_BASE64`: a base64-encoded string of the contents of your development certificate
  - given the certificate is installed in your Keychain Access.app, you must expand it and highlight both the certificate and its associated key, and export it to .p12 format with a secure password
  - from there one can run e.g. `base64 -i Certificates.p12 | pbcopy` and paste it into the secret
- `APPLE_P12_PASSWORD`: the password used to encrypt the exported certificate from the secret above
- `APPLE_DEVELOPER_ID_INSTALLER`: the string identifier for your developer certificate, in my case: "Developer ID Application: Isaac Aronson"
- `APPLE_KEYCHAIN_PASSWORD`: any random, secure string -- only used briefly within the runner in-place

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -304,6 +304,8 @@ jobs:
       - name: Set up Homebrew
         id: set-up-homebrew
         uses: Homebrew/actions/setup-homebrew@master
+      - name: Install modern Bash on macOS
+        run: brew install bash
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0 # all history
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -368,8 +370,67 @@ jobs:
...
+          target_type=${target%-universal}
+          BUNDLE_NAME="Dungeon Crawl Stone Soup - ${target_type^}"
+          MACOS_BINARY=$(if [[ "$target" == "console-universal" ]]; then echo "crawl"; else echo "${BUNDLE_NAME}"; fi)
+          echo "BUNDLE_NAME=$BUNDLE_NAME" >> $GITHUB_ENV
+          echo "MACOS_BINARY=$MACOS_BINARY" >> $GITHUB_ENV
+      - name: "Generate signed .app for macOS"
+        working-directory: crawl-ref/source
+        shell: bash
+        run: |
+          # Create p12 file
```
Cgettys, line 388: Can we put this in a SH file with env variables as input or something?

aronson: I'm not sure what the benefit is exactly. Is this how the other pipelines are doing this?

aronson: Thinking about this more, it would mean debugging the build when the certificate expires or the notarization server is down is not immediately obvious. I don't think this is a wise idea.

Cgettys: Benefit would be local development could use same script to validate this works. I wish bash had nicer support for e.g. scripts that take named arguments. But fair enough

Cgettys: You're probably right that it's not worthwhile.

aronson:
Just mentioning, the checks will fail until someone installs the required secrets into the repo.

paschun:
The homebrew casks have been slated to be removed in a year due to failing the gatekeeper checks

https://github.com/Homebrew/homebrew-cask/commit/82e6ca77519a4ae48f33e7aa9af612d0d8416dd5
https://github.com/Homebrew/homebrew-cask/commit/95bec080fb8d50e9023f33b09d534b9decf6458f

struffl:
There doesn't seem to be much interest in this from maintainers who are both versed in macOS and able to make decisions here. I'm going to close this PR.

PR #4387, opened 2025-03-20
Yaktaur Cleric & Spirit Yak

chucksellick:
(Firstly, this is definitely not ready for merge, but ready for feedback at least. In particular, *where* these should appear is very much an open question, as is how these will mesh with other devs plans around undead, )

Two new yaktaur threats: a necromantic Cleric, and a Spirit Yak psychopomp. The cleric is able to summon "siege wall" ancestor ghosts which can block LoF of even piercing shots, and additionally can sacrifice its own (living, but nearly dead) allies to gain self regeneration. The Spirit Yak applies an aura of "spirit arrow" to allies which gives them a 2nd free shot every time a shot connects (and partially bypassing armour).

These are both very strong for the time when yaktaurs first start showing up, and the spirit yak can be very strong in *any* ranged group. So I am thinking the final state of things would be (the implementation is part way there but needs some tweaking): both cleric and spirit yak can show up in yaktaur captain bands and vault warden bands, as will scribes (this involves adding a separate band type for these stronger bands); and cleric's own more necromantic-flavoured band type to start showing later in the depth range (or perhaps just in Vaults, as a bridge into the undead branches). Then spirit yak can be placed in a few branches with a band of local archers which may or may not be undead (Snake, Elf, Crypt, Tomb, Tar all seem like reasonable candidates ... and perhaps Underworld once that exists?)


chucksellick:
Closing as various aspects of this didn't work in current form, will salvage parts of it later.

PR #4388, opened 2025-03-20
New spell: Detonation Catalyst

Hellmonk:
Playing with buff duration and melee on-hit effects. Detonation Catalyst is a level 5 Fire/Alchemy spell that causes melee hits to do an ignition-style shaped explosion centered on their primary target. Does not work on cleaving attacks, but does work with reach, WJC move attacks, rampage, electric charge, manifold assault. The duration is extremely short (30-50 aut), but attacking prevents the duration from decreasing. Also, a large percentage of the spell's damage comes from the weapon's base damage. Please charge into enemies while wielding dark maul I am begging you, it will be really funny.


pooka109:
Since this adds a Catalyst, shouldn't there also be a description of the Catalyst status in source/dat/descript/status.txt?

crawl-ref/source/melee-attack.cc
```diff
@@ -670,6 +678,9 @@ bool melee_attack::handle_phase_hit()
         return false;
     }
 
+    // Detonation catalyst should trigger even if the defender dies later on.
+    maybe_trigger_detonation();
```
Cgettys, line 682: Should this be before the slimify if statement above?

PR #4392, opened 2025-03-21
Fix swapping between Release and Debug builds in MSVC

WizardIke:
Some of the dlls that crawl depends on in the MSVC build weren't always being rebuilt when swapping between Release and Debug configurations, which resulted in a crash when e.g. loading release crawl.exe with debug libpng.dll.


crawl-ref/source/MSVC/crawl.vcxproj
```diff
@@ -99,6 +99,7 @@ python.exe util/gen-all.py</Command>
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <SubSystem>Windows</SubSystem>
       <TargetMachine>MachineX86</TargetMachine>
+      <AdditionalLibraryDirectories>$(ProjectDir)..\contrib\bin\$(PlatformName)\Debug;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
```
Cgettys, line 102: Can we just parameterize the \Debug bit and not have this be conditional? I don't love the duplication.

crawl-ref/source/contrib/MSVC/SDL.vcxproj
```diff
@@ -84,6 +77,18 @@
     <CodeAnalysisRules Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
+    <CustomBuildAfterTargets>PostBuildEvent</CustomBuildAfterTargets>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <CustomBuildAfterTargets>PostBuildEvent</CustomBuildAfterTargets>
+  </PropertyGroup>
```
Cgettys, line 85: Can just remove the condition and not repeat 4x

crawl-ref/source/contrib/MSVC/SDL_image.vcxproj
```diff
@@ -91,6 +84,18 @@
     <CodeAnalysisRules Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
     <CodeAnalysisRuleAssemblies Condition="'$(Configuration)|$(Platform)'=='Release|x64'" />
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
```
Cgettys, line 87: ditto, can avoid duplication

Cgettys:
Overall, seems reasonable to me.

crawl-ref/source/MSVC/crawl.vcxproj
WizardIke: I don't think we can as we need to strip the Tiles or Console part off crawl's configuration when determining the contrib configuration

crawl-ref/source/contrib/MSVC/SDL.vcxproj
WizardIke: Actually, I vaguely remember reading that all properties need to be in conditionals for editing to work properly in Visual Studio

Cgettys: Ugh... that's silly

Cgettys: I'm used to always writing them manually, so to me, less bloat == better

PR #4399, opened 2025-03-23
bugfix: AUTOFIGHT while immotile should not wait a turn (CrawlOdds)

### LINKED ISSUE #4394: Tab passes a turn with -move
webtiles

trunk

**DESCRIPTION:**
When the player has the -move status, there are no monsters in weapon range, and they press tab, they pass a turn.

The correct behaviour here is probably the "no reachable target in view" warning, or similar.

Cgettys:
If immotile (e.g. from -Move), we should not wait when you try to autofight by default.
No other case that I can see will wait unless AUTOFIGHT_WAIT is set - why is this one different?

This logic was added last year in 52037ea7a925a6ae6b1b67939267cd8f1f4837a4.
I'm guessing because it was confusing to hit tab and have nothing happen? but mpr would make more sense to me to solve that.
Gate the old behavior behind autofight_wait.

Could add another option for it instead, but it seems consistent with the behavior of the autofight_wait option in general.

Closes #4394

### DIFF
--- a/crawl-ref/source/dat/clua/autofight.lua
+++ b/crawl-ref/source/dat/clua/autofight.lua
@@ -204,12 +204,14 @@ local function move_towards(dx, dy)
   local move = choose_move_towards(0, 0, dx, dy, can_move_now)
   if move == nil then
     crawl.mpr("Failed to move towards target.")
-  else
-    if you.status("immotile") then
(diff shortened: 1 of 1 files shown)


Cgettys:
@elliptic - I believe you added the original logic, would welcome feedback from you on this PR.

PR #4401, opened 2025-03-24
feat: Let tornado and the browser handle compression for us

Cgettys:
Even quite older versions of tornado can handle compression for us when supported by client.
 Won't be faster or slower particularly, but less branchy and cleaner code. 

Secondarily, https://www.rfc-editor.org/rfc/rfc7692 has been around for close to a decade now. 
There's wide browser support. 

Let's decompress in well optimized browser code, rather than JavaScript. Less JavaScript to send to clients, more readable websocket messages. 
Lower default compression level while we're here - earlier testing showed default (currently 6) is likely way past diminishing returns for us.


crawl-ref/source/webserver/webtiles/ws_handler.py
```diff
@@ -396,27 +400,12 @@ def __hash__(self):
     def __eq__(self, other):
         return self.id == other.id
 
-    def allow_draft76(self):
-        return True
-
-    def select_subprotocol(self, subprotocols):
-        if "no-compression" in subprotocols:
```
Cgettys, line 403: default implementation will do fine / better

PR #4402, opened 2025-03-24
Mark maxed out misc evokers as useless if not in player's inventory

radio-gra:
Existing code only marks useless misc evokers as such if they're in shops, but if they're just laying on the floor they are currently not considered useless. This is a nuisance in zigs, which have one or more evokers on almost every floor, quickly maxing them out and adding more visual noise to an already overloaded ctrl-X menu.

The hacky inventory check is not ideal, but I couldn't come up with a better solution without going out of my depth and messing with core functions. I checked all edge cases I could think of like standing on +5 evo while having a copy in your inventory

### DIFF
--- a/crawl-ref/source/item-name.cc
+++ b/crawl-ref/source/item-name.cc
@@ -3317,15 +3317,15 @@ bool is_useless_item(const item_def &item, bool temp, bool ident)
     }
 
     case OBJ_MISCELLANY:
-        // Try to discourage players from wasting money on a useless evoker in a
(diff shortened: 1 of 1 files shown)


NormalPerson7:
As a temporary measure until this (or something equivalent) is merged, the rcfile option
`menu_colour += pickup:darkgray:\+5 (phial|tin of tremor|condenser|box of beast|sack of spider|phantom|Gell|lightning rod)`
will do the trick for decluttering Ctrl-X

PR #4406, opened 2025-03-26
Increasingly re-roll low quality shop items deeper in the dungeon

chucksellick:
Adds some fairly simple logic to decide "trash" items and give a chance to re-roll them if picked by a shop, based on your depth.

This doesn't kick in until *after* absolute depth 10, i.e. Orc:$ ... as the loot there is already good for the stage of the game and we don't want to upset early balance.

After this the chance steadily increases until finding a shop in late Depths or Zot means most trash items will get re-rolled.

This works rather differently from normal "item level" code (which still operates in tandem) by more aggressively evicting items which are likely to have been found or are already obselete and would be boring to see again.

This will also be a nice bump to Gozag shops at later XLs which are frequently considered to not be great value after the first few even though they have built in bonus to item level and rolling "good" items.

### DIFF
--- a/crawl-ref/source/dungeon.cc
+++ b/crawl-ref/source/dungeon.cc
@@ -72,6 +72,7 @@
 #include "tilepick.h"
 #include "tileview.h"
 #include "timed-effects.h"
+#include "transform.h"
 #include "traps.h"
 #include "unique-creature-list-type.h"
 #ifdef WIZARD
@@ -6130,6 +6131,132 @@ static bool _valid_item_for_shop(int item_index, shop_type shop_type_,
 }
 
 /**
+ * Is this item reasonably worthless after the early/mid game? Shops get an
+ * increasing chance to reroll such items the further down we go.
+ */
+static bool _item_is_trash(item_def& item)
+{
+    // Never trash! (Maybe we could apply *some* logic to random artefacts, but
+    // it's really hard to say what may or may not be useful to a given player.
+    if (is_artefact(item))
+        return false;
+    switch (item.base_type)
+    {
+    case OBJ_SCROLLS:
+        switch (item.sub_type)
+        {
+        case SCR_NOISE:
+        case SCR_IDENTIFY:
+            return true;
+        default:
+            return false;
+        }
+    case OBJ_POTIONS:
+        switch (item.sub_type)
+        {
+        case POT_MOONSHINE:
+        case POT_LIGNIFY:
+            return true;
+        default:
+            return false;
+        }
+    case OBJ_WANDS:
+        switch (item.sub_type)
+        {
+        case WAND_FLAME:
+        case WAND_DIGGING: // Questionable
+            return true;
+        default:
+            return false;
+        }
+    // Get a bit more choosy. A lot of base jewellery types are boring once
+    // you've seen one, but ones with variable plusses can still be good.
+    case OBJ_JEWELLERY:
+        switch (item.sub_type)
+        {
+        case RING_SLAYING:
+        case AMU_FAITH:
+        case AMU_REGENERATION:
+        case AMU_MANA_REGENERATION:
+            return false;
+        case RING_STRENGTH:
+        case RING_INTELLIGENCE:
+        case RING_DEXTERITY:
+        case RING_PROTECTION:
+        case RING_EVASION:
+        case RING_MAGICAL_POWER:
+            return item.plus < 5;
+        default:
+            // All other rings and amulets
+            return true;
+        }
+
+    case OBJ_WEAPONS:
+    {
+        const int rarity = weapon_rarity(item.sub_type);
+        // Rarest weapons are great whatever the stats. Uncommon weapons need
+        // to have a good plus or a brand.
+        return rarity > 6 || rarity > 3 && item.brand == SPWPN_NORMAL && item.plus < 5;
+    }
+
+    case OBJ_MISSILES:
+        // Get rid of the most boring missiles
+        switch (item.sub_type)
+        {
+        case MI_STONE:
+        case MI_BOOMERANG:
+            return true;
+        case MI_JAVELIN:
+            return item.brand == SPMSL_NORMAL;
+        case MI_DART:
+            return item.brand == SPMSL_POISONED;
+        default:
+            return false;
+        }
+    case OBJ_ARMOUR:
+        switch (item.sub_type)
+        {
+        case ARM_ORB:
+            // Non-artefact orbs.
+            return true;
+        case ARM_BARDING:
+            // Barding is rare enough let's allow it even if mundane
+            return false;
+        default:
+            // For everything else, use a bit of logic.
+            return !armour_is_special(item) && item.brand == SPARM_NORMAL
+                   // Eliminates plain +2 hats or boots, but not +5 plate armour
+                   && item.plus < 5;
+        }
+    case OBJ_TALISMANS:
+    {
+        // Keep the higher-tier non-artefact talismans
+        const transformation form_type = form_for_talisman(item);
+        const Form* form = get_form(form_type);
+        // Max skill seems to be a better indication of tier than min skill.
+        // This trashes blade and serpent talismans. Review once more forms
+        // are added.
+        return form->max_skill < 20;
+    }
+    case OBJ_STAVES:
+        // Non-artefact staves aren't very interesting. They don't even generate
+        // in shops right now anyway.
+        return true;
+    case OBJ_MISCELLANY:
+    case OBJ_BOOKS:
+        // Miscellany could be trash if we tracked how many were generated
+        // so far in the dungeon, but we're not doing that.
+        // Nothing can really be done with books; once single spells are
+        // sold in shops, maybe trash spell level < 5.
+        return false;
+    default:
+        // Should never reach here unless a new item type is added; at least
+        // fail quickly if that happens.
+        ASSERT(false);
+    }
+}
+
+/**
  * Create an item and place it in a shop.
  *
  * FIXME: I'm pretty sure this will go into an infinite loop if env.item is full.
@@ -6153,6 +6280,9 @@ static void _stock_shop_item(int j, shop_type shop_type_,
 
     int item_index; // index into env.item (global item array)
                     // where the generated item will be stored
+
+    int rerolls = 0;
+    bool no_trash = false;
 
     // XXX: this scares the hell out of me. should it be a for (...1000)?
     // also, it'd be nice if it was just a function that returned an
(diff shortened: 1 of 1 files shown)


crawl-ref/source/dungeon.cc
```diff
@@ -6129,6 +6130,132 @@ static bool _valid_item_for_shop(int item_index, shop_type shop_type_,
...
+        case AMU_MANA_REGENERATION:
+            return false;
+        case RING_STRENGTH:
+        case RING_INTELLIGENCE:
+        case RING_DEXTERITY:
+        case RING_PROTECTION:
+        case RING_EVASION:
+        case RING_MAGICAL_POWER:
+            return item.plus < 5;
+        default:
```
NormalPerson7, line 6189: I don't really understand what this is trying to achieve. Str/Int/Dex rings already only spawn at +6, protection at +4 and evasion at +5, so at present this bans rings of protection and keeps everything else.

crawl-ref/source/dungeon.cc
```diff
@@ -6129,6 +6130,132 @@ static bool _valid_item_for_shop(int item_index, shop_type shop_type_,
...
+            return true;
+        case ARM_BARDING:
+            // Barding is rare enough let's allow it even if mundane
+            return false;
+        default:
+            // For everything else, use a bit of logic.
+            return !armour_is_special(item) && item.brand == SPARM_NORMAL
+                   // Eliminates plain +2 hats or boots, but not +5 plate armour
+                   && item.plus < 5;
+        }
```
NormalPerson7, line 6230: How does this interact with antique shops? Currently it's very often the case that you see e.g. a "glowing cloak" in a shop and you buy it hoping for an ego and it's just a boring +1 cloak. This will potentially significantly change decision making around purchasing such items late in the game (possibly in a good way).

crawl-ref/source/dungeon.cc
```diff
@@ -6129,6 +6130,132 @@ static bool _valid_item_for_shop(int item_index, shop_type shop_type_,
...
+    case OBJ_TALISMANS:
+    {
+        // Keep the higher-tier non-artefact talismans
+        const transformation form_type = form_for_talisman(item);
+        const Form* form = get_form(form_type);
+        // Max skill seems to be a better indication of tier than min skill.
+        // This trashes blade and serpent talismans. Review once more forms
+        // are added.
+        return form->max_skill < 20;
+    }
```
NormalPerson7, line 6240: This seems about right for me for talismans.

crawl-ref/source/dungeon.cc
```diff
@@ -6129,6 +6130,132 @@ static bool _valid_item_for_shop(int item_index, shop_type shop_type_,
...
+        case RING_INTELLIGENCE:
+        case RING_DEXTERITY:
+        case RING_PROTECTION:
+        case RING_EVASION:
+        case RING_MAGICAL_POWER:
+            return item.plus < 5;
+        default:
+            // All other rings and amulets
+            return true;
+        }
```
NormalPerson7, line 6192: I'm a bit worried that designating the vast majority of base rings as "useless" here will result in endgame shops being stacked full of slay/str/int/dex/EV/MP rings, which won't be very interesting.

In particular, I would argue rings of rF+, rC+, Will+ as critical resistances that can be stacked (possibly also rN+, rPois, rCorr although those are clearly less important), and wizardry, are also very useful late in the game and shouldn't be regarded as obsolete.

NormalPerson7 (commented):
I definitely like the idea of this change and I have certainly experienced lategame shops (in particular weapon and armour shops) being rather unexciting.

I've left a few specific comments regarding specific items being included/excluded from the list. One other brief point: you mention this not affecting Orc:2 but Orc:2's absdepth is 12, so if I understand the code correctly, bad items already have a 1/3 chance to be rerolled on Orc:2, so you may want to tweak the depth threshold there.

crawl-ref/source/dungeon.cc
chucksellick: Obviously this was a first pass and plenty needs tweaking. I'm actually wondering if it should go *further* and class all base ring types as "trash" because this gives you rolls at *artifact* rings instead which can contain those and other properties. But I'll test the theory and see what the shops look like.

crawl-ref/source/dungeon.cc
chucksellick: What happens is that (for the very late artifact shops), *most* of the items are glowing something or other ... so yes it becomes more of a gamble, but you have a few items to choose from, so maybe buying 2 or 3 things you'll find something nice (and obviously boring +1 cloaks are mostly evicted).

crawl-ref/source/dungeon.cc
chucksellick: This was mistaken as I though these rings had variable plusses generally but obviously that's only artifacts. But see my reply to your other comment on rings below.

crawl-ref/source/dungeon.cc
```diff
@@ -6129,6 +6130,132 @@ static bool _valid_item_for_shop(int item_index, shop_type shop_type_,
...
+            return true;
+        default:
+            return false;
+        }
+    case OBJ_WANDS:
+        switch (item.sub_type)
+        {
+        case WAND_FLAME:
+        case WAND_DIGGING: // Questionable
+            return true;
```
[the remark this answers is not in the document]
chucksellick: Yeah maybe this was premature (along with some other decisions) altho if we merge my gadget shops PR then it doesn't matter so much if you find less of certain wands in *general* stores later.

chucksellick:
> I definitely like the idea of this change and I have certainly experienced lategame shops (in particular weapon and armour shops) being rather unexciting.
> 
> I've left a few specific comments regarding specific items being included/excluded from the list. One other brief point: you mention this not affecting Orc:2 but Orc:2's absdepth is 12, so if I understand the code correctly, bad items already have a 1/3 chance to be rerolled on Orc:2, so you may want to tweak the depth threshold there.

Yeah I need to adjust the probabilities a bit further down I think.

chucksellick:
Some example shops on Depths:4 with this change:

image
image
image
image
image

It might be worth reducing the overall number of items *in* the shops, as can be seen these are very good (but not every shop is necessarily this good, I just picked some nice ones)

NormalPerson7:
Okay, those shops largely look reasonable. I agree that perhaps a somewhat reduced number of items in endgame armour/weapon/(jewellery) shops could be beneficial.

I'm still not sure jewellery shops need every item to be an artefact - artefact rings are already the best artefacts in the game because they're swappable and don't need a high enchantment roll to be good - so I fear this could make jewellery shops too good. What do you think?

How does reducing Depths max shops to 3 work with those vaults that place 4 shops? Are they just banned altogether or does one of the shops just place as abandoned? (The latter seems like the better option but I have no idea how that piece of vault code works.)

crawl-ref/source/dungeon.cc
```diff
@@ -6333,15 +6305,36 @@ static void _stock_shop_item(int j, shop_type shop_type_,
...
-
-            if (!no_trash && level_number > 10 && is_trash)
-                no_trash = x_chance_in_y(level_number - 10, level_number - 6);
+            item_def& item = env.item[item_index];
+            bool is_trash = _item_is_trash(item);
+            const bool boring_consumable =
+                consumable_rarity(item) >= RARITY_RARE;
+            // Start trashing after Orc:$ depth; starts at a 3/20 probability,
+            // jumps to 50% at Depths:1, 90% by Zot:$.
+            if (!no_trash && level_number > 12 && (is_trash || boring_consumable))
```
NormalPerson7, line 6314: This comment isn't quite correct. (absdepth-10)/20 gives the following probabilities

30% on Shoals:1/Snake:1/Elf:1 (absdepth 16)
50% on Vaults:1 (absdepth 20)
65% on Depths:1 (absdepth 23)
100% by Zot:3 (absdepth 30)

These probabilities do seem much more reasonable to me now, though.

crawl-ref/source/dungeon.cc
Cgettys: Yes please, more artifact rings :)

chucksellick:
@NormalPerson7 

> I'm still not sure jewellery shops need every item to be an artefact - artefact rings are already the best artefacts in the game because they're swappable and don't need a high enchantment roll to be good - so I fear this could make jewellery shops too good. What do you think?

It has the *potential* to be a bit extreme, certainly - although as you can see in the example above, I only actually got 3 artefact rings ... ironically the algorithm seems to make talismans *more* likely to appear in the jewellery stores, which is maybe not a bad thing with new talismans about to land; although this would not be the case if we use my gadget shop PR too. The artefact amulet there was also nothing amazing. Artefacts are not *necessarily* going to be *much* better than base rings, but I do think it's *appropriate* to have that kind of decision in a late-game shop anyway.

Two adjustments I could possibly make: the probability could scale differently for jewellery, so we keep more base rings but leave the other shops alone; or, the probability could be capped for all shops at around 80-90% so we never reach (and this latter option seems probably a good idea). These sort of tweaks are easy to make and adjust later based on further feedback as well.

> How does reducing Depths max shops to 3 work with those vaults that place 4 shops? Are they just banned altogether or does one of the shops just place as abandoned? (The latter seems like the better option but I have no idea how that piece of vault code works.)

Those vaults are floating vaults that place by themselves and are nothing to do with this change (although now it's brought more to my attention, I'm slightly wondering if they are part of the problem of high variance of shop numbers). This change is only for the serial_vaults placement which decides how many *single shop* vaults to place (it was introduced to replace the old random shop placement).

crawl-ref/source/dungeon.cc
chucksellick: Was adjusting the numbers quite a bit and probably wrote the comment before the last adjustments :) but yeah, these numbers seem quite reasonable.

chucksellick:
> Ah right, I see. From my experience it's mainly Depths that seems to have a particularly high likelihood to place multi-shop vaults, so that might be something to look at.

Looking deeper into shops.des, I realised it's possible to have *both* multiple random shops (via serial_shops) *plus* one of the "floating multi shop" vaults (tagged luniq_floating_shops) which is almost certainly where we're seeing extremely high rolls for shop counts on a single level or over a whole branch. And most of these vaults have Depths in their DEPTH spec. There's also a multi vault missing the luniq_floating_shops tag which could theoretically spawn *in addition to* those two, as well as some themed single shop vaults with no tags.

I'm now contemplating implementing a buniq_* tag so we can at least limit that particular groups of vaults to only show up once *per branch*, as well as putting luniq_floating_shops on serial_shops itself so we can't get such crazy high rolls (11+ shops on a single floor is a bit much however you look at it). V:$ is also a bit of a target for having crazy amounts of shops (in the public objstats google sheet, it certainly got 11 shops on one generation).

regret-index:
Having reviewed over the balance intent of this (and having taken out a mostly unrelated component of it in cba18e0, I'm not convinced as to many of the choices done here are particularly safe for balance (plain javelins are quite crucial for throwing characters throughout the game, handing out nothing but randart jewellery in jewellery shops serves as an immensely strong player boost that also reduces the deployment of those as rewards in endgame branches). I'm also not the happiest with guaranteeing shops in Depths to try and even out the number of shops per game- it's both a required branch when the endgame of 3-runers has a lot of optional branches, and the branch is overflowing with smaller vault placement as is. This PR also noticeably predates b171341, which already provides an endgame money sink that has helped that optional branch.

I would instead assess this PR as a reasonable means to improve the average quality of Gozag shops (possibly after they're made a bit more expensive), kept separate of regular lategame shops. Without particular attempts to make those shops better, they tend to not be particularly worthwhile for more than a small amount of gambling on a god with other, stronger uses of gold, and isolating such rerolling for those shops would serve fine for a god that's already balanced by being weak earlier on.

PR #4407, opened 2025-03-27
feat: rework packed_cell::add_overlay

Cgettys:
Might be faster in typical case. not sure yet.

### DIFF
--- a/crawl-ref/source/tilecell.cc
+++ b/crawl-ref/source/tilecell.cc
@@ -93,7 +93,7 @@ void packed_cell::add_overlay(int tileidx)
     // Clang 19.1.0 happily auto-vectorizes this on -O2
     // auto end = dngn_overlay.begin() + num_dngn_overlay;
     // or something like that might be more efficient for gcc though?
-    auto end = dngn_overlay.end();
+    auto end = dngn_overlay.begin() + num_dngn_overlay;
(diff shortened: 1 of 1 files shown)


geekosaur:
I'm rebuilding crawl with different debug flags to hopefully improve the output (speedscope is currently showing most time in a function that `perf` couldn't resolve), but the _named_ function it spends the most time in is now `map_cell::operator==(map_cell&&)`. The second largest is a `string` allocator, which we probably can't do much about.

geekosaur:
Okay, with some other stuff it seems to have mostly gone away; there's still a lot of time going into something it can't identify, and the earlier trace seems to have been mostly from new areas of the map being revealed. I haven't managed to get enough active on screen yet to see the delay in action.
[perf.data.txt](https://github.com/user-attachments/files/19542752/perf.data.txt)

Cgettys:
Does the same behavior repro in say .32?
#4264 might be responsible; it fixes some bad undefined behavior, but it's possible it's a perf regression on older compilers

geekosaur:
I first noticed it well before 0.32, and tbh it's possible that it was actually fixed then since I haven't yet reproduced it — but I also haven't gotten a bunch of stuff on screen yet so I can't yet distinguish the long pause at the end of a bunch of movement.

PR #4411, opened 2025-03-28
feat: Add staff of earth's ac bonus and update description

tianmaster:
The Staff of Earth currently lacks distinctiveness. Its special physical damage doesn't offer any standout advantages, and players instead fixate on its trivial weakness against flying creatures (a largely irrelevant concern). As one of the four elemental magic schools' foundational staff types, it lacks the signature resistance to basic elemental damage—a feature present in the Elemental Staff. I propose repurposing this resistance mechanic for the base Staff of Earth. Given the staff's inherent rarity in early gameplay, I don't anticipate this adjustment significantly altering early-game balance or progression.

### DIFF
--- a/crawl-ref/source/artefact.cc
+++ b/crawl-ref/source/artefact.cc
@@ -362,7 +362,7 @@ static void _populate_staff_intrinsic_artps(stave_type staff,
     if (prop)
     {
         // Staffs of earth give AC, not resistance.
-        if(*prop == ARTP_AC)
+        if (*prop == ARTP_AC)
             proprt[*prop] += 3;
         else
             proprt[*prop] += 1;
--- a/crawl-ref/source/artefact.cc
+++ b/crawl-ref/source/artefact.cc
@@ -341,7 +341,8 @@ static map<stave_type, artefact_prop_type> staff_resist_artps = {
     { STAFF_ALCHEMY, ARTP_POISON },
     { STAFF_DEATH,   ARTP_NEGATIVE_ENERGY },
     { STAFF_AIR,     ARTP_ELECTRICITY },
-    // nothing for conj or earth
+    { STAFF_EARTH,   ARTP_AC },
+    // nothing for conj
 };
 
 static map<stave_type, artefact_prop_type> staff_enhancer_artps = {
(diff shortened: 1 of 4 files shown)


PleasingFungus:
I approve of the quest to make Crawl's items feel more distinctive and unique.

It's worth noting that we removed something very similar, the old protection ego, almost a decade ago (740ae02a9c630a6a69edd81f8aaf6b194c4d01e2/9d3f7082a8e28e3f5d69b42a311a389155c2454a).

As you noted, staves of earth are relatively rare and hence are unlikely to have the same early game impact that protection weapons did. However, they still share the same gameplay of being useful against the overwhelming majority of threats, and thus being something that otherwise-uninterested players should swap to by default. (Other elemental staves aren't perfect, but their resistances are at least only relevant to a small minority of enemies.)

Moreover, it's very unclear that this would make the staff of earth more distinctive in a way that would interest or excite players. +3 AC is certainly good, but in a very low-key way. It is not the sort of thing that will make people shout 'whoopee!', or that will make the playstyle of 'earth mages' feel unlike that of other 'casters'.

If you're interested in trying again, I would recommend thinking about what would make *using* an earth staff feel different. A further tweak to the melee damage? An on-cast trigger?

Regardless, thank you for contributing!

tianmaster:
> I approve of the quest to make Crawl's items feel more distinctive and unique.
> 
> It's worth noting that we removed something very similar, the old protection ego, almost a decade ago ([740ae02](https://github.com/crawl/crawl/commit/740ae02a9c630a6a69edd81f8aaf6b194c4d01e2)/9d3f7082a8e28e3f5d69b42a311a389155c2454a).
> 
> As you noted, staves of earth are relatively rare and hence are unlikely to have the same early game impact that protection weapons did. However, they still share the same gameplay of being useful against the overwhelming majority of threats, and thus being something that otherwise-uninterested players should swap to by default. (Other elemental staves aren't perfect, but their resistances are at least only relevant to a small minority of enemies.)
> 
> Moreover, it's very unclear that this would make the staff of earth more distinctive in a way that would interest or excite players. +3 AC is certainly good, but in a very low-key way. It is not the sort of thing that will make people shout 'whoopee!', or that will make the playstyle of 'earth mages' feel unlike that of other 'casters'.
> 
> If you're interested in trying again, I would recommend thinking about what would make _using_ an earth staff feel different. A further tweak to the melee damage? An on-cast trigger?
> 
> Regardless, thank you for contributing!

On-cast trigger is indeed a good idea, but I am worried that if set in this way, the actual effect may not be significantly different from a flat AC gain?

PleasingFungus:
Sorry for confusion. I agree that on-cast AC is similar (though could perhaps be somewhat larger, or scale with spell level..?) But I was not suggesting that the hypothetical on-cast trigger would be AC specifically.

PR #4413, opened 2025-03-30
make packed_cell smaller still

Cgettys:
The typical case is zero icons. every single packed_cell uses 48 bytes nonetheless.

### DIFF
--- a/crawl-ref/source/tilecell.h
+++ b/crawl-ref/source/tilecell.h
@@ -16,7 +16,7 @@ enum halo_type: uint8_t
 // for 8 bytes fixed on every cell
 // + 32 bytes on cells with icons.
 struct packed_icons {
-    unique_ptr<FixedBitVector<256>> icons;
+    unique_ptr<set<tileidx_t>> icons;
     // constructor
     packed_icons() = default;
     // destructor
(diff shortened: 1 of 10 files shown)


crawl-ref/source/tilecell.h
```diff
@@ -48,8 +93,9 @@ struct packed_cell
     tile_flavour flv;
     tileidx_t fg;
     tileidx_t bg;
+    // TODO: seems weird this isn't just part of dngn_overlay?
     tileidx_t cloud;
-    set<tileidx_t> icons;
+    packed_icons icons;
```
WizardIke, line 98: If I'm not mistaken, dungeon overlays are things like water shorelines and ice floors from frozen ramparts which draw under monsters where as icons are things like the berserk status and draw over monsters

crawl-ref/source/tilecell.h
```diff
@@ -48,8 +93,9 @@ struct packed_cell
...
```
WizardIke, line 99: If I'm not mistaken, dungeon overlays are things like water shorelines and ice floors from frozen ramparts which draw under monsters where as icons are things like the berserk status and draw over monsters

crawl-ref/source/tilecell.h
WizardIke, line 98: If I'm not mistaken, dungeon overlays are things like water shorelines and ice floors from frozen ramparts which draw under monsters where as icons are things like the berserk status and draw over monsters

Cgettys: The comment refers to cloud. I believe the cloud effects also sometimes end up in dungeon overlays too, so we may have the same info in two places.

WizardIke: Are that makes sense, although most clouds draw over monsters with transparency. I think there are a few exceptions to this like ink clouds, so maybe these are the clouds being added to the dungeon overlays?

chucksellick: Clouds render twice - once *underneath* the foreground, and once over it with the transparency mask, to give the effect of the sprite being *inside* the cloud

Cgettys: Good to know. Even so, it seems weird we only pick one cloud to render underneath?
Seems like it should be the same list twice, or something.

PR #4427, opened 2025-04-01
Don't waste a turn when summoning with no space (NormalPerson7)

### LINKED ISSUE #3837: Summon Lightning Spire wastes a turn when terrain clearly prevents spire placement

WizardIke:
When trying to summon a creature where you can't see any available space to place it, don't spend a turn or any resources.

Fixes #3837

### DIFF
--- a/crawl-ref/source/mon-place.cc
+++ b/crawl-ref/source/mon-place.cc
@@ -122,6 +122,26 @@ static bool _habitable_feat(habitat_type ht, dungeon_feature_type feat)
 }
 
 /**
+* What dungoen features can have it least one of the monster types on them?
+*
+* @param mon_types the array of monster types to check.
+* @param types_count the number of monster types in the array to check.
+* @return the set of dungeon features that can hold at least one of the
+*         moster types.
+*/
+habitat_type habitat_for_any(const monster_type* mon_types,
+                             std::size_t types_count)
+{
+    habitat_type features = HT_NONE;
+    for (std::size_t i = 0; i < types_count; ++i)
+    {
+        habitat_type feat = mons_class_habitat(mon_types[i]);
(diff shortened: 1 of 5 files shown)


Hellmonk:
Extremely good pull request, thank you.

crawl-ref/source/mon-place.cc
```diff
@@ -112,6 +112,135 @@ static bool _hab_requires_mon_flight(dungeon_feature_type g)
     return g == DNGN_LAVA || g == DNGN_DEEP_WATER;
 }
 
+enum habital_feature_type_flag
```
chucksellick, line 115: I think this should be "habitable" not "habital".

I think this is a really good improvement, separately to the summoning changes, as the old system was rather clunky; I am just not sure on a first look at this whether this slightly affects the way e.g. amphibious creatures are placed w.r.t trying to place them in their preferred habitat?

WizardIke: Thanks that is a spelling mistake I've fixed It. The code is still using the old habitat code to decide where to place monsters so this shouldn't affect that. Although, we could refactor everywhere is to use this new habitat code if that would be helpful (in a separate PR of cause).

DracoOmega:
I've merged your general habitat changes from #4432. If you could update this PR to use that system, that would be great. (And then I'll do a final review on it in a timely fashion.)

WizardIke:
Merged in commit f2a3531

PR #4429, opened 2025-04-01
Add a cachebusting key to lobby static assets

chucksellick:
Adding a version string to all the lobby assets ensures that, if the server rebuilds and there is an update to client scripts, players will definitely get the update when they reload the browser, and not get a locally-cached copy of the file.

This enables use of more aggressive caching strategies, longer cache times, in-memory caching with nginx/Apache, or even pushing static assets over to CDN, if server admins wish to configure thing way, without worrying about the client browser potentially serving stale assets.

### DIFF
--- a/crawl-ref/source/webserver/templates/client.html
+++ b/crawl-ref/source/webserver/templates/client.html
@@ -5,9 +5,10 @@
     <link rel="icon" href="/static/stone_soup_icon-32x32.png" type="image/png">
     <script type="text/javascript">
       var socket_server = "{{ socket_server }}";
+      var game_version = "{{ game_version }}";
     </script>
-    <script src="/static/scripts/contrib/require.js" data-main="/static/scripts/app"></script>
-    <link rel="stylesheet" type="text/css" href="/static/style.css">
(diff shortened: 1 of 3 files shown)


crawl-ref/source/webserver/templates/client.html
```diff
@@ -5,9 +5,10 @@
     <link rel="icon" href="/static/stone_soup_icon-32x32.png" type="image/png">
     <script type="text/javascript">
       var socket_server = "{{ socket_server }}";
+      var game_version = "{{ game_version }}";
     </script>
-    <script src="/static/scripts/contrib/require.js" data-main="/static/scripts/app"></script>
-    <link rel="stylesheet" type="text/css" href="/static/style.css">
+    <script src="/static/scripts/contrib/require.js?v={{ game_version }}" data-main="/static/scripts/app"></script>
+    <link rel="stylesheet" type="text/css" href="/static/style.css?v={{ game_version }}">
```
WizardIke, line 11: Do we need to URI encode game_version here to? Maybe server.py could send it through already encoded?

chucksellick: Actually in practise it doesn't needed as all chars in the version are URL-safe; did so in app.js purely out of habit

PR #4432, opened 2025-04-03
Rework habitats

WizardIke:
Store habitats as a bit mask of the different terrain categories monsters of that habitat can survive on. This lets us easily check of one habitat is more restrictive than another or if a monster can survive on a category of terrain e.g. land, water, shallow water or deep water etc.

### DIFF
--- a/crawl-ref/source/dat/mons/eldritch-tentacle-segment.yaml
+++ b/crawl-ref/source/dat/mons/eldritch-tentacle-segment.yaml
@@ -11,7 +11,7 @@ hp_10x: 1200
 ac: 13
 ev: 0
 intelligence: animal
-habitat: amphibious
+habitat: eldritch_tentacle
 speed: 12
 size: giant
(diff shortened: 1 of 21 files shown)


crawl-ref/source/dat/mons/eldritch-tentacle-segment.yaml
```diff
@@ -11,7 +11,7 @@ hp_10x: 1200
 ac: 13
 ev: 0
 intelligence: animal
-habitat: amphibious
+habitat: eldritch_tenticle
```
chucksellick, line 14: small typo, tenticle -> tentacle

WizardIke: Thanks, I've fixed it

DracoOmega:
Merged (with a few minor fixes/adjustments). Thanks!

PR #4441, opened 2025-04-04
Damnation Targetter (and allow unrand ranged targetters in general)

chucksellick:
Also fixes some existing issues with the beam targetter and uses it as a default for ranged weapons (instead of vanilla direction chooser). Opens the door to new and exciting ranged unrands.


crawl-ref/source/art-func.h
```diff
@@ -883,7 +899,7 @@ static void _DAMNATION_launch(bolt* beam)
     expl->name   = "damnation";
     expl->tile_explode = TILE_BOLT_DAMNATION;
 
-    beam->special_explosion = expl;
```
Cgettys, line 886: this should still be a -> 
???

crawl-ref/source/target.cc
```diff
@@ -208,8 +208,6 @@ targeter_beam::targeter_beam(const actor *act, int r, zap_type zap,
     beam.target = aim;
     beam.ex_size = min_ex_rad;
     beam.aimed_at_spot = true;
-
-    penetrates_targets = beam.pierce;
```
Cgettys, line 212: Is this removal intentional?

chucksellick: Yes. This value was copied to a private variable meaning that if I wanted to adjust the piercing property of the beam I couldn't (and for ranged, we need to set piercing based later based on the type of the launcher).

Now the logic looks directly at beam.pierce in is_affected.

crawl-ref/source/art-func.h
chucksellick: Thanks. I shuffled things around a few times and beam stopped being a pointer at one time, before I reverted most of it back, but missed this.

PR #4442, opened 2025-04-04
Fix incorrect max magic after using artifact MP+ talisman

WizardIke:
Activating an artifact talisman that gives max magic points wouldn't increase your max magic points until the game was reloaded and deactivating it would incorrectly increase your max magic points until the game was reloaded.


DracoOmega:
PR4443 was merged, which as you say supersedes this one.

PR #4443, opened 2025-04-04
Fix swapping between talisman of the same type

### LINKED ISSUE #4412: Evoking artifact talisman when already transformed does not apply unique effects
Running webtiles on akrasiac via macOS.

trunk (0.33-a0-1149-g9fdd3ef)

https://crawl.akrasiac.org/rawdata/geewiz/geewiz.txt

**DESCRIPTION:**

'pode transformed into snake form via a normal/default serpent talisman.

When I found an artifact talisman:

 G - the serpent talisman of Hospitality (active) {Str+2}
   (You took it off Aizul on level 4 of the Snake Pit)   

...evoking this talisman did not change the character's strength. Untransforming and then evoking the talisman did correctly add +2 STR relative to the "base" serpent talisman's snake form.

WizardIke:
If one or both of the talisman are artifacts we need to change the
artifact properties applied to the player which we weren't doing.

Fixes #4412

This PR includes the changes from PR #4442 as it depends on them to function

### DIFF
--- a/crawl-ref/source/transform.cc
+++ b/crawl-ref/source/transform.cc
@@ -1635,13 +1635,13 @@ static void _enter_form(int pow, transformation which_trans, bool scale_hp = tru
     if (you.has_innate_mutation(MUT_MERTAIL))
         merfolk_check_swimming(env.grid(you.pos()), false);
 
-    if (is_artefact(you.active_talisman))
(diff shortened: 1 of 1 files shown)


DracoOmega:
Merged, thanks!

PR #4461, opened 2025-04-10
Fix wizmode not being able to place electric eels

WizardIke:
When placing an electric eel using wizard mode, if there isn't any water near by it should place water and then place the eel in the water. However, it was placing a floor tile instead.


WizardIke:
@DracoOmega That would also work and is probably more robust in case we ever decide to add dry land only monsters

DracoOmega:
Thanks!

PR #4465, opened 2025-04-10
Speed up copying map knowledge

WizardIke:
Web tiles makes multiple copies of the map knowledge each frame (about 5) so by making it faster to copy we can get a roughly 10% reduction in total CPU time in web tiles.


crawl-ref/source/viewmap.cc
```diff
@@ -1044,7 +1043,7 @@ map_control_state process_map_command(command_type cmd, const map_control_state&
                 break;
             if (env.map_forgotten)
                 _unforget_map();
-            MapKnowledge *old = new MapKnowledge(env.map_knowledge);
+            map_knowledge_type *old = new map_knowledge_type(env.map_knowledge);
```
chucksellick, line 1046: I'm not too sure about this class rename. I know the codebase is a bit inconsistent when it comes to naming conventions, but classes are frequently PascalCase. However if I see a `_type` suffix my immediate assumption is this is probably some kind of enum so I think this name is particularly confusing

WizardIke: Good point, I'd based the name on sweep_type but that seems to be a bit of an exception when it comes to names. I've changed the name back

crawl-ref/source/catch2-tests/test_tags.cc
```diff
@@ -213,35 +214,41 @@ TEST_CASE( "Basic marshalling/unmarshalling works correctly.", "[single-file]" )
...
 
             auto r = reader(buf);
-            map_cell roundtrip_cell;
-            unmarshallMapCell(r, roundtrip_cell);
+            MapKnowledge roundtrip_map_knowledge;
+            unmarshallMapCell(r, gc, roundtrip_map_knowledge);
 
+            const map_cell& cell = map_knowledge(gc);
+            const map_cell& roundtrip_cell = roundtrip_map_knowledge(gc);
             // TODO: Compare other members.
```
Cgettys, line 230: maybe a good time to address this todo?

crawl-ref/source/map-cell.h
```diff
@@ -67,100 +68,41 @@ struct cloud_info
...
-        _cloud = o._cloud ? make_unique<cloud_info>(*o._cloud) : nullptr;
-        _item = o._item ? make_unique<item_def>(*o._item) : nullptr;
-        _mons = o._mons ? make_unique<monster_info>(*o._mons) : nullptr;
-
-        return *this;
-    }
-
-    // move constructor
-    map_cell(map_cell&& o) noexcept = default;
-    // move assignment
```
Cgettys, line 116: make sure gcc 4.7 is ok with the new version

crawl-ref/source/map-cell.h
```diff
@@ -171,61 +113,23 @@ struct map_cell
...
-        _feat = nfeat;
+        _feat = static_cast<uint8_t>(nfeat);
         _feat_colour = colour;
-        _trap = tr;
-    }
-
-    item_def* item() const
-    {
-        return _item.get();
+        _trap = static_cast<uint8_t>(tr);
```
Cgettys, line 118: COMPILE_CHECK for trap count somewhere?

Cgettys:
Looks very nice overall.

crawl-ref/source/map-cell.h
WizardIke: std::make_unique was added in gcc 4.9.0 and we use it in a few places, so I think support for gcc 4.7 has been dropped

crawl-ref/source/map-cell.h
WizardIke: It's in the `trap` member function, although I could move it somewhere else if that would be clearer to read

crawl-ref/source/catch2-tests/test_tags.cc
WizardIke: Properly comparing all members is probably a decent size change as `item_def` and `monster_info` don't define equality operators. I'd rather leave this to a separate PR

crawl-ref/source/map-cell.h
Cgettys: as long as it's somewhere

PR #4469, opened 2025-04-11
Fix no warning when rampaging over trap (CrawlOdds)

### LINKED ISSUE #4459: No warning when rampaging into a trap
webtiles

trunk

**DESCRIPTION:**
In the layout <player><trap><empty space><monster>, if the player has rampage, they are not warned when stepping into the trap. This is especially sad because it's not implausible that rampage would let you skip over the trap.

WizardIke:
When you rampage over a trap, the trap will be set off, so you should receive a warning. This was broken in 18e1206

Fixes #4459

### DIFF
--- a/crawl-ref/source/movement.cc
+++ b/crawl-ref/source/movement.cc
@@ -632,6 +632,21 @@ monster* get_rampage_target(coord_def move)
     return nullptr;
 }
 
+static bool _check_rampage(bool attacking, coord_def rampage_destination,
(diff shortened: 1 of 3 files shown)


Hellmonk:
Something related to this change seems to cause rampaging "over" a sigil of binding, saying yes to the prompt, and getting stuck to take zero time.

WizardIke:
@Hellmonk I tested it and it also takes zero turns in unmodified 0.33.1 just without the prompt

PR #4482, opened 2025-04-14
Revise the description from 'Max blocks/turn' to 'Max blocks/action'

tianmaster:
I found that the reset of shield-blocks value occurs after each action, rather than after each turn (in my understanding, turn refers to the impression of 10 aut). I think the description should be modified from blocks/turn to blocks/action. This makes it easier for readers of the description to understand the working principle. @PleasingFungus I hope you can review it:)

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -2372,7 +2372,7 @@ static string _describe_armour(const item_def &item, bool verbose, bool monster)
                         + to_string(property(item, PARM_AC));
             description += "     Encumbrance rating: "
                         + to_string(-property(item, PARM_EVASION) / 10);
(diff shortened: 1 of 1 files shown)


NormalPerson7:
This raises the question of whether it _should_ work this way. It seems a bit strange to me that I get to block twice as many attacks if I'm attacking with a quick blade than if I'm casting spells.

tianmaster:
> 这就提出了一个问题，即它是否_应该_以这种方式运作。对我来说，如果我用快速的刀片攻击，我可以阻止的攻击是我施法时的两倍，这对我来说似乎有点奇怪。

I think this may be due to the lack of a content for handling public turns in the game, which has led to the adoption of such compromise measures

PR #4489, opened 2025-04-19
Fix monster backup weapons giving halos/umbras (NormalPerson7)

### LINKED ISSUE #4487: Monsters unwielding Eos keep their halo
CXC 0.33-a0-1200-gece6b70641

I came across a monster wielding the +10 morningstar Eos, which gives a halo. However, the monster then switched to a ranged weapon, but kept their halo despite the morningstar no longer being equipped. See below screenshots.

Image
Image

[Note: I realise I am playing about 240 commits behind HEAD at the moment so it's possible this has been fixed, though I couldn't see any indication that it has been.]

WizardIke:
When a monster was using a ranged weapon but had Eos or Brilliance as its backup melee weapon it would incorrectly get a halo/umbra.

Fixes #4487

### DIFF
--- a/crawl-ref/source/areas.cc
+++ b/crawl-ref/source/areas.cc
@@ -636,8 +636,11 @@ int monster::halo_radius() const
         size = max(size, 3);
 
     item_def* alt_wpn = mslot_item(MSLOT_ALT_WEAPON);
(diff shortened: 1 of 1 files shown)


pooka109:
Applied; thanks.

PR #4497, opened 2025-04-20
Fix quote for microbat

croikle:
An extra blank line caused the microbat's quote to be literal `<bat>` rather than copying the basic bat quote.


pooka109:
Applied; thanks.

PR #4501, opened 2025-04-21
Place a statue on `nicolae_hells_hurtin_island`

thewilsonator:
... so people don't accidentally walk into it.

This could be made an Ice statue in Doc, but I don't know how to do that.


yrdzrfxndfvh (commented):
: if you.in_branch("Coc") then
KMONS: > = ice statue
:  else
KMONS: > = obsidian statue
: end

orange crystal statue could be used in Dis too

thewilsonator:
Done. Ping.

thewilsonator:
Thanks!

Issue #4513, opened 2025-04-26
Jelly/eyeball-shaped shapeshifter is hostile to Jiyva worshipper but cannot be attacked.


powerbf:
trunk 0.34 webtiles

As a Jiyva worshipper, if a shapeshifter is in jelly/eyeball-form, it will be hostile, but you will be unable to attack it.

caiman-dorohedoro:
same in 0.33

PR #4518, opened 2025-04-29
Don't let Jiyva ban attacking hostile slime-shaped shifters

NormalPerson7:
Jiyva currently doesn't turn slime-shaped shifters neutral, but she would previously prevent attacks on them anyway. This behaviour leaves the player (temporarily) with a monster they cannot attack.

I have elected to effect this by not banning attacks on hostile slimes. This may need to be adjusted if there are other ways for hostile slimes to appear under Jiyva; I am not aware of any.


NormalPerson7:
Completely forgot to mention, this will fix #4513.

PR #4523, opened 2025-05-01
Fix a test requiring long to be 64 bit

WizardIke:
It was trying to store 2,000,000,000,000 in a long which won't work on 32 bit builds or MSVC builds.

### DIFF
--- a/crawl-ref/source/fixedp.h
+++ b/crawl-ref/source/fixedp.h
@@ -136,6 +136,10 @@ public:
                                                 static_cast<BaseType>(Scale))
     {
     }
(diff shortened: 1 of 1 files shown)


neilmoore:
Mightn't `int64_t` be better than `long long`, particularly on platforms where `long` is already 64-bit and `long long` 128-bit?

WizardIke:
@neilmoore Unfortunately, int64_t is allowed to be a typedef for long if long happens to be 64 bits, so if I added a constructor taking type int64_t we could get a duplicate definition error on some compilers with our constructor taking type long. We could convert all the constructors to fixed width types but that might cause other problems and I don't think we use anything but the int constructor outside of tests anyway

PR #4525, opened 2025-05-01
Make diamond sawblades splatter more blood (GnomeSayin)

WizardIke:
Shredding enemies with sawblades feels like it should result in a lot of blood, which it wasn't producing before. I limited each sawblade to only spraying blood in a small area around its victims as its hard to tell where the blood is coming from otherwise.


pooka109:
Applied; thanks.

PR #4554, opened 2025-05-18
Monkooky's Gastronomic Expanse (updated)

pooka109:
This is an update of Monkooky's PR #3775 to current 0.34 trunk.

Changes from that:
* Merged some checkwhite-related whitespace removals earlier to avoid warnings when applying patches
* Removed an unused variable from the Gastronomic Expanse spell targeter
* Left out the reordered variables in packed_cell (this adds a lot of warnings, I don't quite understand packed_cell enough to fix the warnings, and it seems to work without the reordering)


regret-index:
Monkooky has assented to also closing this for being out of date, in the same spirit as closing #3775, and will submit a new PR with a new draft.

PR #4560, opened 2025-05-21
Make `more_gem_info` show unacquired gems

thewilsonator:
... so that *ahem* [certain people](https://crawl.project357.org/morgue/DispaterIsMyDaddy/morgue-DispaterIsMyDaddy-20250518-155010.txt), do not forget to pick them up once they have the orb of zot.

~~Please let me know if I need to assign to `inf.light_colour` here as well~~ done.

### DIFF
--- a/crawl-ref/source/status.cc
+++ b/crawl-ref/source/status.cc
@@ -962,10 +962,11 @@ static void _describe_gem(status_info& inf)
     if (time_taken >= limit)
         return; // already lost...
 
-    if (!gem_clock_active())
+    if (!gem_clock_active() && !you.gems_found[gem])
     {
         // player has picked up the orb, but the gem has not yet shattered
-        inf.light_text = "Gem not acquired";
+        inf.light_text = "No Gem";
+        inf.light_colour = CYAN;
         return;
     }
     const int d_aut_left = (limit - time_taken + 9) / 10;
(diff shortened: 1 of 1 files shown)


crawl-ref/source/status.cc
```diff
@@ -962,6 +962,12 @@ static void _describe_gem(status_info& inf)
     if (time_taken >= limit)
         return; // already lost...
 
+    if (!gem_clock_active())
+    {
```
NormalPerson7, line 966: I think you also want to check that the player doesn't yet have the gem. As currently written, "Gem not acquired" will always be visible, even when the player has the gem!

crawl-ref/source/status.cc
```diff
@@ -962,6 +962,12 @@ static void _describe_gem(status_info& inf)
...
+        // player has picked up the orb, but the gem has not yet shattered
+        inf.light_text = "Gem not acquired";
+        return;
```
NormalPerson7, line 969: This status text is a bit long - perhaps "No Gem" or "Gem*" or something like that could work. You should also assign `inf.light_colour` as you suggest, otherwise the status will show up in black text (which is invisible) - white is probably an appropriate colour, although something like cyan might make it stand out a bit more.

thewilsonator:
Thanks, done.

thewilsonator:
Ping

thewilsonator:
Thanks!

PR #4578, opened 2025-05-27
Feat: parchments

Hellmonk:
Book changes in 0.27 reduced spell counts per book and increased number of 
books generated in order to make spell progression more varied and interesting.
This approach largely succeeded, but there's still some friction: 1) small
books have to be rebalanced every time spells are added or removed; 2) spell
frequencies are still not independent from one another, which makes it hard to
do things like "rare" spells; 3) the interface for checking which spells 
various books contain (for example, in a shop) is kind of bad. 

Parchments are a new type of book containing exactly one spell which serves as 
the name of the parchment. Most existing book generation is moved to 
parchments, with appropriate weighting changes so characters see roughly the
same number of total spells. Fixed and randart books are retained for Sif Muna 
gifts and as special rewards from sources where one spell isn't sufficient, 
such as portal vault loot and acquirement.

TODO BEFORE MERGING THIS PR:

- compensate for increased floor book generation (out of gold stack number?)
- add a robust weighting function for parchment spells to mimic current behavior
- new tiles


Hellmonk:
closing because this feature is in trunk now.

PR #4586, opened 2025-05-31
Handle ?-key spell description from cast menu

shmup:
This change adds `[?] help` to the cast menu, similar to the memorization menu.

image


Hellmonk:
Seems like a good feature to me, and worked fine in my brief testing. I will let someone else look at it more closely before merging.

PR #4594, opened 2025-06-05
Fix fair_adjacent_iterator returning duplicate locations

WizardIke:
Instead of returning the 8 adjacent locations in a random order, it would sometimes repeat a location and miss another location.


DracoOmega:
Merged, thanks!

PR #4596, opened 2025-06-05
Fix rampage hints showing on levels you aren't on

WizardIke:
When examining a level that you weren't on, rampage hints would still show up leaking information about where you were relative to the map borders.


pooka109:
Applied; thanks.

PR #4597, opened 2025-06-05
Fix the cloud targeter avoiding immune players

### LINKED ISSUE #4592: Freezing Cloud (spell) ignores Frostbite immunity for targeter and self-harm warning
0.34-a0-100-g98ae272499 (webtiles)

**DESCRIPTION:**
Wielding the artifact axe Frostbite makes you immune to freezing clouds from all sources. The spell Freezing Cloud has the only effect of generating freezing clouds (no impact damage), and thus cannot harm a player wielding Frostbite.

However, the targeter and the warning prompt for self-harm seem to ignore Frostbite's immunity. From what other players reported, this issue does not occur with other sources of cloud immunity, such as Qaz or Robe of Clouds, but I could not verify this myself.

Repro:
1. Wield Frostbite
2. Start casting Freezing Cloud
3. Notice the targeter going out of its way not to hit the caster
4. Try casting in a way that hits the caster
5. See warning prompt

Expected behavior:
The targeter does not mind overlapping the caster, and there is no warning prompt for casting freezing cloud on the caster (it should still warn for other non-immune allies).

WizardIke:
When casting a cloud spell such as freezing cloud, the targeter would try to avoid hitting the player even when they were immune to the cloud.

Fixes #4592

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -3376,10 +3376,8 @@ bool bolt::harmless_to_player() const
     case BEAM_PETRIFY:
         return you.res_petrify() || you.petrified();
 
-#if TAG_MAJOR_VERSION == 34
     case BEAM_COLD:
-        return is_big_cloud() && you.has_mutation(MUT_FREEZING_CLOUD_IMMUNITY);
-#endif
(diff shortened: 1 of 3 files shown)


pooka109:
Applied; thanks.

PR #4599, opened 2025-06-05
Fix agony not saying when you are immune

WizardIke:
When examining an enemy with the agony spell as a torment immune character, it would show the chance of the spell effecting you instead of saying that you are immune. Change it to say that you are immune for consistency with the sleep and confusion spells.


pooka109:
Applied; thanks.

PR #4601, opened 2025-06-06
Add several new skill titles

antimachina:
Listing them all here for ease of viewing:

Octopode 27 UC from Crusher to Kraken to feel more epic and more unique to Op
Octopode/Merfolk 27 Invo Lugonu gets Abyssopelagic (https://en.wikipedia.org/wiki/Abyssal_zone)
Minotaur 27 Invo Ash gets Lord of the Labyrinth (Navigating a labyrinth feels very appropriate for Ash to me)
Merfolk 27 Hexes gets Siren
Octopode 27 Invo with an evil god gets Leviathan (Leviathan is one of those loosely defined mythological creatures, but is definitely a sea monster of some sort, was considered demonic pretty often. Crawl's tainted leviathans are not aquatic at all, so I would understand if this adds unwanted confusion.)
Human of Yred 27 Fighting gets Death Knight (Hell Knight is also specifically a Fighting title, not an Invocations title, so I kept symmetry)
Demigod 27 Spellcasting gets Ascendant
Demigod 27 Stealth gets Thief of Immortality

### DIFF
--- a/crawl-ref/source/skills.cc
+++ b/crawl-ref/source/skills.cc
@@ -1938,6 +1938,8 @@ string skill_title_by_rank(skill_type best_skill, uint8_t skill_rank,
                 result = "Pharaoh";
             else if (species == SP_FELID)
                 result = claw_and_tooth_titles[skill_rank];
+            else if (species == SP_OCTOPODE && skill_rank == 4)
+                result = "Crusher";
             else if (species == SP_OCTOPODE && skill_rank == 5)
                 result = "Kraken";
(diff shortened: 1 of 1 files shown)


riverfiend:
The Octopode Crusher title is there on purpose -- in ye olde crawl days, there was a late-game monster called an Octopode Crusher, and this title references that.

antimachina:
Ah, good to know! I still think there's a reasonable case to remove a title referencing a monster that seemingly last existed in 0.18, but I'm not all that attached to doing so.

PR #4605, opened 2025-06-09
Move off of windows 2019 GHA

wheals:
@Cgettys This may be of interest to you

Context: actions/runner-images#12045
> The Windows 2019 Actions runner image will begin deprecation on 2025-06-01 and will be fully unsupported by 2025-06-30 #12045

There are two obvious ways to fix the issue noted in the comment; either upgrade the toolset or install an older toolset on a new image. The good thing is that upgrading to v142 shouldn't drop support for anyone's setup because [VS 2019's sysreqs](https://learn.microsoft.com/en-us/visualstudio/releases/2019/system-requirements) (works back to windows 7) are the same as [VS 2017's](https://learn.microsoft.com/en-us/visualstudio/releases/2017/vs2017-system-requirements-vs). But this might not even work, that's why I'm opening a PR.

### DIFF
--- a/crawl-ref/INSTALL.md
+++ b/crawl-ref/INSTALL.md
@@ -388,9 +388,13 @@ These instructions have been successfully tested with Ubuntu only.
 This build process is currently unsupported, and unlikely to be straightforward
 in versions of MSVC besides those explicitly mentioned here.
 
-This build is tested on Visual Studio 2017 15.9.60 on Windows 10 and 11.
-Tested configurations are `Debug/Release;Console/Tiles;Win32/x64`, Python and
-Lua support for editing are untested, and a webtiles build is not available.
+This build is tested on Visual Studio 2022 17.14.4 on Windows 10. However, the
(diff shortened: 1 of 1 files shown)


Cgettys:
I don't actively use msvc for crawl, ironically - but use msvc very extensively at work.

@WizardIke is the person who might have a personal interest afaik.

If you want my personal opinion, upgrading to v142 in my opinion is the obvious right call - MSYS2 etc are always options if anyone out there is pre-Windows 7.

Also note that Visual Studio 2017 is officially out of mainstream support and into extended support (which of course, presumably crawl folks aren't paying for for a hobby project):
https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2017
For that matter, Visual Studio 2019 is out of mainstream support too, but less so:
https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2019
So ideally we'd be talking about Visual Studio 2022. But VS 2019 is at least _less_ out of support.

Cgettys:
Other than that, looks reasonable :)

PR #4622, opened 2025-06-20
Fix old player marker being left on minimap after tele

WizardIke:
After a teleport goes off, the old player marker wouldn't be cleared from their minimap.


nlavsky:
Merged, thanks for the fix!

PR #4629, opened 2025-06-28
add missing words to section 6-a of the options guide

wtodom:
I was checking something in the options guide and noticed a typo. This PR simply adds a reasonable option for the missing word(s).


gammafunk:
Thanks for the PR!

PR #4633, opened 2025-07-03
allow using hive form under Oka; explicitly mention when bees would not spawn (Sac love and now Oka)

radio-gra:
it doesn't make sense to me that Oka would make hive talisman unusable, because "no allies" is not really a conduct thing, just something he enforces as a matter of fact (cf. Ds w/ guardian mutation, Oka does not excommunicate you or prevent you from joining if you have it, nor does he prevent it if you already worship him), hence all ally makers being gray and not red. if Ru's Sac love interaction with hive talisman is intentional (can use hive and get regen, but will not get swarm), then it seems that nectar-making bees aren't really allies, so there's no reason for Oka to intervene with their work, only with the swarm spawning part.

not sure about the wording of "The swarm will not protect you as you can not gain allies.", feel free to change if it doesn't fit

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -7568,7 +7568,11 @@ static string _describe_talisman(const item_def &item, bool verbose)
 
         if (is_useless_item(item, false))
             _uselessness_desc(description, item);
-        else if (item.sub_type != TALISMAN_PROTEAN)
+
+        if (item.sub_type == TALISMAN_HIVE && you.allies_forbidden())
+          description << "\n\n" << "The swarm will not protect you as you can not gain allies.";
+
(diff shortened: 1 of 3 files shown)


DracoOmega:
If you are allowed to use the talisman with Sac Love (which I did not realize, actually), I would consider *that* the bug.

I felt like not being able to benefit from the regeneration with Oka (that you are getting *because* of the ambrosia the bees are making inside you!) is consistent with not being able to use ?butterflies, even if the wind effect would sometimes be useful with no butterflies attached (since the butterflies are apparently what is generating the wind also!)

pooka109:
Since the loveless' being able to use a hive talisman is the bug, I've fixed that.

PR #4641, opened 2025-07-06
Changes with display of current attack delay in weapon descriptions

RypoFalem:
2 commits, one to display weapon delay in item descriptions (even if you're at the base delay) and one to include your current delay on throwing weapons as well.

---
https://github.com/crawl/crawl/blob/3cd4677b91f90b0ed0b777eed8f9afef84451c52/crawl-ref/source/describe.cc#L2254

I wanted to ask about this `crawl_state.needsave`. This check was performed elsewhere https://github.com/crawl/crawl/blob/3cd4677b91f90b0ed0b777eed8f9afef84451c52/crawl-ref/source/describe.cc#L1506 and I copied it over but I am not sure what it's guarding against exactly. I thought it would be easier to ask about it if I provided the context. Needs attention!

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -1372,16 +1372,13 @@ static int _get_delay(const item_def &item)
 
 static string _desc_attack_delay(const item_def &item)
 {
-    const int base_delay = property(item, PWPN_SPEED);
-
     // Hide speed/heavy brand from unidentified weapons.
     item_def dummy = item;
     if (!item.is_identified())
         dummy.brand = SPWPN_NORMAL;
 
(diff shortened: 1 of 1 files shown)


RypoFalem:
If desired, I could also bring over the "slowed by your shield" from the `@` command. The game already tells you if your armour is impeding your ranged weapons in the weapon's item description but there might be a slight tradeoff for space. (whereas I felt these first two commits didn't really have any meaningful tradeoffs since weapon descriptions were already using that space, just not for every weapon)

WizardIke:
The crawl_state.needsave check is needed because in offline crawl you can look up item and spell etc descriptions from the ?/ menu from the startup menu before resuming or creating your character.

PR #4650, opened 2025-07-11
Make the lua function dgn.vault_at return a pointer

WizardIke:
The lua function dgn.vault_at was returning a copy of the vault_placemnt at a location. Unfortunately, it is slow to copy a vault_placement and because the object returned to lua is a wrapper around a c++ object lua doesn't know how big it is and won't trigger garbage collection until you have about 8000 of them (about 17 MB).

The only place using dgn.vault_at is the disco Pan colour function which can work just fine with a pointer rather than a copy and runs considerably faster.

This is also consistent with the c++ equivalent that returns a pointer.

### DIFF
--- a/crawl-ref/source/l-dgn.cc
+++ b/crawl-ref/source/l-dgn.cc
@@ -1569,7 +1569,10 @@ LUAFN(_dgn_map_parameters)
 
 static int _dgn_push_vault_placement(lua_State *ls, const vault_placement *vp)
 {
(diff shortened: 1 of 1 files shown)


gammafunk:
We saw a crash recently from particleface that illustrated the problem of this expensive copy, since it was an OOM:

https://cbro.berotato.org/morgue/particleface/crash-particleface-20260728-135654.txt

Hence it was a good time to take a closer look at this fix and merge. Using a pointer without the full copy seems great here.

PR #4659, opened 2025-07-15
Dismiss monsters that are lured too far.

Hellmonk:
Experimental change. Store the origin level of each monster in its structure, and track how far away from the origin we are on transit to a new floor. If the monster is 3 or more floors away after transit, it goes away forever, yielding no items or exp. This only applies to hostile monsters in connected branches, and is turned off for the orbrun.

This is a thing that shouldn't be noticeable for most players. It prevents the sort of degenerate exploits that happened with eg. silent spectres, without having to either tag every monster with an exploitable mechanic to not use stairs or to avoid using those mechanics altogether. So now making an earlygame digging monster, or monsters with double-edged auras or other symmetrical mechanics should be possible.

I left in a bunch of testing messages for easier debugging. Those should be removed and an actual message should be added before merge.

### DIFF
--- a/crawl-ref/source/dungeon.cc
+++ b/crawl-ref/source/dungeon.cc
@@ -5403,6 +5403,8 @@ monster* dgn_place_monster(mons_spec &mspec, coord_def where,
     for (const mon_enchant &ench : mspec.ench)
         mons->add_ench(ench);
 
+    mons->origin_level = level_id::current();
+
     return mons;
 }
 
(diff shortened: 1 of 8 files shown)


NormalPerson7:
Doesn't this provide an easy way to deal with some monsters? I'm particularly thinking of uniques like Mennas, Parghit, Nobody, who are difficult to kill but not so difficult to endure for enough turns to move them 3 floors, and who are often skipped by players.

Implojin:
Squashed and merged in 5afa24a926.

PR #4661, opened 2025-07-16
 visual update for Uskayaw's passive.

fbwer:
I've attempted a visual update for Uskayaw's passive skills. 

It's purely about adding visual effects. 

This aims to provide a more visually interesting sense of rhythm to the Rhythm God while playing with o-tab, 
and also offers a visual cue for when Pain Bond activates. 

Personally, I felt that Uskayaw, being a god of music, was too quiet and serene, which seemed unfitting. I deliberately wanted to give it a flashy, "club or party" vibe, because dancing isn't quiet. 

dcss usk visual_update

### DIFF
--- a/crawl-ref/source/god-passive.cc
+++ b/crawl-ref/source/god-passive.cc
@@ -2219,7 +2219,9 @@ void uskayaw_prepares_audience()
     if (count > 0)
     {
         simple_god_message(" prepares the audience for your solo!");
-        apply_area_visible(_prepare_audience, you.pos());
+        targeter_radius hitfunc(&you, LOS_NO_TRANS);
+	flash_view_delay(UA_PLAYER, MAGENTA, 300, &hitfunc);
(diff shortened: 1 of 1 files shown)


regret-index:
Flashing the screen in similar hues for two different passive Uskayaw effects will cause undesired degrees of constant visual noise, desensitizing their worshippers to more important flashes (like stuns, Mark, orb of winter Refrigeration). In terms of comparative regular player-sourced flashes, fe3fe7f removed this from Ignite Poison, I'd like to revise it completely out of Drain Life, and it currently heavily worsens using the Singing Sword.

A PR submitting much more subtle vfx than a bright flash- something like checker-boarded translucent sparks of alternating colours on every other tile?- would be highly preferred, which is complicated enough (and in need of new graphics) that I would prefer such a different arrangement to be submitted in a new PR instead.

PR #4662, opened 2025-07-16
[Feedback Required] Command palette concept

akukhta:
Hi, dcss folks!

It would be fair to say that amount of the hotkeys that player should remember for the comftable experience is quite overwhelming for the beginners. 

In my draft PR i have attempted to implement a QOL feature that may ease the such experience.

However, before diving deeper into optimizations and reducing the number of modification i've made to dcss core framework i would like to gather some feedback:
1. Do you find this feature usable?
2. What hotkey could be assign to the command palette?
3. What commands to add to the command palette?

If this feature has a potential, i will continue my work on:
1. Optimization of the substr search
2. Add "Select target" step for commands that require a target (if there will be any)

Demo:

Screen Recording 2025-07-17 at 00 36 59


akukhta:
I have tried multiple approaches to reduce the complexity of substring search in the code. 

Things i've tried:
- strstr
- std::string::find
- std::boyer_moore_searcher

Given the size of the input data (26 commands currently supported by the command palette), i'd stick to std::string::find function as it has better search time and more predictable time consumption (at least on my machine).

I am not sure if it is a good idea to introduce an ovekill solution implementing structures like tries and suffix trees to speed-up a search within a vector containing 26 strings.

Currently i'm trying to test the feature within Web Tiles version.

Meanwhile i'm still waiting for a feedback regarding what things to add and what hotkey could be actually assigned.

From my point of view, the best hotkey that it could have is cmd + p and moving 'Replay messages' command to the palette itself, but  obviously this replacement of a command is not ideal.

fslottner:
I like the concept! I think the palette should show the hotkey so that you can eventually learn them. for example:

> Cast spell (z)
> Go upstairs (<)

Issue #4653, opened 2025-07-14
Bane of Dilettante doesn't lower UC


PaperTar:
**PLATFORM and OS or BROWSER:**
webtiles

**DCSS VERSION AND GAME INFORMATION:**
0.34-a0-363-g3f269bfe75

**DESCRIPTION:**
Got Bane of Dilettante from a treasure trove, but instead of lowering my UC (~15 skill lvl) it lowered my Short Blades (2.4 skill lvl).

Issue #4674, opened 2025-07-22
Skill training panel shows `Skills reduced by [nothing] are in magenta` when bane of the dilettante is active


code2828:
**PLATFORM and OS or BROWSER:**
local console build

**DCSS VERSION AND GAME INFORMATION:**
Newest commit by the time of reporting (561764f)

**DESCRIPTION:**
I was stricken with the bane of the Dilettante:
<img width="1313" height="736" alt="Image" src="https://github.com/user-attachments/assets/4b7fc222-743b-4d66-bb21-1e5746f819ec" />
Traning panel shows:

<img width="972" height="707" alt="Image" src="https://github.com/user-attachments/assets/ce840ed5-d36f-48d1-8910-b2bd747a81c9" />

Very easy to reproduce.
1. Get the bane of the dilettante.
2. Press `m` to view current training.

**EXTRA FILES:**

PR #4678, opened 2025-07-23
fix bane of dilletante not lowering UC and not showing up in the 'skills are reduced by' line

code2828:
### fix: #4653, #4674
commit msg of cda74be:
```
I am aware that it is now possible for skill enhancement/reduction
messages to take up *4* lines (e.g. if one has all of Heroism, the
Charlatan's Orb, the Hermit's pendant, Ashenzari's wrath and the Bane of
the Dilettante active), but there are only 3 lines avaliable for these.
Not sure how to deal with this (sorry).
```


code2828:
Fixed by 9ab3afe and cbb4ff4

Issue #4682, opened 2025-07-26
Aphotic Marionette leaks illusion info


radio-gra:
**DCSS VERSION AND GAME INFORMATION:**
trunk (0.34a)

**DESCRIPTION:**
if a hostile illusion is created (rakshasa/Mara), then hovering Aphotic Marionette will reveal which one is real, and using it on the real enemy after the illusions were created will only apply shadowless status on the real enemy and not illusions

PR #4683, opened 2025-07-26
make Aphotic Marionette fail on illusions

radio-gra:
#4682 fix, instead of providing illusion info for free, make it auto-fail on illusions with custom message and apply shadowless status


Implojin:
Thank you for the PR and the bug report, I've gone with a simplified version of this fix in 26e05f8.

PR #4685, opened 2025-07-28
make fire and ice dragon scales have the same base AC

radio-gra:
due to alternative orb variants. 

(could also go with the reverse (fds AC 8 -> 9), personally i find fds/ids to be the worst "special" armour type and i rarely go for it)


Implojin:
I think there's reasonable consensus that we're not interested in this.

You've mentioned "due to alternative orb variants" as your reasoning here: Unfortunately, reducing item differentiation with the goal of reducing run-to-run variance, is not actually a gameplay change that we would want. (If we wanted all runs to play similarly wrt orbs, it would have been quite easy to not add new orb types.)

By meaningfully differentiating things, we increase combinatoric variance of potential gamestates from game to game, which increases replayability. (If anything, I'd prefer to make fire dragon armour and ice dragon armour _less_ similar.)

Thank you for the PR, but I'm going to close this as won't do.

PR #4686, opened 2025-07-28
Let Lua check the type of an ident

Aliscans:
Crawl didn't offer a way to check which subtype an identified item has. `item.subtype()` returned the same name ("body") for robes as for plate mail. `item.name("base")` returned different names for artefact and non-artefact items. Change each function as follows:

* Give `item.subtype()` a way to return the item type for armour.
* Give `item.name()` a way to return the `DBNAME` of an item.

Also delete the "choose the description level by number" code from `item.name()`. It doesn't work, and I don't know if it ever has. It could be made to work, but I don't know if anyone would actually find that useful.

### DIFF
--- a/crawl-ref/source/l-item.cc
+++ b/crawl-ref/source/l-item.cc
@@ -303,26 +303,30 @@ static int l_item_do_subtype(lua_State *ls)
         return 1;
     }
 
-    const char *s = nullptr;
-    string saved;
+    bool armour_slots = true;
+    if (lua_isboolean(ls, 1))
+        armour_slots = lua_toboolean(ls, 1);
 
-    // Special-case OBJ_ARMOUR behavior to maintain compatibility with
-    // existing scripts.
-    if (item->base_type == OBJ_ARMOUR)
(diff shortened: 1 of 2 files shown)


Implojin:
To touch base here: I've cherry-picked the dbname commit (which is now merged), and I've implemented the other two slightly differently.

We previously *had* an item slot lua bind, but apparently it was so poorly named that nobody was using it, as far as I could tell! I've cleaned this up slightly in 2fdb7e1e2 and changed the return signature such that you can now call item.equip_type and get a string like "weapon", "boots", "gizmo", etc.

Re: removing the "dead" bit of code: We have places in the game's code that are *intended* to use the number version of the call, so I've fixed that up instead of removing it. This may cause other, unknown issues since these codepaths were in error previously. Exciting!

Implojin:
Thank you for the PR btw, even though it took us a while to get to it.

Implojin:
As a hopefully final note I've also merged the subtype commit itself as 2fdb7e1e2 was doing something adjacent and missed the point of being able to _properly get the subtype from the subtype bind_. Thanks again for your patience, and for the PR. :)

PR #4702, opened 2025-08-04
Dysonsfear:add more arrival vaults

fslottner:
Add three new, relatively simple arrival vaults.

First time contributor, feedback welcome!

### DIFF
--- a/crawl-ref/source/dat/des/arrival/simple.des
+++ b/crawl-ref/source/dat/des/arrival/simple.des
@@ -2641,3 +2641,67 @@ MAP
              .c.
              ...
 ENDMAP
+
+NAME:   dysonsfear_arrival_boulevard
+TAGS:   arrival no_monster_gen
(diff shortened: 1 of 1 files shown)


regret-index:
This all look fine in principle, though:
* you've left in the extra-high testing weight in `dysonsfear_arrival_boulevard`,
* and for future submissions, it is preferred for commits covering very similar concepts (e.g. the first and fourth commits) or ones adding and removing things in sequence (e.g. the second and third commits) be squashed into one commit or squashed out before submission.

I can handle these matters myself before merging in this pull request, but if you wish to fix the first of these, then go ahead.

Is there any particular name you'd prefer to be listed under in the credits.txt?

fslottner:
Thanks for catching that. I'll handle the tweaks. Go with dysonsfear for the name.

fslottner:
Sorted, rebased as well!

PR #4705, opened 2025-08-06
fix: xom speech @head@ (55-ENN)

RypoFalem:
image
as reported by 55-ENN in discord


RypoFalem:
Amended the commit title since github treats `@head@` as referencing a github user

pooka109:
Applied; thanks.

PR #4712, opened 2025-08-07
Add a title splash by YLam

mainiacjoe:
This is a commissioned art piece I hired a co-worker to do.  She would like to be credited as "YLam",  I prefer that my username not be associated with it, except unavoidably as PR author.


PleasingFungus:
(Still retired, but wanted to pop in to say that this is sweet.)

PR #4713, opened 2025-08-07
Adds a handful of small entry vaults

ms813:
Adds 5 entry vaults.
* 3 cones of doors with varying sizes
* one square with some statues and doors on the corners
* a larger one that uses floor tiles to look a bit like an eye. It's a nice oval shape on console too

### DIFF
--- a/crawl-ref/source/dat/des/arrival/simple.des
+++ b/crawl-ref/source/dat/des/arrival/simple.des
@@ -2707,3 +2707,65 @@ x...x...x...x
 x...x...x...x
 xxaxxxaxxxaxx
 ENDMAP
+
+NAME:   ms813_arrival_square_statues
+TAGS:   arrival no_monster_gen
+ORIENT: float
+SUBST:  ? : GIAtTUVY:1
+MAP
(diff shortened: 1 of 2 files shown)


regret-index:
For the most part, these look fine. There are a few parts to fix, though:
* The three cone vaults should probably all be [lowered in weight as extensions of the same one vault](https://github.com/crawl/crawl/blob/master/crawl-ref/docs/develop/levels/introduction.txt#L246).
* The newline at the end of simple.des was lost in these vault additions.
* The preferred style for handling floor tile specifications would be to use grammatical symbols similiar to `.`s that have no inherent glyph assignment, like `'"-` or \`. Using glyphs that already have other meanings (like `g` being in the default item array and `o` being permanent transparent rock) which are then substituted back to regular floor are harder to read for future vault maintenance.

ms813:
@regret-index thanks for the review, I've made the changes you've suggested.
The blank lines at the end of the files aren't showing up in the diff, but are there locally, and if I try to add a second, the linter breaks the build.

PR #4726, opened 2025-08-13
fix: A temple with the wrong number of altars

### LINKED ISSUE #4725: dpeg_three_leaves_temple too many altars?
https://github.com/crawl/crawl/blob/50ae0a3e3163bfef7423b04b6d7df82899f99ba7/crawl-ref/source/dat/des/branches/temple.des#L3454

It's set to use 9 or 18 altars but in the 18 case there's actually 21 altars.

Kalcos in discord seems to have this temple and got this error message 

Image

RypoFalem:
dpeg_three_leaves_temple places [9 or 21] not [9 or 18] altars.

Hopefully fixes #4725


Implojin:
Thanks for the fix!

Issue #3415, opened 2023-12-23
FR: excluded staircases should block travel


CrawlOdds:
**PLATFORM and OS or BROWSER:**
webtiles

**DCSS VERSION AND GAME INFORMATION:**
0.31

**DESCRIPTION:**
A single-tile exclusion on a staircase does not block travel (e.g. autoexplore). This is deliberate; it was implemented in https://github.com/crawl/crawl/commit/56266ee012d54c6dc9021abb1deebc51da3539df (credit to Darby for digging this out). But in my opinion it should be reconsidered - using exclusions to block travel is common and those failing to work for stairs is surprising (and potentially rather dangerous).

CrawlOdds:
Done in the linked commit

PR #4727, opened 2025-08-13
Let single-tile exclusions block destination stairs. (CrawlOdds, Darby, riverfiend splats spriggans, Zewo)

### LINKED ISSUE #3115: Excluding stairs should warn for travel from the other side
webtiles

0.31-a0-1-g4198be8af4

**DESCRIPTION:**

- Exclude a staircase tile
- Go to the unexcluded end of the staircase
- Navigate the staircase; there is no warning that you are going on to an exclusion
It would be nice if there was such a warning, since we are walking into an excluded area (and I think there was, once upon a time)

RypoFalem:
For some reason single-tile exclusions on the destination side of stairs didn't give warnings or prevent autotravel while multi-tile exclusions did. This was done intentionally according to code comments but it's not clear what benefits this had.

On the other hand it caused several people confusion over the years when their exclusions didn't stop them from going into dangerous situations that they had marked.

Closes #3115 and #3415

### DIFF
--- a/crawl-ref/source/travel.cc
+++ b/crawl-ref/source/travel.cc
@@ -5324,15 +5324,12 @@ bool stairs_destination_is_excluded(const stair_info &si)
             return false;
         }
 
-        // Check for exclusions that cover the stair destination, but ignore
-        // those that have radius 1: those exclude travel in the _other_
-        // direction only (from the destination to here, not from here to the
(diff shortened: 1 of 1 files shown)


Implojin:
LGTM. Thanks again. :)

PR #4732, opened 2025-08-16
Recommend Tab instead of ! in spell library

coraxioU9:
Pairing `!` with the spell library  is a rational decision, given that it is normally accessed via `M`, which is shift+m for most players. But the main flaw in pressing `!` for navigation is that one can only navigate forwards (i.e. to the right). For a screen with as many  submenus as the spell library, this is could be better.

The good news is that the `Tab` key performs the same function as `!` here, but does not require shift -- even more critically, `Shift+Tab` will navigate the spell library backwards. Along with being slightly more conveniently placed than `!`, the `Tab` key may encourage more active use of hiding and showing spells, as well as the `auto_hide_spells` option.

(Usefully for me and for players, there is no underlying change: `!` can still be used to navigate between the pages if you want)

### DIFF
--- a/crawl-ref/source/dat/database/help.txt
+++ b/crawl-ref/source/dat/database/help.txt
@@ -286,13 +286,14 @@ Apart from (de)selecting skills, you also have the following toggles:
 %%%%
 spell-library
 <h>Spell Library</h>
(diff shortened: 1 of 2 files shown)


DracoOmega:
Merged, thanks.

PR #4738, opened 2025-08-19
Give Command to scales of the Dragon King

RypoFalem:
I think the new Command heavy armour ego is fitting thematically since kings give commands and the Command ego would allow many characters to cast Dragon's Call in golden dragon scales with extreme training.

I removed will+ because it's boring. Although an identical artefact could be still be randomly generated, hopfully this is more interesting thematically and has more impact on players' strategic decisions.


RypoFalem:
Possibly the Plus could be brought down from 9? I don't have a strong opinion

PR #4742, opened 2025-08-22
Make Mass Market shop vault heavier

NotAJumbleOfNumbers:
Add the Heavy ego to all weapons, plus the granite talisman and book of the Fortress (it literally says it's unreasonably heavy!)


NotAJumbleOfNumbers:
Added Morningstar/Eveningstar, Broad Axe, and Double Sword after concerns were floated about there not being any 1hand support.
Also added in Triple Crossbow, Dark Maul, and Sniper while I was at it

regret-index:
As [our vault documentation states](https://github.com/crawl/crawl/blob/master/crawl-ref/docs/develop/levels/syntax.txt#L114), the maximum length lines for .des files should tend to take is 80 characters; while we're not 100% consistent with this across all .des files, this has been mostly cleaned up over the past few years. (I wouldn't worry too much over the exact previous vault formatting, though, as shops.des is somewhat inconsistent with shop content formatting otherwise; I prefer to not waste lines for several different items). There are also added empty lines between the header and the vault contents, which we use almost nowhere in all vaults.

I can clean this up myself (rather than type out the exact problems with vault PRs multiple times) with your permission, or you may go ahead and fix this yourself.

NotAJumbleOfNumbers:
You have permission to fix it.  Thank you.

PR #4745, opened 2025-08-24
List the conditions Sniping ego does bonus damage in the description

RypoFalem:
The conditions are essentially "when SH would be set to 0". Many players assumed that any status that would cause a monster to be vulnerable to stabs would also cause a monster to be vulnerable to sniping but that isn't the case.


pooka109:
Ported to current trunk and applied; thanks.

PR #4746, opened 2025-08-24
fix: a trove map appearing about half as often

RypoFalem:
Dread Knight's Derelict Chapel was the only trove with a weight set. Without the weight, mapstat suggests it now spawns as often as other troves.


regret-index:
The trove has lower weight than other troves because its theme and contents are narrower than other troves (with zero spellcasting support for a namesake that has -Cast on it, and reaping weapons producing allies as some slice of both players and character archetypes clash with). I probably should have just left a comment as to why it's lower weight than other troves as a result.

(If this feels inconsistent, probably Rutra's Hidden Sanctum could also be lowered in weight, since less characters care about hexes and stealth, and the wall mimic gimmick is awkward to lift up to nearly as much prominence without as many other trove maps.)

I can handle both myself in a future vault review.

RypoFalem:
Oh thanks! (I don't mind if the pr is closed but wasn't sure if it was kept open for a reason)

PR #4749, opened 2025-08-25
Settings: add "default_skill_cost" & modify "default_manual_training"

bucketss:
The skill training menu has two main options: manual/auto and cost/training (aka ratio/percentage).

The setting _default_manual_training_ currently controls both of those options, and doesn't support all four possible configurations.

I split off the cost/training setting into its own variable—_default_skill_cost_—so that players can have better control over it via their init/crawlrc file.

default_skill_cost  
* true = begin new characters in cost mode (ratio)
* false = begin new characters in training mode (percentage)

default_manual_training  
* true = begin new characters in manual skill mode
* false = begin new characters in auto skill mode

In addition to covering all possible configurations, I believe this is also more intuitive behavior for _default_manual_training_. It doesn't make sense for any setting to do a secret second thing that the user has to dig into the options file to read about.

I tested all four configurations, and they seem to work as intended.

### DIFF
--- a/crawl-ref/settings/init.txt
+++ b/crawl-ref/settings/init.txt
@@ -1,0 +1,71 @@
+##### Crawl Init file ###############################################
+# For descriptions of all options, as well as some more in-depth information
+# on setting them, consult the file
+#    options_guide.txt
+# in your /docs directory. If you can't find it, the file is also available
+# online at:
+# https://github.com/crawl/crawl/blob/master/crawl-ref/docs/options_guide.txt
+#
+# Crawl uses the first file of the following list as its option file:
+#  * init.txt in the -rcdir directory (if specified)
+#  * .crawlrc in the -rcdir directory (if specified)
+#  * init.txt (in the Crawl directory)
+#  * ~/.crawl/init.txt (Unix only)
+#  * ~/.crawlrc (Unix only)
+#  * ~/init.txt (Unix only)
+#  * settings/init.txt (in the Crawl directory)
+
+##### Some basic explanation of option syntax #######################
+# Lines beginning with '#' are comments. The basic syntax is:
+#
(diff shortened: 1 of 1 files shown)


bucketss:
made a typo in the help file description 🥲

WizardIke:
The problem is that training percentage is basically useless information for someone with `default_manual_training = true` as having one skill active will show 100% for that skill and having two skills active will show 50% for each of them, 3 skills active will show about 33% each etc. Training cost on the other hand is very useful, for example, if you are choosing between training two skills and one is about 75% as good per level as the other but half the cost it is the correct skill to train. But you wouldn't have known this without knowing the training costs.

Unless we want to force everyone with `default_manual_training = true` in their rc file to also add `default_skill_cost = true` we would need to make `default_skill_cost = true` the default. However, this would easily confuse players using auto training as they would turn on a magic still to get a spell castable but the skill wouldn't train as the hidden training percentage had been set to 0% as they weren't using the skill due to the spell not being castable. I suppose we could make `default_skill_cost = auto` the default which preserves the current behaviour while allowing the player to choose between all four possible configurations.

I also don't understand why you describe skill training cost as a ratio, what is it a ratio between?

bucketss:
>  I suppose we could make default_skill_cost = auto the default which preserves the current behaviour while allowing the player to choose between all four possible configurations.

this would be fine, as I really only care about being able to choose between all possible configurations

and by ratio, I meant the relative training cost that's displayed on the skill menu

> If you press m to go to the skill menu, you can look at relative costs for each of your skills. These are given in multiplies of the "baseline cost", which is the cost of taking a skill from level 0 to level 1 with a +0 aptitude.[[2]](http://crawl.chaosforge.org/Skill_point#cite_note-2)

PR #4751, opened 2025-08-26
New species: Moon Moths

SentientSupper:
Moon moths are a species focused around a unique spellcasting gimmick that differs from those already present in the game. They are entirely unable to cast spells manually, instead having a chance to cast a chosen spell without consuming MP whenever they make an attack in melee. This chance rises with their spellcasting skill, which no longer increases their MP reserves, and decreases with the level of the spell being cast. Spells can be cast this way even when under the effects of silence. This can only happen once per turn.

They have a unique second trait that supports the automatic use of spells in melee by making them immune to damage from their own spells.

They come with a unique weakness to fire and electricity damage; whenever they are hit by the aforementioned damage types, they have a good chance of catching fire. This damage scales with XL to remain a threat unless mitigated by resistances. This makes less-popular amulet brands like acrobat and dissipation more desirable on this species.

They have mostly-balanced aptitudes and stats to support a hybrid playstyle. They have poor aptitudes with ranged weapons and shields as well as ally-summoning schools of magic that don't work as well with their spellcasting gimmick. Their positive evocations aptitude encourages the use of orbs over shields in the off-hand. Having previously undergone metamorphosis, they are good shapeshifters.


DracoOmega:
I want to start by saying that I really love the tiles (and am partial to mothfolk on an aesthetic level), and I think that the design space of 'species who recontextualizes how one interacts with spells' still has room for new ideas. However, I feel that the concept presented here not only has numerous implementation issues (some of which are fixable), but is also fundamentally troubled in a way that likely *cannot* be fixed without fundamentally rethinking it.

Being able to cast spells solely through a (relatively low) chance on melee attack has some pretty profound effects on how usable different spells are. Some of this is interesting (I think 'a species who values things differently than most species' is a positive point), but other aspects are deeply troubled:
 * Spells that rely on being able to be used at specific appropriate moments are fairly unusable (ie: if Borgnjor's Revivification has only a ~20% chance to cast when you want it to, its utility nearly vanishes. Even things like Blink become very bad for the same reason.)
 * Quite a few spells interact awkwardly with this system in other ways (eg: Vhi's Electric Charge being literally only usable with polearms or aqua form, Boulder only casting successfully in most situations if the triggering attack also kills the monster to make room for the boulder where it was standing, Paragon immediately destroying itself before fully charged if you don't remember to swap off the spell and on again while tabbing, Putrefaction draining the player repeatedly if they don't toggle it off immediately after it triggers, for the same reason, etc.)
 * Worse of all, a very large number of summon/buff/setup spells (ie: things a player usually wants to cast *before* fully engaging with monsters) aren't technically impossible to cast pre-combat, but instead are *deeply* fiddly and annoying to do so. The player is highly incentivized to carry some harmless monsters around (or possibly make some of them themselves!) so they can repeatedly 'fish' for spellcasts ahead of combat. While many of these sorts of 'micro-optimizations' that are periodically raised as concerns for other things in Crawl are sometimes theoretical, I think a fundamental difference here is that the 'annoyingly optimal' way to do this is the *literally only possible way to do this*. If you want to pre-cast Passage of Golubria, you *can*, but you might need to keep swinging at that zombie of yours over and over again until it actually happens (dealing with ally prompts on every swing) and then hope you roll the game randomly putting the portals in a useful spot. It frequently makes simple actions for every other species into lengthly ordeals.

Tabcast spells are also subject to normal miscasts, which means that warning prompts while tabbing are common.

While theoretically intended for hybrid play (and obviously requiring the player be willing to spend time in melee), spells are subject to full armour penalties and the species' apts are average-to-lousy, meaning that getting anything beyond intro-level spells online can take a level of investment that likely is not cost-effective compared to the melee/defense skills the character already wants. (Earlier iterations of revenant had the same issue, which is how Enkindle ended up the way it is.)

Mage starts are pretty dire in general. Conjurer starting with a ~30% chance to cast magic dart only *when they punch something in melee* is miserable. And your potion of magic doesn't even do anything! Now, I think it's fine for a species to be worse at some things and better at others, and obviously minotaur casters are pretty bad too (though I legitimately think minotaur mages have the better start of the two, which is saying something!), but I feel like mage starts being so bad on a species whose identity *centers* around magic is very awkward.

Beyond this, it's not clear to me that even under 'ideal' circumstances (ie: lategame, with both high spell and melee skills trained and strong AoE spells that work well with tabcasting known) tabcasting is especially strong. Sure, you can cast big spells with no MP cost, but you also can't frontload them, which is a significant part of what makes many of those spells strong. By the time you account for having to close the gap to enemies and attack the predicted number of times to cast a normal mage's MP worth of spells, in many cases the fight is either already won or you're too beat up to continue.

Basically, I think the species is in many ways treating tabcasting as if it were an upside when I feel like in practice it's a downside, and their magic aptitudes could handle being *massively* improved in compensation. (However, even then we're left with the much more profound issues I mentioned above of the awkwardness of how most non-damage spells interact with tabcasting, which is not something that can be addressed numerically.)

(As an addendum, I'm also not particularly fond of them getting sticky flamed in response to fire/elec as a downside, either. At a typical XL one is fighting orc wizards, they inflict meaningfully worse sticky flame than bombardier beetles do *on top of* their already meaningful direct damage. Feels too punishing on top of the species' other earlygame struggles. (And then eventually you get rF+ for it and it goes back to just a nuisance instead.))

In any case, the consensus among myself and multiple other devs is that it would not be suitable to merge any version of this species that has not undergone significant rethinking to address the biggest issues I have mentioned here.

SentientSupper:
To address the problems:
Situational/awkward spells:
- I'm not sure if it's possible to solve these issues while tabcasting remains the species gimmick. One option would be to let them cast normally, but with their nerfed MP pool from spellcasting not giving MP and a possible additional MP malus. I'm a bit reluctant about this approach because the inability to cast spells normally is an intended drawback for the species but it's doable. 

Miscasts:
- I take it that this is related to some rc-file options that warn you if the miscast chance is too high? I don't know what to do about this because I don't use those options.

Hybridization:
- I see three ways non-mutually exclusive of solving this.
  1. Increased tabcast chance so training up magic as an aux-attack is way more cost-effective.
  2. Simply increase their aptitudes to make it easier.
  3. Some new passive effect that makes casting easier/stronger. I'm wary about overloading their mechanics when they already have tabcasting which is why I'll likely take the other solutions.

Mage starts:
- The manual casting/increased cast chance solutions I mentioned earlier could solve this.

Big spells/tabcasting drawback:
- I have my concerns too with how if you could cast powerful spells like firestorm, you would have a much easier time if you had played another species instead and blasting everything from the start. Simply increasing the chance to tabcast spells could work in addition to the return of manual casting.

High flammability:
- The entire species would function just as well even if this were removed and I mostly tacked this on for flavor. If removed, I feel like some sort of drawback should be present so the species doesn't feel like a human+ especially if I restore manual casting and improve their aptitudes. Perhaps nerfing its damage would be the way to go until I come up with a better idea.
  
I'll wait a while to see if any other devs have anything else to add before I make any revisions and to think over how to improve the species.  (suggestions welcome!)

regret-index:
The miscast prompts are not any particular custom RC setting, but instead have to do with our default Crawl game settings. A miscast chance prompt will show up each swing if it's at a light red miscast rate or higher, which creates constant prompts when otherwise fighting normally. Spells can be at a passable cast rate but still be indicated as highly dangerous due to one's low health early on (e.g. casting Mephitic Cloud at xl 4 on some current low health good apts species) or due to needing to only be cast sparingly early into the fight (like much of Necromancy and other-such ally spells) or both. This is, admittedly, an awkward consequence of our currently overly-cautious miscast arrangements (with miscast damage having extremely rare chances to actually reach notable damage), but it is also very awkward to foreground in a central species gimmick without further adjustments.

I'm not the biggest fan of the flavour in this precise naming schema, even if insect directions and moths flavours are both aesthetically appealing. We've got three distinct moths in-game that definitely aren't anthropomorphic, and any excuse of "being from mystic space" falls a bit flat when nearly every space association we have besides the extremely rare moon troll (already something anthropomorphic!) or orbs of stardust are all malformed / abstract horrors. Maybe something along the lines of `"Moon Mothfolk"` would be fine? (I'm not enthusiastic about engaging in lengthily bikeshedding this out before the power level and playability matters of this PR are addressed, of course.)

(I'm also unenthusiastic about it also being unavoidably conjurations-focused either as works best for randomly firing midcombat, as while it may be fairly less safe than ally builds this is also strongly suggested on quite a few species gimmicks and apts (Djinn, Mountain Dwarves, Tengu, Revenants) plus other common / popular arrangements (Deep Elves, Gargoyles). The obvious counterpart to these in sun moths might be damage-focused, but the other two moths are fairly more indirect, more common, and more famously dangerous. Some of those mentioned apts changes specifically helping the non-destructive schools might help?)

I've not played a lot of serious games testing through it myself, but after having looked through the numbers as provided, the species definitely needs some sorts of substantial buffs to make a very complicated upside (both in execution and codebase burden!) avoid being almost immediately in practice a very complicated significant downside.

SentientSupper:
Miscasts:
- Turns out I had a rc file setting that disabled miscast chance warnings entirely so I never ran into prompts. I think it'd be a bad idea to force this on all players though. 
- Reducing cast chance proportionally to miscast chance or making miscasts not have any effect feels a bit too powerful. Not sure what else to do here.

Flavor:
- Personally I'm not too attached to either the Moon or the Moth part of the species. I just haven't thought of anything better. I've considered alternatives like Astral/Cosmic moths or simply Mothmen. I'm open to suggestions here.

Current in-dev changes are as follows:
- Spells can be cast normally again at a +1 MP cost.
- New ability to choose your tabcast spell that requires 5 turns like spellspark servitor imbue.
- Channeled spells cost MP to channel again.
- +1 aptitudes to conj, hex, summon, necro, forge, tloc and alch.
- Tabcast chances multiplied by 1.5x.
- High flammability damage reduced.

pooka109:
Of the alternate species name suggestions, Mothfolk and Astral/Cosmic Moth sound like the best options to me. (Also, Cosmic Moth especially would allow Cosmothnaut as a title, for something like Translocations.)

SentientSupper:
I'm thinking of settling with Mothkin (Mo/Mk) because Mothfolk is a bit too close to Merfolk and Astral/Cosmic/Whatever Mothfolk is a bit of a mouthful.

A possible downside to replace high flammability that I'm gonna experiment with is giving them permanent Sapped Magic which prevents them from manually casting multiple spells in a row.

SentientSupper:
I reckon moths are ready for another review now.
Changes:
- You can now cast again normally, but this prevents you from doing so again until you spend a few turns without moving.
- High flammability removed.
- Slight aptitude increase.
- Tabcast spell is chosen with a new ability which takes 5 turns.
- Tabcast chance increased, but is now reduced by high failure chance.
- Tabcasting no longer miscasts.

kuniqsX:
Will Moon Moths get special benefits from wearing the helm of light?

SentientSupper:
They can't wear helmets because of antennae.

Monkooky:
While the -Mcast status seems like a better system than outright banning manual casting, I think it will still be rather annoying for a reasonably large number of characters. 
The duration is short enough that if you find a sleeping monster you can reasonably duck out of LOS, summon something, wait off the delay, summon a second thing, wait off the delay, repeat two more times, and then move back in to LOS.

SentientSupper:
-Mcast is intended to be a downside that dissuades the regular caster gameplay of spamming p at enemies and forcing players to engage with the species upside which in my opinion is busted enough to make up for it.
As for summoners vs sleeping monsters, I can think of a couple possible ways to address the problem.

- Increased spell noise for manual casting. Mr optimal man will probably just walk farther away or just worship dith and keep doing the same thing so not a great idea.
- Make -Mcast expire slower with no visible monsters so the summons time out. Not really a big fan of this one because it could increase the resting time after each battle by a lot. Also promote keeping rat kiting.
- Killing moth stealth as if the -2 aptitude wasn't enough. Negatively impacts builds who weren't gonna summon anyway, especially stabbers. Will probably take this approach if it becomes a big issue.

Maybe the real issue is letting players pre-summon out of combat?

SentientSupper:
Important changes:

- Tabcasting now works with ranged attacks and manual spellcasting with -33% and +100% modifiers to tabcast chance respectively. Having a tabcast chances above 100% now gives a chance of tabcasting more than once.
- +2 apt to ranged and throwing. +1 apt to invo and evo.

crawl-ref/source/ability.cc
```diff
@@ -2911,6 +2916,7 @@ bool activate_talent(const talent& tal, dist *target)
         case spret::fail:
             if (!testbits(abil.flags, abflag::quiet_fail))
                 mpr("You fail to use your ability.");
+            you.apply_berserk_penalty = true;
```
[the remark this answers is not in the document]
SentientSupper: There was a bug with berserk that would cause berserk to not run out if you found a way to cast abilities/do other things during it which bypassed the berserk penalty thing such as interacting with the menu through the local tiles menu. Mothkin used to make use of the berserk penalty to determine if a channeled spell should continue channeling when attacking but I forgot to revert this when moving away from melee-only tabcasts.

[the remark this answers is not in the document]
SentientSupper: Done.

crawl-ref/source/beam.cc
```diff
@@ -4540,6 +4547,13 @@ bool bolt::ignores_player() const
     if (flavour == BEAM_DIGGING)
         return true;
 
+    if (agent() && agent()->is_player()
+    && you.has_mutation(MUT_SAFE_SPELLS)
+    && is_player_book_spell(origin_spell))
```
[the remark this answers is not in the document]
SentientSupper: Yes. This is intended.

crawl-ref/source/item-use.cc
```diff
@@ -3279,6 +3279,7 @@ void tile_item_pickup(int idx, bool part)
         if (quantity > env.item[idx].quantity)
             quantity = env.item[idx].quantity;
     }
+    you.apply_berserk_penalty = true;
```
Implojin, line 3282: more berserk things

crawl-ref/source/item-use.cc
```diff
@@ -3301,11 +3302,14 @@ void tile_item_drop(int idx, bool partdrop)
         if (quantity > you.inv[idx].quantity)
             quantity = you.inv[idx].quantity;
     }
+    you.apply_berserk_penalty = true;
```
Implojin, line 3305: and here

crawl-ref/source/item-use.cc
SentientSupper: For this I will make a separate pr instead of reverting it here because it's an actual abuseable bug with berserk in tiles.

crawl-ref/source/melee-attack.cc
```diff
@@ -1222,6 +1226,15 @@ void melee_attack::handle_spectral_brand()
     spectral_weapon_fineff::schedule(*attacker, *defender, mutable_wpn);
 }
 
+void melee_attack::maybe_trigger_tabcast()
+{
+    if (!attacker->is_player() || !you.has_mutation(MUT_AUXILIARY_CASTING) || !defender->is_monster() || is_projected)
```
Implojin, line 1231: I think you might want more restrictive conditions here, to sidestep some cases of the player tabcasting on things that would cause problems. One suggestion at least is `!mons_aligned(attacker, defender)` to prevent the player from tabcasting on summons they're hauling around behind them. You might want more than that!

SentientSupper: Done. Can't think of any other conditions that aren't covered yet.

crawl-ref/source/mon-project.cc
```diff
@@ -588,6 +588,18 @@ bool iood_act(monster& mon, bool no_trail)
         if (mons && mons->type == MONS_ORB_OF_DESTRUCTION
                  && mon.type == MONS_ORB_OF_DESTRUCTION)
         {
+            //if tabcasting, kick it forward instead if they are heading
+            //in the same direction to prevent instant collisions
```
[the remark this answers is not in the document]
SentientSupper: I think it's better than disappointing players with fizzling every time you cast more than 1 orb at a time even if the implementation isn't exactly the best. I'm concerned about potential infinite loops but I've tested a full game with this and hopefully the requirement for the orbs to be heading in the same direction is enough to prevent that.

crawl-ref/source/ranged-attack.cc
```diff
@@ -835,6 +839,18 @@ bool ranged_attack::player_good_stab()
     return false;
 }
 
+void ranged_attack::maybe_trigger_tabcast()
+{
+    if (!attacker->is_player() || !you.has_mutation(MUT_AUXILIARY_CASTING) || !defender->is_monster()
```
Implojin, line 844: same alignment check as the melee_attack version, please

crawl-ref/source/nearby-danger.cc
```diff
@@ -130,7 +130,10 @@ static bool _mons_is_always_safe(const monster *mon)
           || mon->type == MONS_BUTTERFLY
           || (mon->type == MONS_BALLISTOMYCETE
               && !mons_is_active_ballisto(*mon))
-          || mon->type == MONS_TRAINING_DUMMY && !mon->weapon();;
+          || mon->type == MONS_TRAINING_DUMMY && !mon->weapon()
+          || (you.has_mutation(MUT_SAFE_SPELLS)
+              && (mon->type == MONS_BALL_LIGHTNING || mon->type == MONS_FULMINANT_PRISM)
```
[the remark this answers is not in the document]
SentientSupper: Fixed. Didn't know you could check for peripheral monsters back when I wrote this. I blame my lack of crawlcode familiarity.

crawl-ref/source/player.cc
```diff
@@ -4092,7 +4109,7 @@ int get_real_mp(bool include_items)
     // the last 4 give no mp
     int enp = min(23 * scale, scaled_xl);
 
-    int spell_extra = spellcasting; // 100%
+    int spell_extra = you.has_mutation(MUT_AUXILIARY_CASTING) ? 0 : spellcasting; // 100%
```
Implojin, line 4112: From a balance perspective, what's the motivation for this?

crawl-ref/source/ranged-attack.cc
Implojin: Ok, thank you.

crawl-ref/source/player.cc
SentientSupper: Mothkin have a strong incentive to train more spellcasting for a higher tabcast chance, but doing so would also give them a lot of mp that they don't have a use for. I don't want guardian spirit effects to be too good on mothkin to the point of making other options worthless since it would give them a ~33% increase to effective hp with little drawback otherwise.

SentientSupper:
While brainstorming ideas to alleviate the code burden of automatic casting, I have decided to categorize existing spells and how the pr currently deals with them.

type: it just works
	stone arrow, fireball etc
	spells that just work
	solution: not needed

type: indirect targeting
	paragon, lrd, prism, tempering
	stuff that doesn't work when directly targeting a monster and needs to target somewhere close by
	solution: special case targeting a random nearby tile instead for those spells

type: counterproductive recasting
	teleport other, necromancy spells, summonings, paragon, confusing touch
	teleport other can cancel itself
	necromancy spells like animate dead, death channel, fugue lose their effects when recast
	summon spells constantly reposition active summons which is undesirable
	paragon may not want to recast before charging up
	solution: check if doing so would be counterproductive and cancel the cast automatically if so

type: mixed recast
	searing ray sometimes wants to be recast to retarget but not always
	solution: special case it to retarget even when not recasting

type: it just doesn't work
	passwall
	solution: don't

type: noisy
	freeze, jinxbite, ozocubu's armor, beckoning, vhi's, haunt
	work as expected but will print a lot of messages that flood the log meaninglessly when succeeding or failing
	solution: silence them or cast them only as frequently as is needed

type: rigid targeting
	gavotte, splinterfrost shell, boulder
	spells must be aimed by targeting a position adjacent to you even if the target monster is far away from you
	solution: clamp the targeted location

crawl-ref/source/spl-cast.cc
```diff
@@ -3408,6 +3464,330 @@ void do_demonic_magic(int pow, int rank)
...
+    }
+
+    you.attribute[ATTR_TABCASTING] = 1;
+    cast_a_spell(false, spell, &target);
+    you.attribute[ATTR_TABCASTING] = 0;
+}
+
+static bool _is_channeled_spell(spell_type spell)
+{
+    return spell == SPELL_SEARING_RAY || spell == SPELL_FLAME_WAVE
```
Implojin, line 3633: Would it make sense to add this as a spflag over in spl-data, so that we can just check that instead of hardcoding the currently-existing channeled spells?

crawl-ref/source/spl-cast.cc
```diff
@@ -859,8 +875,19 @@ spret cast_a_spell(bool check_range, spell_type spell, dist *_target,
...
+    {
+        if (is_tabcasting())
+            return spret::abort;
+
+        crawl_state.zero_turns_taken();
+        return spret::abort;
+    }
+
+    //this is here instead of can_cast_spells() to avoid interrupting channels
+    if (you.duration[DUR_NO_MANUAL_CAST] && !is_tabcasting() && !you.divine_exegesis)
```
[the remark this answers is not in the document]
SentientSupper: Killed -MCast.

crawl-ref/source/player.cc
SentientSupper: This has been removed in the latest patch.

Implojin:
I want to start this off by saying thank you for your interest in Crawl. The game wouldn't be what it is without community contributions!

Unfortunately, this PR has been open for a while now, and nobody on the devteam wants to merge it.

We can all see that you've put a lot of time into this, but both the design and the implementation here have problems that seem unfixable while the core gimmick remains "tabcasting":

On the implementation side, the manual design assessment and code handling your tabcasting implementation requires for every new spell we might add to crawl in the future is simply a non-starter.

This imposes design constraints that will have a chilling effect on *all future spells* we might add to the game; if we merged this tabcasting implementation, we'd forever need to ask ourselves, "How will this work with moths?"

Many spells simply aren't intended to work within the conditions that tabcasting imposes. Passwall is a good example, but also spells like Passage of Golubria, Death's Door, Borg's Revivification, or even simpler things like Slow -- any spell that wants a non-monster target, or is intended to be used at a specific moment of combat rather than at a random point in the fight, is going to have problems with this. That latter category includes buffs and debuffs! These are huge design restrictions to be imposing onto *every future spell in crawl*. DracoOmega mentioned all of this in her very first response to this PR, and although you followed up with a flurry of commits, *none of them addressed this basic problem* because it's inherent to your tabcasting design.

This PR is also overengineered: You have all of these fiddly little special cases sprinkled everywhere that, frankly, should be simplified out. https://en.wikipedia.org/wiki/KISS_principle

You have the immunity to self-damage which opens up a lot of problems: Multiple people (including players) have pointed out that autocasting Conjure Ball Lightning on the player is powerful, but it isn't good gameplay. (Note that even straightforward conjurations have problems with this design!)

Then you have the immunity to silence, which is just, why? If this is intended to be a hybrid species, let them lean on their hybrid abilities to deal with silence. Statuses that affect the player, except when they don't because the player is special, create power spirals where the game's complexity budget gets destroyed. That isn't a path we need to go down. Any time we add something to the game, we need to explain it to players in a way they can understand and remember. Keep it simple.

You've said repeatedly when receiving feedback that you don't want to simplify some bits of this because you want them as they are, and that compromise is needed (by which you seem to mean you want us to merge this despite our many misgivings). Unfortunately, that's not how this works. You want us to upstream this, we're going to be responsible for maintaining it, and it doesn't meet our standards.

In a theoretical world, if the Mothkin design were impressive enough that we were excited about it, we might be willing
to take this on and find ways to work around it in the future. The blunt truth here, though, is that at least 5 separate devs (and one contributor) have independently concluded this PR doesn't work.

The underlying concept of interacting with spells in non-standard ways *is* interesting, and I'd like to see alternate (not-tabcasting) explorations of that concept space, but this isn't it.

If you want to take this design and put it in a fork, you can do that, but we're not going to upstream it.

PR #4763, opened 2025-08-31
fix: correct typo 'Mahkleb' to 'Makhleb'

Steinardarri:
Noticed this when making the typo myself when searching the repo


Implojin:
Thank you for the patch.

PR #4765, opened 2025-09-01
Give Demigods Clarity at XL 13

Steinardarri:
Demigods now start with innate Mutation Resistance 1.

Since Halflings were removed back in 0.27, no species has had built-in mutation resistance. Demigods feel like a good candidate for it thematically, they’re made in a divine image and should be less vulnerable to chaotic corruption than ordinary mortals.

On the gameplay side, it also fits their role as a straightforward “big stats, no god” species — solid and reliable, but still an advanced challenge. Mutation resistance doesn’t really come into play until the mid- to late game, so it shouldn't strengthen them too much.

This idea was discussed earlier in #1728, where ideas for the last Demigod rework came from.

### DIFF
--- a/crawl-ref/source/dat/descript/species.txt
+++ b/crawl-ref/source/dat/descript/species.txt
@@ -22,7 +22,7 @@ They regenerate magical power rapidly.
 Demigod
 
 Demigods refuse to worship any god, but make up for this with divine attributes,
-mutation resistance and high reserves of health and magic.
+clarity of mind and high reserves of health and magic.
 %%%%
 Demonspawn
 
--- a/crawl-ref/source/dat/descript/species.txt
+++ b/crawl-ref/source/dat/descript/species.txt
@@ -21,8 +21,8 @@ They regenerate magical power rapidly.
 %%%%
 Demigod
 
-Demigods refuse to worship any god, but make up for this with divine attributes
-and high reserves of health and magic.
+Demigods refuse to worship any god, but make up for this with divine attributes,
+mutation resistance and high reserves of health and magic.
 %%%%
 Demonspawn
 
--- a/crawl-ref/source/dat/species/demigod.yaml
+++ b/crawl-ref/source/dat/species/demigod.yaml
@@ -49,7 +49,8 @@ mutations:
     MUT_HIGH_MAGIC: 1
     MUT_FORLORN: 1
     MUT_DIVINE_ATTRS: 1
-    MUT_MUTATION_RESISTANCE: 1
+  13:
+    MUT_CLARITY: 1
 recommended_jobs:
   - shapeshifter
   - conjurer
--- a/crawl-ref/source/dat/species/demigod.yaml
+++ b/crawl-ref/source/dat/species/demigod.yaml
@@ -49,6 +49,7 @@ mutations:
     MUT_HIGH_MAGIC: 1
     MUT_FORLORN: 1
     MUT_DIVINE_ATTRS: 1
+    MUT_MUTATION_RESISTANCE: 1
 recommended_jobs:
   - shapeshifter
   - conjurer


regret-index:
With neqoxecs having lost permanent malmutation in 0e930c6, and orbs of fire being set to only appear in a third of all non-zig games in eee5569, rMut is fairly less relevant than it was previously. I'm rather disinclined to staple onto a species with otherwise straightforwardly restrained and minimal mechanics a starting passive that starts having relevance mostly in Elf and Slime- while the three undead species have a stronger version innately, they can also see upsides of other factors of their innate stat like rPois and rN pre-Lair. (There's something to be said for how Death Form is highly prominent in extended for rMut alongside rTorment, but matters like [Hell-Pan Roulette](https://github.com/crawl/crawl/wiki/0.34-Development-and-Release-Plan#hell-pan-roulette-regret-index-dracoomega) should also help with this.)

I would personally be open for giving them some mild additional upside that could, for example, help with high strength being more interesting to take compared to the other two stats, or to help them through the earlier part of the game where the stats haven't done a lot and one could be leaning on an earlier god.

Implojin:
I don't think we're likely to do this.

From a mechanical perspective, rMut makes fewer things happen. Having less stuff going on makes the game less interesting. We really don't use rMut much these days: where it exists it has been inherited from earlier design eras and it is only currently used on one unrand that is not Zin-themed. rMut is also not very impactful: the player has positional ways to avoid mutators, and mutations tend to not be very relevant throughout the course of a 3 rune game. rMut *sort of* makes more sense where it's used on Zin, as that's at least a significant part of the god identity -- but even there, Zin is not the most interesting of gods.

We're not trying to maintain feature parity with removed species -- [we had reasons for removing them, you know](https://github.com/crawl/crawl/commit/0520e8ecd4ae51adf44cfb3dd49c21dd23d5a276)?

We've also recently had a set of changes in trunk that replaced some temporary malmutation sources with Doom effects (eddbedc2, dd4ca101, 3cc1a7f2c), so adding this would be even less impactful than before.

I'll leave this open for a little while in case anyone wants to chime in on the discussion, but more broadly you may also find it beneficial to discuss potential gameplay changes with the team in #crawl-dev on Libera IRC.

Steinardarri:
I could've started a discussion first, but the change is small enough that I decided to just chuck in a pull request.

I didn't mean that "there must always be an rMut species", rather that the door was open and I've felt that Demigods could benefit from it.

It is a rather minor change so it might not be worth it to add, since changes are also being made to reduce mutations (didn't realise that implication)

Regarding rMut reducing variety, when picking (or randomly getting) Demigods you aren't expecting variety anyways.

### Final note

Is there some space to make Demigods a little bit more interesting though? What about gaining clarity at XL 13 for example? Or immunity from getting drained at XL 13.

Implojin:
>  the change is small enough that I decided to just chuck in a pull request

Sure, that's fine! I just mentioned #crawl-dev above because it can be difficult for devs who are new to the project and aren't familiar with crawl's extensive history to navigate what kinds of gameplay changes might be received positively, and talking with us helps on that front.

> Is there some space to make Demigods a little bit more interesting though?

We're always open to gameplay changes that add meaningful decisions or exciting gameplay.

I don't know that passive clarity or drain resistance would sound all that interesting to me -- these kind of have the same pitfall as rMut above, in that giving out resistance to gameplay effects reduces the decision space for players when interacting with monsters that have those effects. I think we would be more open to effects that are *more impactful* and *add* to the decision space instead of *reducing* it, if that makes sense?

(I'm just one dev and the team's decisions are not monolithic, it's always possible someone may have a different take.)

Steinardarri:
> I don't know that passive clarity or drain resistance would sound all that interesting to me -- these kind of have the same pitfall as rMut above, in that giving out resistance to gameplay effects reduces the decision space for players when interacting with monsters that have those effects.

Doesn't that apply to every resistance though? Like with Tengu having innate acrobatic or the species that innately have rPois. The latter means that you don't have to sideline your potential build just to cover off the valuable resistance in the 3 lair poison branches.

_"I really want to use this unrandart cloak with stealth and slaying I found but the only source of poison resistance I have for Lair and it's branches is a cloak with rPois, so I am inclined to use it instead"_

I don't see removing these kind of scenarios limiting, but rather freeing. The same could be said with regards to clarity & will+ for Shoals for example.
Knowing that the threat of getting disabled is reduced, with no god powers to help bail you out, will free you to try more unconventional playstyles without gimping yourself. For example: using more niche spells with your high int or be okay with using a hand cannon in a heavy armor with your high strength and dex combo.

-----

I think I might actually change the PR to add clarity instead.

Implojin:
Well, it's been a couple of weeks and nobody has merged this or commented, so I think it's safe to say we're not interested in this one. If you have any other ideas feel free to open a new PR!

PR #4769, opened 2025-09-02
feat: Rework Mask of the Dragon

LiciTheCrawler:
Mask of the Dragon is one of the boring unrands. This changes it to have a 10% chance to summon a dragon on using an Invocation with the same rules as the ephemeral shield mutation + there being enemies in LoS. It stays at +0 and loses sInv, Will+ and +3 Slay.


oktalk:
So then, you cannot wear this if your god doesn't allow it.

SentientSupper:
Not really a fan of this change because of how some gods are entirely useless with it and others are overwhelmingly better. Some gods have spammable invocations like beogh, makhleb, yred and hep while others either have none or piety-costly invocations that combined with the very low 10% chance make it rather unappealing. I think it should either keep its old stats or have a tweak to its triggering condition that makes it more generally-useful.

Implojin:
Also it looks like the linter caught an unneeded pair of braces, you can run util/unbrace and do a new commit to fix that

crawl-ref/source/ability.cc
```diff
@@ -2882,21 +2938,21 @@ bool activate_talent(const talent& tal, dist *target)
...
+            if (_not_free_religious_ability(abil.ability)
                 && you.has_mutation(MUT_EPHEMERAL_SHIELD))
             {
                 you.set_duration(DUR_EPHEMERAL_SHIELD, random_range(3, 5));
                 you.redraw_armour_class = true;
             }
 
+            if (_not_free_religious_ability(abil.ability)
+                && you.unrand_equipped(UNRAND_DRAGONMASK)
+                && there_are_monsters_nearby(true, false, false))
```
Implojin, line 2950: `there_are_monsters_nearby(true, true, false)` , please. This doesn't strictly change anything because the mask has sinv already, but we do want visibility-to-the-player to be enforced here in case anyone ever does a drive-by-change of the unrand's artprops in the future.

Implojin:
Merged with some minor style tweaks in 089dd02ed7. Thank you, Lici :)

PR #4780, opened 2025-09-10
Fix tile mismatch on monster lookup menu

WizardIke:
When typing part of a monster's type into the ?/m menu, it will list all monsters who's type is similar to what you typed in. In this list, monsters that usually have a randomized tile will always show the same tile. However, after selecting one of these monsters to view its description, it will have a randomized tile. To fix this tile mismatch, also don't randomize the tile when viewing the description.


pooka109:
Applied; thanks.

PR #4792, opened 2025-09-19
Add more Spanish translations

person50605:
Add more Spanish translations for various things. This adds a few files, and I'm not sure if they need to be integrated or will just work.
This also edits the German description for Lair to say it has five floors, and not eight.

### DIFF
--- a/crawl-ref/source/dat/descript/es/unrand.txt
+++ b/crawl-ref/source/dat/descript/es/unrand.txt
@@ -1,8 +1,8 @@
 %%%%
 amulet of Elemental Vulnerability
 
-Un dije de acero azúl con quemaduras borrosas. Su creadora, una mujer muy sensata, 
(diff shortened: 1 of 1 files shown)


pooka109:
Applied; thanks.

What name would you like to be known by in the credits?

(Also, if you didn't know, running checkwhite removes all unneeded whitespace from files in one shot; you don't have to do it manually.)

person50605:
I'd like to be known as person50605 in the credits, please. Also, how does one run checkwhite?

pooka109:
Done.

As for checkwhite, it's an executable Perl script  in the source/util/ directory that you can run.

pooka109:
And, importantly, checkwhite needs to be run inside the directory you want it to work on.

So, since it's in source/util, if the files you want it to run on are in source/dat/descript/es, you'll need to run it as:

`./../../util/checkwhite`

PR #4793, opened 2025-09-19
Stop debug_xom_effects() (&^X) from trying to run -1 times.

Aliscans:
prompt_for_int() signals an error to debug_xom_effects() by returning -1, which caused it to create a nonsensical xom_debug.stat file.

Fix this by only generating a file if the user requests a positive number of iterations.


Implojin:
Thanks for the fix!

Aliscans:
No problem. Iit's just something I noticed when I was looking at something else.

PR #4802, opened 2025-09-22
Update !Lignification's description to reference turning bloodless

person50605:
See pr title

Fixes crawl#4752


pooka109:
Applied; thanks.

PR #4803, opened 2025-09-22
Update the translation guide

person50605:
The Transifex project has been deleted, and we now edit the text files directly, so the translation guide needs to be updated.

Fixes crawl#4624


pooka109:
Applied; thanks.

PR #4804, opened 2025-09-23
Fix merfolk fish-tail clipping in Quill form

tianmaster:
Quill form is not a full-body polymorph like Dragon or Fortress Crab; it only alters certain body parts, similar to Hive or Maw. Therefore the tile should preserve the player's original species silhouette.
Genies and nagas already have bespoke lower-body tiles for this purpose, but merfolk were left wearing a humanoid "legs" layer on top of their fish tail. Copy the approach used for Hive: when a merfolk is in water (fishtail active) simply crop the bottom of the Quill armour tile so the tail remains the only visible lower extremity. This produces a clean, species-consistent sprite.

After fix:
quill_merfolk_on_land
_on land_
quill_merfolk_in_water
_in water_


WizardIke:
Thanks for this fix. It should be merged now

PR #4805, opened 2025-09-24
feat: Add selection menus for Gozag’s potion petition and call merchant abilities

tianmaster:
Until now, both Gozag’s Potion Petition and Call Merchant abilities printed their choices directly to the message window and only accepted selection via hot-keys.
On tiles/web builds, long random shop names could be truncated (especially for Call Merchant) and mouse selection was impossible.
This patch introduces a proper shop-style menu for both abilities:

1. Works like a one-time “pop-up shop”

2. Supports both hot-keys and mouse clicks

3. Guarantees all options remain visible regardless of font/screen size

Because the menu re-uses the existing icon, Call Merchant now also shows the shop-type icon next to each candidate, giving players an extra visual cue.

potions
potion petition

shop
call merchant

### DIFF
--- a/crawl-ref/source/shopping.cc
+++ b/crawl-ref/source/shopping.cc
@@ -2823,7 +2823,7 @@ int gozag_shop_petition_menu(const vector<shop_offer> &offers)
 {
     GozagPetitionMenu<shop_offer> menu(
         offers, &_shop_line, &_shop_tile, "gozag-shop",
-        "Which merchant should call?");
+        "Fund which merchant?");
     menu.show();
     return menu.selected_index();
 }
--- a/crawl-ref/source/shopping.cc
+++ b/crawl-ref/source/shopping.cc
@@ -2729,18 +2729,94 @@ class GozagPetitionMenu : public InvMenu
     const vector<T> &_items;
     string (*_line_func)(const T &);
     tileidx_t (*_tile_func)(const T &);
+    int (*_price_func)(const T &);
+    bool _confirmed = false;
+
+    void _update_footer()
+    {
+        string top_line;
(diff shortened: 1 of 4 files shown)


crawl-ref/source/shopping.cc
```diff
@@ -2716,3 +2717,113 @@ string ShoppingList::item_name_simple(const item_def& item, bool ident)
...
+ * Display a menu for Gozag's shop petition.
+ *
+ * @param offers A vector of shop offers.
+ * @return The index of the selected offer, or -1 on abort.
+ */
+int gozag_shop_petition_menu(const vector<shop_offer> &offers)
+{
+    GozagPetitionMenu<shop_offer> menu(
+        offers, &_shop_line, &_shop_tile, "gozag-shop",
+        "Which merchant should call?");
```
randomdevpete, line 2826: "Call which merchant?" would be nicer phrasing i think

tianmaster:
There is currently an issue here. After changing to menu display, if there are insufficient coins when purchasing the effect, it will cause a crash. I forgot to handle this issue, I will make the necessary changes later.

tianmaster:
Okay, I have created this ability interface by completely imitating ShopMenu. At the same time, it also solved the crash.
Now, items that cannot be purchased will be visually grayed out. When you select an item that your golds cannot afford, it will prompt you that you cannot purchase and display the current price difference (completely consistent with ShopMenu's processing logic)
Meanwhile, confirmation is required after shopping to take effect. At this point, you can exit the menu and cancel the selection. (It won't crash anymore!)
gozag_petition

tianmaster:
I will initiate a new PR for a cleaner submission record

PR #4808, opened 2025-09-25
Correct Lua Regex:matches() doc/name

brianfaires:
Lua API lists `Regex:find()` when it's actually `Regex:matches()`.  Updated the docstring to reflect this. Updated cpp function name to follow existing pattern.

Built and tested with `make -j4 TILES=y BUILD_PCRE=y` on MacOS tiles.

Tested in lua interpreter. Works as intended.
Screenshot 2025-09-24 at 10 09 55 PM

Screenshot 2025-09-24 at 10 10 57 PM

### DIFF
--- a/crawl-ref/source/l-crawl.cc
+++ b/crawl-ref/source/l-crawl.cc
@@ -785,9 +785,9 @@ static int crawl_regex(lua_State *ls)
  * passed.
  * @tparam string s
  * @treturn boolean|nil
- * @function Regex:find
+ * @function Regex:matches
  */
-static int crawl_regex_find(lua_State *ls)
+static int crawl_regex_matches(lua_State *ls)
 {
     text_pattern **pattern =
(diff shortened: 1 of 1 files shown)


Implojin:
~So the thing is, this is a player-facing (clua) function, and if we change its function signature, it seems likely to break a bunch of downstream user scripts. Historically we haven't worried too much about breaking the clua API, but find is pretty basic and seems likely to have lots of pre-existing use.

I'd be happy to take a patch that updates the documentation to clarify that it internally uses regex matches (despite the function name), but just flatly replacing this particular API call without concern for backwards compatibility seems quite likely to end with a bunch of players asking us why their rcfiles don't work anymore.~

edit: My bad, I see that the clua bind was already `matches` and you've left the player-facing portion of this as-is. This seems fine in that case.

Implojin:
Thanks for the patch. :)

PR #4809, opened 2025-09-25
feat: Add selection menus for Gozag’s potion petition and call merchant abilities

tianmaster:
Until now, both Gozag’s Potion Petition and Call Merchant abilities printed their choices directly to the message window and only accepted selection via hot-keys.
On tiles/web builds, long random shop names could be truncated (especially for Call Merchant) and mouse selection was impossible.
This patch introduces a proper shop-style menu for both abilities:

- Works like a one-time “pop-up shop”(Interface imitates ShopMenu)

- Supports both hot-keys and mouse clicks

- Guarantees all options remain visible regardless of font/screen size

Because the menu re-uses the existing icon, Call Merchant now also shows the shop-type icon next to each candidate, giving players an extra visual cue.
gozag_petition

### DIFF
--- a/crawl-ref/source/god-abil.cc
+++ b/crawl-ref/source/god-abil.cc
@@ -111,7 +111,8 @@
 #include "traps.h"
 #include "viewchar.h"
 #include "view.h"
-
+#include "menu.h"
+#include "shopping.h"
 static bool _player_sacrificed_arcana();
 
 // Load the sacrifice_def definition and the sac_data array.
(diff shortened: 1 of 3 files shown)


pooka109:
Applied; thanks.

What name do you want to be known by in the credits (since you don't seem to be in them)?

tianmaster:
> Applied; thanks.
> 
> What name do you want to be known by in the credits (since you don't seem to be in them)?

Just use the name tianmaster, thx

pooka109:
Done.

tianmaster:
sad， Already rolled back. The reason for rollback is stated on the official DC: 'You can cancel abilities after seeing the list, so you can prepare a good set of potentials before you encounter dangerous situations.',I currently have no good way to solve this problem from the perspective of UI framework, so I can only roll back

pooka109:
It's unfortunate. Hopefully there'll be a way to solve it at some point.

tianmaster:
> It's unfortunate. Hopefully there'll be a way to solve it at some point.

In the future, we may consider processing on the UI framework, adding only the UI without changing the original design intent

PR #4815, opened 2025-09-28
Fix fortress talisman description in small screens

robertxgray:
The fortress talisman description popup breaks with narrow screen size. Smartphones are most affected by this, but also any other device with the default configuration when the game window is shrunk.

I have shortened the Rust Breath Size column title and also reduced the columns spacing, but of course it's not the only valid solution.

BEFORE
before

AFTER
after

### DIFF
--- a/crawl-ref/source/ui.cc
+++ b/crawl-ref/source/ui.cc
@@ -1701,7 +1701,10 @@ void Popup::_allocate_region()
             m_region.x + m_region.width, m_region.y + m_region.height,
             VColour(0, 0, 0, 150));
     const int pad = base_margin() + m_padding;
-    region.width -= 2*pad;
+    if (tiles.is_using_small_layout())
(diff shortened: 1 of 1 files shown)


regret-index:
I would substantially prefer this sort of adjusting of all talisman text displays to be singularly part of the `tile_use_small_layout` for the mobile UI rather than have the default text interface for all others to be less cleanly distanced apart. People not using the mobile UI and shrinking their output to anything smaller than console already are making much bigger UI sacrifices (like cutting off default LOS radius at default zoom) to begin with.

robertxgray:
Thanks for the suggestion. I changed to a different solution by reducing the popup horizontal margins when the small layout is enabled.

Captura de pantalla -2025-10-05 01-00-53

robertxgray:
This is the way, but I'm going to include this patch in another group of tweaks that I want to add to the next Android release.

PR #4821, opened 2025-10-02
Show failure rates of abilities in chardump

code2828:
Shows failure rates of abilities in the chardump `a:` list, e.g. `a: Oozemancy (3%), Slimify (12%)`, similar to the passive effects list in format.

There were a couple of times that either me or my friends wanted to know the failure rates of some abilities. I'm not sure if its intuitive enough so that most people won't mistake it for success rate. Maybe showing a table like the spell library can be better?


Hellmonk:
gonna go ahead and merge this, will see what the feedback looks like.

PR #4822, opened 2025-10-02
fix: Prevent acquiring a few things with certain sacrifices

LiciTheCrawler:
Prevents Mask of the Dragon with Love sac, Death ego with Necro sac, Command ego with Armour, Summoning or Love sac and Resonance ego with Forgecraft sac.


pooka109:
Applied; thanks.

PR #4829, opened 2025-10-09
Fix Spanish floor description

person50605:
Oops, I didn't do my Lua comments correctly in #4792, so this fixes that. Also, adds a line break before the branch-specific part of the floor's description


pooka109:
Applied; thanks.

PR #4831, opened 2025-10-12
equip_slot_by_name(): Remove duplicated list and catch "Body Armour"

brianfaires:
Allows Lua API's item.equip_type to work with:
- l_item_equipped_at()
- l_slot_is_available()

Previously didn't work for body armour. equip_type = "Body Armour",
Both functions only check for "Armour"

s_equip_slot_names was only created to support this function afaict,
and just duplicates `enum equipment_slot`

Backward compatible: equip_type returns the same strings, but functions
will now check for both varieties of names.

Tested locally and everything works as before but catches body armour.

### DIFF
--- a/crawl-ref/source/output.cc
+++ b/crawl-ref/source/output.cc
@@ -1999,17 +1999,17 @@ int update_monster_pane()
 }
 #endif
 
-static const char *s_equip_slot_names[] =
-{
-    "Weapon", "Offhand", "Armour", "Helmet", "Gloves", "Boots",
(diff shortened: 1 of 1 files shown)


WizardIke:
Have you been added to the credits yet? If not what name should I put you under?

brianfaires:
> Have you been added to the credits yet? If not what name should I put you under?

oh yeah, nope not yet! Thanks @WizardIke  Could you add me as `Brian 'buehler' Faires`

PR #4832, opened 2025-10-12
feat: Adjust Body Armour acquirement

LiciTheCrawler:
Currently body armour acqurement is dominated by robes, ring mails,
troll leather, plate and crystal plate. Higher armour skill makes plate
and crystal plate more likely, but the chance to get one of those five
is about the same even at 27 skill.
Additionally scale mail and chain mail are very rare. This version
brings several new body armour egos that spawn on scale, chain or plate
so it makes sense to make those slightly more likely.
Okawaru gifts also tend to leave a bad feeling after being offered a
robe or troll leather on your heavy armour using char.

This commit changes the weights of ring, scale, chain, plate and cpa.
It also changes the formula for body armour acquirement so low base AC
body armours become significantly less likely at higher armour skill.
For example: Body size restricted species still had ~27% to acquire
robes or TLA at 27 Armour skill. With the new formula at 27 skill
that chance is ~2.5%. For normal size species robe + tla + ring
mail went from ~23% at 27 with the old formula to ~4% with the new
formula.

### DIFF
--- a/crawl-ref/source/acquire.cc
+++ b/crawl-ref/source/acquire.cc
@@ -219,7 +219,7 @@ static armour_type _acquirement_shield_type()
 /**
  * Determine the weight (likelihood) to acquire a specific type of body armour.
  *
- * Weighted by Armour skill, though not particularly strongly.
+ * Weighted by Armour skill.
  *
  * @param armour    The type of armour in question. (E.g. ARM_ROBE.)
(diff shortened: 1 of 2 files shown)


Implojin:
Thanks for the gameplay tweak :)

I might prefer to do something that accounts for Dodging skill here, eventually, but it seems reasonable to take this for now.

PR #4834, opened 2025-10-14
Fix the Spanish floor desc

person50605:
The function I was calling in crawl#4792 for the floor descriptions wasn't defined in the ../es/features.txt file, so this changes the function used to one already used in other /es files: you.in_branch("Branch").


pooka109:
Applied; thanks.

(It does appear to be using the proper function now, since the French translation and other Lua-defined levels that want to get the current player branch use the same one.)

PR #4841, opened 2025-10-20
Wizmode bane

unutbu:
Add the ability to toggle banes in wizmode. To use, type `&}` then the name of the bane.


pooka109:
Applied; thanks.

What name do you want to be known by in the credits?

PR #4847, opened 2025-10-25
Prevent hiding from your past

CrawlOdds:
(Edited this description to match the changes)

Fix friendly monsters not to target our pet_target if it is invalid.

This changes the behaviour in this case from:
- Set the foe to pet_target, then later clear it to MHITNOT, to
- Set the foe to the nearest monster, or the player if none exists

The motivating bug is one with Hep, where if we run from a fight and then our ancestor respawns, they can end up wandering. However, the main noticeable change to monster behaviour is in the case where we have allies and a monster on the screen, but have not begun hostilities or have only used certain AoE effects (and hence not set our pet target). Previously, the behaviour in this case depends on whether we have a previous pet_target - our allies would take the initiative and begin the fight if and only if our pet_target is empty.

This could also be quite bad if there is any case in which we fail to set pet_target. In that case, we could have an entire fight in which our allies just don't really do anything.

### DIFF
--- a/crawl-ref/source/mon-behv.cc
+++ b/crawl-ref/source/mon-behv.cc
@@ -59,29 +59,29 @@ static void _guess_invis_foe_pos(monster* mon)
         mon->target = dgn_random_point_from(mon->pos(), guess_radius);
 }
 
-static void _mon_check_foe_invalid(monster* mon)
+static bool _is_valid_foe(monster* mon, unsigned short foe)
 {
     // Assume a spectral weapon has a valid target
     // Ideally this is not outside special cased like this
     if (mons_is_avatar(mon->type))
-        return;
+        return true;
 
-    if (mon->foe != MHITNOT && mon->foe != MHITYOU)
-    {
-        if (actor *foe = mon->get_foe())
-        {
-            const monster* foe_mons = foe->as_monster();
-            if (foe_mons->alive() && monster_los_is_valid(mon, foe)
-                && (mon->has_ench(ENCH_FRENZIED)
-                    || mon->friendly() != foe_mons->friendly()
-                    || mon->neutral() != foe_mons->neutral()))
-            {
-                return;
-            }
-        }
+    if (foe == MHITNOT || foe == MHITYOU)
+        return true;
 
+    const monster* foe_mons = &env.mons[foe];
+    return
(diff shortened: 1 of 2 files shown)


DracoOmega:
I am not sure what might be causing the bug you speak of (regarding ancestors wandering off), but I'm fairly sure it isn't this. In fact, Infernal Servant demons are fairly demonstrably already summoned seeking the player if there's no enemies around. (Yes, technically at some point in the creation process they are BEH_WANDER, but this gets converted to BEH_SEEK already *before the game even prints the message about summoning them* (as a automatic consequence of them having a 'foe' in their LoS while wandering, via handle_behavior() which is called immediately after creation, and is completely unrelated to player stealth value.) 

So the alert events added there by this PR seem entirely superfluous, and thus likely don't affect whatever the root cause might be.

CrawlOdds:
Thanks - for clarity, I do sometimes get infernal servants wandering on summon. But I agree that the code intends to convert them to BEH_SEEK (I'd missed that the "foe" needn't be a foe), so this isn't the right fix. I'll investigate further.

DracoOmega:
I have never seen (or heard of, prior to now) such summons arriving wandering. I don't suppose you can you think of any special circumstances that might have been involved?

CrawlOdds:
Yeah - still confirming, but I _suspect_ it's that I have a pet_target set but not in view, and this is messing up the handle_behaviour. Something like that, anyway :)

DracoOmega:
pet_target is *all sorts of dubious* in a lot of way, so I guess that wouldn't be terribly surprising if it was involved. (The whole system has clearly needed ripping out and replacing for a while, but monster behavior code is one of the more tangled and subtle things that exists in the codebase and the current implementation *sorta* works, so it's easy to put it off ^^; )

CrawlOdds:
Yep - what happens is that we set the foe to pet_target, then notice that the monster now has an invalid foe, and set the foe to HITNOT. So I need to move that validity check to before we set the foe to pet_target.

CrawlOdds:
Sorry, the fixed version is not ready for review

CrawlOdds:
This is now updated with (I think) the correct fix and should be ready for review.

DracoOmega:
Merged (with some tweaks), thanks!

PR #4849, opened 2025-10-26
Separate hidden spells for divine exegesis

CrawlOdds:
Change the divine exegesis menu so that it maintains a separate list of hidden spells from memorisation.

The player often wants these lists to be very different, as exegesis uses spells the player will never be able to cast.


CrawlOdds:
Merged in 064b58b

PR #4850, opened 2025-10-26
Fix log on zero contamination

CrawlOdds:
The contamination level for small amounts of contamination is zero, so this was not triggering.


pooka109:
Applied; thanks.

PR #4855, opened 2025-10-27
Remove piety decay

CrawlOdds:
Removes piety decay from all non-Usk gods, and rebalances piety gain
with the intent of making the overall effect neutral.

Piety decay acts as a soft clock, which applies to most gods. While the
effect is relatively minor, this leads to some bad incentives - it's
generally not that fun to try to minimise turns. The upside is limited;
these days there aren't many mechanics that necessitate a soft clock,
and at any rate there are several gods who don't have piety decay.

To balance this change, I played a few test games through to
championing various gods, seeing how much piety was decayed vs
increased (counting only actual increases and hysteresis offsets, not
piety that is swallowed by gifts or by the high-piety stepdowns). The
results were reasonably consistent; ~20% piety loss for gods with high
piety decay, and ~10% for gods with low decay. We therefore decrease
piety gain by these amounts.

We then decrease gift timers correspondingly, which leads to the
same rate of gifts.

While I don't think we log enough information to get the numbers right
using online games, we can check post-hoc and rebalance if needed by
looking at the time between worship and champion milestones per god;
if this works as intended, these should not change dramatically for any
god except Sif.

### DIFF
--- a/crawl-ref/source/religion.cc
+++ b/crawl-ref/source/religion.cc
@@ -83,8 +83,6 @@
 #    define DEBUG_PIETY
 #endif
 
-#define PIETY_HYSTERESIS_LIMIT 1
-
 #define MIN_IGNIS_PIETY_KEY "min_ignis_piety"
 #define YRED_SEEN_ZOMBIE_KEY "yred_seen_zombie"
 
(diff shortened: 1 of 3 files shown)


CrawlOdds:
Alas, this has two issues (thanks to DracoOmega for pointing these out):

1. I misunderstood piety gain, and the changes to on-kill piety don't in fact decrease it as I thought. This is rather hard to fix because of the complexity of piety.
2. Because of piety stepdowns, it tends to slow early piety gain and accelerate late piety gain. This could probably be fixed by twiddling stepdowns

CrawlOdds:
While I still intend to remove piety decay, I think it will look substantially different to this, so closing.

PR #4856, opened 2025-10-27
Don't offer Ash curses while ostracised from full piety

CrawlOdds:


WizardIke:
Thanks, this makes sense to me and is similar to what we do for Ru.

CrawlOdds:
Thanks - I'll go by "Odds", please

PR #4857, opened 2025-10-27
Interrupt for monsters on their turn

CrawlOdds:
This fixes a bug whereby seen_monster interrupts were sometimes not triggering, because:
- On the first turn an already seen monster comes into view, if it is the monster's turn when we check for the interrupt, we don't interrupt in case it wanders off.
- On subsequent turns, the player has already seen it so the interrupt doesn't trigger.

We simply remove the first check. It's probably better to give the player a chance to interrupt for monsters who wander onto the screen even if they might wander off again.


CrawlOdds:
Fixed (along with many other interrupt issues) elsewhere

PR #4862, opened 2025-10-29
Update Vehumet's gift weighting to consider averages

CrawlOdds:
It seems wrong that we are more likely to offer multischool spells; e.g. offering Chain Lightning more than MCC to a mage with 20 air magic and 5 conjurations feels off. So we update the weighting to consider the average skill of (revelant) schools, rather than the sum.

Since Veh's spells average around 1.6 schools, we then increase the weight of the preference by a similar amount so that on average we should get about as many relevant gifts.

Also do a light cleanup and rename of destructive_elemental_preference, since we probably won't want the elemental-only version of this function again.


Implojin:
Thanks for the PR!

PR #4863, opened 2025-10-29
Fix pathfinding through doors

CrawlOdds:
Monsters that can open doors should pathfind through them.

This was causing a bug whereby a monster on the other side of a closed translucent door did not stop the player feeling safe (and perhaps other problems).

We move a "feature is solid" check only to be triggered when we don't have an explicit monster; in the case of an actual monster this is triggered anyway by mons_traversable (specifically, in the habitability check).

We also move it to after the door check, to fix this in the case where we are pathfinding for a default monster.

### DIFF
--- a/crawl-ref/source/mon-pathfind.cc
+++ b/crawl-ref/source/mon-pathfind.cc
@@ -444,9 +444,6 @@ bool monster_pathfind::traversable(const coord_def& p)
     if (traverse_no_actors && actor_at(p))
         return false;
 
-    if (feat_is_solid(feat) && (!mons || !mons->can_pass_through_feat(feat)))
-        return false;
-
     if (monster* mon_at = monster_at(p))
     {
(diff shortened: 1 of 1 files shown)


pooka109:
The problem appears to have been fixed in a different way by 6ed2157ae40adfe3c0db995c722062c6baaab4bf.

CrawlOdds:
That's right; that commit fixes the headline bug. I still think this is worth merging, for a couple of reasons:
- I think that commit leaves a smaller bug - in the no monster case, if traverse_doors is false we will still pathfind through doors.
- The feat_is_solid check that both commits touch is duplicating logic in mons_is_traversable, so I think it's better to only make it when we don't have an explicit monster.

But of course @DracoOmega is very likely to know better than me.

DracoOmega:
I agree that this is a minor additional improvement and have merged it. Thanks!

PR #4866, opened 2025-10-30
Allow igniting poisoned allies

### LINKED ISSUE #4858: Ignite Poison cannot find targets with a poisoned ally in LOS
While attempting to ignite poison while a bribed ally kobold was standing in toxic bog but not yet poisoned, I got a warning 'are you sure you want to do this'. Presumably this was in case of flame clouds, and a correct warning. 

After the bog expired, but while there was still a poisoned cyan ugly thing in LOS, attempting to cast Ignite Poison did not warn me, it simply failed to cast with a 'no susceptible monsters in range'

Image

Image

CrawlOdds:
The tracer codepath in cast_ignite_poison was being used in two different ways - to decide if the spell was useless, and to check if monsters should cast it. These wanted different criteria, and so the uselessness check was wrong. We split the check by adding a separate function to do the check for monsters.

Fixes #4858

### DIFF
--- a/crawl-ref/source/mon-cast.cc
+++ b/crawl-ref/source/mon-cast.cc
@@ -9604,8 +9604,7 @@ ai_action::goodness monster_spell_goodness(monster* mon, spell_type spell)
         }
 
     case SPELL_IGNITE_POISON:
-        return ai_action::good_or_bad(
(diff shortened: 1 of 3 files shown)


pooka109:
Applied; thanks.

PR #4867, opened 2025-10-31
Tweak putting on orb text

srcrip:
Hi there,

This is my first crawl PR so let me know if theres anything special I'm supposed to do.
this changes the text when you equip an orb from `"you begin putting on your orb"` to `"you begin holding your orb"`.

I would like to change the removing text too but I'm not sure what the best verbiage is. 'you begin removing your orb' also sounds weird but not as bad as 'you begin putting on your orb'.

### DIFF
--- a/crawl-ref/source/delay.cc
+++ b/crawl-ref/source/delay.cc
@@ -172,6 +172,8 @@ const char* EquipOnDelay::get_verb()
         return "haunting";
     else if (equip.base_type == OBJ_ARMOUR && you.form == transformation::fortress_crab)
         return "fusing with";
+    else if (equip.base_type == OBJ_ARMOUR && equip.sub_type == ARM_ORB)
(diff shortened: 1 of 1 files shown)


randomdevpete:
"You begin holding your orb" still sounds a bit awkward (although better than what it replaces obviously).

"You hold the orb aloft" or some other more natural phrasing could be even better.

randomdevpete:
(i guess it just sounds weird because obviously putting on a shield is something that does conceptually take time, but when you start picking up an orb you are already holding it. Maybe the first text should be "you start taking out your orb" followed by "you are now holding the orb aloft" or along those lines)

srcrip:
I think that would be a good flavor improvement... I'll see how difficult that would be cause right now this func just kinda inserts this word into a broader sentence

PR #4868, opened 2025-10-31
Add check for targets in plasma beam

### LINKED ISSUE #3337: Plasma beam can't target a spatial vortex
webtiles

0.31-a0-968-g24f9ace8b5

**DESCRIPTION:**
Plasma beam can't target spatial vortices. When cast with one on the screen, it is ignored. If it is the only monster on the screen, the mana is wasted.

(If this is somehow expected, then plasma beam should only be castable with Z in this situation)

CrawlOdds:
Fixes a bad behaviour where casting plasma beam with some monsters but no targettable monsters could waste a turn and MP.

Fixes #3337


pooka109:
Applied; thanks.

PR #4869, opened 2025-10-31
Improve ranged combat tabbing

CrawlOdds:
Update automatic shooting of ranged weapon - via tab or p - so that we pick shots with more chance of hitting monsters other than the primary target and less of hitting allies.

Specifically, we calculate the same primary target as before, but instead of shooting straight at it we search for the beam which (in order of priority):
- Hits the primary target
- Hits as few friendly monsters as possible (if there are any, tabbing isn't actually going to work)
- Hits as many enemies as possible
- Is the primary target (to retain predictability in the simple case)

The main effect is that when we miss the main target, or are firing a penetrating weapon, we are more likely to hit other targets.

### DIFF
--- a/crawl-ref/source/dat/clua/autofight.lua
+++ b/crawl-ref/source/dat/clua/autofight.lua
@@ -292,33 +292,6 @@ local function compare_monster_info(m1, m2)
   return false
 end
 
-local function is_candidate_for_attack(x,y, no_move)
-  m = monster.get_monster_at(x, y)
-  --if m then crawl.mpr("Checking: (" .. x .. "," .. y .. ") " .. m:name()) end
-  if not m then
-    return false
-  end
-  if m:name() == "butterfly"
-      or m:name() == "orb of destruction" then
-    return false
-  end
-  if m:is_firewood() then
-  --crawl.mpr("... is firewood.")
-    if string.find(m:name(), "ballistomycete") then
-      return true
-    end
-    return false
-  end
-  if m:is_damage_immune() and (no_move or not m:is("warding")) then
-    return false
-  end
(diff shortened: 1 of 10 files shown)


crawl-ref/source/dat/clua/autofight.lua
```diff
@@ -340,7 +340,62 @@ local function get_target(no_move)
   return bestx, besty, best_info
 end
 
+local function target_score(monster_x, monster_y, target_x, target_y)
+  -- TODO: this is a hack using an arbitrary beam spell. Replace with a
+  -- dedicated path function for "normal" projectiles.
+  path = spells.path("Plasma Beam", target_x, target_y, 0,0, true)
```
Implojin: As you've mentioned, we wouldn't want to merge this using Plasma Beam's targeter as a hack. Would suggest a custom beam if there's nothing more appropriate in the quiver code. (There might be?)

Implojin:
Regarding the basic premise: I'm broadly in favor of autofight targeting improvements where possible, but I do wonder whether this code is going have any difference compared to simply setting `aimed_at_spot = false` which iirc is *already* autofire's default behavior?

Implojin:
For reference, here's a screenshot of the existing behavior when autofiring boomerangs at the nearest target and missing it:
autofire-miss-miss-hit-target-nearest

Implojin:
Oh, I see, this would be maximizing autofire beam targets on noncardinal lines behind the primary target.

CrawlOdds:
> Oh, I see, this would be maximizing autofire beam targets on noncardinal lines behind the primary target.

That's right - I think this comes up very regularly in a normal launcher game

PR #4871, opened 2025-11-01
Fix phantom tentacles

CrawlOdds:
These were broken in the case of cloning an eldritch tentacle which had moved, because the base position was being cloned.


pooka109:
Applied; thanks.

Issue #4873, opened 2025-11-01
Calculated time to train to talisman level does not take into account amulet of wildshape


srcrip:
<img width="1174" height="1094" alt="Image" src="https://github.com/user-attachments/assets/5dd7daff-432a-404e-ad65-60bb0ead7c27" />

see here, I have an amulet of wildshape on and taking it off or putting it on does not update the calculation of XL levels to get to 23 Shapeshifting at 100% training. it appears to me like it always does *not* take into account the amulet.

srcrip:
I'm going to try and fix this one myself

PR #4874, opened 2025-11-02
Fix wildshape skill estimation for talismans

srcrip:
Resolves: https://github.com/crawl/crawl/issues/4873

The calculation for estimated XLs to train up to talisman min Shapeshifting requirements took into account other sources of XP boosting but not the new wildshape amulet. This resolves that.


pooka109:
Applied; thanks.

PR #4880, opened 2025-11-06
art: Make exclusions more readable

RypoFalem:
As someone that is partially colorblind, exclusions were very difficult for me to read when placed over certain tiles. Shoals was the biggest culprit since the exclusions are hard to make out on top of sand and shallow/deep sea water tiles. To address this, exclusions were given a transparent black outline.

Exclusions were also very difficult to see when placed over lava tiles. The black outline doesn't help so much here unfortunately. I've turned up the opacity slightly. It's still difficult to read but there is some improvement.

Exclusion tiles connected with each other in a way that made it difficult to see where the center of a tile was, since the corners had the same crisscross pattern as the centers. I've seperated the corners so they are no longer connected. It's possible this might be more of an issue with the background tiles not having clear edges in some cases but this at least makes excluded zones easier to quickly evaluate.

Shoals Before
Shoals After

Lair Before
Lair After

Volcano Before
Volcano After


Implojin:
Merged with tweaks to the tiles in b9efe5695e. Thanks for the PR!

PR #4887, opened 2025-11-08
fix: Hepliaklqana's ancestor names checking the wrong database

RypoFalem:
They were generated every time. :c


pooka109:
Applied; thanks. Good catch!

PR #4891, opened 2025-11-08
text: consider corrosion in some descriptions (#4885 and more)

### LINKED ISSUE #4885: Description of quivered action doesn't take corrosion into account
local tiles & console

0.33.0 official release, 0.34 trunk (as of a6fc723709)

**DESCRIPTION:**
The `Fire:` line is unaffected by corrosion. Fsim confirms the corrosion is applied in combat, though.

Image

**EDIT:** and here

Image

**EDIT*2:** and here

Image

I'm not sure if the latter two are intended or not, but the first case does look weird.

code2828:
The idea is: if the description cares about the weapon you are wielding *now* and your current status, take corrosion into account (e.g. damage rating, hit chance); if the description only cares about your *possession* of the weapon, don't (e.g. inventory, wielding/unwielding message). Sorry for my unfamiliarity to the crawl codebase if I happen to miss out some occurrences.

A brief summary:
* describe.cc: consider corrosion in damage rating (rating may be less than 0 if heavily corroded, but since damage can never be negative, we want a non-negative value here) and when describing hit chance (although it doesn't affect the number)

* item-def.h: adjust declaration accordingly

* item-name.cc: add parameters regarding corrosion

* melee-attack.cc: consider corrosion when notifying which weapon did what on Coglins

* output.cc: remove unused function

* player.cc: consider corrosion in damage rating

* quiver.cc: make corroded quiver red; consider corrosion in quivered weapon (#4885)

* tags.cc: irrelevant; noticed these when randomly browsing the code

* wiz-fsim.cc: consider corrosion when describing the weapon used in fsim; use ternary operators

Closes #4885 and #5181.

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -1454,7 +1454,8 @@ string damage_rating(const item_def *item, int *rating_value)
 
     const int slaying = slaying_bonus(thrown, false);
     const int ench = item && item->is_identified() ? item->plus : 0;
-    const int plusses = slaying + ench;
+    const int corr = you.corrosion_amount();
+    const int plusses = slaying + ench - corr;
 
     const int DAM_RATE_SCALE = 100;
     int rating = (base_dam + extra_base_dam) * DAM_RATE_SCALE;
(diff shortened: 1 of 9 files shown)


code2828:
Checkwhite.

CrawlOdds:
Closing - in the end we went the opposite direction here in b51359a, and made corrosion never show up as part of a weapon

PR #4892, opened 2025-11-08
Create lua func you.movement_cost()

brianfaires:
Super simple. No params used, to keep the intent obvious. Parameterizing is a rabbit hole. Even base speed is based on current form.

Tested local tiles, MacOS.  Shallow water, !haste, merfolk/barachi, etc. Works as intended 👍


CrawlOdds:
Merged with some tweaks to make it deterministic in 15110aa. Thanks!

PR #4897, opened 2025-11-11
Add revenant's first spell to their library

adelrune:
This fixes an issue where revenants could have messages like "You add the spell X to your library" when picking up parchment and books for a spell that they already had learned. This also affected the book inspection ui where the spell was both greyed out but not considered known.


pooka109:
Applied; thanks.

PR #4901, opened 2025-11-13
Add testing vault for redefined monster names.

pooka109:
This will more easily allow testing of whether names of redefined monsters are set correctly after they're zombified. It covers a lot of possible options (especially "nocorpse"), many of which currently produce incorrect results.

[Edited to consistently set "always_corpse" on redefined monsters, so the corpse names can be tested as well.]


pooka109:
I've made one last adjustment: it's about testing redefined monster names (of which testing zombified monster named is only a subset), so I've renamed the vault and updated its associated comment; the contents of the actual vault are unchanged from before.

PR #4902, opened 2025-11-13
fix: Faulty feedback when paralyzing sleeping enemies  (qwarkbark)

RypoFalem:
You would get "Nothing appears to happen" which sounds like a failure message. Code comments indicate that "monster stops moving!" was the intended paralysis message even for sleeping targets. That intention was broken years later when mons_is_immotile checked if the monster was sleeping or petrified.

https://tavern.dcss.io/t/paralysis-wand-and-status-check-minor-bug/1468

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -1962,11 +1962,9 @@ void bolt::apply_bolt_paralysis(monster* mons)
 {
     if (mons->paralysed() || mons->stasis())
         return;
-    // asleep monsters can still be paralysed (and will be always woken by
-    // trying to resist); the message might seem wrong but paralysis is
-    // always visible.
+
     if (!mons->is_firewood()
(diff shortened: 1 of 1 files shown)


RypoFalem:
Arguably it should say something else if the monster was asleep or petrified and is now paralyzed but I'm don't know what that would be.

Implojin:
Could we change this messaging to something like `simple_monster_message(*mons, " becomes paralysed!"))`, and get rid of the comment?

I think the historical issue here was that crawl didn't internally distinguish paralyse as a separate status from other tier 1 stabs until surprisingly recently (dc6bdc143), so a lot of these bits of oldcrawl paralyse messaging were intentionally vague about what exactly was happening.

Implojin:
Merged with some edits in b3fd7c6d4. Thanks for the fix!

PR #4907, opened 2025-11-14
Make demon weapon randart generation more interesting

yrdzrfxndfvh:
Currently, demon weapon randarts have a 7/9 chance to generate with a brand chosen from an unweighted list of mostly non-evil brands. Only 3 out of the 7 brands are actually evil-aligned, and there is about a 30%* chance for any given demon weapon randart to be flaming or freezing. This commit changes demon weapon randart generation to only apply the 'special' brand override 1/3 of the time, but guarantees that the chosen brand will be of an evil nature (or antimagic). This should make demon weapon randart generation a little more thematic and varied.


pooka109:
Applied (with some number adjustments); thanks.

PR #4912, opened 2025-11-15
Make teleport other prefer not to cancel teleports

### LINKED ISSUE #3124: Teleport other should prefer to target monsters without the teleport status
webtiles

0.30

**DESCRIPTION:**
With two monsters on screen, one in the process of teleporting, Teleport Other should prefer to target the non-teleporting one. You usually don't want to cancel the teleport.

CrawlOdds:
Fixes #3124


pooka109:
Applied; thanks.

PR #4916, opened 2025-11-18
Fix IMB uselessness check

CrawlOdds:
Fixes the case where we want to cast IMB with a friendly monster between ourself and a hostile one; this should be a warning not a useless cast.


pooka109:
Applied; thanks.

PR #4917, opened 2025-11-18
Make spell name lookup prefer spells that exist

CrawlOdds:
This makes it easier to memorise Freezing Cloud in wizmode, for example.


CrawlOdds:
> This change looks good to we. Just a few minor formatting nitpicks

Thanks, fixed all of these (sorry for the volume of those, will be more careful!)

PR #4920, opened 2025-11-20
Fix a tiny exclusion related info leak

CrawlOdds:
It was possible to detect the positions of unknown, stationary monsters behind transclucent walls on levels where you had read revelation.


CrawlOdds:
Example before this:

image

PR #4921, opened 2025-11-20
Stop stationary monsters pathfinding

CrawlOdds:
This is causing a bug where you can't rest behind a translucent wall from a stationary monster.


DracoOmega:
Merged, thanks!

PR #4924, opened 2025-11-23
Warn for riding crocodiles into traps

CrawlOdds:
The crocodiles were sometimes *too* surprising.


WizardIke:
This looks good to me. I don't seem to be able to get it to leak any information to the player, so I've gone ahead and merged it

Issue #4926, opened 2025-11-25
Crash on meeting orb of fire


powerbf:
**PLATFORM and OS or BROWSER:**
Webtiles (CPO)

**DCSS VERSION AND GAME INFORMATION:**
Trunk

**DESCRIPTION:**
Walking around Zot:1. Orb of Fire comes into view. Crash.

**EXTRA FILES:**
https://crawl.project357.org/morgue/MrDizzy/crash-MrDizzy-20251125-110136.txt

powerbf:
Have raised a PR for the fix. It was just a null pointer dereference. Easy fix.

pooka109:
This should be fixed as of cc7c8d0d9bc2f557f98c24aca2b87fac6e403b1f.

PR #4927, opened 2025-11-25
Fix crash (null pointer dereference) in _ench_animation (issue #4926)

powerbf:
Fix for crash reported in issue #4926 
Check mon is not null before calling mon->pos()


wheals:
Thank you!

PR #4935, opened 2025-12-01
Don't chaos lace or fire champion derived undead.

### LINKED ISSUE #4279: harlequin traps do not affect derived undead
local tiles

0.33-a0-802-g62b7471ca3

**DESCRIPTION:**
derived undead monsters can gain the "interlaced with chaos" status, but will not actually have chaotic attacks. this includes zombies and skeletons which do not have any special flavor melee attacks. i also tested this with the ignis fire champion status and it also does not affect derived undead

Hellmonk:
These statuses were not functional, so don't bother trying. 
Fixes #4279.


regret-index:
Managed in a different way in 3440744.

PR #4939, opened 2025-12-01
Rework death ego.

Hellmonk:
Make it work like command, which was a more successful design.


Hellmonk:
superseded by other commits

PR #4941, opened 2025-12-01
Preview attack delay in equip (SentientSupper)

Hellmonk:
Updated for the past ~2 years of changes and fixed some bugs. It may be better to expand the existing armour skill delay message to unworn equipment instead.

Closes #3715.


CrawlOdds:
I missed this one and did something very similar in 2f29fc7 - sorry!

PR #4948, opened 2025-12-04
Updated options_txt with talismans for autopickup

SenorGato:
I'm finally sitting down to write my own RC file, and I had to verify and validate that talismans are in fact %, and it works in the RC file.

I'm going through the entire options_guide, so there may be more PRs but I figured it was easiest to just send the simple one liner first to establish the process.


pooka109:
Applied; thanks.

PR #4949, opened 2025-12-04
Sludge Elves

Hellmonk:
Sludge elves were removed in version 0.13 for largely being worse than other species in all of their supposed niches. This is an attempt to return them in a more distinct role.

Sludge elves are slightly good at unarmed combat and shapeshifting as before, with mostly flat aptitudes elsewhere. However, they begin the game with the protean grace mutation and a "natural shifter" mutation that provides AC and intelligence while in any form. Additionally, they get one random good mutation at level 4 and receive an "innate" high-skill form at level 10. They can freely transform into their innate form without a talisman and receive a skill bonus with that form (whether from a talisman or not).

My hope is that this species provides a different take on shapeshifting than species like octopode and draconian, while remaining somewhat flexible overall.

(TODO before merging: refactor some of this messy code, extensive playtesting, and tiles.)

((Also, if we don't want this species to be a throwback to ancient crawl, it's reasonable to retheme them to 
"proteans" or something, I think.))

### DIFF
--- a/crawl-ref/source/player.cc
+++ b/crawl-ref/source/player.cc
@@ -6546,7 +6546,7 @@ int player::armour_class_scaled(int scale) const
         AC += you.props[PHALANX_BARRIER_POWER_KEY].get_int();
 
     if (has_mutation(MUT_NATURAL_SHIFTER) && you.form != transformation::none)
-        AC += 200 + 200 * get_mutation_level(MUT_NATURAL_SHIFTER);
+        AC += 100 + 200 * get_mutation_level(MUT_NATURAL_SHIFTER);
 
     AC -= 100 * corrosion_amount();
 


pooka109:
Interesting.

Although the flavor that they're not true elves, but they're called such because of the grace they conduct themselves with doesn't fit with their getting the "Elf" genus (and calling them Sludge Elves but _not_ giving them the Elf genus is a thematic problem; for example, if they're not really elves, should "Elf" even be used in their titles?). I'd be in favor of changing the name to something else, and Protean does seem to fit .

Or, if not, there are a few lore references to sludge elves in the database that need to be adjusted to account for the differences between the old sludge elves and this new version of them.

regret-index:
To summarize feedback given by DracoOmega on the official dev discord around ten or so days ago:
* The single random good mutation at XL 4 is still capable of rolling a lot of duds compared to the eventual full levels in multiple different ways demonspawn get, and is readily lost in drinking potions of mutation as Protean Grace promotes.
    * Protean Grace itself is likely to need more reworking than its current state to better deal with the gap between having mutation gods or not, but this is troubled regardless of the state of that. 
* Despite the obvious thematic synergies, the number of forms that suppress mutations makes any incentive to gain many mutations mix poorly with picking up forms.
* The odds of a single high-than-protean-tier form being relevant to a given non-unarmed character by the time the one form is gained are somewhat low, which feels a bit sad for a core species gimmick.
    * The implementation in this PR also doesn't allow previewing anything about the form stats versus one's current skill and the innate bonus Sludge Elves get for their innate form, though this is obviously fixable.
* The innate bonus provided for being in a form feels somewhat unlikely to matter without starting as a shapeshifter background, since talismans aren't reliably provided by many gods or locations until several branches past Lair at best.

As such, this particular draft of Sludge Elves is probably not likely to work out. A new PR should be opened up for other approaches towards these pressing issues.

Species concepts floated by multiple devs as better directions included:
* Something closer to a more literal "djinn of shapeshifting", where one gets multiple innate forms, an innate advantage to mixing them up (like a cooldown to instant swap forms as one needs), and an inability to use talismans. This is likely to need fairly further form variety and build support (like more magic-supporting talismans or shared randart talisman properties) and more distinction between runs (like a longterm plan of talisman randarts having per-form properties, so Shapeshifting varies more between games), before it can really be pushed forward as a species.
* A species that can allow racking up lots of mutations at once without Xom or Jiyva support, by reducing the effect of the default innate mutation caps and being able to work off bad mutations (from potions) with e.g. gaining experience. This is likely to want yet further reviews of the good mutation pool in general to be less bland than the general pool of randart properties, even beyond what 0.34 did, but it does present its own unique points of interest.

(There are also quite a few other plans sketched out in the future that could also help add more variety to shapeshifting, like chaotic talismans in the Xom bazaar, retooling the Lugonu distortion branding into talisman randarts with unique Abyssal properties and innate distortion brand, backgrounds beyond Shapeshifter with new early forms that can nudge one towards other builds like magic and stealth, and more portals that can provide talismans as item rewards. Nearly all of these are large projects and longterm prospects, though.)

PR #4978, opened 2025-12-17
Prevent concussion player movement when in tree form

poifra:
I noticed during gameplay that I could move around in tree form while having a concussion weapon. I don't think that's intended, is it? This is my first PR to dcss, let me know if there's anything I should address.


pooka109:
Good attempt, but a problem like this should be fixed more generically in the actor class, since there could be monsters that can't move but can wield concussion weapons.

In any case,  the problem has already been fixed as of 6c3c90b89a6c28a8efbb886e587b120548e8c965.

PR #4982, opened 2025-12-20
Add messages for firing without anything quivered

CrawlOdds:
Fixes some code which intended to give a message when autofiring without a quivered action, and adds the same message for CMD_FIRE.

Also tidy up quiver_description code to commonise a validity check, and improve the consistency of how l-you.cc quiver functions work.


WizardIke:
Thanks for the PR

PR #4983, opened 2025-12-20
Fix preview of talisman attributes between integer skill levels.

CrawlOdds:
Between integer values of shapeshifting, these were wrong - e.g. showing -10% HP until hitting minimum shapeshifting, when in fact things are much more continuous than that.


DracoOmega:
Merged, thanks

PR #4984, opened 2025-12-21
Fix monster javelin friendly fire

CrawlOdds:
This fixes a couple of issues with monsters calculating the likely effect of throwing javelins:
- They didn't realise these were penetrating
- When the player is invisible to them, penetrating tracers counted the player multiple times


DracoOmega:
Merged (with some additions/tweaks), thanks

PR #4986, opened 2025-12-22
fix: missing newline in spell descriptions

RypoFalem:
When viewing spell descriptions from the main menu, there was a missing newline when the description ended early. If the spell had a quote, a string of underscores would be appended to the spell school instead of on a new line.

image


DracoOmega:
Merged, thanks.

PR #4989, opened 2025-12-22
Display non-integer values for EV and AC in forms

CrawlOdds:
Display fractional EV and AC gains from forms; these already provide randomly rounded bonuses, but we were displaying them rounded down.

This tweaks how EV scaling in forms works so that it is randomly rounded in the main EV calculation rather than in the form itself.

Also, allow variable numbers of decimal places in such displays - most want 1, but regeneration wants 2.


DracoOmega:
Merged, thanks!

PR #4990, opened 2025-12-22
Make Brom's Barrelling Boulder autotarget

CrawlOdds:
By changing its range. This should have no other effect except changing the displayed range.

The true range of BBB is a matter for philosophers; displaying  LoS-radius is consistent with OOD, which is another sphere-summoning spell with projectiles that can go to LoS and beyond.

### DIFF
--- a/crawl-ref/source/spl-data.h
+++ b/crawl-ref/source/spl-data.h
@@ -3770,7 +3770,7 @@ static const struct spell_desc spelldata[] =
     spflag::target | spflag::not_self,
     4,
     100,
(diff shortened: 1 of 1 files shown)


CrawlOdds:
<s>This isn't quite right; ally warnings depend on whether you fire before or after them, which is awkward.</s>

EDIT: while I could have sworn I saw an ally warning after this change, I now think there aren't any before or after it, and that's probably fine.

DracoOmega:
Seems reasonable to me. Merged, thanks.

Issue #4968, opened 2025-12-12
Yes/No prompt with buttons not passable with mouse


MIC132:
**PLATFORM and OS or BROWSER:**
local tiles on Windows 11

**DCSS VERSION AND GAME INFORMATION:**
0.34-a0-1763-g76bf294775 (built under msys2)

**DESCRIPTION:**
The Yes/No prompt shown in the attached image can't be passed with the mouse, despite clearly having clickable buttons.
Clicking the buttons gives the same red warning (shown in the image) as trying to use "y" or "n" instead of "Y" or "N".
If this is intentional then I think it should just have some text about the options instead of hoverable/clickable buttons.

**EXTRA FILES:**
<img width="741" height="292" alt="Image" src="https://github.com/user-attachments/assets/6f3fde8b-3239-40bb-a135-21bcf210d9dc" />

code2828:
I've encountered (probably) the same issue in #4590, and it looks like #4090 & #3074 a lot.

MIC132:
> I've encountered (probably) the same issue in [#4590](https://github.com/crawl/crawl/issues/4590), and it looks like [#4090](https://github.com/crawl/crawl/issues/4090) & [#3074](https://github.com/crawl/crawl/issues/3074) a lot.

Yeah, seems like the same problem. My bad, should have looked harder.
It's really not reassuring that one of those is from 2023. At least it's not really a major issue.

tianmaster:
I fix that. [https://github.com/crawl/crawl/pull/4994](url)

PR #4994, opened 2025-12-23
Fix: prompt handling of mouse clicks on yes/no popups issue #4968

tianmaster:
A bug was causing mouse clicks on yes/no popups to be ignored on Windows. Root cause: negative mouse keycodes (for example CK_MOUSE_B1 == -9997) were sometimes passed to C library character functions (such as isalpha/toupper). Those functions are defined only for EOF or values representable as unsigned char; passing negative non-EOF values invokes undefined behaviour and corrupted the prompt logic. The fix replaces use of the C library character test with the project’s ASCII-only helper [isaalpha] when checking popup input. Using [isaalpha] avoids invoking undefined behaviour on negative keycodes and ensures mouse clicks are correctly accepted. This change only replaces the character test; no other logic was changed. Verified locally on Windows 10.

### DIFF
--- a/crawl-ref/source/prompt.cc
+++ b/crawl-ref/source/prompt.cc
@@ -238,7 +238,7 @@ int yesno(const char *str, bool allow_lowercase, int default_answer, bool clear_
                 if (actual_key != EOF && actual_key > UCHAR_MAX)
                     actual_key = tmp;
 
-                if (isalpha(actual_key) && actual_key != tmp)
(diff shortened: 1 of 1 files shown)


tianmaster:
Related issues #4968

DracoOmega:
While I wasn't able to reproduce the original issue myself, the logic here seems sound, so I am optimistically merging it nonetheless. Thanks!

PR #4995, opened 2025-12-23
Fix display for sphinx form airstrike damage

CrawlOdds:
This was missed in 15d368e.


CrawlOdds:
a) This doesn't work - this method doesn't work when not in the right form.
b) Walking scroll skill boosts have the same issue.

EDIT: should now be fixed.

DracoOmega:
Merged, thanks!

PR #4997, opened 2025-12-23
Fix digging targetters leaking diggability

### LINKED ISSUE #4509: Hellfire Mortar spoils whether an unseen wall is diggable or not
When casting Hellfire Mortar into a revealed but unseen wall (scroll of revelation, Ashenzari powers, etc), the targeter spoils whether the unseen walls are diggable or not.

0.33-a0-1486-g8df79cbc47 (webtiles)

**DESCRIPTION:**
I am a Coglin Forgewright worshipping Ashenzari. I am at the Spider:4 rune vault, which contains undiggable walls. Ash reveals the layout of the vault, but does not show you the type of walls. However, the Hellfire Mortar targeter can spoil whether the wall is diggable or not.

Trying to cast into unseen undiggable wall:
Image

Trying to cast into unseen diggable wall:
Image

As a reference, here's what wand of digging shows:
Image

Wand of Digging does not spoil any information - it shows all tiles as diggable until proven otherwise.

CrawlOdds:
This affected walls which had been revealed by revelation.

The change to can_affect_wall is a no-op for other callers, who all have p = pos() if they use map_knowledge.

Fixes #4509, although that issue claims it doesn't affect digging.

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -2974,7 +2974,7 @@ bool bolt::can_affect_wall(const coord_def& p, bool map_knowledge) const
 
     // digging might affect unseen squares, as far as the player knows
     if (map_knowledge && flavour == BEAM_DIGGING &&
-                                        !env.map_knowledge(pos()).seen())
+                                        !env.map_knowledge(p).seen())
     {
         return true;
     }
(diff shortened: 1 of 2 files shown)


CrawlOdds:
Likely superceded by #5004 (though if we decide not to do that, this still makes sense)

DracoOmega:
As said above, superseded by #5004 (which is now merged).

PR #5000, opened 2025-12-27
Disambiguation submenus for the drop and inventory menus.

unutbu:
When `show_paged_inventory = false`, and an overloaded hotkey is pressed, the submenu asks the player to choose among the possible items.  Control returns to the original menu when the player exits the submenu.

Currently, the drop menu is always paged. This PR changes the drop menu to a single non-paged format when `show_paged_inventory = false`.

When `show_paged_inventory = true`, the current behavior is preserved.


unutbu:
One thing that concerns me about this user interface is that the submenu only pops up when the hotkey is overloaded. It doesn't pop up if there was no ambiguity. The inconsistency might surprise players. The alternative -- making the submenu always pop up -- could be annoying too.

PR #5002, opened 2025-12-28
feat: Grand Finale can destroy a wall (hellmonk)

RypoFalem:
This fixes a crash when using Grand Finale on wall-dwelling monsters. Instead of crashing or aborting the teleport, attempt to destroy the wall when targeting a monster inside a wall. Don't teleport if the wall couldn't be destroyed for whatever reason.

Technically this adds Uskayaw to a short list of gods capable of destroying walls. (The others being Fedhas, Lugonu, and Sif Muna with Shatter.) Hopefully this won't encourage luring.


DracoOmega:
I have merged the other fix to this issue instead.

PR #5003, opened 2025-12-28
fix: don't outline some tiles (acrobat)

RypoFalem:
Felid and Octopode dragon transformation tiles used a shadow with transparent pixels which made the generated outline a noticeably strange.

The issue revealed a couple of stray almost-transparent pixels so I made those completely transparent.

For reference here's the oddities
Screenshot from 2025-12-28 02-09-36-export
Screenshot from 2025-12-28 02-09-24-export


nlavsky:
Merged, thanks for the fix!

PR #5004, opened 2025-12-28
Make revelation reveal most squares exactly.

CrawlOdds:
Make revelation reveal most squares exactly.

This removes aliasing from features such as floors, walls, water, statues, and fountains, which previously showed as base types when mapped.

The aliasing wasn't really doing much - vaults are already often quite identifiable from shapes. Magic mapped levels now look prettier, and we don't tell small lies to players.

Altars and portals are still mapped to UNKNOWN_ALTAR and UNKNOWN_PORTAL, because running after unknown altars and portals is fun.

Small consequences:
- Exclusions placed in unseen terrain now respect the opacity of that terrain, since we know it from mapping.
- Digging targeters now know whether mapped squares are diggable, fixing #4509.
- We note any notable features when we map them, if the aliased feature is notable. This includes orb statues revealing the orb enemy.

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -2974,7 +2974,7 @@ bool bolt::can_affect_wall(const coord_def& p, bool map_knowledge) const
 
     // digging might affect unseen squares, as far as the player knows
     if (map_knowledge && flavour == BEAM_DIGGING &&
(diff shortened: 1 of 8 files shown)


WizardIke:
Merged as commit 19414c7

PR #5005, opened 2025-12-28
Do not lose LoS while trying to step towards target

CrawlOdds:
Monsters move towards targets by:
a) selecting a direction that moves them towards the target, irrespective of walls etc.
b) selecting a direction near that one which is closer to the target than they are now

This has a failure mode where they step out of LoS, typically in the configuration below (w=wall, m=monster, t=target, b=blocker, a=air):

```
awt
mba
```

This happens with friendly and hostile monsters alike; it is quite annoying with Hep ancestors, who do this when the player is the blocker and they can fire through them. But it is also a rather silly thing for an orc priest to do when blocked by a melee enemy.

We update step b) of this algorithm so that it will not select directions which lose LoS on the target square.


CrawlOdds:
image
image

Typical cases where this changes behaviour for enemies and allies.

PR #5007, opened 2026-01-02
Simplify piety and remove decay

CrawlOdds:
This PR makes multiple related changes to piety. We do them all at once to minimise the testing and monitoring needed.

1. Simplification of monster-related piety formulae

Previously, how piety scaled with level and how much you got were interrelated. This made it surprisingly hard to do operations like "increase piety by x%", and different conducts scaled piety differently by monster and player level (with good gods in particular ignoring player level for no clear reason, and monster holiness mattering slightly).

We simplify this, so that monster-related conducts have a fixed scaling component `f(monster_level, player_level)`, matching the most common one before. Different conducts can now have different magnitudes of piety by providing an explicit scaling factor, so the overall piety is `g(conduct) * f(monster_level, player_level)`.

2. Remove piety decay and hysteresis

Piety decay acts as a soft clock, which applies to most gods. While the effect is relatively minor, this leads to some bad incentives - it's generally not that fun to try to minimise turns. The upside is limited; these days there aren't many mechanics that necessitate a soft clock, and at any rate there are several gods who don't have piety decay.

Piety hysteresis existed to stop thrashing around level boundaries due to piety decay, so we remove that too.

3. Fix a bug with TSO

TSO was previously giving piety for seeing undead monsters, but given code comments this was a clear bug. It also seems undesirable to have a god who is _so_ skewed towards piety gain in extended. We fix this. TSO remains the fastest piety gainer in evil-dominated settings.

4. Balance the above

None of the above changes are piety-neutral. Using logs from real games, we balance these changes as best we can. This ends up meaning:
- Decrease piety gain rates by between 5 and 25% to account for piety decay
- Raise good gods' kill piety 20%, and other gods' minutely, to account for level scaling changes
- Increase stepdown probabilities from 33% to 40%, to account for the fact that piety decay affected stepped-down piety more.

The balance is not perfect; in particular different gods and levels want different stepdown probabilities (even assuming we've done everything else perfectly, which we obviously haven't). But everything should be _fairly_ balanced - within perhaps 10% in the worst case.

For details of how we got to these numbers, see https://docs.google.com/document/d/1G7-EvOMX8yOKw2EuyDPSm6fUAaNl8dUXhWRwGc5Pp3c/edit?tab=t.0#heading=h.48gpjrrr7g5v.

Testing: we tested this by playing games with a bunch of gods of representative piety styles (Oka, Zin, Hep, Fed, TSO) to around championing, making sure they felt OK and roughly matched the typical times for piety milestones. This isn't a way to check fine balance - the natural variance in this from piety RNG and spawn RNG is much higher than 10% - but it would have flagged up very large errors.

### DIFF
--- a/crawl-ref/source/god-conduct.cc
+++ b/crawl-ref/source/god-conduct.cc
@@ -466,19 +466,10 @@ struct like_response
     const char* desc;
     /// Whether the god should be described as "especially" liking it.
     bool really_like;
-    /** Gain in piety for triggering this conduct; added to calculated denom.
-     *
-     * This number is usually negative. In that case, the maximum piety gain
-     * is one point, and the chance of *not* getting that point is:
-     *    -piety_bonus/(piety_denom_bonus + level - you.xl/xl_denom)
-     * (omitting the you.xl term if xl_denom is zero)
-     */
-    int piety_bonus;
-    /// Divider for piety gained by this conduct; added to 'level'.
-    int piety_denom_bonus;
-    /// Degree to which your XL modifies piety gained. If zero, do not
-    /// modify piety by XL; otherwise divide player XL by this value.
-    int xl_denom;
+    /// Numerator for gain in piety (before level-based scaling).
+    int piety_numerator;
+    /// Denominator for gain in piety.
+    int piety_denominator;
     /// Something your god says when you trigger this conduct. May be nullptr.
     const char *message;
     /// Special-case code for weird likes. May modify piety bonus/denom, or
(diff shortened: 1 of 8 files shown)


Cgettys:
Does Uskayaw's piety still decay normally?

CrawlOdds:
> Does Uskayaw's piety still decay normally?

Yes

CrawlOdds:
Merged (with many additions, fixes and improvements) by DracoOmega

PR #5013, opened 2026-01-06
fix: Usk grand finale on wall creatures crash

RypoFalem:
Don't let the player target creatures inside of walls!

I still really like hellmonk's idea I implemented in #5002 but if that's not desired this PR just forbids the edge(heh)-case.


DracoOmega:
Merged (with a wording tweak). Thanks!

PR #5015, opened 2026-01-07
add stop/pickup grouping name for talismans

robbiemu:
- Teach autopickup_exceptions to recognize talismans via the new talisman/talismans prefixes emitted by _autopickup_item_name. Regexes like `>talismans` now match every talisman subtype (including rand/unrands) without listing them individually.
  - Extend explore_stop with an `ES_TALISMAN` flag, wire it through initfile.cc, and update `explore_discoveries::found_item` so both greedy and non-greedy auto-explore can stop for talismans independently of the items bit.
  - Document the new keyword in docs/options_guide.txt and mention the change in the user-facing docs.

### sample usage

From my .crawl/init.txt (summarized so reviewers see an intended workflow):

```
  ae := autopickup_exceptions

  # Default: ignore talismans
  ae += >talismans
  explore_stop -= talismans

  # Turn talismans back on for species/backgrounds that care about shapeshifting
  : if table_has({"Felid","Octopode","Troll","Oni","Poltergeist","Revenant",
  :               "Demonspawn","Demigod","Naga"}, you.race())
  :    or you.class() == "Shapeshifter"
  :    or (table_has({"Gnoll","Merfolk","Tengu","Barachi","Deep Elf","Formicid"},
  :                 you.race()) and not table_has(pure_melee, you.class()))
  : then
      ae -= >talismans
      explore_stop += talismans
  : end
```

This snippet defaults to ignoring talismans, then selectively reenables autopickup/auto-explore stops for characters with talisman synergy. The new engine support makes the ae += >talismans / explore_stop += talismans lines actually work.


RypoFalem:
Thanks for the contribution! You asked me to look at this so here's my feedback. 

Talismans already have the `talisman` prefix because item base types are already added to prefixes.

`explore_stop` can already stop based on if the item would be autopickup-able  (using `item`, `greedy_pickup`, etc) and `explore_stop_pickup_ignore` is there for exceptions.

I think it's strange to add talismans as a `explore_stop` category because talismans are just regular items that can be treated like other items with respect to explore_stop. I don't see the need for a new category. If there is something that can't already be done, why not make it doable with all items instead of only talismans?

I believe altering your snippet slightly should already work. I changed "talismans" to "talisman" (the prefix isn't plural). I also explicitly added `ae+= <talisman` so this snippet should work in an otherwise empty initfile (if you have `%` in your `autopickup` option, that line isn't needed). This assumes `explore_stop` contains `greedy_pickup_smart` (it does by default).

```lua

  ae := autopickup_exceptions

  # Default: ignore talismans
  ae += >talisman

  # Turn talismans back on for species/backgrounds that care about shapeshifting
  : if table_has({"Felid","Octopode","Troll","Oni","Poltergeist","Revenant",
  :               "Demonspawn","Demigod","Naga"}, you.race())
  :    or you.class() == "Shapeshifter"
  :    or (table_has({"Gnoll","Merfolk","Tengu","Barachi","Deep Elf","Formicid"},
  :                 you.race()) and not table_has(pure_melee, you.class()))
  : then
      ae -= >talisman
      ae += <talisman
  : end
  ```

RypoFalem:
So overall I think the PR is trying to enable something that is already possible? If I misunderstand please let me know what it's trying to accomplish exactly.

robbiemu:
> Thanks for the contribution! You asked me to look at this so here's my feedback.
> 
> Talismans already have the `talisman` prefix because item base types are already added to prefixes.
> 
> `explore_stop` can already stop based on if the item would be autopickup-able (using `item`, `greedy_pickup`, etc) and `explore_stop_pickup_ignore` is there for exceptions.


Thank you for taking the time to take a look, especially since you do not contribute in the code base (it was my mistaken impression that you did!). It's very kind of you. Yes that about describes it.

Re: explore_stop. The existing flags all key off autopickup eligibility. That works fine if you leave items or greedy_items in explore_stop, but it doesn’t give you a way to say “stop for talismans even though I don’t stop for other non-autopickup items” (talisman itself isnt a prefix [prior to this PR], it is a raw string match). In my rc, I remove items (to cut down on explore spam) and rely on specific categories for the few things I care about. Without ES_TALISMAN, there was no knob to get talisman stops back unless I reintroduced items (thereby stopping for every miscellaneous scroll/weapon again) or forced talismans into autopickup and relied on greedy_pickup_*, which has different semantics (it only triggers after the pickup happens, not at LOS reveal). The new bit simply mirrors the existing glowing_items / artefacts / runes flags so talismans can be toggled independently.

Re talisman prefix. _autopickup_item_name already appends the terse base-type string, but that string is the singular "talisman". I just, for consistence/leniency, adapt it to also accept the plural.

PR #5016, opened 2026-01-07
Correctly set tile flavours with multiple terrain changes

CrawlOdds:
This fixes a bug whereby reverting one of a stack of terrain changes would show the flavour of the base tile. For example:
- Cast bog in shoals, covering an upstair (which has a flavour tile).
- Have a water nymph cover the upstair tile.
- Kill the nymph

This would display a stairs tile (but not allow the player to go up it)


DracoOmega:
Merged, thanks!

PR #5017, opened 2026-01-08
Remove duplicated -Cast description

person50605:
-Cast in Status.txt is duplicated, so one should be removed.


WizardIke:
Thanks, I've merged the change

PR #5019, opened 2026-01-09
Don't re-enter shops after they reappear (Flugkiller)

CrawlOdds:
When a shop was covered by a temporary terrain change, it was marked as unnoticed when it came back, and therefore reentered.

This was happening because even though temporary changes don't call unnotice_feature, their reversions do.

We fix this by making unnotice_feature only work on features which are present (which was already the case for stairs), and fixing the one caller of this who changed features before unnoticing them.

Also make a function with an unused return type void.


WizardIke:
Thanks for the fix

PR #5024, opened 2026-01-14
Make the Cost column show the full cost of reaching skill targets

unutbu:
- What it does
  - When a target is set, the Cost column shows the total cost of
    reaching the target skill level

  - When no target is set, the Cost column shows the cost of adding 1.0
    to the skill level.

    - For integral skill levels, there is no change in the current
      behavior.

- Why is it useful
  - Suppose a player wishes to compare the cost of:

    - Training Shields from 1.1 to 10
    - versus training Dodging from 1.5 to ??

  - Currently, it is difficult (without a wizmode test) to determine how
    many levels of Dodging could be trained for the same amount of XP as
    the first option. This PR makes this easier to do in-game.
cost
target

  - As a side effect of computing total cost to reach decimal targets,
    this PR makes the Cost column change more granularly, as
    skill levels (or targets) change by tenths.

    - For example, currently the Cost column does not change
      as long as the floor of the skill level stays the same. With this
      PR, the cost changes as either the skill level or the target
      changes by tenths:
```
      | skill level | target | master cost | PR cost |
      |-------------+--------+-------------+---------|
      |         1.0 |    2.0 |         3.4 |     3.4 |
      |         1.0 |    1.1 |         3.4 |     0.3 |
      |         1.5 |    2.0 |         3.4 |     1.7 |
      |         1.5 |    2.5 |         3.4 |     4.2 |
```
- Handling modified skill levels (like Wildshape or Ash skill boosts)
  - Normally, a player sets a target based on (possibly) modified skill
    levels. But the cost of improving that skill should be based on the
    unmodified skill level. So (for the purpose of computing the cost)
    the target is adjusted (generally downwards) by the amount the skill
    is currently boosted. Note that this adjustment may be roughly
    correct but not exactly accurate for Ash skill boosts, since the
    size of the skill boost is itself affected by the skill level.

  - Here is an example of the PR's behavior when using Wildshape. For
    each pair of lines below, the first line shows what cost is reported
    when using wildshape. The second line confirms the computed costs
    using unmodified skill and target levels.

```
    | wildshape | raw skill | target | master cost | PR cost |
    |-----------+-----------+--------+-------------+---------|
    | yes       |         0 |      6 |         1.4 |     1.4 |
    | no        |         0 |      1 |         1.4 |     1.4 |
    |-----------+-----------+--------+-------------+---------|
    | yes       |         0 |      8 |         1.4 |     8.5 |
    | no        |         0 |      3 |         1.4 |     8.5 |
    |-----------+-----------+--------+-------------+---------|
    | yes       |         5 |     12 |         8.5 |    18.4 |
    | no        |         5 |      7 |         8.5 |    18.4 |
```

### DIFF
--- a/crawl-ref/source/skills.cc
+++ b/crawl-ref/source/skills.cc
@@ -189,17 +189,31 @@ int skill_cost_baseline()
 }
 
 /**
- * The skill cost to increase the given skill from its current level by one.
+ * The skill cost to increase the given skill from its current level to
+ * the target or by one if no target has been set.
  *
  * @param sk the skill to check the player's level of
  * @returns the XP needed to increase from floor(level) to ceiling(level)
  */
-int one_level_cost(skill_type sk)
+int target_cost(skill_type sk)
 {
     if (you.skills[sk] >= MAX_SKILL_LEVEL)
         return 0;
-    return skill_exp_needed(you.skills[sk] + 1, sk)
-           - skill_exp_needed(you.skills[sk], sk);
+
+    // Normally, a player sets a target based on (possibly) modified
+    // skill levels. But the cost of improving that skill should be
+    // based on the unmodified skill level. So I adjust the target
+    // (generally downwards) by the amount the skill is currently
+    // boosted. Note that this adjustment is not exactly accurate
+    // for Ash skill boosts.
+    int you_deciskill = you.skill(sk, 10, true, false);
+    int you_deciskill_modified = you.skill(sk, 10, false, false);
+    int deciskill_diff = (you_deciskill_modified - you_deciskill);
+    int target = you.training_targets[sk] - deciskill_diff;
+    target = (target <= you_deciskill) ? you_deciskill + 10 : target;
+
+    return static_cast<int>(deciskill_exp_needed(target, sk)
+                            - deciskill_exp_needed(you_deciskill, sk));
 }
 
 /**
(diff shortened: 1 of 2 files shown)


unutbu:
This PR creates a new function, `deciskill_exp_needed` which is similar
to the already existing function `skill_exp_needed` -- it can compute
everything `skill_exp_needed` computes but can also handle tenths of
skill levels. 

I would like to replace `skill_exp_needed` with `deciskill_exp_needed`
and then perhaps rename `deciskill_exp_needed` to `skill_exp_needed`,
but I wanted to keep this PR as small as possible.

If you think that `deciskill_exp_needed` should replace
`skill_exp_needed`, I could do the refactoring here or in a separate
PR.

DracoOmega:
I agree that being able to see the cost of training a skill all the way to a desired level would be very useful to have in-game, but I do have some concerns that making the semantics of the cost column quite different depending on whether or not a skill target is set (which itself isn't even displayed unless the player goes into skill target setting mode) will be confusing to many players. It's not like relative cost-per-level stops being a useful or relevant piece of information just because this also is...

We don't exactly have a lot of spare room in this interface, but I wonder if 'cost of reaching X skill' should be highlighted in a different color than 'cost of gaining 1 level of skill' and the key at the bottom should explicate the difference between these? Though even then, one would still need to press = to even see how many levels this was displaying the cost *for*. I wonder if there's a better way than either of these (though I don't know what it is, just yet).

CrawlOdds:
I think b1462f3 (seeing XLs to target on the help for a skill) covers this piece of information, so closing this one

PR #5026, opened 2026-01-14
Lucky is as lucky was

unutbu:
The message log uses present tense to describe events so let's change "was" to "is" in the message "Lucky! The ABC was actually the XYZ!".


pooka109:
Applied; thanks.

PR #5033, opened 2026-01-16
Quickset skill target to maximum skill on talismans

CrawlOdds:
Allow (s) to set a skill target to the max skill, if we are above the min skill


DracoOmega:
Merged, thanks!

PR #5034, opened 2026-01-17
fix: Blade-form status text (Undo)

RypoFalem:
It was simply "hands" (or whatever similar limbs you possessed). Now it defaults to "blade-form" since eels are the hands-replacer now.


WizardIke:
Thanks for the fix

PR #5035, opened 2026-01-17
Bind right click to Describe when in Display Map to match Look Around

Jewelots:
I kept getting thrown off when trying to right click examine things in Display Map mode ('X') because it currently does nothing, requiring a 'v' press instead, so I matched the typical behavior of right click in standard gameplay. Right clicks in map mode/`MOUSE_MODE_NORMAL` were just being thrown away, so it's just using a button that was previously unused.

I also moved describing into a seperate function with a position argument instead of calling `process_command(CMD_MAP_DESCRIBE)` because `CMD_MAP_DESCRIBE` requires you move the cursor first, which jarringly moves the view as well, unlike the typical right click examine. 

I agree with the comments about "this really seems like it shouldn't be here" for the click handling code, but it's above my expertise so I didn't do any further refactors; hopefully my change is small enough that it's not growing the existing slightly-messy code too badly.

### DIFF
--- a/crawl-ref/source/tilereg-dgn.cc
+++ b/crawl-ref/source/tilereg-dgn.cc
@@ -474,8 +474,7 @@ int DungeonRegion::handle_mouse(wm_mouse_event &event)
         return 0;
 
     if (mouse_control::current_mode() == MOUSE_MODE_NORMAL
-        && event.event == wm_mouse_event::PRESS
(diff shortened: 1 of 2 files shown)


WizardIke:
Thanks for the PR. It looks like you haven't been added to the credits yet, what name would you like me to add you under?

Jewelots:
Thank you for the merge! Jewel, please ^_^

PR #5036, opened 2026-01-17
Fix a bolt of flesh redirection crash

CrawlOdds:
When firing a Zykzyl bolt of flesh with Ru redirection or Dith's Marionette, we would sometimes crash. This is because the tracer beam for the attack is aimed at the Zykzyl itself, and the beam sets stop_at_allies. This causes us to attempt to regress a ray with no position or direction, which does not work.

We fix this by not regressing self-aimed rays when stop_at_allies is set.


DracoOmega:
Merged, thanks!

PR #5045, opened 2026-01-23
feat: add estimated XL to skill target in the skill ? menu

RypoFalem:
Because sometimes you want to know how long it will take to reach some arbitrary skill target that isn't a weapon's mindelay or a talisman's min/max skill.

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -4334,6 +4334,28 @@ string get_skill_description(skill_type skill, bool need_title)
 
     result += getLongDescription(lookup);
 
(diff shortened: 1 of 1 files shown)


RypoFalem:
image
It looks like this.

Cgettys:
This is a fantastic idea :)

Cgettys:
I'd also love to see how much MP and HP added to FIghting and Invo and so on, but I should do that myself rather than just complaining if I can find time...

CrawlOdds:
Very nice, thanks! Fixed (with minor commonisation) in b1462f3

PR #5046, opened 2026-01-24
Handle feature mimics and terrain change better

CrawlOdds:
We ignore feature mimics which are currently covered by a terrain change, and don't discard feature mimics on terrain reversion.

This fixes a few silly issues:
- If you covered a mimic and didn't step next to it, it became a real feature. This awesome tech can be used to get 1% more shops using hellfire mortar or similar.
- If you cover a mimic with temporary terrain other than water (e.g. a bog or hellfire mortar), and then stepped next to it, it was revealed as a lava mimic or similar.

This means that if temporary mimics previously worked (which the author would bet against), they definitely don't now.

### DIFF
--- a/crawl-ref/source/mon-util.cc
+++ b/crawl-ref/source/mon-util.cc
@@ -1010,6 +1010,10 @@ void discover_mimic(const coord_def& pos)
 
     const dungeon_feature_type feat = env.grid(pos);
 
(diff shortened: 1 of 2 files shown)


WizardIke:
This doesn't quite work as we still need to remove the mimic status when changing from one type of permanent terrain to another. I going to close this as I did a similar fix in 4d7c644

PR #5053, opened 2026-01-27
fix: correct crawl.grammar re: qual vs. qualname

dossy:
See: crawl-ref/source/libutil.cc description_type_by_name()


dossy:
https://github.com/crawl/crawl/actions/runs/21416360265/job/64614158806 failed because it needed https://github.com/crawl/crawl/commit/511dc2e70d5548748bd4992796731a204faa7df3

WizardIke:
Thanks for the fix. I don't think you've been added to the credits yet. If you haven't, want name would you like me to add you under?

dossy:
@WizardIke `Dossy Shiobara` would be great, thanks.

PR #5054, opened 2026-01-28
feat: alert monsters affected by clouds to the clouder

RypoFalem:
Since f96b50d monsters are no longer alerted automatically to cast spells. That made apparent that clouds weren't alerting monsters to the cloud creator; monsters would wander through player-created freezing clouds without taking notice of the player. Silly! So change that.


DracoOmega:
Merged, thanks!

PR #5057, opened 2026-02-01
Don't summon trees under monsters who can't live in them

CrawlOdds:
This is especially awkward for creeping plasmodia, who "flop about on dry land" in response. Earth elementals cope better, but it still seems like a bad idea to violate habitability rules like this.


DracoOmega:
Merged, thanks!

PR #5061, opened 2026-02-03
fix: Prompt for talisman warning inscriptions and excommunication

RypoFalem:
Talisman !P (Put on) and !R (Remove) inscriptions are now respected. (As well as !*, implicitly.)

Talismans that would excommunicate you now give you a warning. This technically also affects rings and amulets. Though, as far as I know putting on rings and amulets that your god hates merely give you penance each turn. I suspect the warning for penance was intentionally absent so I didn't change that but if there's a ring or amulet that would instantly excommunicate you, there will now be a prompt for that.

The items were already highlighted a scary red by default which was warning enough for the issue to go unnoticed.

### DIFF
--- a/crawl-ref/source/invent.cc
+++ b/crawl-ref/source/invent.cc
@@ -1687,7 +1687,8 @@ bool needs_handle_warning(const item_def &item, operation_types oper,
         return true;
     }
 
-    if (oper == OPER_EVOKE || oper == OPER_PUTON && god_hates_item(item))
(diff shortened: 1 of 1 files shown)


RypoFalem:
Hold up. There's an issue with flux baubbles

RypoFalem:
Ok should be good to go when the tests pass.

DracoOmega:
Merged, thanks!

PR #5062, opened 2026-02-05
text: minor tutorial message tweak

RypoFalem:
You can pretend this had a misleading commit title like "replace stones with boomerangs" if you like.


gammafunk:
The current text is actually correct. This message is referring to the training dummies in the previous section, not the player.

PR #5072, opened 2026-02-11
feat: Add options to limit longwalking aka shift running

RypoFalem:
I often use shift+direction instead of autoexplore because it lets me explore manually with significantly fewer key presses and lets me act fast with the assurance that it will stop if a monster comes on screen. However, sometimes longwalking can unexpectedly put you dozens of tiles from where you started in unexplored territory leaving a lot of opportunity for monsters to flank you. Avoiding such situations is usually why I opted for longwalking over autoexplore in the first place!

There's a new option, longwalk_range, that gives you some control over the maximum distance shift+direction can take you. It is documented in options_guide.txt but I will summarize the possibilities here:
* los       - travel as far as your character's current line of sight range.
* visible   - travel only to tiles you can see when you start.
* unlimited - no distance limitation. Just as it was before.
* constant  - a set distance the user can provide using the new longwalk_range_constant option.

-- end commit message

I've wanted this feature for years but got lost trying to make sense of the travel code every time I tired haha.

Some bike shedding potential:
* is longwalk a good name for it? That's the name given in the in-game control menu `??` and I preferred using that name at least in the codebase because I felt "run" might be ambiguous. The player-facing option doesn't need to keep that name, I think.
* What's the best default mode? The current default is `los` which is a character's los radius, which changes with species, equipment and god abilities. A `constant` 7 or 10 might confuse players less. My favorite is `visible` which only travels to tiles you could currently see (so up to 2 tiles with heavenly storm active) but I am not sure if others would appreciate that.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -1144,6 +1144,23 @@ travel_open_doors = (avoid | approach | open)
             open = Autoexplore/travel will open doors. (default)
         This does not affect runed doors, which are always avoided.
 
+longwalk_range = (los | visible | unlimited | constant)
+        Set the maximum distance for longwalking (shift+direction by default).
+          los = Your current line of sight range. This is 7 for most species and
+                is affected by items such as scarf of shadows.
+          visible = Only visit visible tiles. This is limited by line of sight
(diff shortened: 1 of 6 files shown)


CrawlOdds:
Thanks! Merged in 4de224a.

Issue #3346, opened 2023-11-09
Ely penance ambiguity, unrewarding gameplay


robbiemu:
**PLATFORM and OS or BROWSER:**
webtiles at crawl.kelbi.org

**DCSS VERSION AND GAME INFORMATION:**
0.31 "trunk" at crawl.kelbi.org

**DESCRIPTION:**

tl-dr; when completing penace for a good god, which you might do for gameplay reasons, you original piety should be restored.

I did a bad thing, and I saw the error in my ways. I killed a bunch of butterflies in vault:2 with immolation and scroll of butterflies. I was placed under penance. After some continued play, I brought it up on discord, and was told I can just leave and rejoin. It brought to mind the natural question "if I wait out penance, will my piety be at zero or will I get my old piety back?" And the same person said "the later". I looked in the source a tiny bit and noticed there is a MAX_PENANCE of 200. So I thought, well, I've already played through probably 50 piety (which was an overestimate, hindsight is 20/20), I'm basically just as far either way, I'll just play out penance.

I completed vault 2, and 3 and 4. I went on to elf and after completing elf 2, I finally saw the first change in my penance status message. So I went on to crypt and completed crypt. Even after crypt, I still was in the final stages of penance. So I went into Depths and in depths 1 I finally completed penance. My piety was at 0 pips. 

I sent screenshots on discord and we talked about it, someone noted that while it should restore piety, piety decay still happens. 

Of course, it takes so long to work off penance when you immolate butterflies that by then, my piety had decayed to zero, so the discord idea goes.

To me this is _horribly wrong_. I am chatting with contributors and players who go on 15 rune streaks and the answers they had were natural enough and inspired good gameplay. Yet this failing consequence is stunning.

**EXTRA FILES:**

PR #5083, opened 2026-02-14
Except piety decay of good gods during penance

Coaxy:
During penance, the piety of the good gods would no longer decay.

Saw https://github.com/crawl/crawl/issues/3346 and agreed with it. While the good gods ability to change gods without wrath is interesting, having your piety decay during penances (especially during long ones) makes the decision for you.


RypoFalem:
Should point out that there's a plan to remove piety decay in general! #5007

WizardIke:
Closing as piety decay has now been removed

PR #5090, opened 2026-02-18
Bugfix range description for two monster spells

anderspapitto:
These two spells both use _foe_not_nearby
https://github.com/crawl/crawl/blob/511dc2e70d5548748bd4992796731a204faa7df3/crawl-ref/source/mon-cast.cc#L473

https://github.com/crawl/crawl/blob/511dc2e70d5548748bd4992796731a204faa7df3/crawl-ref/source/mon-cast.cc#L537

which enforces a minimum range of 3, not 2
https://github.com/crawl/crawl/blob/511dc2e70d5548748bd4992796731a204faa7df3/crawl-ref/source/mon-cast.cc#L3590

I notice that beckoning gale also uses _foe_not_nearby, maybe it should also receive special treatment at these two locations?


DracoOmega:
Merged (with the suggested inclusion of Beckoning Gale). Thanks for noticing!

PR #5093, opened 2026-02-19
Restore old raven quote

damerell:
Removed with ravens. More pleasing than the obvious Poe selection.


damerell:
FWIW on Discord nicolae suggested that, since ravens now bring DOOM, the obvious Poe poem or something from Norse mythology would be preferable. I rather like this one from Alice, but up to you really.

PR #5113, opened 2026-03-01
tweak the morgue upload setup

geekosaur:
improve documentation/comments
only upload if we had an actual crash, per DracoOmega
reduce retention time to 7 days


geekosaur:
Away this afternoon but I again want to force a test crash to make sure `hashFiles` is getting the right file before I take it out of draft.

gammafunk:
Thanks for the PR!

PR #5114, opened 2026-03-01
fix: ^Fragile equipment being destroyed before unequip effects

RypoFalem:
I got to use the bug for my benefit and now I am pulling up the ladder behind me!
---
The player could dodge negative unequip effects on a ^Fragile item. This means Amulets of Faith wouldn't have a piety penalty for removal and weapons of Distortion wouldn't do their distortion unwield effects. This is apparently unintended and it turns a negative property into a positive one which will surprise players and vault makers.

This commit breaks ^Fragile gear only after other unequip effects have happened.


WizardIke:
Thanks for the fix

PR #5130, opened 2026-03-11
Allow autoaiming explosions at targets we can't hit directly

CrawlOdds:
When targets can be hit by an explosion, but not by a direct shot, we were marking spells as useless and not autoaiming. This fixes this by changing some visibility checks to use default visibility.


DracoOmega:
Merged, thanks!

PR #5134, opened 2026-03-14
Update map correctly when Ashenzari reveals portals

CrawlOdds:
19414c7 changed how we handle revelation, drawing mapped cells immediately when revealed. However, it neglected to do this for other methods of mapping, such as when Ashenzari reveals a portal.

We fix this so that whenever we set the magic mapping flag, we also draw the mapped tile.


WizardIke:
Thanks for the fix

PR #5136, opened 2026-03-15
Check skill training targets when cursing items

CrawlOdds:
This isn't terribly important - without this we still check the target next time we gain experience, and disable the skill then - but is a bit clearer and is consistent with other strange skill gainings, like draconian apts changes.


WizardIke:
Thanks for the PR

PR #5137, opened 2026-03-15
Fix Ru drink and words thresholds slightly

CrawlOdds:
This makes them both obey their wording "below 50% of its maximum".

Previously, words allowed scrolls at exactly 50% HP, while drink banned potions just below 50% if you had odd max HP.


WizardIke:
Thanks for the fix, it should be merged now

PR #5142, opened 2026-03-19
Improve barding drop chances

planckenstein:
Improved barding drop chances on all monsters that can drop them.


regret-index:
Superseded by db5e642.

PR #5148, opened 2026-03-21
fix: Elven twins didn't get haste/berserk if they can't see the player

RypoFalem:
The twins get a buff if the other twin is dead when they first see you. If they were blind or you were invisible, they wouldn't get their buffs until the moment they could see you. It's unclear if this was once intended behavior but it's spoilery tech if so.

Now they get buffed when they are your line of sight, regardless of who can see whom. This should ensure that the player gets a message when the buff is applied. (unless the twins are invisible somehow)

### DIFF
--- a/crawl-ref/source/mon-act.cc
+++ b/crawl-ref/source/mon-act.cc
@@ -4178,7 +4178,7 @@ void seen_monsters_react()
         if (!mi->visible_to(&you))
             continue;
 
-        if (!mi->has_ench(ENCH_FRENZIED) && mi->can_see(you))
+        if (!mi->has_ench(ENCH_FRENZIED) && you.see_cell(mi->pos()))
         {
             // Trigger Duvessa & Dowan upgrades
             if (mi->props.exists(ELVEN_ENERGIZE_KEY))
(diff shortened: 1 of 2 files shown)


DracoOmega:
At first, I was hesitant whether it was a good idea to let them react to the player if they hadn't even *noticed* the player (ie: if they come into LoS after an off-screen death without being aware of you), but the more I look through this code, the more I am uncertain why we're doing their death enrage in such a deferred fashion. It may have historical roots in when Duvessa's berserk wasn't endless (so it actually mattered whether she did it in sight of your or not), but at this point I am wondering if a better approach would be to just do all of their empowerment upon the twin's death and leave the rest of this out.

(As it stands, *part* of Dowan's upgrade is even done in a different place already, specifically because they considered elven_twin_energize() unreliable >.>)

RypoFalem:
I had assumed it was so the player got a message saying the monster was buffed (giving info and opportunity for force mores to trigger). I don't feel strongly about it but that's the only reasoning I could imagine so I went with it haha

CrawlOdds:
Fixed (empowering them immediately unless frenzied) in 1b0f8bd

PR #5150, opened 2026-03-21
fix: Amulet of Dissipation now removes -Swift faster (aeshnacyanea)

RypoFalem:
-Swift is dispellable and apparently negative. However, it uses DUR_SWIFTNESS which is also used by the positive Swift status and so it is not flagged D_NEGATIVE.

We now check for -swift when checking if the effect is negative so dissipation will speed up the slow down effect.

---

(I know the parenthesis isn't necessary but thought it would be good for clarity)

### DIFF
--- a/crawl-ref/source/status.cc
+++ b/crawl-ref/source/status.cc
@@ -114,7 +114,8 @@ bool duration_dispellable(duration_type dur)
 
 bool duration_negative(duration_type dur)
 {
-    return _lookup_duration(dur)->duration_has_flag(D_NEGATIVE);
+    return _lookup_duration(dur)->duration_has_flag(D_NEGATIVE)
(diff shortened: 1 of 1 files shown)


DracoOmega:
I remain on the fence about whether this is appropriate or not. It's true that -Swift is bad, but it feels closer to a cooldown than other negative effects do, and Dissipation doesn't work on those. Which would make this potentially the single case of a spell being usable more rapidly because of this amulet.

I'm not 100% opposed, but hesitant. (And decided to actually make a note of this here rather than having it appear to be ignored altogether ^^; )

CrawlOdds:
Went the opposite way to resolve this inconsistency in 28ce492; -swift is now undispellable

PR #5167, opened 2026-03-31
Tree description and tile bugs

CrawlOdds:


DracoOmega:
Testing this out, there still seem to be some oddities with vault renames. Or rather, if one casts Summon Forest in, say, the dead forest Crypt end, the trees that are created will all say 'tree' instead of 'dead tree', but the *tiles* for some of those trees will be those of dead trees (and some won't, for reasons which aren't immediately obvious to me.). And the preexisting trees which are awakened will still say 'dead tree' while looking the same as many of those that claim to be alive. I am not actually certain this behavior is an improvement?

(The other two commits look good, from what I could see, but I'll wait on pulling them out of this PR)

CrawlOdds:
I haven't been able to reproduce that - when I cast Summon Forest in the dead forest, I get existing dead trees awoken, and live trees created - in both cases they have descriptions which match the tiles (here are a couple of screenshots of the forest with and without my magic)
image
image

DracoOmega:
Weird. I tested again and still get the same result, as seen here:
crypt_forest1
crypt_forest2

CrawlOdds:
Ah, I can reproduce your dead trees by destroying existing trees and then recreating them with forest.

Unfortunately the tile changing stuff (third commit here) here is no good at all - we can't use tile overrides for temporary terrain because it ruins the tile flavours for out-of-LoS tiles when we reload the game.

Overall I think this stuff to do with tiles and temporary terrain needs a more root and branch approach - other than the tiny change to describe trees as flammable I don't think this worth merging.

WizardIke:
Creating temporary terrain doesn't change the grid colour of a square, so if you create a temporary tree on a brown square (such as the brown path squares in the picture above) you will get a brown tree and the tile for a brown tree is the dead tree tile

CrawlOdds:
It is worth noting that the above means that summon forest trees don't necessarily get the correct descriptions when the tile is overridden - e.g. in disco pan we get all sorts of crazy trees with normal tree descriptions (before or after any of these changes)

CrawlOdds:
Merged the flammability description commit in f8e1fdb. The tile issues have been better dealt with elsewhere, and the description overrides are only sometimes right (summoned trees can be dead if they appear on paths), and should probably await a more comprehensive solution.

PR #5177, opened 2026-04-05
Enchantment on mundane items fix

dllllb:
Ensure that newly enchanted mundane weapons and armor utilize an enchanted tile variant


pooka109:
Implemented, but with different code. Thanks anyway.

PR #5178, opened 2026-04-05
Fix autoexplore leaving unexplored diagonal tiles at map edges

dllllb:
Autoexplore sometimes leaves unexplored diagonal tiles at map edges

Fix: Add in_bounds(*ai) — boundary tiles can never hold monsters or interesting content; only playable in_bounds tiles are candidates for a lurking threat. The redirect now only fires for genuinely unexplored playable tiles.


CrawlOdds:
I believe the bug this was aimed at fixing was addressed in 7ecd7e9

PR #5191, opened 2026-04-08
Display Status Icons on Monster Tile in Title of Describe View

### LINKED ISSUE #5132: Monster status icons do not show in title tile of describe view (xv) in local tiles

local tiles only

trunk

**DESCRIPTION:**

The monster title tile at the top of the describe view (xv screen) does not display active monster status like sleep ZZZ or ally green halo in local tiles. This seems to have been broken in commit 322ac23. 

To resolve, it may be possible to move this `fake_cell` code to its own function so that it doesn't run each time `add_monster` is ran, and then use that function to display the describe title tile. Another method may be to just put the `fake_cell` code directly into `describe.cc`.

sandertyu:
The function DungeonCellBuffer::add_monster() is used once in describe.cc to display the monster tile when viewing a monster description, and is used nowhere else. Local Tiles has not displayed status icons nor foreground flags on the monster tile in the title of the describe view since commit 322ac23. A call to function pack_foreground() was removed from add_monster() in that commit, and that led to issue #5132.

Add a function call to pack_foreground() in DungeonCellBuffer::add_monster() so that status icons and foreground flags display on the monster tile in the title of the describe view. This effectively reverts 322ac23, while incorporating the tile_with_flags_t refactoring done in 9a5c35d, and fixes #5132.

### DIFF
--- a/crawl-ref/source/tiledgnbuf.cc
+++ b/crawl-ref/source/tiledgnbuf.cc
@@ -106,6 +106,7 @@ void DungeonCellBuffer::add_monster(const monster_info &mon, int x, int y)
 {
     tile_with_flags_t t = tileidx_monster(mon);
     tileidx_t t0   = t.tile();
+    tile_flag_t flag = t.flags();
 
     // Copied from _tile_place_monster()
     if (!mons_class_is_stationary(mon.type) || mon.type == MONS_TRAINING_DUMMY)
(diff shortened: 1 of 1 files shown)


CrawlOdds:
Thanks for the fix!

CrawlOdds:

sandertyu:

Thank you for the merge! You can credit me as Lyla Sanders.

PR #5194, opened 2026-04-10
Fix crash in monster::attempt_escape() when constrictor is gone

### LINKED ISSUE #5192: [Trunk] Game crashes when reentering Zot:2
webtiles, mozilla firefox 

Trunk on Xtahua, username: Troq, ver. ""0.35-a0-263-g9a5c35dbd8". 
morgue: https://crawl.develz.org/wiki/doku.php?id=mantis:savebackup 

**DESCRIPTION:**
Game crashes instantly after trying to reenter Zot:2 via stairs or hatch. 

**EXTRA FILES:**
Crash report: https://crawl.xtahua.com/crawl/morgue/Troq/crash-Troq-20260409-092431.txt

OlivierCavadenti:
When re-entering a level, attempt_escape() can be called on a monster whose constrictor no longer exists; handle this gracefully instead of asserting.

Closes https://github.com/crawl/crawl/issues/5192

### DIFF
--- a/crawl-ref/source/monster.cc
+++ b/crawl-ref/source/monster.cc
@@ -6541,7 +6541,12 @@ bool monster::attempt_escape()
     else
     {
         const monster* themonst = monster_by_mid(constricted_by);
-        ASSERT(themonst);
+        if (!themonst)
+        {
+            // The constricting monster is gone (e.g. died off-level).
+            stop_being_constricted(true);
+            return true;
+        }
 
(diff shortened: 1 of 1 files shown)


CrawlOdds:
Fixed in a different way in be839c1 (this PR would also have fixed the immediate issue, but it's nice to keep the invariant that monsters being constricted have a constrictor, which this assert enforces)

PR #5195, opened 2026-04-10
Do not power up Wyrmbane twice when killing dragons with convulsion

CrawlOdds:
Since 2bee245, when we killed a dragon with convulsion wyrmbane powered up twice - once in the on-hit melee effects call, and once in the on-kill one.

The reason this worked _before_ that commit is that when wyrmbane killed a monster via the hurt() method, the monster was cleaned up and therefore did not register as a dragon in handle_phase_killed() call. Now that we use inflict_damage, this has changed. So we make sure we never power up wyrmbane in the on-hit melee effects, and leave it to handle_phase_killed in all cases.


DracoOmega:
Merged, thanks!

PR #5198, opened 2026-04-10
Don't use randart tiles for enchanted hammers

CrawlOdds:
Fixes #5165

I couldn't reproduce this (though I have seen it on webtiles), but this seems very likely to be the cause.


DracoOmega:
Merged, thanks!

PR #5200, opened 2026-04-12
Fix Eringya's Surprising Crocodile movement effects

CrawlOdds:
When the spell moved the player 1 tile and created shallow water there, movement effects like barbs and icy armour breaking weren't triggering.

This happened because the terrain change was calling trigger_movement_effects, clearing the deferred move.

This does mean that if there were any movement effects that should be suppressed by the terrain change, we'd trigger them anyway. But since we only change terrain if it is plain floor, this shouldn't be a problem in practice.


Hellmonk:
It seems that barbs doesn't trigger with the movement finalization in either position, despite the prompt saying that it will.

CrawlOdds:
Merged with an additional comment in 8c32f14340b1

PR #5201, opened 2026-04-13
Fix item method docs

aeshna-cyanea:
Ldoc was generating item methods with a self param (e.g. `item:inscribe(param)` notation), when the correct usage is `item.inscribe(param)`. Changing the annotations to reflect this makes ldoc generate the correct docs.


CrawlOdds:
Thanks for the fix, merged in 6b10917 (with a few zero-parameter functions added).

PR #5203, opened 2026-04-14
Let tracers enchantment chain

CrawlOdds:
This fixes petrify not targetting monsters who were petrifying but could set off enchantment chains, following 8e16f14


WizardIke:
Thanks for the fix

PR #5210, opened 2026-04-17
Fix a bug with exclusions for new enemies on descent

CrawlOdds:
Flagging that there may be better ways to fix this - particularly we could make map_knowledge in date so that auto exclusions don't need to look at ground truth monster data.

---

When we traverse stairs and see a new monster which should be autoexcluded, we were previously placing an exclusion without the monster name in its description, and then immediately removed it before the player could even see it.

This is happening because the autoexclusions happen before our map_knowledge is updated. An alternative fix would be to update map_knowledge before add_auto_excludes, or on changing level.


CrawlOdds:
Fixed in a different way in e965272cd008

PR #5216, opened 2026-04-19
Implement a bias for exploring near stairs

CrawlOdds:
This penalises squares in autoexplore by their distance to the nearest known upstair.

The formula is
adjusted_distance(x) = distance(x) + bias*stair_distance(x)/10

From some testing, a bias of x adds approximately x% to explore time for x=0,5,10,20,100. The exploration clearly hugs stairs, sensible so at biases from 0 to 20 and tediously so at 100.

Performance-wise, this adds one extra floodfill to autoexplore, which very roughly doubles how long it takes; on my laptop from 1.1ms to 1.8ms in a typical case on a fully-explored floor.


CrawlOdds:
This is very much a draft - it seems to work but I haven't thought hard enough about some bits of it (e.g. the scarily global travel_point_distance).

Issue #5138, opened 2026-03-15
Sigil of Binding does not give swiftness to player


petete1:
**PLATFORM and OS or BROWSER:**
Webtiles

**DCSS VERSION AND GAME INFORMATION:**
0.34.1

**DESCRIPTION:**
Sigil of Binding locks the target in place, then gives swiftness. This works for monsters but the player doesn't get the swiftness.

fmaddaxumich:
I took a look into this and can confirm, the Sigil of Binding doesn't work when the player walks into it. (Version 0.35-a0-241-g5b54ce7fc3)

After digging through some code, I found the function responsible for handing the logic, it outright doesn't even attempt to put swiftness on the player before returning:
`if (actor.is_player()) {`
`        mprf(MSGCH_WARN, "You move over the binding sigil and are bound in place!");`
`        you.increase_duration(DUR_NO_MOMENTUM, random_range(3, 6));`
`        revert_terrain_change(you.pos(), TERRAIN_CHANGE_BINDING_SIGIL);`
`        return;`
`    }`

A working method I explored was to just give the player the same level of swiftness an enemy would after triggering a trap (duration * 2), which compiled and worked fine. Using the pre-existing systems, I replicated the functionality of the swiftness spell, which gives swiftness then sluggishness. 

I will try to implement that solution fully and make a pull request sometime in the near future! The description **does** say 

> "Anyone who passes over them will be bound, able to act and dodge normally but unable to move away. When the effect wears off, their movement speed will rebound, giving them temporary swiftness.

PR #5217, opened 2026-04-19
fix(spells): Make binding sigil speed up players

fmaddaxumich:
The sigil's description says "Anyone who passes over them"
would receive swiftness once it wears off.
However, this is not the current behavior.

This change mirrors the swiftness spell after the player is bound.
The amount is the same as monsters, 2x the duration of immobilization.

Resolves issue #5138


Hellmonk:
Letting the player use sigils for swiftness is not something we're interested in (too abusable by placing sigils out of line of sight), so I think we will have to update the spell description instead.

CrawlOdds:
Fixed the description in a2a04dd, so closing this one

Issue #5218, opened 2026-04-20
I am a poisoned scroll


riffraff:
**PLATFORM and OS or BROWSER:**
webtiles on https://crawl.xtahua.com trunk, Version 147.0.7727.56 (Official Build) (arm64), macOS

**DCSS VERSION AND GAME INFORMATION:**

```
This is Dungeon Crawl Stone Soup 0.35-a0-290-g6208fce011 (Console/Webtiles)
Report bugs to: https://github.com/crawl/crawl/issues

Features
--------
 * Lua user scripts
 * Webtiles support
 * Wizard mode
 * PCRE regexps
 * Milestones
```

**DESCRIPTION:**

I'm not sure how I triggered this, but I think I put on an inkwell talisman while poisoned.. and the poisoned status remained. I am not sure this is a bug, but since I have infinite poison resistance I think it should expire.

**EXTRA FILES:**

<img width="56" height="56" alt="Image" src="https://github.com/user-attachments/assets/ffa14c57-5229-4bc4-a420-d70ce9e596d2" />

https://crawl.xtahua.com/crawl/morgue/riffraff/morgue-riffraff-20260420-145135.txt

```
%: reflection
@: mildly poisoned (69 -> 69), scroll-form
A: scribal knowledge, magic absorption, novice magic, weak attacks, tiny,
unfitting armour, (horns 1), (tough skin 1), double potion healing, drunken
brawling
a: no special abilities
```

PR #5219, opened 2026-04-20
Time out poison when poison immune

CrawlOdds:
While intentional and reasonable flavourwise, poison going inert when in an unliving form is consistently surprising to players, and doesn't seem important from a gameplay perspective when all unliving forms are slow to swap into.

Fixes #5176, #5218


CrawlOdds:
Done slightly differently in 0203234

PR #5221, opened 2026-04-22
Apply coglin's rev penalty to melee attacks.

### LINKED ISSUE #5220: Coglin's rev penalty may not be applied to melee damage
**PLATFORM and OS or BROWSER:** N/A

**DESCRIPTION:**

While playing as coglin, I got curious about exactly how much the REV mechanic affects damage output. I asked AI about the numbers, and it flagged a potential bug in the code. I then asked several other AI tools to look at it independently, and they also identified the same issue.

Since there's no way to see the exact numbers in-game. I couldn't verify the behavior directly. so, after adding some debug logging and building the game. the logs suggest that the REV penalty is not being applied correctly in melee. 

---
(The code analysis below was provided by Claude Code)

The suspected issue is in `melee-attack.cc`, where `player_apply_final_multipliers` calls `apply_rev_penalty` but discards its return value:

```cc
  apply_rev_penalty(damage);  // return value discarded — damage unchanged
```

  Since `apply_rev_penalty` returns the adjusted value rather than modifying in-place, the penalty would have no effect on melee. The ranged-attack equivalent does this correctly:

```cc
  damage = apply_rev_penalty(damage);
```

  If this is indeed the case, coglin melee damage would be unaffected by REV level — dealing the same damage at 0 REV as at full REV.

Apologies if this turns out to be a false alarm — and always thanks for keeping the project going either way.

**EXTRA FILES:**

CrawlOdds:
Fixes #5220

### DIFF
--- a/crawl-ref/source/melee-attack.cc
+++ b/crawl-ref/source/melee-attack.cc
@@ -2774,7 +2774,7 @@ int melee_attack::player_apply_final_multipliers(int damage, bool aux)
     if (you.duration[DUR_WEAK])
         damage = div_rand_round(damage * 3, 4);
 
-    apply_rev_penalty(damage);
+    damage = apply_rev_penalty(damage);
 
     if (you.has_mutation(MUT_RECKLESS) && weapon
         && hands_reqd(&you, weapon->base_type, weapon->sub_type) == HANDS_TWO)


CrawlOdds:
(Ongoing discussion on the solution here)

PR #5231, opened 2026-04-28
Make LRD hit plasmodia

CrawlOdds:
LRD cast directly at a creeping plasmodium was not hurting it. This was because the logic for LRD explosions was "don't affect the centre unless exploding terrain with a non-solid tile".

We change the logic to "don't affect the centre when exploding a monster", which should be correct in all cases.


Hellmonk:
This seems to work correctly, but draws the beam tile on the wall in all cases which is kind of ugly. Merging for now, but may want to change to avoid drawing the beam unless a monster is actually hit in the explosion's center.

PR #5236, opened 2026-05-01
Don't allow knockback to take monsters out of bounds

CrawlOdds:
This could cause rare crashes with wall walkers being trampled


WizardIke:
Thanks for the fix

PR #5237, opened 2026-05-01
Extend and improve "unusual_monster_items" option functionality and clarify its guide

radio-gra:
Right now the list for brand vulnerability is first assembled before the values from .rc are read, and is not reset, which makes it impossible to disable or narrow the default "vulnerable:holy_wrath" and "vulnerable:entangling" options. The vulnerability check or it's XL adjustment also isn't documented anywhere, nor is the default value of the option present in the options_guide, breaking the convention.


radio-gra:
(p.s. i'm sorry if some things were done in a clumsy or inefficient manner and need to be redone, i did the best i could with what knowledge of c++ and this game's codebase)

PR #5240, opened 2026-05-02
Revert the removal of "merged slime creature" pseudo-monster.

radio-gra:
With this removal, there is currently no way to distinguish base slime monsters from their merged variants. Merged variants are by design near-guaranteed to appear alongside their regular counterparts, are prone to subtly changing their state and/or position and can be very punishing to have been confused with base slimes. This makes tracking them in console a headache, especially if you can't distinguish well between green and lightgreen due to terminal theme or color blindness. I have used the merged_slime_creature mon_glyph override since I found about it and have no alternative solution now that it's gone. I reversed the removal of this pseudo-monster only and added the ability to override its color as well as the glyph (previously it would always be lightgreen, ignoring color overrides for both base and merged slimes).

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -3080,6 +3080,7 @@ mon_glyph -= <monster name or symbol>
              player
              sensed monster
              {trivial,easy,tough,nasty,friendly} sensed monster
+             merged slime creature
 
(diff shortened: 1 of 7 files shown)


regret-index:
Obsoleted by the combo of 10c3989 and 9cf51ca.

PR #5242, opened 2026-05-04
Fix some spells aiming at monsters they can't hit

CrawlOdds:
Following b0ab3d4, spells with no targeter (e.g. haunt) would sometimes aim at monsters they can't hit, because we now consider targets even if they are blocked by transparent obstructions.


WizardIke:
Thanks for the fix

PR #5249, opened 2026-05-07
Fix summon forest vine tile bugs

CrawlOdds:
Summon forest vines, like all tentacles, pick their tile based on their parent and child tentacles. For vines, the parent can be a tree.

Before this commit, the first vine in the chain finds its parent by looking for an adjacent tree. This has interesting consequences when there isn't such a tree, which can happen in two ways: 1) If the tree is destroyed, or
2) If the spell has ended (but the tentacle is out of LoS so still assumed to be there).

When this happens, the vine gets a half-submerged kraken tile (because this is the only tentacle with no parent, or at least it was when submerging was a thing), and then adjusts it by shifting by <START_VINE>-<START_KRAKEN> tiles. However, because vines don't _have_ half-submerged tiles, this takes us out of the vine section of tiles altogether, leading to starspawn tiles.

We fix this by making vines know what tree they came from. It is worth highlighting two edge cases with this:
- If the tree is destroyed, the vine will still appear to come from it (whereas before it would have displayed some silly tile)
- If the tree is destroyed, but another tree is next to the vine is still there, the vine will appear to come from the destroyed tree (whereas before it would have switched to the still-present one)

### DIFF
--- a/crawl-ref/source/beam.cc
+++ b/crawl-ref/source/beam.cc
@@ -964,6 +964,49 @@ void bolt::burn_wall_effect()
         }
     }
 
+    // When burning a tree, check for any vines it is supporting. These must
+    // find a new home or die.
+    for (adjacent_iterator ai(pos()); ai; ai++)
+    {
+        monster *m = monster_at(*ai);
+        if (!m || (m->type != MONS_SNAPLASHER_VINE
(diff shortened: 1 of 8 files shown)


CrawlOdds:
image
An example of this bug with a tree destroyed

CrawlOdds:
Merged in 063b2be after squashing and a little tidying

PR #5252, opened 2026-05-10
Improve ccache reuse for trunk builds with revision-specific install paths

refracta:
## Summary

Live WebTiles trunk builds get almost no ccache reuse between adjacent revisions when install paths include the revision hash.

The Makefile currently puts `SAVE_DIR_PATH`, `SHARED_DIR_PATH`, `DATA_DIR_PATH`, and `WEB_DIR_PATH` into global `ALL_CFLAGS`. On live servers these paths usually include something like `crawl-git-<sha>`, so every new trunk revision changes the compiler command line for almost every object file.

In practice, those macros are only used by:

- `files.cc`
- `initfile.cc`
- `tilesdl.cc`
- `tileweb.cc`

This patch moves those macros into `INSTALL_PATH_CFLAGS_L` and applies them only to the corresponding objects with target-specific Makefile flags.

## Why

A few live server operators reported trunk build times around 24-30 minutes. Builds can also cause WebTiles latency while they run, and on busy commit days the build queue can keep the server compiling for hours.

With ccache, rebuilding the same revision is fast. The problem is that adjacent revisions change the install path flags, which invalidates ccache for almost everything.

## Test

I tested this on a live-like test server with WebTiles + dgamelaunch build options enabled.

I built three adjacent trunk revisions with a fresh `CCACHE_DIR` before the first build, then reused the same cache for the next two builds.

| Build | Commit | Time | ccache result |
|---|---|---:|---|
| Cold build | `3daca8ecd0` | 6064s / 101m04s | 0 / 299 hits |
| Next revision | `b832844d6f` | 399s / 6m39s | 294 / 299 hits |
| Next revision | `b15e5d74c6` | 387s / 6m27s | 294 / 299 hits |

The remaining 5 misses are expected, since the install-path-dependent objects still need to rebuild when those paths change.

### DIFF
--- a/crawl-ref/source/Makefile
+++ b/crawl-ref/source/Makefile
@@ -925,7 +925,7 @@ endif
 
 ifneq ($(strip $(SAVEDIR)),)
   ifneq ($(filter ~%,$(SAVEDIR))$(ANDROID),)
-    CFOTHERS_L += -DSAVE_DIR_PATH=\"$(SAVEDIR)\"
+    INSTALL_PATH_CFLAGS_L += -DSAVE_DIR_PATH=\"$(SAVEDIR)\"
     savedir_fp :=
     shareddir_fp :=
   else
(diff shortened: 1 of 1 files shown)


refracta:
Please also consider #4243.

gammafunk:
Merged, thanks for the PR! I've taken a look at #4243 and will probably merge that with a couple of tweaks later on.

PR #5254, opened 2026-05-11
New spell, Forge Soul Syphon

Hellmonk:
A level 5 necromancy/forgecraft spell that summons an immobile ally which abjures summons and drains living enemies in a radius around it. If the ally survives to time out "naturally," it will deal a large amount of negative energy damage to a random nearby enemy depending on how much draining and abjuring it got to do.

This commit also removes the anti-summon component of Yara's Violent Unravelling, a spell which some players consider "overloaded" in terms of effects. It will once again require actual buffs to be useful.

(Still needs balance testing and some tiles work)


Hellmonk:
Closing as I'm no longer convinced this is the best approach.

PR #5259, opened 2026-05-21
Coatl Monster

AmeliaGryffon:
Added a new monster for snake pits! 

The coatl is a powerful servant of Qazlal that uses air magic like mephitic cloud, silence, and air strike.


crawl-ref/source/dat/descript/monsters.txt
```diff
@@ -4094,3 +4094,8 @@ You hear the echoes of planets behind the deafening roar of stars as it
 approaches. It is the jaws of the abyss. It brings forth crawling chaos from
 within. It is the last sight you will ever see.
 %%%%
+coatl
+
+Holy serpent and servant to Qazlal. It utilizes it's power over the air to
```
geekosaur:
```suggestion
Holy serpent and servant to Qazlal. It utilizes its power over the air to
```

PR #5283, opened 2026-06-03
Don't spend a turn when aborting channeling.

### LINKED ISSUE #3506: Saying no to the prompt to hit an ally when channeling a spell waits a turn

webtiles

Version: 0.31

**DESCRIPTION:**

If you cast flame wave and an ally walks into it while you are channeling it with a god that doesn't like you hurting allies and say no to the prompt, the game waits a turn without channeling instead of just canceling channeling the spell.

**EXTRA FILES:**

Screenshot 2024-02-05 172547
Screenshot 2024-02-05 172607

Hellmonk:
When channeling stops due to the player saying no to a prompt, don't take time.
Closes #3506.

### DIFF
--- a/crawl-ref/source/main.cc
+++ b/crawl-ref/source/main.cc
@@ -1321,12 +1321,13 @@ static void _input()
         you.turn_is_over = false;
 #endif
 
+    if (you.turn_is_over && !handle_channelled_spell())
+        you.turn_is_over = false;
+
     if (you.turn_is_over)
     {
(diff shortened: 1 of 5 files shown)


CrawlOdds:
I wonder if it would make more sense for the player saying no to this prompt make nothing happen at all? Usually when you say no to a "do you really mean this" prompt there's no state change, so having it instantaneously cancel the spell seems a bit surprising.

Issue #3474, opened 2024-01-14
Shield slowdown should be listed in melee weapons


NotAJumbleOfNumbers:
It's already done for ranged weapons, so why not melee ones too?

PR #5285, opened 2026-06-06
Addressed issue #3474 by generalizing weapon encumbrance descriptions.

TrevorShelton:


CrawlOdds:
Thanks! Merged (with minor tweaks) in a2a5072.

PR #5287, opened 2026-06-06
Refactor passage of golubria

Hellmonk:
I think this code is very bad, but it is hard to handle the edge case of exactly one valid location for a portal that is also valid for the other portal.


CrawlOdds:
Merged with a bugfix and other minor changes in 1b8bb01

PR #5289, opened 2026-06-08
Replace many ways to get penance with forbidding items/actions.

CrawlOdds:
This updates most bad conducts (except attacking and killing certain kinds of creature) to make them utterly forbidden, rather than give penance. This means the forbidden thing can never be wielded/worn/evoked/read/quaffed/thrown/cast, so the previous acts that gave penance are impossible.

There are a few reasons for this:
* The rare cases that accepting penance or passively wielding a banned item are correct are rather unflavourful.
* Penance and piety loss are confusing, and the distinction is poorly explained (many warnings about this lie to the player).
* Codewise the current situation is quite fragile; warnings and actual penance have to be kept in sync in many places.

In doing this, there were a couple of somewhat-related changes that were sensible:
* In various menus, display useless items/spells - this was a bit inconsistent before, and it can be quite confusing when your stuff disappears.
* Add uselessness text to more items.
* Some refactoring to skill training/menu code, which I found somewhat hard to understand


CrawlOdds:
Largely merged in https://github.com/crawl/crawl/compare/92050278c1c0...1e03fe1c1a01, with some tidying.

The only functional change here that was NOT merged there is banning phantom mirroring hated monsters. Considerations on that were a) what about other ways of getting hated allies (charming, summoning scroll) and b) is it sufficiently explained/obvious which monsters your god hates. It might still be good to something in this direction, but for now I've left it.

PR #5293, opened 2026-06-11
fix spell key test

geekosaur:
Uppercase letters come before lowercase, not after.
And it's irrelevant anyway, because it ends up at in the loop's own `isaalpha()` check either way.

### DIFF
--- a/crawl-ref/source/adjust.cc
+++ b/crawl-ref/source/adjust.cc
@@ -146,11 +146,7 @@ static void _adjust_spell()
         // without spells from this menu.
         // XX this does not really work well with new menu code
         if (keyin == '?' || keyin == '*')
-        {
             keyin = list_spells(true, false, false, false, "adjust it to");
-            if (keyin < 'a' || keyin > 'Z')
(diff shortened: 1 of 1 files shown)


crawl-ref/source/adjust.cc
```diff
@@ -148,7 +148,7 @@ static void _adjust_spell()
         if (keyin == '?' || keyin == '*')
         {
             keyin = list_spells(true, false, false, false, "adjust it to");
-            if (keyin < 'a' || keyin > 'Z')
+            if (keyin < 'A' || keyin > 'z') // XXX exclude brackets and such? (isalpha())
                 continue;
```
CrawlOdds: This is a continue at the end of the loop body, which is surely a no-op; so I think this conditional doesn't matter at all, and we should just delete it.

geekosaur: Redesigned. Love crawlcode….

CrawlOdds:
Thanks! Merged in fc95cd1

PR #5295, opened 2026-06-13
Do not show breath status if player can't use it in current form

Predelnik:
The information about how many charges breath has seems irrelevant if we cannot use it.
Note that for dragon form breath status will still be shown due to second line of the condition.


CrawlOdds:
Thanks! Merged with a trivial formatting fix in 5612391

PR #5304, opened 2026-06-22
Fix missing space between prefix and text in menu_colour()

### LINKED ISSUE #5303: correct menu_colour regex not matching for "a stone"
**PLATFORM and OS or BROWSER:** local tiles 
**DCSS VERSION AND GAME INFORMATION:** crawl_tiles-0.35-a0-542

**DESCRIPTION:**
```menu_colour += darkgrey:^(un)?identified missile.*\b(stones?)\b```
I was trying this menu color setting to match stones (either "a stone" or "5 stones") and wasn't working as I was expecting against "a stone" (singular 1 stone)
After some testing I found that this regex is the only one that matches the "a stone"
```menu_colour += magenta:^(un)?identified missilestone```

Seems like in this case the space between the prefixes and the item name is not added.
I was looking at the code and found menu_colour() doesn't ensure or add an space between prefix and name when creating tmp_text. This is assuming prefix ends with a space or name starts with one.

https://github.com/crawl/crawl/blob/eed3790eb1321a95cb28589f234a3539f246931a/crawl-ref/source/menu.cc#L3491-L3496

Vjio:
menu_colour() initially didn't check if there was a space between prefix and text before jamming them together. this pr makes menu_colour() first check if there is at least one space between the two (and if not adds a space)

fixes #5303

### DIFF
--- a/crawl-ref/source/menu.cc
+++ b/crawl-ref/source/menu.cc
@@ -3490,7 +3490,14 @@ void Menu::webtiles_write_item(const MenuEntry* me) const
 
 int menu_colour(const string &text, const string &prefix, const string &tag, bool strict)
 {
-    const string tmp_text = prefix + text;
+    bool extra_space = false;
+    if ((!text.empty() && text.front() != ' ') 
+        && (!prefix.empty() && prefix.back() != ' '))
+    {
+        extra_space = true;
+    }
+
+    const string tmp_text = prefix + (extra_space ? " " : "") + text;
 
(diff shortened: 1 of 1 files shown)


CrawlOdds:
Thanks for the fixed! Merged in 7e1f3d5.

I don't think you're in the credits yet - what name would you like to be added under?

Vjio:
"Victor 'Vji' Davidescu" works. thank you

PR #5305, opened 2026-06-22
Rework lightning rod damage

CrawlOdds:
The current damage from lightning rod sheet lightning is quite counterintuitive; it does reduced damage to each enemy according to the number of tiles hit at the same distance from the player.

This is quite hard to think about, and it means that using the lightning rod to hit a large area is usually less effective than using it for single beam damage. It also has fiddly targetting shenanigans where it is good to decrease the number of squares hit.

We update this so that:
* The main beam does full damage every cast
* Squares off the main beam take half damage

This is intended to incentivise more use of AoE; as long as the player has sensible targets for the main beam they should want to move the beam in a wide arc to deal some damage to more targets. It also makes moving the beam for single target damage more effective, e.g. if the first target is dead.

Since this should be a buff for lightning rod, we nudge the single-beam damage numbers down to compensate.

Also make the displayed damage deterministic.

### DIFF
--- a/crawl-ref/source/dat/descript/items.txt
+++ b/crawl-ref/source/dat/descript/items.txt
@@ -932,8 +932,8 @@ lightning rod
 A rod that allows its wielder to fire unavoidable blasts of lightning.
 Consecutive evocations cause the discharge to be sustained, which increases the
 power level up to four times and allows the blast to be redirected to affect an
-area. The strength of the blast increases with Evocations skill and, for
-redirected blasts, decreases with the size of the affected area.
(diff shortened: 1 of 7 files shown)


CrawlOdds:
This is an alternative to https://github.com/crawl/crawl/pull/3352, which also makes lightning rod damage much simpler. I mildly prefer this version, because I think even the simplified arc length formulation is quite complex and leads to targetting fiddliness, and this *should* still lead to interesting decisions about where to target it and cool lightning sheets being sometimes optimal.

PR #5312, opened 2026-06-25
Rip out Xom's moods

CrawlOdds:
This removes Xom's mood and boredom, replacing the former with a roll per-effect on the same scale, and the latter with nothing at all. Xom's piety meter becomes a tensionometer. Xom still laughs at the player's foolishness, but this now has no mechanical effect.

This PR is not a good change on its own; it leaves Xom as extremely scummable. It is intended to be combined with a change to move Xom's strategic effects to an explore-based timer and put an end to Xomscumming.


Zireael07:
> It is intended to be combined with a change to move Xom's strategic effects to an explore-based timer and put an end to Xomscumming.

Why not make that other change in the same PR then?

PR #5320, opened 2026-06-29
Fix teleport closets

CrawlOdds:
This PR adds a check to explicitly search for teleport closets - regions the player can be teleported into and not out of.

Teleport closets are not rare - the typical game generates several, though most of them are very small and so teleports into them are rare. They arise in two main ways:
- Vaults which have made a mistake and failed to mask dangerous squares
- Vaults which do procedural generation that can generate isolated squares (e.g. in water)

The way we fix these is mask found teleport closets. This is preferable to vetoing their generation, because this would silently veto many vaults, only some of which are fixable.

We also support a mapstat mode -veto-closets, which vetos these closets instead of masking. Running this is useful for tracking down vaults which generate closets.

There are a few vaults which clearly deliberately generate closets, generating items for their escape. For these vaults, we add a tag "allow_tele_closets", which we apply to the instances we found.

We attach a CSV of vaults found by this mapstat, along with a diagnosis of whether the teleport closets are real and how they arise:
[closet_vaults.csv](https://github.com/user-attachments/files/29470158/closet_vaults.csv)

As well as the deliberate closets, there are a few cases where the masking is harmless but unnecessary, e.g. because of permanent clouds on the relevant squares.

### DIFF
--- a/crawl-ref/docs/develop/levels/syntax.txt
+++ b/crawl-ref/docs/develop/levels/syntax.txt
@@ -353,6 +353,12 @@ TAGS:     Tags go on a TAGS: line and are space-separated. You can have several
              have a vault that can be used more than once, use allow_dup to tell
              the dungeon builder that the vault can be reused. This is tracked
              separately for the abyss and the rest of the dungeon.
+          * "allow_tele_closets": Exempts the vault from teleport-closet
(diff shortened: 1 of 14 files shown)


CrawlOdds:
Merged

PR #5329, opened 2026-07-05
Give a description to items the player has detected but not seen.

Aliscans:
Let the "examine tile" function describe items which has been detected (for instance) by a gnoll.

Give the items a name and quantity, and let full_describe_square() describe items which aren't in a stash.

While the significance of detected items isn't that mysterious, I think it would be better if Crawl told people what the symbol meant.

### DIFF
--- a/crawl-ref/source/dat/descript/items.txt
+++ b/crawl-ref/source/dat/descript/items.txt
@@ -1873,3 +1873,7 @@ young poisoner's handbook
 An introductory book for the magic arts of poison. Readers are recommended to
 wear a pair of very thick, non-absorbent gloves.
 %%%%
+detected item
+
+You have detected an object here, but you do not know what it is.
+%%%%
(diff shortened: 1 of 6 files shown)


CrawlOdds:
This feature seems like a good idea to me, but I'm worried about the implementation - it makes detected items return true for defined() when they didn't before, and that function is used in a lot of places. I wonder if it would be better to leave them as-is and make the necessary exceptions for them in the places related to description?

Aliscans:
Crawl uses item_def::defined() a lot, but not so much for things in env.map_knowledge as that uses null pointers for its purposes.

I'll upload some changes which uses non-standard "is this item defined" tests for these items instead of setting item_def::defined(),  but I think they just make the code harder to understand.

CrawlOdds:
Thanks - merged in 9975744. I do slightly prefer the version that leaves the detected item undefined; we use that in at least one lua function that is deliberately excluding detected items because various methods on them will fail.

PR #5330, opened 2026-07-05
Describe things under the cursor on the level map in console.

Aliscans:
Describe the item/monster/terrain displayed under the cursor in the level map (only list terrain if the monster_item_view_features option is set for it), and show when there are other things in the square.

To make space, hide the list of stairs whenever something is described like this. This will normally only affect a small minority of cells. Add an "always show stair list" key.

...

My main reason for writing this was to make it less of a chore to find out what's in a gauntlet or necropolis after I arrive. I'm not familiar with tiles, so I don't know if this is an issue there.

I added an "always show stair list" command for the map screen because it's easier to say "press = to show stairs" than to say "move the cursor to somewhere with less stuff on it".


CrawlOdds:
Thanks, this is great - merged with tiny tweaks in 2149569 (and this inspired me to do the tiles version, that gauntlet and necropolis use case has been annoying me for years as well!)

PR #5331, opened 2026-07-05
Show the in-game colours in the monster help lookup.

Aliscans:
As things stood, monsters were set to be shown using the character used in the game, but with the default colour for that monster (if you set "mon_glyph += rat : white U", a rat would appear as a white U in-game but as a brown U in the help).

Make it reflect the preferences.


CrawlOdds:
Thanks for the fix!

PR #5334, opened 2026-07-09
Don't wait infinitely to regenerate HP/MP in &O

### LINKED ISSUE #4028: Wizmode exploretime crash
Locally compiled on Ubuntu 22.04.

Dungeon Crawl Stone Soup 0.32-a0-1845-gfed0f14cea (SDL Tiles)

**DESCRIPTION:**
Possible to crash in wizmode by setting XL+skills to 27, and then using &O to measure explore time. It doesn't do this every time I perform those actions, but has done it 3 times on characters which did perform those actions, with various species, classes, and mons on level. Save and crash files are attached.

**EXTRA FILES:**
[wizmode_crash.zip](https://github.com/user-attachments/files/17002968/wizmode_crash.zip)

code2828:
The wizmode version of autoexplore doesn't regen (in fact, it does nothing other than exploring), but start_explore is called without skipping autorest, causing the game to wait indefinitely trying to recover HP to full and crash upon reaching 2000 turns.

Fixes #4028.

(Duh, made a dumb mistake. force-pushed.)

### DIFF
--- a/crawl-ref/source/wiz-dgn.cc
+++ b/crawl-ref/source/wiz-dgn.cc
@@ -645,7 +645,7 @@ static int _debug_time_explore()
 {
     viewwindow();
     update_screen();
-    start_explore(false);
+    start_explore(false, true);
 
     unwind_var<int> es(Options.explore_stop, 0);
 


CrawlOdds:
Thanks for the fix!

PR #5335, opened 2026-07-09
Add a flag to friendly battlespheres if they can see a target.

Aliscans:
If you summon a battlesphere, it only fires if it can see a target you also see when you cast a spell, and if it can reach a spot where it can make a direct shot at it.

Add some text to mons_conditions_string() to show that a battlesphere has a target. This text appears in the (console-only) monster list and the character dump, and is available to Lua.

### Notes

As written, update_monster_pane() calls trigger_battlesphere() with the just_check parameter whenever there's a friendly battlesphere around. This is quite involved and could be cached, but I don't know if it would be worthwhile, and I note that calling update_monster_pane() less often would be another approach.

I changed what trigger_battlesphere() returns, but nothing looked at it before this.

This could be adapted for tiles, but I don't know how the tile-specific code works.

### DIFF
--- a/crawl-ref/source/mon-info.cc
+++ b/crawl-ref/source/mon-info.cc
@@ -2130,6 +2130,12 @@ static string _condition_string(int num, int count,
         return make_stringf("%d %s", num, word.c_str());
 }
 
+static bool _is_your_battlesphere_with_target(const monster_info &mi)
+{
+    return MONS_BATTLESPHERE == mi.type && MID_PLAYER == mi.summoner_id
+           && trigger_battlesphere(&you, true);
+}
+
 void mons_conditions_string(string& desc, const vector<monster_info>& mi,
                             int start, int count, bool equipment)
 {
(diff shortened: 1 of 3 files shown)


DracoOmega:
I must admit that while it's *probably* inconsequential in practice, I'm slightly dubious about repeatedly performing the sometimes fairly involved battlesphere aiming code when, say, the player moves a cursor around (in a fairly realistic scenario, it can perform pathfinding and fire tracers from ~20 different locations each time).

But I'm also a little unclear why it gets the distinction of a UI indicator for 'Can perform something useful from its present position' when allies in general do not (even some of those with similarly triggered effects). Is it just because this one is more technically feasible to do than many others?

Aliscans:
> I must admit that while it's _probably_ inconsequential in practice, I'm slightly dubious about repeatedly performing the sometimes fairly involved battlesphere aiming code when, say, the player moves a cursor around (in a fairly realistic scenario, it can perform pathfinding and fire tracers from ~20 different locations each time).

I agree that it's not ideal, but, by the same token, it isn't ideal to redraw the monster pane from scratch when moving a cursor around anyway. A c_message hook could change a relevant option, maybe, but that can be handled separately.

I could imagine rearranging the battlesphere code to trace paths from the monsters rather than the battlesphere's possible locations, but similarly I wonder if anyone would notice.

> 
> But I'm also a little unclear why it gets the distinction of a UI indicator for 'Can perform something useful from its present position' when allies in general do not (even some of those with similarly triggered effects). Is it just because this one is more technically feasible to do than many others?

If you move to where you can just see an enemy oklob and pelt it with spells, a summoned rat (say) will rush towards the enemy. A battlesphere will sit behind you until it times out. To make things worse, battlespheres don't follow verbal orders.

As such,  you often have to choose between moving or attacking. It's easier to make that choice if the game tells you when the battlesphere is already in a useful position.

PR #5342, opened 2026-07-17
Fix Mad Mage Mauler align

Aliscans:
_format_dbrand() didn't calculate the width of unicode characters accurately. padded_str() does, so let that handle the width.

This means that the two DESCR lines for the Mad Mage's Maulers are now aligned in the same way.


CrawlOdds:
Thanks for the fix!

PR #5345, opened 2026-07-22
Show owned counts for potions, scrolls, wand charges and throwables in shops

hatvik:
When looking through a shop, it is useful to know how
many of an item are already in your inventory. This adds that count to
the shop entry when it is non-zero.

For example:
potion of haste  (owned: 2)

The count is  shown for potions, scrolls, wand charges and throwables. It also respects
identification state, so it cannot reveal the type of an unknown item.

Tested with make catch2-tests, and manually in wizard mode with spawned shops.

### DIFF
--- a/crawl-ref/source/catch2-tests/test_shopping.cc
+++ b/crawl-ref/source/catch2-tests/test_shopping.cc
@@ -1,0 +1,67 @@
+#include "catch_amalgamated.hpp"
+
+#include "AppHdr.h"
+
+#include "items.h"
+#include "item-status-flag-type.h"
+#include "player.h"
+#include "shopping.h"
+
+#include "test_player_fixture.h"
+
+TEST_CASE_METHOD(MockPlayerYouTestsFixture,
+                 "Shop owned count sums matching scroll stacks", "[shopping]")
+{
+    item_def fear_scroll;
+    get_item_by_exact_name(fear_scroll, "scroll of fear");
+    fear_scroll.quantity = 3;
+    you.inv[0] = fear_scroll;
+
+    item_def identify_scroll;
+    get_item_by_exact_name(identify_scroll, "scroll of identify");
+    identify_scroll.quantity = 2;
+    you.inv[1] = identify_scroll;
+
+    item_def another_fear_scroll = fear_scroll;
+    another_fear_scroll.quantity = 2;
+    you.inv[2] = another_fear_scroll;
+
+    REQUIRE(shop_owned_consumable_count(fear_scroll) == 5);
+}
+
+TEST_CASE_METHOD(MockPlayerYouTestsFixture,
+                 "Shop owned count sums matching potion stacks", "[shopping]")
+{
+    item_def haste_potion;
+    get_item_by_exact_name(haste_potion, "potion of haste");
+    haste_potion.quantity = 2;
+    you.inv[0] = haste_potion;
+
+    REQUIRE(shop_owned_consumable_count(haste_potion) == 2);
+}
+
+TEST_CASE_METHOD(MockPlayerYouTestsFixture,
+                 "Shop owned count does not reveal unknown item types", "[shopping]")
+{
+    item_def inventory_scroll;
+    get_item_by_exact_name(inventory_scroll, "scroll of fear");
+    inventory_scroll.quantity = 3;
+    inventory_scroll.flags &= ~ISFLAG_IDENTIFIED;
+    you.inv[0] = inventory_scroll;
+
+    item_def shop_scroll;
+    get_item_by_exact_name(shop_scroll, "scroll of fear");
+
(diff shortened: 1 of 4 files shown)


crawl-ref/source/shopping.cc
```diff
@@ -862,6 +862,23 @@ static int _count_identical(const vector<item_def>& stock, const item_def& item)
...
+    if (item.base_type != OBJ_POTIONS && item.base_type != OBJ_SCROLLS)
+        return 0;
+
+    int count = 0;
+    for (const item_def& inv : you.inv)
+    {
+        if (inv.base_type == item.base_type && inv.sub_type == item.sub_type
+            && inv.is_identified() == item.is_identified())
+        {
+            count += inv.quantity;
```
CrawlOdds: Since we can't have two inventory items with the same consumable, we can just return here

crawl-ref/source/catch2-tests/test_shopping.cc
```diff
@@ -0,0 +1,67 @@
...
+
+#include "items.h"
+#include "item-status-flag-type.h"
+#include "player.h"
+#include "shopping.h"
+
+#include "test_player_fixture.h"
+
+TEST_CASE_METHOD(MockPlayerYouTestsFixture,
+                 "Shop owned count sums matching scroll stacks", "[shopping]")
```
CrawlOdds: This test sets up an impossible inventory (you can't have two stacks of fear scrolls) - I think it's better not to have this (or to make it a simple test like the potions one)

crawl-ref/source/shopping.cc
```diff
@@ -1021,7 +1038,11 @@ class ShopEntry : public InvEntry
         const string keystr = colour_to_str(keycol);
         const string itemstr =
             colour_to_str(menu_colour(text, item_prefix(*item, false), tag, false));
-        return make_stringf(" <%s>%c %c </%s><%s>%4d gold   %s%s</%s>",
+        const int owned = shop_owned_consumable_count(*item);
+        const string ownedstr = owned > 0
+            ? make_stringf("  <cyan>(owned: %d)</cyan>", owned)
```
CrawlOdds: I tend to think this would look better as matching the colour of the item (like we do for cost)

CrawlOdds (commented):
Thanks for the PR! I like the overall idea - I've left a few comments on the details.

I think for consistency this would be worth extending to wand charges and throwables before submitting

crawl-ref/source/shopping.cc
hatvik: Thanks, fixed

hatvik:
@CrawlOdds Thank you for the review! I added throwables and wand charges now

CrawlOdds:
Ah, actually I think we can replace everything except the wands with a single items_stack check, along the lines of

```
    for (const item_def& inv : you.inv)
    {
        if (items_stack(inv, item))
            return inv.quantity;
    }
```

This also adds in baubles, and anything else that later becomes stackable, which I think seems correct?

(Apologies for not spotting this the first time round)

crawl-ref/source/shopping.cc
```diff
@@ -1021,7 +1057,14 @@ class ShopEntry : public InvEntry
         const string keystr = colour_to_str(keycol);
         const string itemstr =
             colour_to_str(menu_colour(text, item_prefix(*item, false), tag, false));
-        return make_stringf(" <%s>%c %c </%s><%s>%4d gold   %s%s</%s>",
+        const int owned = shop_owned_consumable_count(*item);
+        const string ownedstr = owned > 0
+            ? item->base_type == OBJ_WANDS
+                ? make_stringf("  (owned: %d charge%s)", owned,
```
CrawlOdds: I think this should be one space not two for consistency with " (unknown)"

crawl-ref/source/shopping.h
```diff
@@ -26,6 +27,8 @@ int item_price(const item_def& item, const shop_struct& shop);
 // curse scrolls are worthless for everyone, most potions aren't worthless
 // for mummies, etcetera.
 bool is_worthless_consumable(const item_def &item);
+// Return how many matching potions or scrolls are currently in the inventory.
```
CrawlOdds: Now "matching items"

hatvik: Thanks, fixed

crawl-ref/source/shopping.cc
hatvik: Yes, agree

CrawlOdds:
Merged in a221df3 (with a tiny tweak to remove the wording "charges" for wands, since the count for wands shows a number without any extra word)

CrawlOdds:

hatvik:

Thanks!
You can add as: Eivin Hatvik

PR #5346, opened 2026-07-24
Mapstat improvements and dungeon connectivity checks

CrawlOdds:
In this PR, we:

- Improve mapstat, to make it log seeds with behaviour matching a seeded game. This is helpful for understanding dungeon generation.
- Add a new connectivity check, that aims to prevent areas the player will never see at all without mapping etc
- Move stair culling before connectivity checks that rely on those stairs

For the new connectivity check, the veto rate is rather low - approximately 0.03% of all levels, or one every 20 or so games. In a random sample of 10 of these, 8 seem good and the other 2 are acceptable false positive from opaque vaults.

Here are a few examples of things it vetoes, which are roughly representative of the distribution:

Tiny disconnected area:
image

Large disconnected area:
image

Disconnected Zig:
image

False positive - appears disconnected to level gen due to an opaque vault:
image

### DIFF
--- a/crawl-ref/source/dbg-maps.cc
+++ b/crawl-ref/source/dbg-maps.cc
@@ -13,8 +13,11 @@
 #include "dbg-objstat.h"
 #include "dungeon.h"
 #include "env.h"
+#include "files.h"
 #include "initfile.h"
+#include "item-prop.h"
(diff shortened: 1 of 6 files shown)


CrawlOdds:
Merged

PR #5351, opened 2026-07-28
Hide tile-only keys in the level map help outside tile mode.

Aliscans:
The [ and ] keys had been included in the help for the level map command with "(tiles only)" in their description. Drop this, and only show the text in tile modes.

This seems like an obvious improvement to me, but there may be a reason why it wasn't done years ago.


CrawlOdds:
Thanks for the fix!

PR #5354, opened 2026-08-01
Add an `explore_stop` option to stop at traps when they enter your LOS

Zigmound:
This gives players the option to automatically stop the first time a trap is encountered while autoexploring. It is turned on by adding `explore_stop += traps` to the rc file.

### DIFF
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -916,7 +916,7 @@ explore_item_greed = 10
         further away, if negative) when deciding whether to visit a square with
         an item or explore a new square.
 
-explore_stop  = items,stairs,shops,altars,portals,branches,runed_doors,traps
+explore_stop  = items,stairs,shops,altars,portals,branches,runed_doors
 explore_stop += greedy_pickup_smart,greedy_visited_item_stack
         (List option)
         Explore will stop for one of these conditions. Whatever you
--- a/crawl-ref/docs/options_guide.txt
+++ b/crawl-ref/docs/options_guide.txt
@@ -916,7 +916,7 @@ explore_item_greed = 10
         further away, if negative) when deciding whether to visit a square with
         an item or explore a new square.
 
-explore_stop  = items,stairs,shops,altars,portals,branches,runed_doors
+explore_stop  = items,stairs,shops,altars,portals,branches,runed_doors,traps
 explore_stop += greedy_pickup_smart,greedy_visited_item_stack
         (List option)
         Explore will stop for one of these conditions. Whatever you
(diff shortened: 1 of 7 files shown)


CrawlOdds:
A couple of thoughts on this:

1) I think we should probably announce the reason for stopping in explore_discoveries::stop_explore, like we do for other stop reasons
2) I wonder if trap_is_safe is the most useful rule here... perhaps the option would be better stopping only for traps which are potentially dangerous from afar? What do you think?

crawl-ref/docs/options_guide.txt
```diff
@@ -916,7 +916,7 @@ explore_item_greed = 10
         further away, if negative) when deciding whether to visit a square with
         an item or explore a new square.
 
-explore_stop  = items,stairs,shops,altars,portals,branches,runed_doors
+explore_stop  = items,stairs,shops,altars,portals,branches,runed_doors,traps
```
CrawlOdds: This line is the default, so shouldn't be updated here

Zigmound: I've added code to announce traps when found, and removed "traps" from the default line in options_guide.txt. (Sorry, I didn't intend that to be the effect.)

Zigmound: Regarding the second comment, I'm looking at the list of traps in
`dungeon-feature-type.h`, and the only types of traps that strike me as
ignorable from a distance are `DNGN_TRAP_SHAFT`, and `DNGN_TRAP_WEB`. Do you agree?  If so, I could add conditions to ignore those.

Zigmound: Do you think it is okay to keep the `!trap_is_safe(feat)` condition? My only
goal was to allow net-immune characters to ignore `DNGN_TRAP_NET`s.

CrawlOdds: Yeah I think that's probably fine... I think the other way would be a new category of "traps that are safe unless you step in them", including teleportation traps. But I'm not sure that's better than the simple thing you've gone with.

CrawlOdds: > I've added code to announce traps when found, and removed "traps" from the default line in options_guide.txt. (Sorry, I didn't intend that to be the effect.)

(To be clear this line didn't change the default, it's just that the example line in the documentation is generally supposed to correspond to the default)

Zigmound: I added a `trap_is_safe_from_afar` function. Maybe this will be better than adding conditions in `explore_discoveries::found_feature` in case a new type of trap is added in the future. Hopefully the presence of this function will help alert the developer that the logic may need updating.

PR #5365, opened 2026-08-06
fix: Blorkula mitosis bug (Nnyef)

RypoFalem:
### Note: I reached out to .deekay in the roguelikes discord a few days ago to ask how they would like to be credited. They haven't responded to me. Nnyef is the online crawl account name they were using when they encountered the bug.

---

If Blorkula the Orcula split into bats and one of the bats was shafted, that vampire bat would turn into a copy of Blorkula on another floor complete with duplicated gear!

Make Blorkula's bats shaft immune to prevent this.

### DIFF
--- a/crawl-ref/source/monster.cc
+++ b/crawl-ref/source/monster.cc
@@ -5769,8 +5769,8 @@ bool monster::do_shaft()
     if (!is_valid_shaft_level())
         return false;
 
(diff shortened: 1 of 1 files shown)


pooka109:
Applied; thanks.

(It's been over a month; if Nnyef does respond to you at some point, let us know.)

PR #5366, opened 2026-08-08
i18n: HUD captions

powerbf:
This pull request is intended to be the first in a series that will ultimately bring full internationalisation (i18n) to Crawl (as discussed on #crawl-dev).

The solution will be based on that used in the Alphabet Soup project, but with some modifications based on lessons learned in the course of that project.

This PR handles a fairly simple use case - that of translating the HUD captions - but serves to bring in some of the required infrastructure.

### From the translators' POV:
- translations will be in plain text (UTF-8) files under dat/translate/{langcode}
- these files will use the same dbm format used in dat/descript and dat/database
- this has the advantage that the game will automatically detect any changes on restart, and there's no need to compile the translations into some intermediate format like the gettext .po format
- the keys for the entries will be the English strings to be translated rather than internal ids. This approach was chosen in order to avoid the massive amount of code changes which would be required to replace every string in the code with an identifier.
- Since there's no need to translate English into English, the "English" versions of the files in the top-level directory (dat/translate) will not be used by the game, and will contain keys only (no values). They will serve as templates for the translators to fill in and place in the relevant sub-directory.

### From the programmers' POV:
- in most cases, translation will be achieved by calling localise(str)
- wherever possible, this call will be embedded at a low level in functions like mprf, in order to minimise code changes
- there will also be a function localise_contextual for some special cases (see below)

### A note about regex
This solution will make heavy use of regular expressions (regex). Unfortunately, Crawl builds with different regex implementations based on OS - PCRE on Windows, POSIX otherwise. This is annoying because they are not functionally identical and use somewhat different syntax. Also, the POSIX implementation is notorious for performance issues. Ideally, we would standardise on PCRE across all OSes, but it's not a mandatory requirement. Also, while PCRE supports UTF-8, Crawl's fork of PCRE builds with UTF-8 support disabled. We will need to change that.

## More Details

There are a number of challenges when translating strings for an application. I will go through some of the main ones and show how they will be addressed.

### The problem of context
Sometimes the translation of a particular string will depend on context. For example, the word "fire" as in flames would likely have a different translation to the command "fire" meaning to shoot.

This will be addressed by allowing an optional context to be provided for translations. For the translator, it would look like this (German example):
```
%%%%
fire
Feuer
%%%%
{cmd}fire
Schießen
%%%%
```
In this case, the default translation of "fire" is the German word meaning fire as in flames, but we provide a specific contextual translation for the command "fire" (as in shoot). In the case of the command, the code would have to call:
```
localise_contextual("cmd", "fire");
```

The context mechanism is a key part of the solution, and allows us to solve a number of other problems.

### Grammatical Gender

In many languages, nouns have gender. This often affects the associated articles (the/a) and adjectives.

Handling the articles is fairly straightforward. Example (French):
```
%%%%
the bat
la chauve-souris
%%%%
the butterfly
le papillon
%%%%
```
We can even handle exceptions:
```
%%%%
the orc
l'orc
%%%%
```

Of course, there are situations where we would want these nouns with the indefinite article (a/an) or no article. It would be tedious to manually define all the possible forms. For this reason there will be the capability to generate the other forms using regular expression rules. Any exceptions to the rules can be explicitly defined (explicit definitions always override regex-generated translations).
Simplified example (French):
```
# generate indefinite form from definite form
%%%%
PREGENERATE:1

SELECT_BY_KEY:/^the /

# transform key (English)
KEY:/^the /a /

# transform value (French)
/^le /un /
/^la /une /
%%%%
```

But then how do we handle adjectives? By setting a context, indicating the noun gender, in the translation of the noun, and then translating the adjective with that context.
```
%%%%
the bat
{f}la chauve-souris
%%%%
the butterfly
{m}le papillon
%%%%


%%%%
{m}charmed
charmé
%%%%
{f}charmed
charmée
%%%%
```
Note that the contexts {f} and {m} are not hardcoded. The translator is free to use whatever they want.

### Adjective placement

French also has another challenge where some adjectives come before the noun and some after.
We indicate this by putting a space after adjectives that come before the noun, and before adjectives that come after the noun.
This requires quoting the strings so that the whitespace doesn't get stripped.
```
%%%%
"{m}charmed "
" charmé"
%%%%
"{f}charmed "
" charmée"
%%%%
```

### Grammatical Case
In some languages, like German, the form of a noun and its associated articles and adjectives is affected by grammatical case, which is related to the role that the noun plays in the sentence. For example, in German, if a monster is hitting you then it's in the nominative case, but if you're hitting the monster then its in the accusative case. This can also be solved with the context mechanism:
```
%%%%
# nominative
the orc
der Ork
%%%%
# accusative
{acc}the orc
den Ork
%%%%

%%%%
@the_monster@ hits you
@the_monster@ trifft dich
%%%%
You hit @the_monster@ 
Du triffst {acc}@the_monster@ 
%%%%
```

Note that I have deliberately not used a context for the nominative case - this makes it the default.
And, as per the gender example, the context {acc} is not hardcoded. The translator can use whatever they want.

## Limitations/Outside the inital scope
- ~~some languages have more than just singular and plural. For example, Russian uses the "dual" form for 2-4 (and 12-14). I implemented a solution to this in Alphabet Soup, but it was rather clunky. I will have to rethink this.~~ Edit: This is implemented, so now in scope.
- character set: The local tiles build uses a font that only has Latin, Greek and Cyrillic scripts (and Arabic, but that has another issue... see below), which means other scripts like Chinese/Japanese/Korean are not currently supported by local tiles. For the console version, it depends on system fonts. I'm not sure what the story with the fonts in web tiles is.
- Text direction: Some scripts like Arabic and Hebrew are written right to left. I don't have a solution for this. Handling this would require rejigging the layout of the HUD and other components.

### DIFF
--- a/crawl-ref/source/MSVC/crawl.vcxproj
+++ b/crawl-ref/source/MSVC/crawl.vcxproj
@@ -463,6 +463,7 @@ python.exe util/gen-all.py</Command>
       <PrecompiledHeader>Use</PrecompiledHeader>
     </ClCompile>
     <ClCompile Include="..\loading-screen.cc" />
+    <ClCompile Include="..\localise.cc" />
     <ClCompile Include="..\los.cc" />
     <ClCompile Include="..\los-def.cc" />
     <ClCompile Include="..\losparam.cc" />
@@ -807,6 +808,7 @@ python.exe util/gen-all.py</Command>
     <ClInclude Include="..\libutil.h" />
     <ClInclude Include="..\libw32c.h" />
     <ClInclude Include="..\loading-screen.h" />
+    <ClInclude Include="..\localise.h" />
     <ClInclude Include="..\lookup-help.h" />
     <ClInclude Include="..\los-def.h" />
     <ClInclude Include="..\los-type.h" />
--- a/crawl-ref/source/MSVC/crawl.vcxproj.filters
+++ b/crawl-ref/source/MSVC/crawl.vcxproj.filters
@@ -639,6 +639,9 @@
     <ClCompile Include="..\l-spells.cc">
       <Filter>cc</Filter>
     </ClCompile>
+    <ClCompile Include="..\localise.cc">
+      <Filter>cc</Filter>
+    </ClCompile>
     <ClCompile Include="..\losparam.cc">
       <Filter>cc</Filter>
     </ClCompile>
@@ -1477,6 +1480,9 @@
       <Filter>h</Filter>
     </ClInclude>
     <ClInclude Include="..\loading-screen.h">
+      <Filter>h</Filter>
+    </ClInclude>
+    <ClInclude Include="..\localise.h">
       <Filter>h</Filter>
     </ClInclude>
     <ClInclude Include="..\lookup-help.h">
--- a/crawl-ref/source/Makefile
+++ b/crawl-ref/source/Makefile
@@ -1596,6 +1596,7 @@ endif
 	for LANG in $(LANGUAGES); \
 		do $(COPY) dat/descript/$$LANG/*.txt $(datadir_fp)/dat/descript/$$LANG; \
 	done
+	$(COPY_R) dat/translate $(datadir_fp)/dat/
 	mkdir -p $(datadir_fp)/dat/dist_bones
 	$(COPY) dat/dist_bones/* $(datadir_fp)/dat/dist_bones/
 	$(COPY) ../docs/*.txt $(datadir_fp)/docs/
--- a/crawl-ref/source/Makefile.obj
+++ b/crawl-ref/source/Makefile.obj
@@ -127,6 +127,7 @@ lang-fake.o \
 lev-pand.o \
 libutil.o \
 loading-screen.o \
+localise.o \
 lookup-help.o \
 los.o \
 los-def.o \
--- a/crawl-ref/source/android-project/jni/src/Android.mk
+++ b/crawl-ref/source/android-project/jni/src/Android.mk
@@ -146,6 +146,7 @@ LOCAL_SRC_FILES := $(SDL_PATH)/src/main/android/SDL_android_main.c \
     $(CRAWL_PATH)/lev-pand.cc \
     $(CRAWL_PATH)/libutil.cc \
     $(CRAWL_PATH)/loading-screen.cc \
+    $(CRAWL_PATH)/localise.cc \
     $(CRAWL_PATH)/lookup-help.cc \
     $(CRAWL_PATH)/los.cc \
     $(CRAWL_PATH)/los-def.cc \
--- a/crawl-ref/source/cio.cc
+++ b/crawl-ref/source/cio.cc
@@ -10,6 +10,7 @@
 #include <queue>
 
 #include "libutil.h"
+#include "localise.h"
 #include "macro.h"
 #include "message.h"
 #include "options.h"
@@ -241,6 +242,7 @@ void nowrap_eol_cprintf(const char *s, ...)
     va_list args;
     va_start(args, s);
     string buf = vmake_stringf(s, args);
+    buf = localise(buf);
     va_end(args);
 
     cprintf("%s", chop_string(buf, max(wrapcol + 1 - wherex(), 0), false).c_str());
@@ -326,6 +328,7 @@ void wrapcprintf(int wrapcol, const char *s, ...)
     va_list args;
     va_start(args, s);
     string buf = vmake_stringf(s, args);
+    buf = localise(buf);
     va_end(args);
     wrapcprint_skipping(0, wrapcol, buf);
 }
@@ -345,6 +348,7 @@ void wrapcprintf(const char *s, ...)
     va_list args;
     va_start(args, s);
     string buf = vmake_stringf(s, args);
+    buf = localise(buf);
     va_end(args);
     wrapcprint_skipping(0, cgetsize(get_cursor_region()).x, buf);
 }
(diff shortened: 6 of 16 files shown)


perryprog:
Two initial thoughts:

- How does this deal with translation key invalidation? Is there a way to automatically update translation keys for when a string is changed in English? 
- This doesn't come up in this PR so it's not too critical, but relying on regexes like you suggest feels a little weird to me, though I'm also not sure of a better solution. Generally it's better to avoid string interpolation for i18n as much as possible, as that allows things like differing word order to be handled by the translator, but obviously that isn't really possible here. (IMO it's better to repeat yourself translation side rather than trying to do things the instinctually-DRY way.) I'm curious what other games with things like an attack log do to handle i18n; I'm not sure there's many examples. (CDDA maybe? I know they use gettext but I've never looked into how they handle stuff like attack messages.)

powerbf:
> How does this deal with translation key invalidation? Is there a way to automatically update translation keys for when a string is changed in English?

This will be a challenge. For Alphabet Soup, I have a python script that extracts all the strings from the source code. There are lots of special cases though, so it's kind of messy.

powerbf:
There are two ways that I propose to use regexes. 

The first way is to generate inflected forms for nouns, adjectives, etc. When I initially started with Alphabet Soup, I was defining all these forms explicitly, but it was a maintenance nightmare. For example, for German nouns, I had:
- definite nominative
- indefinite nominative
- no article nominative
- definite accusative
- indefinite accusative
- no article accusative
- definite dative
- indefinite dative
- no article dative
- definite genitive
- indefinite genitive
- no article genitive

And that was just for singular. There were a bunch of plural forms as well.

powerbf:
The second way I propose to use regexes is to match a finished message to a parameterised string provided by the translator.

I didn't do this in Alphabet Soup - it requires the correctly-structured parameterised string to be passed in by the code. The result was a large amount of code changes. Many of the format strings used for mprf, for example, are structured in a way that only works for English, so they had to be changed.

perryprog:
Hmm. It is good to know that worked well for German, but what about languages that don't decline regularly? In English for instance you can have identical plural and singular forms, or other irregular plurals like child/children, foot/feet, men/man and so on. So unless I'm misunderstanding how exactly this looks translation-side, I'm not sure how that would be done. (Another example is Polish which has famously complicated pluralization rules.)

---

For some context on what some i18n systems (read: the two that I'm familiar with) do to solve this, just for anyone reading who's unfamiliar:

gettext handles this by having you pass the relevant number to one of their plural-specific functions (e.g., `ngettext`) and the translators add a header to the PO file of the form
```
Plural-Forms: nplurals=X; plural=Y;
```
where X is the number of plural forms possible and Y is a C expression that returns an integer for what "type" of plural it is (for English and German it'd be `n != 1` while Irish which has a dual plural would be `n==1 ? 0 : n==2 ? 1 : 2;`) You then provide a translation string for each possible case (e.g., for German, assuming this is for the inventory label or something):

```
msgid "a throwing net"
msgid_plural "%d throwing nets"
msgstr[0] "case 0 translation goes here"
msgstr[1] "case 1 translation goes here"
```

What Fluent does is it has you use a DSL for each translation string to break out the different cases. For English this usually looks like
```
throwing-net-inventory = {$count ->
    [one] a throwing net
    *[other] {$count} throwing nets
}
```
where "one" and "other" are defined by the [CLDR](https://cldr.unicode.org/index/cldr-spec/plural-rules) and the asterisk denotes the default case. Other languages could have more cases for that message ID, such as `[two]` for the dual plural.

powerbf:
> Hmm. It is good to know that worked well for German, but what about languages that don't decline regularly? In English for instance you can have identical plural and singular forms, or other irregular plurals like child/children, foot/feet, men/man and so on.

I didn't use regex to generate plurals from singular in Alphabet Soup, but it's feasible (edit: by that I mean, the current Alphabet Soup implementation would handle it). As with the generation of other forms, we can define a regex rule for the regular cases, and explicitly define the exceptions. For German, we would probably always explicitly define all the plurals because German is pretty much all child/children, foot/feet, men/man.

In regards to dual, etc. I propose to leave that out of scope for the initial implementation, but thinking about it, we could extend the context mechanism to handle this. Allow the translator to define a rule that says, for example:
n == 2 -> {dual}.

powerbf:
BTW, I had a look at Fluent last night after you mentioned it. It's a very interesting project. Are you involved with that?

perryprog:
Haha, no, I just worked on a large project with it (34k lines across all its .ftl files) and initially got very annoyed at how it did translation keys compared to gettext. (It is objectively much more annoying as a developer!) Then when I went to research why they designed it that way I was very much convinced of its advantages, and then much more annoyed at the way gettext implemented things (most of which is just a result of it being old and having backwards compatibility to worry about).

powerbf:
I was using gettext in the initial stages of the Alphabet Soup project, but I was not very impressed by it.

For one thing, you have to compile your translations into .po files, which is annoying when you're actively working on translations and want to test as you go.

The way it handles multiple plural forms (dual, etc.) is also less than ideal: msgstr[0], msgstr[1] is not very intuitive. The way Fluent does it with named forms is much nicer.

Edit: Not very intuitive for the translators, I mean. They are likely not going to be programmers, and so will be unfamiliar with array indexing.

powerbf:
@perryprog I have implemented handling of different plural forms for different numbers and tested with Russian which has a distinct plural form for numbers ending in 2-4 (but not 12-14).

This is how it is specified:
https://github.com/powerbf/alphabet-soup/blob/i18n-master/crawl-ref/source/dat/translate/ru/test.txt

So this is now in scope.

Issue #5371, opened 2026-08-10
Sound conditional in `initfile.cc` has been broken for ~3 years


geekosaur:
**PLATFORM and OS or BROWSER:**
All platforms

**DCSS VERSION AND GAME INFORMATION:**
All versions since ffccbfb393

**DESCRIPTION:**
There is a conditional for sound support in `initfile.cc`, in 0.32-a0-6604-gf98660601d it's line 3494.
```c
#if !defined(USE_SOUND) || !defined(WINMM_PLAY_SOUNDS) || !defined(SOUND_PLAY_COMMAND) || !defined(USE_SDL)
```
This requires all four macros to be defined, but only one of the three later ones is needed (Windows console, POSIX console, SDL/tiles). It should be
```c
#if !defined(USE_SOUND) || !(defined(WINMM_PLAY_SOUNDS) || defined(SOUND_PLAY_COMMAND) || defined(USE_SDL))
```

geekosaur:
Alternately and probably more clearly,
```c
#if !(defined(USE_SOUND) && (defined(WINMM_PLAY_SOUNDS) || defined(SOUND_PLAY_COMMAND) || defined(USE_SDL)))
```

nerdcorenet:
I can confirm that this fixed the issue which I experienced, and sound works in POSIX console version after this change.

CrawlOdds:
Thanks, fixed by merging #5372

PR #5372, opened 2026-08-10
fix sound conditional in initfile.cc

geekosaur:
As currently written, all four defines must exist. But three of them are intended to be platform-specific and it's not really expected that more than one of them be defined.


geekosaur:
I guess I don't have permission to link #5371 ?

Issue #3006, opened 2023-02-20
Pain Bond+Poison dart one-shot death yaks 


brianfaires:
**PLATFORM and OS or BROWSER:**
local tiles, Win10 version 22H2
DCSS 0.30-a0-788-ga5e9a8b7f5

**DESCRIPTION:**
Worshipping Uskayaw; ~5 yaks + 2 death yaks got paralyzed. I stabbed 2 of them, Pain Bond kicked in and some of the paralyzes wore off. Everything was full HP, no poison, no debuffs (except 1-2 of them were still paralyzed).
I threw 2 poison darts, then 2 death yaks and 2 yaks died from close to full HP. I had +10 slaying, but this seems like a bug. Idk how to reproduce or test my game state. Just figured I'd report it.

Message history:
You throw a poisoned dart. The poisoned dart hits the death yak. The death yak is poisoned.
The death yak is lightly wounded.
You kill the yak! You kill the death yak! You kill the yak! You kill the death yak!

[Tester.txt](https://github.com/crawl/crawl/files/10788247/Tester.txt)
[Tester.zip](https://github.com/crawl/crawl/files/10788249/Tester.zip)

PleasingFungus:
Repro'd this once, but my next few attempts didn't succeed. Very odd.

PR #5376, opened 2026-08-13
Prevent pain bond dealing infinite damage

CrawlOdds:
In sufficiently tight clusters of monsters (e.g. a square), pain bond's total damage diverged, so that a 1-damage attack could continue radiating until at least one monster died. This was a bit silly. More generally, the tightness of the clustering of monsters affected pain bond's damage to a fairly unreasonable and surprising extent. For example, with n mutually-adjacent enemies, the expected damager multiplier on pain bond was 1,2.5,25,infinity for n=1,2,3,4.

Also, pain bond had a separate probably-bug whereby the subsequent monsters in a single call to radiate took less damage, making the order of iteration over monsters important.

Here, we change pain bond to effect each enemy only once (and not to damage the original target at all), and to deal the same damage to each enemy in a single pass. It still radiates to hit enemies beyond the original radius. We up the numbers a little bit to balance this; pain bond will be weaker against tight clusters of monsters and stronger against loose ones.

### DIFF
--- a/crawl-ref/source/mon-util.cc
+++ b/crawl-ref/source/mon-util.cc
@@ -5548,42 +5548,53 @@ int max_mons_charge(monster_type m)
     }
 }
 
-// Deal out damage to nearby pain-bonded monsters based on the distance between them.
-void radiate_pain_bond(const monster& mon, int damage, const monster* original_target)
+void radiate_pain_bond(const monster& mon, int damage)
 {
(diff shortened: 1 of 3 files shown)


yrdzrfxndfvh:
was this the culprit behind #3006 ?

CrawlOdds:
> was this the culprit behind #3006 ?

Yes it was, forgot to mention that in the commit

PR #5385, opened 2026-08-23
Improve some spell QoL

CrawlOdds:
This adds three things:
- A flag for "this spell only does damage", which prevents autotargeting warded monsters
- A flag for "this spell needs a target", making spells not suggest Z-casting when it won't work
- Resists and defenses checked are added to spell descriptions (deriving this from zap information where possible)


CrawlOdds:
Merged (incorporating review comments from DracoOmega via another channel)

PR #5389, opened 2026-08-26
Spelling: Change AmE Duelist to AusE Duellist

powerbf:
In Australian (and British, etc.) English, duel follows the same rule as other words ending in -el where the l is doubled when adding a suffix (e.g. traveller, cancelled, modelling - in contrast to AmE traveler, canceled, modeling).

Here is a photo of the relevant entry from my Pocket Macquarie Dictionary (Macquarie is the standard Australian dictionary, equivalent to Oxford in the UK or Mirriam-Webster in the US):

duellist2


pooka109:
Applied; thanks.

PR #5390, opened 2026-08-26
Feat - Add android morgue stats

hatvik:
Adds a Morgue stats screen that reads completed morgues and the local logfile.

Tests: Android unit tests and the buildTest APK build pass locally.

UI:

Screenshot_20260826_164508_Dungeon Crawl Stone Soup
Screenshot_20260826_164513_Dungeon Crawl Stone Soup
Screenshot_20260826_164449_Dungeon Crawl Stone Soup

### DIFF
--- a/crawl-ref/source/android-project/app/build.gradle.in
+++ b/crawl-ref/source/android-project/app/build.gradle.in
@@ -29,6 +29,8 @@ android {
         debug {
         }
         buildTest {
+            applicationIdSuffix ".morguestatsdev"
+            signingConfig signingConfigs.debug
             ndk {
                 //noinspection ChromeOsAbiSupport
                 abiFilters "arm64-v8a"
--- a/crawl-ref/source/android-project/app/src/main/AndroidManifest.xml
+++ b/crawl-ref/source/android-project/app/src/main/AndroidManifest.xml
@@ -50,6 +50,10 @@
             android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
             android:windowSoftInputMode="adjustResize" />
         <activity
+            android:name=".DCSSMorgueStatsActivity"
+            android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
+            android:windowSoftInputMode="adjustResize" />
+        <activity
             android:name=".DCSSMods"
             android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
             android:windowSoftInputMode="adjustResize" />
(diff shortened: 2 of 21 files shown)


RypoFalem:
Looks cool!

Two notes: 

Is this something that needs to be platform-specific to android? 
---
 Can most of this not be implemented in c++ so that only a small amount of display code would be in java?

I think it would be much better to parse `score` files rather than morgues.
---
Morgues are configurable (I think you can leave out entire sections!),  aren't meant to be parsed by a machine, and changes are made to the morgue output without consideration for how a machine would parse it. The `score` has all the information you need, is meant to be machine-parsable, and changes to the structure should be minimal with machine-parsing in mind. They are also much smaller than morgues so should be able to be much faster to parse than morgues! (One player has 100k online games!)

In modern `score` files, every game is summarized on a single line with `key=value` pairs separated by `:`s.  `::` is used for colon literals so if a value is `D:6`, then `D::6` is written. Fields are sometimes omitted if in a default state (if you have 0 piety, it's won't write `piety=0`). That should be much easier to parse! The file is located in the `saves` directory by default and there are separate files for seeded/sprint games. There's already code to help parse it, see `hiscores_print_all()` in `highscores.cc` for an example of loading and parsing the `score` file.

We may wish to compile with `DGL_EXTENDED_LOGFILES` if we want access to death message strings which are usually excluded.

For reference, these are available by parsing the `score` file. Note that the keys might be different from the variable names here.
https://github.com/crawl/crawl/blob/0549b1e323f280d89c5cb1d326614f583922a91e/crawl-ref/source/hiscores.h#L77-L150

hatvik:
> Looks cool!
>
> Two notes:
> 
> ## Is this something that needs to be platform-specific to android?
> Can most of this not be implemented in c++ so that only a small amount of display code would be in java?

Good point. No, it does not need to be android only. It makes more sense to do as you suggest. 

I will try to change this.

> 
> ## I think it would be much better to parse `score` files rather than morgues.
> Morgues are configurable (I think you can leave out entire sections!), aren't meant to be parsed by a machine, and changes are made to the morgue output without consideration for how a machine would parse it. The `score` has all the information you need, is meant to be machine-parsable, and changes to the structure should be minimal with machine-parsing in mind. They are also much smaller than morgues so should be able to be much faster to parse than morgues! (One player has 100k online games!)
> 
> In modern `score` files, every game is summarized on a single line with `key=value` pairs separated by `:`s. `::` is used for colon literals so if a value is `D:6`, then `D::6` is written. Fields are sometimes omitted if in a default state (if you have 0 piety, it's won't write `piety=0`). That should be much easier to parse! The file is located in the `saves` directory by default and there are separate files for seeded/sprint games. There's already code to help parse it, see `hiscores_print_all()` in `highscores.cc` for an example of loading and parsing the `score` file.
> 
> We may wish to compile with `DGL_EXTENDED_LOGFILES` if we want access to death message strings which are usually excluded.
> 
> For reference, these are available by parsing the `score` file. Note that the keys might be different from the variable names here.
> 
> https://github.com/crawl/crawl/blob/0549b1e323f280d89c5cb1d326614f583922a91e/crawl-ref/source/hiscores.h#L77-L150

I wasn't aware of this and this is much better indeed. I will also take a look at this. 



Thank you for the feedback!

PR #5391, opened 2026-08-27
Allow capital letters in an "escape from constriction" message.

Aliscans:
Stop turning "Erica" into "erica" for messages such as "You blink free of Erica."

`actor::end_constriction()` no longer includes capital letters in the constant strings, so it doesn't need to remove them again.


pooka109:
Applied; thanks.

PR #5397, opened 2026-08-29
fix: clarify cloud disappearance description

hatvik:
The previous sentence was a bit hard to read.

- Clarify the generic cloud disappearance description.
- Use direct wording and commas around the conditional phrase


DracoOmega:
Regardless of opinions about phrasing, the new wording is semantically different and factually incorrect.

hatvik:
Care to elaborate what is incorrect?

pooka109:
Not the one you asked the question, but...

You changed it from "These clouds, when made by the user, will disappear when outside the user's sight." to "These clouds will disappear when outside the user's sight." The part you dropped is important.

PR #5402, opened 2026-08-31
Spell description: add showing nondamage effects

SlackingSleeper:
This adds a place to display variable effects to the spell description.
For Passwall and Ozocubu's Armour this can show the bonus AC.
For Summon Imp it can show the imps weapon quality.
For Borgnjor it could show the potential loss of max hp.
For Blink it could show the average or range of -Blink.

### DIFF
--- a/crawl-ref/source/describe.cc
+++ b/crawl-ref/source/describe.cc
@@ -4398,6 +4398,7 @@ static string _player_spell_stats(const spell_type spell)
         failure = failure_rate_to_string(raw_spell_fail(spell));
     description += make_stringf("        Fail: %s", failure.c_str());
 
+    const string effect_string = spell_effect_string(spell);
     const string damage_string = spell_damage_string(spell);
     const string max_dam_string = spell_max_damage_string(spell);
     const int acc = spell_acc(spell);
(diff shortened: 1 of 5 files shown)


SlackingSleeper:
Screenshot from 2026-08-30 12-38-31

Hellmonk:
I like the concept here and would be good to expand on. Ideally I think instead of 'effect' it might be nice to customize the string depending on the spell (so Ozo can say "bonus AC", imp can say "imp weapon", something like that). 

For spells like passwall where the effect is randomized at some power levels, we will also want to use a function that accepts a "random" parameter which is true when we actually cast the spell and false when called for spell descriptions, and returns an unrounded result when random is false (otherwise you will see the parameter randomly change when reexamining the spell). See dice_def base_fragmentation_damage or many of the other damage functions in spl-damage.cc for an example of how we usually do this.

Definitely interested in implementing this in the future, so if you don't get around to the rest of it I will eventually.

SlackingSleeper:
> Ideally I think instead of 'effect' it might be nice to customize the string depending on the spell (so Ozo can say "bonus AC", imp can say "imp weapon", something like that).

Are you thinking this would be in a separate section above or below the power/range/noise/etc section? The reason I put it as "Effect" was to not mess with the indentation of the other ones.

Hellmonk:
No, I think the location is fine, but it might be nice to be a bit more descriptive than "effect". I haven't thought out the best way to do the layout for that, so maybe it's not worthwhile.

PR #5419, opened 2026-09-14
new welcome messages

Rosstin:
a handful of new welcome messages


pooka109:
Nice.

Although having the one message in the first person (If I were the Orb of Zot...) is a little odd.

Maybe put it in quotes, as though the player's talking to themselves?

Rosstin:
> Nice.
> 
> Although having the one message in the first person (If I were the Orb of Zot...) is a little odd.
> 
> Maybe put it in quotes, as though the player's talking to themselves?

"Hmm... If I were the Orb of Zot, I'd be down here."

PR #5420, opened 2026-09-15
Prevent ungrammatical randart names

powerbf:
Two randart name patterns can lead to ungrammatical names

1) `of @_adjective_@ @_power_or_anger_@` because @\_power\_or\_anger\_@ can resolve to a string starting with "the" (via @\_anger\_name\_@ -> @\_anger\_possessive\_@), which means we can end up with an adjective before "the".

2) `of the @_verbing_@ @_battle_or_war_@` because some of the possible values for @\_battle\_or\_war\_@ start with "the", so we can end up with a result containing two instances of "the" (e.g. "of the Baking the Warrior")

To fix this, I removed these two options and rolled their weights back into the versions without @\_adjective\_@/@\_verbing\_@ (their weights were reduced when the versions with @\_adjective\_@/@\_verbing\_@ were added in this commit: https://github.com/crawl/crawl/commit/230a35465e1d75633806e52badab1ca7706c5b13)


DracoOmega:
I'm not certain how much effort it would be in this case (or how important), but randart book name generation does a little extra work to make sure to strip leading articles out of strings in cases where that would be inappropriate. And I wonder if that might be usually generalised in some way for this sort of issue.

PR #5422, opened 2026-09-15
Use exact version for homebrew in gh actions

Predelnik:
Apparently @master or @main is unsupported now, see https://brew.sh/2026/09/13/homebrew-7.0.0/
Replacing with the most recent version.


robertxgray:
You are right, I pointed to main but this is better. Thank you very much.

PR #5423, opened 2026-09-15
New unrands for the 'new' armour egos

Hellmonk:
- +8 Arcane Splint Mail, a chainmail unrand that grants Archery and coronas 
  enemies hit by projectiles (even thrown items, if you are allergic to using 
  archery).
- +6 bone scales, a Death pds modified to have plate level encumbrance (so 
  it gives plate stats while being wearable by offsize species, sort of like 
  armour of Talos but in a more useful way). Also grants a bunch of other
  properties, but banes on equip.
- +12 Forgewarden's cuirass, a Resonance plate that comes with an enhancer
  but reduces piety gain because it was blasphemously devoted to a certain
  god that does not exist.

### DIFF
--- a/crawl-ref/source/dat/descript/egos.txt
+++ b/crawl-ref/source/dat/descript/egos.txt
@@ -171,8 +171,10 @@ It protects you from the effects of acid and corrosion.
 %%%%
 death (Death) armour ego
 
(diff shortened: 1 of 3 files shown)


pooka109:
Cool.

One minor nit regarding the last one: it should be "cuirass" instead of "curiass".

PR #5426, opened 2026-09-18
Monster configuration: minor YAML fixes, documentation fixes, and documentation improvements

desiderautumn:
See each individual commit. For any fixed or expanded comments, such as the new documentation on energy and resists, the correctness was confirmed manually via wizmode.


pooka109:
Looks good to me. Applied; thanks.

(The test spawner is the one monster which goes over the resist limits, but it's an odd case that needs to be flat-out immune to some things and shouldn't be placed except in testing. In any case, your documentation changes are a definite improvement.)

desiderautumn:
Yes, I remarked on IRC about some of the weirdest curiosities that I found in the config files, among those being that resists go up to 4 but functionally stop at 3, making rF+++ impossible to grant. I was tempted to tweak test spawners to make them fall in line with every other monster, but I figured it was basically harmless and in any case the mon-gen.py script properly validates the range of values. Thanks for the review!

PR #5432, opened 2026-09-19
Update Russian descriptions for current content

SeijiOokami:
Brings the Russian description files back in line with the current
English source. They had not been updated in several versions, so the
game fell back to English in the places a new player sees first:
character creation, the skill menu, and every portal branch.

Two commits, both scoped to `crawl-ref/source/dat/descript/ru/`. No
code, no English text, no other language.

**1. Remove obsolete entries and match English order.** 18 keys for
content that no longer exists (the merged/renamed Bows, Crossbows,
Slings, Transmutations and Poison Magic skills; the Centaur, Deep
Dwarf, Halfling, Hill Orc, Ogre and Vampire species; the Abyssal
Knight, Arcane Marksman, Assassin, Transmuter and Venom Mage
backgrounds; `CMD_PRAY`). Entries are also reordered to follow the
English files, so translation diffs from here on show only text that
actually changed. No description text is touched by this commit.

**2. Translate 45 new entries.** Five files reach full coverage:

| file | before | after |
|---|---|---|
| `skills.txt` | 25/29 | 29/29 |
| `species.txt` | 19/27 | 27/27 |
| `backgrounds.txt` | 18/26 | 26/26 |
| `commands.txt` | 87/96 | 96/96 |
| `branches.txt` | 21/37 | 37/37 |

Notes on the translation itself:

- Terminology follows the existing translation wherever it had already
  made a choice, so the new entries read consistently with the old
  ones rather than introducing a second vocabulary.
- The `Ice Cave` Lua conditional is carried over structurally
  unchanged; only its two string literals are translated.
- `quotes.txt` is deliberately left alone, as
  `docs/develop/translation.txt` asks.

Checked with `util/checkwhite` and the `commit-msg` hook. Verified that
every key in the five files now matches the English source exactly,
with no leftover or missing keys, and that the files are UTF-8 with no
BOM.

One thing I could not check myself: how this looks in a running game.
I do not have a build set up, only the release binary, so this has been
verified structurally rather than visually. Happy to fix anything that
reads badly in context.

---

Unrelated to this PR, I noticed two formatting bugs in the **English**
`commands.txt` while working on it, and can send a separate PR if
useful:

- line 26: `CMD_EXPLORE_NO_REST` appears twice; the second is clearly
  meant to be `CMD_EXPLORE_NO_REST verbose`, so that text is currently
  unreachable.
- line 39: `CMD_IGNORE_INVISIBLE verbose` sits inside the preceding
  entry, missing its `%%%%` separator, so it renders as part of the
  short description.


pooka109:
Yes, please send the separate PR for the English commands.txt.

SeijiOokami:
Sent as #5433. It also notes three more duplicate keys I found while scanning the rest of `dat/descript/` — left untouched there, since each has two entries with different bodies and someone would need to decide which is intended.

pooka109:
Applied; thanks.

(Although, in the future, could you please keep the first lines of your commits at a maximum of 50 characters? Thanks in advance.)

PR #5433, opened 2026-09-19
Fix two unreachable entries in commands.txt

SeijiOokami:
As requested in #5432.

Two entries in `commands.txt` exist in the file but are never served,
because of how they are delimited.

**`CMD_EXPLORE_NO_REST` appears twice.** The second occurrence is
plainly the verbose description — it explains the
`explore_auto_rest` / `explore_auto_rest_status` behaviour — but it
repeats the plain key rather than using `CMD_EXPLORE_NO_REST verbose`.

**`CMD_IGNORE_INVISIBLE verbose` is missing its `%%%%`.** It sits
inside the `CMD_IGNORE_INVISIBLE` entry, separated only by a blank
line, so its text reads as part of the short description instead of
being an entry of its own.

Both are one-line fixes; no text is reworded.

---

While scanning for this pattern across all of `dat/descript/`, three
more duplicate keys turned up. I have left them alone since they are
outside what was asked for here, but I am happy to send a follow-up if
you want them fixed — each has two entries with *different* bodies, so
someone would need to decide which one is intended:

- `hints.txt`: `HINT_WIELD_WEAPON` at lines 814 and 1215
- `monstatus.txt`: `damage-immune at range monstatus` at lines 174
  and 742
- `mutations.txt`: `+LOS mutation` at lines 739 and 817


pooka109:
Applied; thanks.

Yes please on the follow-up for the other duplicate entries.

What name would you like to be listed under in the credits?

SeijiOokami:
Thanks!

SeijiOokami, please.

Follow-up is #5434. Note that only two of the three turned out to be real duplicates — `HINT_WIELD_WEAPON` is a key collision between two different call sites, and the Ranger hint from `fight.cc` is currently unreachable because of it. Details in that PR; happy to send the code fix separately if you want it.

pooka109:
Done.

PR #5434, opened 2026-09-19
Remove two shadowed duplicate description entries

SeijiOokami:
Follow-up to #5433, as requested.

`monstatus.txt` defines `damage-immune at range monstatus` twice and
`mutations.txt` defines `+LOS mutation` twice. Each pair describes the
same thing in different words, and each key names exactly one game
object — `MB_WARDING` in `mon-info-flag-name.h`, and a single `"+LOS"`
in `mutation-data.h` — so these really are duplicates rather than two
things sharing a name.

`_add_entry()` stores descriptions with
`dbm_store(..., DBM_REPLACE)`, so the later entry wins and the earlier
one is dead text no player can reach. In both cases the later entry is
also the more detailed one, so this removes the earlier entry and
leaves what the game already shows unchanged.

The duplicates came from later descriptions being written without
noticing the key was taken: e78965e76f for the monster status,
ef7f3e67aa for the mutation.

---

The third duplicate I mentioned, `HINT_WIELD_WEAPON` in `hints.txt`,
turned out **not** to be a duplicate, so it is not in this PR. The two
bodies belong to two different call sites:

- `hints.cc:1489`, reached from `fight.cc:1052` when a Ranger is about
  to melee while wielding a launcher, wants *"You can easily switch
  back to your weapon in slot a…"*
- `hints.cc:1878`, in the description of an unequipped weapon, wants
  *"You can wield this weapon with…"*

They collided in f0ccbfa403 ("Externalise remaining hint text", 10 Sep
2026), which externalised the second text under a key the first was
already using. Since `DBM_REPLACE` means the last entry wins, the
Ranger hint is currently unreachable.

Fixing that needs a new key and a one-line change in `hints.cc`, so it
is a code change rather than a text one. I can send it as its own PR
if you would like — just say which way you would prefer the key split
(rename the newly added one, or the original).


pooka109:
Applied; thanks.

As for the hints issue, my preference would be to rename the first hint's key to HINT_WIELD_MELEE_WEAPON.

SeijiOokami:
Thanks! Sent as #5435 — renamed to `HINT_WIELD_MELEE_WEAPON` as you preferred.

Flagging one thing there: I don't have a build set up, so it's verified by inspection, not compiled. It's a mechanical rename across five lines, but worth a compile before merging.

PR #5435, opened 2026-09-19
Rename the melee hint key so the Ranger hint works again

SeijiOokami:
As requested in #5434 — renaming the first hint's key to
`HINT_WIELD_MELEE_WEAPON`.

Two different hints shared `HINT_WIELD_WEAPON`:

- `hints.cc:1489`, reached from `fight.cc:1052` when a Ranger is about
  to melee something while wielding a launcher, wants *"You can easily
  switch back to your weapon in slot a…"*
- `hints.cc:1878`, in the description of an unequipped weapon, wants
  *"You can wield this weapon with…"*

Since `_add_entry()` stores descriptions with
`dbm_store(..., DBM_REPLACE)`, the later entry in `hints.txt` wins —
the weapon description — so the Ranger hint has been unreachable since
f0ccbfa403 externalised the second text under the key the first
already used.

This renames the Ranger hint and its enumerator to
`HINT_WIELD_MELEE_WEAPON`, and leaves `HINT_WIELD_WEAPON` to the
weapon description. Five lines in four files; no text is reworded.

**Saves are unaffected.** `Hints.hints_events` is indexed by the
enumerator's position and the save compatibility check is on
`HINT_EVENTS_NUM` — the rename changes neither the order nor the
count. Hint keys are only ever literal strings, so nothing constructs
`"HINT_…"` at runtime.

A note on verification: I do not have a build environment set up, so
this is checked by inspection rather than compiled and played. The
change is a mechanical rename — `grep` for `HINT_WIELD_WEAPON` now
returns exactly the two call sites that should keep it, and
`HINT_WIELD_MELEE_WEAPON` exactly the five that should have it — but a
compile before merging would be worth it.


pooka109:
Applied; thanks.

(I had to make a few adjustments: adding another hint, having them both actually show up, and keeping binary compatibility when doing so is a bit more involved of a change.)

PR #5436, opened 2026-09-19
Fix Russian entries that describe things the game no longer does

SeijiOokami:
Nine Russian entries that a key-level check calls translated, but whose
text describes things the game no longer does.

**Dead `$cmd[CMD_PRAY]` — 3 places.** The command is gone;
`name_to_command` returns `CMD_NO_CMD`, whose key is `'\0'`, so
`SDL_GetKeyName` yields an empty string and the instruction renders as
empty brackets.

- `tutorial5 sacrificing` and `dissection reminder` describe
  sacrificing corpses, have no English counterpart, and nothing asks
  for them — removed.
- `tutorial5 altar` also still told the player to confirm with `Y`
  instead of `J`/`Enter`, and described an outdated click — now a
  translation of the current English entry. This is a beginner-facing
  step in the tutorial.

`tutorial2 darts` is removed on the same grounds: no English entry, and
no caller in `dat/des`.

**`HINT_CONVERT` was never translated** — and the English it carries is
an old revision that describes excommunication differently from the
current text and hardcodes `^` in place of
`$cmd[CMD_DISPLAY_RELIGION]`.

**Four entries had lost their `{{ }}` conditionals**, so text meant for
one case showed in every case:

| entry | effect |
|---|---|
| `Invocations` | told Djinn about magical reserves |
| `Spellcasting` | same, and still mentioned magic hunger |
| `fungus` | fixed the colour that varies by branch |
| `skin of Zhor` | fixed the smell, and had dropped a whole sentence |

`skin of Zhor` needed an `else` branch where English has none: Russian
adjectives agree with their noun, so the noun has to sit inside the
conditional rather than after it.

---

Found with a script comparing markup, paragraph count and length
between each English entry and its translation. It flags 238 of the 892
entries that exist in both languages; this PR takes only the ones where
the defect is unambiguous. The rest — mostly entries where the Russian
has lost half the English text, such as `potion of berserk rage` losing
every drawback of berserking — need judgement per entry and would be
better as separate, smaller PRs.

Note on merge order: #5432 rewrites `ru/skills.txt`, so whichever of
the two lands second will need a rebase. Happy to do that whenever you
like.


pooka109:
Applied; thanks.

SeijiOokami:
Thanks! I think this one may not have made it to master, though — `tutorial5 sacrificing` is still in `ru/tutorial.txt` and the Spellcasting/Invocations conditionals aren't in `ru/skills.txt`. Possibly lost in the squash with #5432, which also touched `ru/skills.txt`. I've rebased it onto current master (applies cleanly, subject now under 50 chars) as #5445.

And noted on the 50-character subjects — will keep to that from now on.

pooka109:
Thank you.

Sorry about the confusion; having to keep track of multiple PRs can get tricky.

PR #5437, opened 2026-09-19
Restore content missing from Russian item descriptions

SeijiOokami:
Fifteen entries in `ru/items.txt` had lost most of what the English
says. The key is present and the text is Russian, so a key-level check
calls them translated — but the player does not learn what the item
does.

Potions and scrolls are read constantly, and a missing drawback can
cost a character:

| entry | what the Russian omitted |
|---|---|
| `potion of berserk rage` | everything except "causes a rage": no melee-only restriction, no fatigue, no inability to berserk again |
| `scroll of fear` | which monsters are immune |
| `scroll of teleportation` | that reading it again cancels a pending teleport |
| `potion of mutation` | that it removes mutations before granting new ones |
| `scroll of acquirement` | the choice, the skill matching, the gold option |
| `potion of resistance` | which resistances (said "poison and the elements") |
| `scroll of fog` | that the fog takes time to spill out |
| `potion of enlightenment` | the Will increase — only the levitation was kept |
| `staff of conjuration` | the Evocations effect, and the sentence was unterminated |
| `pair of boots`, `pair of gloves` | everything after the first three words |

`potion of invisibility` had also dropped its `[[player invis desc]]`
include — the block explaining how invisibility actually behaves
(accuracy, attacks of opportunity, the short-blade bonus, water). That
block had no Russian entry at all, so it is translated here too.

Terminology follows the existing translation where it had already made
a choice: Conjurations and Evocations keep the names `ru/skills.txt`
uses. `Will` had never appeared in the Russian files before.

---

Found with a script comparing markup, paragraph count and length
between each English entry and its translation — the same one behind
#5436. For `items.txt` it flagged 33 of 77 entries; this PR takes the
15 where content is plainly missing, and the count is now 19, of which
the only remaining length flag is `scroll of acquirement` sitting just
under the threshold with nothing actually missing.

`monsters.txt` (69 flagged), `unrand.txt` (31) and `spells.txt` (24)
have the same problem and would suit separate PRs.


pooka109:
Applied; thanks.

PR #5438, opened 2026-09-19
Restore content missing from Russian spell descriptions

SeijiOokami:
Same treatment as #5437, this time for `ru/spells.txt`. Seventeen
entries had lost most of what the English says, and several had drifted
far enough to be actively wrong.

**Wrong, not merely short:**

| spell | what the Russian said |
|---|---|
| `Invisibility` | contamination building up and making you glow back into visibility — a mechanic the game no longer has. Nothing about the accuracy, EV and blocking penalties the current entry is about. |
| `Fulminant Prism` | a prism draining magic from its surroundings, rather than the unstable gold alloy that explodes; and nothing about destroying it early weakening the blast |
| `Teleport Other` | that the target moves at random, rather than out of the caster's sight after a delay |
| `Force Lance` | knockback conditioned on the target's defences, instead of the extra damage from being slammed into a wall or another creature |

**Omissions a caster needs to know:** that `Freeze` bypasses armour;
that `Sticky Flame` is put out if the victim moves and will not stick
to insubstantial targets; that `Shatter` is weak against flying,
gelatinous and insubstantial creatures, and strong against ice, bone
and petrified ones; that `Sting` poisons even resistant targets and
damages immune ones; that `Static Discharge` bypasses half of armour
and deals less when arcing back; that `Blink` cannot be recast for a
while; that `Cause Fear` ends when the target leaves line of sight;
that `Leda's Liquefaction` shrinks over time and slows the caster even
in flight. Also `Call Imp`, `Charm`, `Inner Flame`, `Paralyse`,
`Summon Small Mammal`.

---

`spells.txt` went from 24 flagged entries to 8, and none of the
remaining ones is missing content — they are length-ratio noise on
short entries.

Remaining files with the same problem: `monsters.txt` (69 flagged) and
`unrand.txt` (31). Happy to keep going at whatever pace suits review.


pooka109:
Applied; thanks.

PR #5443, opened 2026-09-22
Fixing a typo regarding respawning monsters in arena

Chestertonian:
Original (crawl/crawl-ref/docs/arena.txt):
"will make it so that when one rat dies another takes it's place,"

This should read:
"will make it so that when one rat dies another takes its place,"


pooka109:
Applied; thanks.

